PR #7, opened 2025-07-03
refacto reranker

Ahmath-Gadji:


Ahmath-Gadji:
@EnjoyBacon7 ragatouille and fastembed-gpu uneeded now and are deleted

PR #85, opened 2025-09-26
Feat quick start

Ahmath-Gadji:
- **Quick Start**: Created a **quick_start** folder containing docker-compose configurations that use pre-built images for rapid deployment
- **Environment Variables**: Updated variable names to match the new [indexer-ui](https://github.com/linagora/openrag-admin-ui) naming conventions by removing `VITE_` prefixes and updating configuration files accordingly
- **Documentation**: Enhanced README and setup guides with clearer, step-by-step deployment instructions
- **Docker Compose**: Restructured the main docker-compose file and refreshed service templates for better organization
- **Maintenance**: Updated submodules to latest versions and removed deprecated configuration files

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,20 +2,21 @@ include:
   - vdb/milvus.yaml
   - ${CHAINLIT_DATALAYER_COMPOSE:-extern/dummy.yaml}
   - extern/infinity.yaml
-  - ${INDEXERUI_COMPOSE_FILE:-extern/indexer-ui/docker-compose.yaml}
+  # - ${INDEXERUI_COMPOSE_FILE:-extern/indexer-ui/docker-compose.yaml}
 
 x-openrag: &openrag_template
   image: ghcr.io/linagora/openrag:dev-latest
+  # image: linagoraai/openrag:latest
   build:
     context: .
     dockerfile: Dockerfile
   volumes:
-    - ${CONFIG_VOLUME:-./.hydra_config}:/app/.hydra_config
+    - ${CONFIG_VOLUME:-./.hydra_config}:/app/.hydra_config # For dev mode
     - ${DATA_VOLUME:-./data}:/app/data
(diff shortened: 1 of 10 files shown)


docker-compose.yaml
```diff
@@ -123,11 +137,11 @@ services:
       --model ${EMBEDDER_MODEL_NAME:-jinaai/jina-embeddings-v3}
       --trust-remote-code
       --dtype float32
-      --max-num-batched-tokens 32768
+      --max-model-len ${MAX_MODEL_LEN:-16384}
```
dodekapod, line 140: As far as I can remember, this wouldn't work on ARM (see the comment below).

docker-compose.yaml
```diff
@@ -43,19 +44,32 @@ x-vllm: &vllm_template
...
 
   healthcheck:
     test: ["CMD", "curl", "-f", "http://localhost:8000/health"]
-    interval: 15s
+    interval: 20s
     timeout: 5s
-    retries: 3
-    start_period: 60s
+    retries: 4
+    start_period: 90s
```
dodekapod, line 56: For the `vllm-cpu` it's better to put `start_period == 180s`.

docker-compose.yaml
dodekapod: With v0.9.2 this seems to work.

dodekapod:
* I managed to run the quick-start version with vllm `start_period: 120s`.
* I can't index neither pdf via indexer-ui or plain text from command line. This could be an authentication-related problem, or the indexer UI might not be able to find OpenRAG.

Here are the error messages for plain text indexing:

```
INFO:     172.24.0.1:48466 - "GET /q/partition/check-file/simplewiki-500/file/December.txt HTTP/1.1" 404 Not Found
INFO:     172.24.0.1:48472 - "POST /q/indexer/partition/simplewiki-500/file/December.txt HTTP/1.1" 404 Not Found
```

PR #112, opened 2025-10-21
fixed user access to /tasks

htagourti:


EnjoyBacon7:
Why was this PR reverted?

PR #154, opened 2025-11-27
Feat/spoken style answer

Ahmath-Gadji:
This feature lets OpenRAG generate short, direct, spoken-style answers.
It’s especially useful when OpenRAG is used as the backend of a voice chat system in order to have succint answers instead of highly detailed answers.


dodekapod:
LGTM

PR #158, opened 2025-12-02
Update helm

htagourti:


charts/openrag-stack/values.yaml
```diff
@@ -143,8 +212,8 @@ ingress:
 env:
   config:
     # LLM
-    BASE_URL: "https://example.com/v1/"
-    MODEL: "Mistral-Small-3.1-24B-Instruct-2503"
+    BASE_URL: "http://{{ .Release.Name }}-llm-engine-service/v1"
```
paultranvan: I might misunderstand something, but don't we need to have this `llm-engine-service` defined somewhere in the helm chart config ? I don't find it

PR #171, opened 2025-12-10
Add env vars when running backup/restore scripts as their might be fa…

Ahmath-Gadji:
This pull request updates the smoke test workflow to make sure the environment variables from the .env file are available during the backup and restore steps. In some cases, this is much needed
The main change is adding the --env-file .env option to the relevant docker compose run commands.


Ahmath-Gadji:
@dodekapod the documentation and the examples have been updated

Issue #201, opened 2026-01-12
IndexError: list index out of range in chunker contextualization


Ahmath-Gadji:
### Problem
Getting `IndexError: list index out of range` when trying to contextualize chunks.

**Location:** 
https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L136-L140

### Error
```bash
File "/app/openrag/components/indexer/chunker/chunker.py", line 140, in _apply_contextualization
    filename = chunks[0].metadata.get("filename")
               ~~~~~~^^^
IndexError: list index out of range
```

### Cause

This error occurs when indexing a PDF that contains only an image. If the image is not informative:

1. The VLM labels it as **“[Image Placeholder]”**.

2. As a result, `_get_chunks` returns an empty list because irrelevant images are filtered out in `_prepare_md_elements`:

   * [https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L162-L166](https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L162-L166)
   * [https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L195-L199](https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L195-L199)
   * [https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L282-L283](https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L282-L283)

3. Contextualization is then executed with an empty chunk list, which ultimately triggers the error:

   * [https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L289](https://github.com/linagora/openrag/blob/ce7a280ff97ef486fc728bf868329638d3e3a739/openrag/components/indexer/chunker/chunker.py#L289)

PR #202, opened 2026-01-12
correct IndexError caused by empty list of chunks when performing chu…

Ahmath-Gadji:
This PR makes the chunking pipeline more robust by:
- Passing `filename` explicitly to avoid the metadata extraction issue in #201
- Logging warnings when no chunks are generated
- Fixing false-positive filtering of informative images:
  - Uninformative images have the exact format: `"""<image_description>\n\n[Image Placeholder]\n\n</image_description>"""`
  - The old filter checked if `[Image Placeholder]` appeared anywhere in the description, incorrectly removing informative images where the VLM happened to include that phrase
  - The new filter uses exact pattern matching to only remove true placeholders

## Summary by CodeRabbit

* **Bug Fixes**
  * Returns empty results and logs a warning when no chunks are produced, preventing downstream errors.
  * Case-insensitive detection of image placeholders during markdown processing to avoid false positives.

* **Improvements**
  * Filenames are preserved and propagated through the chunking/contextualization pipeline for clearer metadata and messages.
  * More consistent logging and contextualization behavior across the pipeline.

### DIFF
--- a/openrag/components/indexer/chunker/chunker.py
+++ b/openrag/components/indexer/chunker/chunker.py
@@ -10,6 +10,8 @@ from omegaconf import OmegaConf
 from tqdm.asyncio import tqdm
 from utils.logger import get_logger
 
+from openrag.consts import IMAGE_PLACEHOLDER
+
 from ..embeddings import BaseEmbedding
 from .utils import MDElement, chunk_table, get_chunk_page_number, split_md_elements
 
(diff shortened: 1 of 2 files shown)


openrag/components/indexer/chunker/chunker.py
```diff
@@ -158,10 +164,13 @@ def _prepare_md_elements(
         md_elements: list[MDElement] = split_md_elements(content)
 
         tables_and_images, texts = [], []
+        img_placeholder = (
+            """<image_description>\n\n[Image Placeholder]\n\n</image_description>"""
```
paultranvan: This should probably moved in a constant file, e.g. https://github.com/linagora/openrag/blob/main/openrag/consts.py

PR #210, opened 2026-01-15
ci: remove smoke_tests workflow, replace with api_tests

paultranvan:
Smoke tests took too long to execute in CI. The api_tests workflow provides broader endpoint coverage and is faster. Moved the smoke tests and data in tests/data so it is still possible to be  manually run


Ahmath-Gadji (commented):
LGTM

Ahmath-Gadji (approved):

PR #217, opened 2026-01-19
Add `image_captioning_url` config param to control indexing of image URL

paultranvan:
When  `image_captioning` is true, we used to try to caption with VLM any image found as a URL in markdown. 
This can lead to errors if the image is wrong or not reachable. Also, this capability might be not expected, because it is a costly operation and the image could be irrelevant to caption, because it is an icon for example.

So we add a new param `image_captioning_url` to control this behaviour. 

This also make a refactoring of the loaders, to mutualize code and be consistent for image captioning  

## Summary by CodeRabbit

* **New Features**
  * Image captioning is enabled by default for supported file types.
  * Optionally caption HTTP/HTTPS image URLs in markdown via a new IMAGE_CAPTIONING_URL setting.
  * Unified, centralized image captioning flow across PDFs, Word, PowerPoint, and Markdown.

* **Documentation**
  * Added documentation entry for IMAGE_CAPTIONING_URL (default: true).

* **Tests**
  * Added tests and a fixture covering captioning for embedded data-URI and HTTP markdown images.

### DIFF
--- a/openrag/components/indexer/loaders/base.py
+++ b/openrag/components/indexer/loaders/base.py
@@ -9,6 +9,7 @@ from components.prompts import IMAGE_DESCRIBER
 from components.utils import get_vlm_semaphore, load_config
 from langchain_core.messages import HumanMessage
 from langchain_openai import ChatOpenAI
+from openai import BadRequestError
 from PIL import Image
 from tqdm.asyncio import tqdm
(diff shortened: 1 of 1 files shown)


Ahmath-Gadji (approved):
LGTM.
I've tested PDFs files with images and it works fine

PR #267, opened 2026-03-04
fix: only check LLM model availability and improve error handling

Ahmath-Gadji:
This PR restores the previous implementation of `check_llm_model_availability`, which verifies only the LLM’s availability. This logic was originally introduced in PR #225 but was accidentally overridden by commit https://github.com/linagora/openrag/commit/bc98a75a0d9dbc518b48012c712d0be4902872e7 .

There is no need to check the VLM’s availability, as the endpoints using this dependency do not rely on any VLM components.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer HTTP status mapping for downstream failures (timeouts → 504, connection issues → 503, API errors → propagated or 500).
  * Incomplete LLM configuration now returns 500; missing configured model returns 404 and includes available models in the response.
  * Unified, clearer error messages referencing the single-model endpoint and the specific model involved.

* **Refactor**
  * Simplified single-model validation flow with consolidated logging and per-LLM timeout handling for more consistent behavior.

### DIFF
--- a/openrag/routers/utils.py
+++ b/openrag/routers/utils.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from typing import Any
 
 import consts
-import httpx
 import openai
 from config import load_config
 from fastapi import Depends, Form, HTTPException, Request, UploadFile, status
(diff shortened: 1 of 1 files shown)


EnjoyBacon7:
Addressed the three issues raised in the CodeRabbit review. Three separate commits were pushed to this branch:

**1. Fail-fast on missing config** (`4973b82`)

If `base_url`, `model`, or `api_key` is absent/empty the function now raises `HTTP 500` immediately with `"LLM configuration is incomplete"` before attempting any network call. Previously a `None` value would either crash with an opaque error or produce a misleading 404 saying model `None` is unavailable.

**2. Use configured timeout** (`d97b31c`)

Replaced the hardcoded `timeout=30` with `int(llm_param.get("timeout", 30))` so `config.llm["timeout"]` (default 60, env-configurable) is respected. Slow LLM deployments with a higher configured timeout were hitting a premature 504 during model availability checks.

**3. Graceful empty model list** (`c16f6c8`)

If the LLM endpoint returns zero models, `", ".join(available_models)` was producing a blank `"Available models: "` string. It now shows `"(none)"` instead.

PR #272, opened 2026-03-06
Add workspaces

paultranvan:
Workspace is a new concept that can be seen as sub-partition spaces for files.
Inside a partition, we can now organize files into workplaces and search for files for a given workspace.

The main reason to use workspace rather than dedicated partitions is to avoid file duplication: a single file can belong to multiple workplace, while a same file can exist in multiple partitions. 

This is typically useful to clients with 1 partition per user, each user having the possibility to have workplaces with specific files into it, to create specialized context assistants.

## Summary by CodeRabbit

* **New Features**
  * Workspace management within partitions: create/list/get/delete workspaces; add/remove/list files in workspaces; workspace-scoped indexing and search/filtering; public workspace APIs and DB migration.

* **Documentation**
  * New Workspaces guide with examples and diagrams; updated data model and migration instructions with multi-option deployment steps.

* **Tests**
  * Added API tests for workspace CRUD, file associations, search filtering, and cascade behaviors.

* **Chores**
  * .gitignore updated to ignore docs/plans/ directory.


docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+Workspaces let you organize files within a partition into named subsets. When searching or chatting, you can target a specific workspace so that only its files are considered — without affecting the underlying partition structure.
+
+---
+
+## Concepts
+
+- A **workspace** belongs to exactly one partition
+- A file can belong to **multiple workspaces** (or none)
+- Workspaces do not duplicate files — they reference existing partition files
+- Deleting a workspace deletes **orphaned files** (files not in any other workspace) from the partition automatically
```
Ahmath-Gadji, line 15: Is this a good idea?
Workspaces are just some kind of view on specific files, so i think it's safer to not delete files.

docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+---
+
+Workspaces let you organize files within a partition into named subsets. When searching or chatting, you can target a specific workspace so that only its files are considered — without affecting the underlying partition structure.
+
+---
+
+## Concepts
+
+- A **workspace** belongs to exactly one partition
+- A file can belong to **multiple workspaces** (or none)
```
Ahmath-Gadji, line 13: Why this restriction? It might be useful for users to mix files from different partitions, especially if they can have multiple partitions. Otherwise, this limitation could lead users to duplicate files across partitions simply to make them available in a workspace.

docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+**Constraints:**
+- `UniqueConstraint(workspace_id)` — workspace IDs are globally unique
+- `UniqueConstraint(workspace_id, file_id)` on `workspace_files` — a file appears at most once per workspace
+- Cascade delete: dropping a workspace removes its `workspace_files` rows
+
+---
+
+## API Endpoints
+
+All workspace endpoints live under `/indexer/partition/{partition}/workspaces`.
```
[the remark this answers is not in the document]
Ahmath-Gadji: @paultranvan there is indeed a small typo here

docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+| POST | `/partition/{partition}/workspaces/{workspace_id}/files` | Editor | Add files to workspace |
+| GET | `/partition/{partition}/workspaces/{workspace_id}/files` | Viewer | List files in workspace |
+| DELETE | `/partition/{partition}/workspaces/{workspace_id}/files/{file_id}` | Editor | Remove file from workspace |
+
+---
+
+### Create a Workspace
+
+```bash
+curl -X POST "$BASE_URL/indexer/partition/my-partition/workspaces" \
```
Ahmath-Gadji: As pointed before `indexer` is a typo here.

docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+      "created_at": "2026-03-06T10:00:00"
+    }
+  ]
+}
+```
+
+### Add Files to a Workspace
+
+```bash
+curl -X POST "$BASE_URL/indexer/partition/my-partition/workspaces/project-alpha/files" \
```
Ahmath-Gadji: https://github.com/linagora/openrag/pull/272/changes#r2911239348

docs/content/docs/documentation/workspaces.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+```json
+{"status": "added", "file_ids": ["report.pdf", "notes.md"]}
+```
+
+### Delete a Workspace
+
+```bash
+curl -X DELETE "$BASE_URL/indexer/partition/my-partition/workspaces/project-alpha" \
```
Ahmath-Gadji: https://github.com/linagora/openrag/pull/272/changes#r2911239348

openrag/components/indexer/vectordb/utils.py
```diff
@@ -720,3 +761,78 @@ def get_ancestor_file_ids(self, partition: str, file_id: str, max_ancestor_depth
...
+                select(WorkspaceFile.file_id)
+                .where(WorkspaceFile.workspace_id == workspace_id)
+                .where(WorkspaceFile.file_id.notin_(select(subq.c.file_id)))
+            )
+            orphaned_file_ids = [r[0] for r in result.all()]
+
+            # Delete workspace (cascades workspace_files)
+            session.execute(delete(Workspace).where(Workspace.workspace_id == workspace_id))
+            session.commit()
+            return orphaned_file_ids
```
Ahmath-Gadji, line 819: As pointed before when deleting a ws, i think we can leave the files

openrag/routers/search.py
```diff
@@ -81,7 +82,8 @@ async def search_multiple_partitions(
         include_ancestors=include_ancestors,
     )
```
EnjoyBacon7, line 88: **Security: cross-partition workspace data leak**

The `workspace` filter is forwarded to Milvus without validating that the workspace belongs to the requested partition(s). A user with access to partition A can pass `?workspace=<workspace_in_partition_B>` and receive chunks from partition B's files, bypassing partition isolation.

```python
# Fix — validate ownership before filtering
if workspace:
    ws = await vectordb.get_workspace.remote(workspace)
    if not ws or ws["partition_name"] not in partitions:
        raise HTTPException(status_code=404, detail="Workspace not found")
    filter_dict = {"workspace_id": workspace}
```

The same issue exists in `pipeline.py` (~line 152) where `workspace` from request metadata is used without checking it belongs to the target partition.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,110 @@
...
+async def create_workspace(
+    partition: str, body: CreateWorkspaceRequest, request: Request, vectordb=Depends(get_vectordb)
+):
+    await vectordb.create_workspace.remote(
+        workspace_id=body.workspace_id,
+        partition=partition,
+        user_id=request.state.user["id"],
+        display_name=body.display_name,
+    )
+    return {"status": "created", "workspace_id": body.workspace_id}
```
EnjoyBacon7, line 76: **Duplicate `workspace_id` raises unhandled 500 instead of 409**

(Adding to the `IntegrityError` note already raised on `utils.py`)

`workspace_id` has a `UNIQUE` constraint in the DB. If the same ID is submitted twice, a raw `sqlalchemy.exc.IntegrityError` propagates to the client as a 500. The established pattern (see `partition.py`) is a pre-existence check returning **409 Conflict**:

```python
async def create_workspace(partition: str, body: CreateWorkspaceRequest, request: Request, vectordb=Depends(get_vectordb)):
    existing = await vectordb.get_workspace.remote(body.workspace_id)
    if existing:
        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=f"Workspace '{body.workspace_id}' already exists.")
    await vectordb.create_workspace.remote(...)
    return Response(status_code=status.HTTP_201_CREATED, content=...)
```

Also note this endpoint returns HTTP 200 — it should return **201 Created** to match the `partition.py` convention.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,110 @@
...
+    "/partition/{partition}/workspaces/{workspace_id}/files",
+    dependencies=[Depends(require_partition_editor)],
+)
+async def add_files_to_workspace(
+    workspace_id: str,
+    body: AddFilesRequest,
+    vectordb=Depends(get_vectordb),
+    _ws=Depends(require_workspace_in_partition),
+):
+    await vectordb.add_files_to_workspace.remote(workspace_id, body.file_ids)
```
EnjoyBacon7: **Adding a duplicate `(workspace_id, file_id)` raises unhandled 500**

(Complementing the `IntegrityError` note already flagged on `utils.py`)

The `uix_workspace_file` unique constraint means submitting a `file_id` already in the workspace throws a raw `sqlalchemy.exc.IntegrityError` → 500. Two options:

1. **Idempotent upsert** (preferred for a bulk-add endpoint): use `INSERT ... ON CONFLICT DO NOTHING` so re-adding existing files is a no-op.
2. **Pre-check + 409**: fetch current workspace files, raise 409 for any already present.

Option 1 is more ergonomic for callers who don't want to track state on their side:
```python
# utils.py
from sqlalchemy.dialects.postgresql import insert as pg_insert

def add_files_to_workspace(self, workspace_id: str, file_ids: list[str]):
    with self.Session() as session:
        for fid in file_ids:
            stmt = pg_insert(WorkspaceFile).values(workspace_id=workspace_id, file_id=fid)
            stmt = stmt.on_conflict_do_nothing(constraint="uix_workspace_file")
            session.execute(stmt)
        session.commit()
```

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@router.delete(
+    "/partition/{partition}/workspaces/{workspace_id}",
+    dependencies=[Depends(require_partition_owner)],
+)
+async def delete_workspace(
+    partition: str, workspace_id: str, vectordb=Depends(get_vectordb), _ws=Depends(require_workspace_in_partition)
+):
+    orphaned = await vectordb.delete_workspace.remote(workspace_id)
```
EnjoyBacon7: **Partial deletion leaves inconsistent state — missing `return_exceptions=True`**

`vectordb.delete_workspace.remote()` commits to PostgreSQL before this line runs. If any `delete_file` call then raises (Milvus timeout, network error, file already gone), the bare `asyncio.gather` re-raises and the endpoint returns 500 — but the workspace is **already deleted** from the DB. The orphaned Milvus vectors have no corresponding record and there is no way to retry cleanly.

```python
# Fix: collect failures, log them, and still return success
if orphaned:
    results = await asyncio.gather(
        *[vectordb.delete_file.remote(fid, partition) for fid in orphaned],
        return_exceptions=True,
    )
    for fid, result in zip(orphaned, results):
        if isinstance(result, Exception):
            logger.warning("Failed to delete orphaned file from Milvus", file_id=fid, error=str(result))
```

Alternatively, move the Milvus deletion inside the DB transaction (not always feasible) or accept partial cleanup with explicit logging.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,110 @@
...
+    return {"file_ids": file_ids}
+
+
+@router.delete(
+    "/partition/{partition}/workspaces/{workspace_id}/files/{file_id}",
+    dependencies=[Depends(require_partition_editor)],
+)
+async def remove_file_from_workspace(
+    workspace_id: str, file_id: str, vectordb=Depends(get_vectordb), _ws=Depends(require_workspace_in_partition)
+):
```
EnjoyBacon7, line 178: **`remove_file_from_workspace` always returns 200 — no 404 for non-existent association**

The underlying `PartitionFileManager.remove_file_from_workspace` executes a `DELETE` without checking whether the row existed, so this endpoint returns `{"status": "removed"}` even when the `file_id` was never in the workspace.

If idempotency is intentional (acceptable for a DELETE endpoint), that's fine — but it should be documented. If a 404 is expected, a pre-check is needed:

```python
# In utils.py — check before delete
def remove_file_from_workspace(self, workspace_id: str, file_id: str) -> bool:
    with self.Session() as session:
        result = session.execute(
            delete(WorkspaceFile).where(
                WorkspaceFile.workspace_id == workspace_id,
                WorkspaceFile.file_id == file_id,
            ).returning(WorkspaceFile.id)
        )
        session.commit()
        return result.rowcount > 0
```

Then in the router raise 404 if `False` is returned.

openrag/components/indexer/vectordb/vectordb.py
```diff
@@ -595,6 +603,7 @@ async def delete_file(self, file_id: str, partition: str):
                 filter=f"partition == '{partition}' and file_id == '{file_id}'",
             )
 
+            self.partition_file_manager.remove_file_from_all_workspaces(file_id)
```
EnjoyBacon7: **`remove_file_from_all_workspaces` is not partition-scoped — cross-partition collision risk**

`WorkspaceFile` has no `partition_name` column. The cleanup deletes by `file_id` alone:

```python
# utils.py
def remove_file_from_all_workspaces(self, file_id: str):
    session.execute(delete(WorkspaceFile).where(WorkspaceFile.file_id == file_id))
```

The schema allows the same `file_id` string to exist in multiple partitions (there is no global uniqueness constraint on `files.file_id`). If partition A and partition B both have a file with the same `file_id`, deleting the file from partition A will also silently remove it from all workspaces in partition B.

**Fix:** Pass `partition` through and filter via a join or subquery:

```python
def remove_file_from_all_workspaces(self, file_id: str, partition: str):
    with self.Session() as session:
        # Only target workspaces that belong to this partition
        ws_ids = select(Workspace.workspace_id).where(Workspace.partition_name == partition)
        session.execute(
            delete(WorkspaceFile).where(
                WorkspaceFile.file_id == file_id,
                WorkspaceFile.workspace_id.in_(ws_ids),
            )
        )
        session.commit()
```

openrag/components/indexer/vectordb/vectordb.py
```diff
@@ -1086,6 +1095,32 @@ async def get_ancestor_chunks(
             for res in results
         ]
 
+    # --- Workspace methods ---
+
+    def create_workspace(
```
EnjoyBacon7: **Workspace methods are synchronous — inconsistent with other Ray actor methods**

All seven new workspace methods (`create_workspace`, `list_workspaces`, `get_workspace`, etc.) are plain `def`, while comparable actor methods (`create_user`, `delete_user`, `list_partition_members`) are `async def`. Synchronous DB calls block the Ray actor's event loop thread while executing, which can degrade throughput under concurrent load.

Consider making these `async def` (the underlying SQLAlchemy calls can be kept synchronous inside — Ray handles the event loop scheduling — this is consistent with how `update_user_quota` etc. work, but the outer method signature should match the pattern of the class).

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,110 @@
...
+    if not ws or ws["partition_name"] != partition:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Workspace not found")
+    return ws
+
+
+@router.post(
+    "/partition/{partition}/workspaces",
+    dependencies=[Depends(require_partition_editor)],
+)
+async def create_workspace(
```
EnjoyBacon7, line 50: **Inconsistent auth pattern: `dependencies=[]` + `request.state.user` instead of named dependency**

The established pattern in `indexer.py` captures the user from the dependency return value:

```python
# indexer.py — user object returned and used directly
async def add_file(..., user=Depends(require_partition_editor), ...):
    task = indexer.add_file.remote(..., user=user)
```

Here, `dependencies=[Depends(require_partition_editor)]` discards the returned user and forces a secondary lookup via `request.state.user`. This is functionally equivalent but is less explicit and deviates from the codebase convention. Prefer:

```python
async def create_workspace(
    partition: str, body: CreateWorkspaceRequest,
    user=Depends(require_partition_editor),   # captures and reuses the user
    vectordb=Depends(get_vectordb),
):
    await vectordb.create_workspace.remote(..., user_id=user["id"], ...)
```

openrag/components/indexer/vectordb/utils.py
```diff
@@ -720,3 +761,78 @@ def get_ancestor_file_ids(self, partition: str, file_id: str, max_ancestor_depth
...
+
+    def create_workspace(self, workspace_id: str, partition: str, user_id: int | None, display_name: str | None = None):
+        with self.Session() as session:
+            ws = Workspace(
+                workspace_id=workspace_id, partition_name=partition, created_by=user_id, display_name=display_name
+            )
+            session.add(ws)
+            session.commit()
+
+    def list_workspaces(self, partition: str) -> list[dict]:
```
EnjoyBacon7, line 776: **Inconsistent response fields between `list_workspaces` and `get_workspace`**

`list_workspaces` returns: `workspace_id`, `display_name`, `created_at`

`get_workspace` returns: `workspace_id`, `partition_name`, `display_name`, `created_at`

The list response is missing `partition_name` and `created_by`, while `get_workspace` is also missing `created_by`. REST convention expects the list items to be a subset (or identical) to the single-item shape. A caller iterating a list and then fetching individual items will get different shapes.

Suggest aligning both to return the same fields: `workspace_id`, `partition_name`, `display_name`, `created_by`, `created_at`.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 038fa5b: added `return_exceptions=True` to the `asyncio.gather` call so a failure in one delete doesn't prevent the others from completing.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 15e4b04: `remove_file_from_workspace` now returns 404 when the `(workspace_id, file_id)` row doesn't exist in `workspace_files`, rather than always returning 200.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 6ec506e: replaced `dependencies=[]` + `request.state.user` with a proper `user=Depends(require_partition_editor)` named dependency, consistent with every other endpoint in the router.

openrag/components/indexer/vectordb/utils.py
EnjoyBacon7: Fixed in commit 01fbb7a: `list_workspaces` now returns the full workspace dict (including `partition_name`, `created_by`, `created_at`) instead of only `workspace_id` and `display_name`.

docs/content/docs/documentation/workspaces.md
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit b24c96c: removed the `/indexer` prefix from all workspace endpoint references in the docs. The workspaces router is mounted without a prefix (see `api.py:255`), so the correct base path is `/partition/{partition}/workspaces`. The one remaining `/indexer/partition/...` occurrence on line 141 is intentional — it refers to the file upload endpoint which is correctly mounted under `/indexer`.

docs/content/docs/documentation/data_model.md
```diff
@@ -133,6 +151,38 @@ Defines the many-to-many relationship between users and partitions with role-bas
...
+
+### `workspace_files`
+
+Join table linking workspaces to files.
+
+| Column          | Type | Description |
+|------------------|------|-------------|
+| `id`             | Integer (PK) | Internal identifier |
+| `workspace_id`   | String (FK → `workspaces.workspace_id`, CASCADE) | Workspace reference |
+| `file_id`        | String (FK → `files.file_id`, CASCADE) | File reference |
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit 9b13b49: updated `data_model.md` to remove the incorrect FK claim for `workspace_files.file_id`. The actual `WorkspaceFile` model defines `file_id` as a plain `String` column with no `ForeignKey` constraint. The description now reads: 'File reference (no FK constraint — referential integrity enforced at application level)'.

openrag/components/indexer/vectordb/vectordb.py
EnjoyBacon7: Fixed in commit 350500a: all seven workspace methods (`create_workspace`, `list_workspaces`, `get_workspace`, `delete_workspace`, `add_files_to_workspace`, `remove_file_from_workspace`, `list_workspace_files`) are now `async def`, consistent with other Ray actor methods in the class.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: Fixed in commit b24c96c: the `/indexer` prefix has been removed from all workspace curl examples and the endpoint table header. The correct base path is `/partition/{partition}/workspaces`.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: Fixed in commit b24c96c — same as the other /indexer prefix occurrences.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: Fixed in commit b24c96c — same as the other /indexer prefix occurrences.

openrag/routers/search.py
EnjoyBacon7: Fixed in commit 8c8b8b8: both search endpoints (`search_multiple_partitions` and `search_one_partition`) now look up the workspace via `vectordb.get_workspace` and raise 404 if it doesn't belong to the requested partition(s). The same validation is added to `RagPipeline._prepare_for_chat_completion` in `pipeline.py` — there, an invalid workspace is logged as a warning and silently dropped rather than raising (since the pipeline has no HTTP response path), preventing the data leak while still completing the request.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 48a4920: `create_workspace` now calls `vectordb.get_workspace` first and raises `409 Conflict` if the ID already exists, consistent with the pattern in `partition.py`.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 352a98c: `add_files_to_workspace` now uses `INSERT ... ON CONFLICT DO NOTHING` (via `sqlalchemy.dialects.postgresql.insert`) targeting the `uix_workspace_file` constraint. Re-adding an already-present file is a silent no-op — no 500, no error response.

openrag/components/indexer/vectordb/vectordb.py
EnjoyBacon7: Fixed in commit 67b33d2: `remove_file_from_all_workspaces` now accepts a `partition` parameter and deletes only `workspace_files` rows where the workspace belongs to that partition (via a subquery on `Workspace.partition_name`). The call site in `delete_file` already had `partition` in scope and now passes it through.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: The current behavior matches the intended design: files are only deleted when the last workspace that references them is deleted (orphan detection). Files shared with other workspaces are preserved. Files can also be deleted independently of workspaces via `DELETE /partition/{partition}/file/{file_id}` as before. No change needed.

openrag/components/indexer/vectordb/utils.py
EnjoyBacon7: Same as above — the orphan detection in `delete_workspace` ensures only files exclusively owned by the deleted workspace are removed. Files belonging to other workspaces survive.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: By design, workspaces are scoped to a single partition — a workspace belongs to exactly one partition, and only files from that partition can be added to it. Cross-partition workspaces are intentionally not supported: they would break partition isolation (access control, data segregation) and significantly complicate the permission model. Users who need to search across partitions can use the existing multi-partition search (`?partitions=all`) instead.

openrag/routers/indexer.py
```diff
@@ -150,10 +152,37 @@ async def add_file(
...
     # Indexing the file
     task = indexer.add_file.remote(path=file_path, metadata=metadata, partition=partition, user=user)
     await task_state_manager.set_state.remote(task.task_id().hex(), "QUEUED")
     await task_state_manager.set_object_ref.remote(task.task_id().hex(), {"ref": task})
+
+    # Add file to workspaces
+    if parsed_workspace_ids:
+        await asyncio.gather(
+            *[vectordb.add_files_to_workspace.remote(ws_id, [file_id]) for ws_id in parsed_workspace_ids]
+        )
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit c1bf1ae: workspace association is now deferred into `Indexer.add_file()` and only runs after `set_state(COMPLETED)`. The router passes `workspace_ids` as a parameter to `add_file.remote()` and no longer touches `workspace_files` directly — so if indexing fails, no orphaned rows are created. The unused `import asyncio` in the router was also removed.

openrag/routers/search.py
EnjoyBacon7: Fixed in commit 8c8b8b8: both search endpoints and the RAG pipeline now validate that the requested workspace belongs to the target partition before using it as a filter. An invalid workspace now returns 404 instead of silently leaking data from another partition.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 48a4920: `create_workspace` in `utils.py` now catches `sqlalchemy.exc.IntegrityError` and raises an `HTTPException(409)` with a clear message. The router returns 409 Conflict instead of an unhandled 500.

openrag/routers/workspaces.py
EnjoyBacon7: Fixed in commit 352a98c: `add_files_to_workspace` in `utils.py` now uses `INSERT ... ON CONFLICT DO NOTHING` (via SQLAlchemy's `pg_insert(...).on_conflict_do_nothing(constraint="uix_workspace_file")`). Submitting a file already in the workspace is now idempotent — no error raised.

openrag/components/indexer/vectordb/vectordb.py
EnjoyBacon7: Fixed in commit 67b33d2: `remove_file_from_all_workspaces` in `utils.py` now scopes the delete to workspaces belonging to the file's partition. It first looks up the file's `partition_name` from the `files` table, then deletes only `WorkspaceFile` rows where the workspace's `partition_name` matches — preventing any cross-partition collision.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: The current behavior is intentional: when you delete a workspace, files that exist **only** in that workspace (i.e., not referenced by any other workspace) are deleted from the partition as well. Files shared with other workspaces are left intact.

The reasoning: if a file is not in any workspace after the workspace is deleted, it becomes permanently unreachable through the workspace API — users would have no way to discover or clean it up. Silently leaving such orphaned files behind would lead to storage leaks over time.

If a user wants to keep the files, they should move them to another workspace before deleting.

docs/content/docs/documentation/workspaces.md
EnjoyBacon7: This restriction is intentional and enforced by the data model: a workspace belongs to exactly one partition, and files are also partition-scoped. Allowing cross-partition files in a single workspace would require either duplicating files (defeating the reference-only design) or punching holes through partition access controls — which would undermine the multi-tenant isolation that partitions provide.

If a user needs the same file in multiple partitions they already have to upload it to each one; that's a partition boundary, not a workspace limitation. Workspaces are intentionally a sub-partition organizational unit.

openrag/components/indexer/vectordb/utils.py
EnjoyBacon7: See the reply above on the docs thread — only truly orphaned files (those not referenced by any other workspace) are deleted. Files that belong to at least one other workspace are untouched. The intent is to avoid invisible storage leaks from files that become permanently unreachable after their only workspace is removed.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,123 @@
...
+
+class AddFilesRequest(BaseModel):
+    file_ids: list[str]
+
+
+async def require_workspace_in_partition(partition: str, workspace_id: str, vectordb=Depends(get_vectordb)) -> dict:
+    """Validate that a workspace exists and belongs to the given partition."""
+    ws = await vectordb.get_workspace.remote(workspace_id)
+    if not ws or ws["partition_name"] != partition:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Workspace not found")
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit d31092c: all workspace Ray actor calls in `workspaces.py`, `search.py`, and `pipeline.py` are now wrapped with `call_ray_actor_with_timeout()`. A new `vectordb_timeout` config key (default 30s, overridable via env `VECTORDB_TIMEOUT`) was added under `ray.indexer` in `config.yaml`.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,123 @@
...
+logger = get_logger()
+
+
+class CreateWorkspaceRequest(BaseModel):
+    workspace_id: str
+    display_name: str | None = None
+
+
+class AddFilesRequest(BaseModel):
+    file_ids: list[str]
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit e021178: both `CreateWorkspaceRequest` and `AddFilesRequest` now declare `model_config = ConfigDict(extra="allow")`, matching the pattern used in the mock VLLM test server.

openrag/components/indexer/vectordb/vectordb.py
```diff
@@ -462,6 +462,14 @@ async def async_search(
...
 
         if filter:
+            filter = dict(filter)  # don't mutate caller's dict
+            if "workspace_id" in filter:
+                workspace_id = filter.pop("workspace_id")
+                file_ids = self.partition_file_manager.list_workspace_files(workspace_id)
+                if not file_ids:
+                    return []  # Empty workspace → no results
+                id_list = ", ".join(f'"{fid}"' for fid in file_ids)
+                expr_parts.append(f"file_id in [{id_list}]")
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in commit 47a0bf7: `async_search` now calls `get_workspace(workspace_id)` to fetch the workspace's `partition_name`, then strips any outer `partition in [...]` expression and replaces it with an exact `partition == "<ws_partition>"` equality before appending the `file_id in [...]` filter. This pins the search to the workspace's own partition regardless of whether the caller passed `partition=["all"]` or a named set.

openrag/components/indexer/indexer.py
```diff
@@ -116,9 +117,17 @@ async def add_file(
...
+            # Mark task as completed before workspace association so the file
+            # record exists in the DB before we reference it from workspace_files.
             await task_state_manager.set_state.remote(task_id, "COMPLETED")
 
+            # Associate with workspaces only after successful indexing
+            if workspace_ids:
+                vectordb = ray.get_actor("Vectordb", namespace="openrag")
+                await asyncio.gather(
+                    *[vectordb.add_files_to_workspace.remote(ws_id, [file_id]) for ws_id in workspace_ids]
+                )
```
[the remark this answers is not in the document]
EnjoyBacon7: This is a deliberate tradeoff. The COMPLETED state reflects that the file has been fully indexed and its record committed to the DB — that is the operation the user queued. Workspace association is a secondary metadata step; if it fails, the file is still correctly indexed and searchable in the partition. The task status accurately describes the indexing outcome.

The alternative — setting COMPLETED only after workspace association — would mean a transient DB hiccup causes the user to see a FAILED task for a file that is actually indexed. We log a warning if workspace association fails so operators can detect and remediate it manually.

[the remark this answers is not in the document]
EnjoyBacon7: Fixed in 0a2864b. The `asyncio.gather` call is now wrapped in its own `try/except` block. A transient DB error during workspace association will log a warning but will not propagate to the outer `except` handler, so the task state stays `COMPLETED` rather than being overwritten to `FAILED`.

openrag/components/pipeline.py
```diff
@@ -133,16 +138,35 @@ async def _prepare_for_chat_completion(self, partition: list[str], payload: dict
...
+                task_description=f"get_workspace({workspace})",
+            )
+            if not ws or ws["partition_name"] not in partition:
+                logger.warning(
+                    "Workspace not found in partition(s) — ignoring workspace filter",
+                    workspace=workspace,
+                    partition=partition,
+                )
+                workspace = None
+        filter_dict = {"workspace_id": workspace} if workspace else None
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in 4527b63. The condition is now `"all" not in partition and ws["partition_name"] not in partition`, so when the sentinel `["all"]` is used the workspace filter is kept rather than silently dropped.

openrag/routers/workspaces.py
```diff
@@ -0,0 +1,177 @@
...
+                    task_description=f"delete_file({file_id})",
+                )
+                for file_id in orphaned
+            ],
+            return_exceptions=True,
+        )
+        for file_id, result in zip(orphaned, results):
+            if isinstance(result, Exception):
+                logger.warning("Failed to delete orphaned file from Milvus", file_id=file_id, error=str(result))
+    return {"status": "deleted", "orphaned_files_deleted": len(orphaned)}
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in a78023c. The response now counts only successful deletions in `orphaned_files_deleted` and includes an `orphaned_files_failed` list for any file IDs whose `delete_file` call threw an exception.

docs/content/docs/documentation/workspaces.md
paultranvan: I think it's fair to delete files when deleting a workspace, so the client does not have to bother deleting them individually. Otherwise it could lead to orphans files. However, we could add an API parameter, e.g. `delete_files=<true|false>` to let the client decide whether or not the files should be removed

PR #280, opened 2026-03-12
fix(workspaces): preserve file workspace memberships on PUT and PATCH

### LINKED ISSUE #279: PUT (replace file) permanently removes workspace memberships
## Summary

Replacing a file via `PUT /indexer/partition/{partition}/file/{file_id}` permanently removes all workspace memberships for that file. After the replace completes, the file no longer belongs to any workspace it was previously associated with.

## Root cause

The `PUT` handler (`openrag/routers/indexer.py:269`) calls `await indexer.delete_file.remote(file_id, partition)` before re-indexing the new version. `delete_file` calls `remove_file_from_all_workspaces()` (`vectordb.py:617`), which deletes all `workspace_files` rows for that file:

```python
# vectordb.py:617
self.partition_file_manager.remove_file_from_all_workspaces(file_id, partition)
```

The subsequent `add_file.remote(...)` call at `indexer.py:292` does not pass `workspace_ids`:

```python
task = indexer.add_file.remote(path=file_path, metadata=metadata, partition=partition, user=user)
# ^ no workspace_ids — memberships are never restored
```

## Steps to reproduce

```bash
# 1. Upload a file
curl -X POST http://<host>/indexer/partition/p1/file/my-file \
  -F "file=@doc.txt;type=text/plain" -F 'metadata={}'

# 2. Create a workspace and add the file to it
curl -X POST http://<host>/partition/p1/workspaces \
  -H "Content-Type: application/json" -d '{"workspace_id": "ws1"}'
curl -X POST http://<host>/partition/p1/workspaces/ws1/files \
  -H "Content-Type: application/json" -d '{"file_ids": ["my-file"]}'

# 3. Verify membership
curl http://<host>/partition/p1/workspaces/ws1/files
# {"file_ids": ["my-file"]}  ✓

# 4. Replace the file via PUT
curl -X PUT http://<host>/indexer/partition/p1/file/my-file \
  -F "file=@new_doc.txt;type=text/plain" -F 'metadata={}'

# Wait for indexing to complete, then:
curl http://<host>/partition/p1/workspaces/ws1/files
# {"file_ids": []}  ✗ — membership is gone
```

## Expected behaviour

Replacing a file should preserve all its existing workspace memberships. The file still has the same `file_id`; its workspace associations should survive a content replacement.

## Suggested fix

Before deleting the file, snapshot its workspace memberships, then restore them after the new version is indexed:

```python
# In the PUT handler (indexer.py), before delete:
current_workspaces = await call_ray_actor_with_timeout(
    vectordb.get_file_workspaces.remote(file_id, partition), ...
)
await indexer.delete_file.remote(file_id, partition)
# ... re-index ...
# After indexing completes (task-level), restore memberships
if current_workspaces:
    await vectordb.add_files_to_workspace.remote(ws_id, [file_id]) for ws_id in current_workspaces
```

Alternatively, `remove_file_from_all_workspaces` could be skipped in the delete-then-replace flow and instead only called on true file deletions.

## Affected files

- `openrag/routers/indexer.py:269` — `put_file()` handler
- `openrag/components/indexer/vectordb/vectordb.py:617` — `delete_file()` unconditionally removes workspace memberships

EnjoyBacon7:
## Summary

Fixes #279.

`PUT /indexer/partition/{partition}/file/{file_id}` and `PATCH /indexer/partition/{partition}/file/{file_id}` both use a delete-then-reinsert pattern internally. The delete step called `remove_file_from_all_workspaces()`, permanently wiping all `workspace_files` rows for the file even though the `file_id` never changed. After the operation the file existed in the vector DB but had lost all its workspace associations.

## Fix

Snapshot the file's current workspace memberships **before** the delete step, then restore them **after** re-indexing completes.

### Changes

- **`vectordb/utils.py`** — add `PartitionFileManager.get_file_workspaces(file_id, partition) -> list[str]` that queries `workspace_files JOIN workspaces` filtered by partition
- **`vectordb/vectordb.py`** — expose it as `async MilvusDB.get_file_workspaces(...)` (Ray-callable)
- **`indexer/indexer.py`** — `update_file_metadata()` snapshots workspace IDs before `delete_file`, then restores them via `asyncio.gather` after `async_add_documents` completes
- **`routers/indexer.py`** — `put_file()` snapshots workspace IDs before `delete_file.remote`, then passes them as `workspace_ids` to `add_file.remote(...)` (which already handles re-association after indexing)

## Testing

All 205 unit tests pass (`uv run pytest --ignore=openrag/test_token_validation.py`). The skipped `test_token_validation.py` failure is a pre-existing Ray session lock issue unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added an endpoint to list which workspaces a file belongs to within a partition.

* **Bug Fixes**
  * Files now preserve their workspace associations when updated or re-indexed, preventing loss of memberships during file replacement or metadata updates.
  * File removal now correctly clears all workspace-to-file associations within the relevant scope.

### DIFF
--- a/openrag/components/indexer/indexer.py
+++ b/openrag/components/indexer/indexer.py
@@ -195,8 +195,18 @@ class Indexer:
             for doc in docs:
                 doc.metadata.update(metadata)
 
+            # Snapshot workspace memberships before deletion so they can be restored.
+            workspace_ids = await vectordb.get_file_workspaces.remote(file_id, partition)
+
             await self.delete_file(file_id, partition)
             await vectordb.async_add_documents.remote(docs, user=user)
+
+            # Restore workspace memberships that existed before the delete.
+            if workspace_ids:
+                await asyncio.gather(
+                    *[vectordb.add_files_to_workspace.remote(ws_id, [file_id]) for ws_id in workspace_ids]
+                )
+                log.debug("Restored workspace memberships after metadata update.", workspace_ids=workspace_ids)
 
             log.info("Metadata updated for file.")
         except Exception as e:
--- a/openrag/components/indexer/vectordb/utils.py
+++ b/openrag/components/indexer/vectordb/utils.py
@@ -675,6 +675,18 @@ class PartitionFileManager:
             result = session.execute(select(WorkspaceFile.file_id).where(WorkspaceFile.workspace_id == workspace_id))
             return [r[0] for r in result.all()]
 
+    def get_file_workspaces(self, file_id: str, partition: str) -> list[str]:
+        """Return the workspace IDs that contain the given file, scoped to the given partition."""
+        with self.Session() as session:
+            ws_ids = select(Workspace.workspace_id).where(Workspace.partition_name == partition)
+            result = session.execute(
+                select(WorkspaceFile.workspace_id).where(
+                    WorkspaceFile.file_id == file_id,
+                    WorkspaceFile.workspace_id.in_(ws_ids),
+                )
+            )
+            return [r[0] for r in result.all()]
+
     def remove_file_from_all_workspaces(self, file_id: str, partition: str):
         """Remove file from all workspaces in the given partition — called during file deletion."""
         with self.Session() as session:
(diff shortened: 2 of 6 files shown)


paultranvan:
The fix sounds good for the current behaviour, but why do we delete/create the file entry in postgres on PUT/PATCH? Why not a SQL update?

Ahmath-Gadji (commented):
LGTM. I've added a test for workspace membership preservation after put or patch

paultranvan:
I suggest we do not delay the bug fix and merge this PR, but treat this issue afterwards: https://github.com/linagora/openrag/issues/286

PR #283, opened 2026-03-12
fix(workspaces): validate workspace_id rejects URL-unsafe characters

### LINKED ISSUE #276: workspace_id accepts URL-unsafe characters (slash, empty string) making workspaces inaccessible
## Summary

There is no validation on `workspace_id` when creating a workspace. Characters such as `/` are accepted and stored, but the resulting workspace is permanently inaccessible via the REST API because the slash is interpreted as a URL path separator.

## Root cause

`CreateWorkspaceRequest` in `openrag/routers/workspaces.py:24` declares `workspace_id: str` with no constraints. FastAPI/Pydantic accept any string, including empty strings and strings containing slashes.

```python
# workspaces.py:24 — no validation on workspace_id
class CreateWorkspaceRequest(BaseModel):
    workspace_id: str
    display_name: str | None = None
```

## Steps to reproduce

```bash
# Create a workspace with a slash in the ID
curl -X POST http://<host>/partition/default/workspaces \
  -H "Content-Type: application/json" \
  -d '{"workspace_id": "ws/slash"}'
# Returns 201 Created

# Try to access it — 404 because the router parses the slash as a path separator
curl http://<host>/partition/default/workspaces/ws/slash
# Returns 404 Not Found (route not matched)
```

An empty string `""` is also accepted, producing a workspace that cannot be referenced in any subsequent API call.

## Expected behaviour

`workspace_id` should be validated to contain only URL-safe characters (e.g. alphanumerics, hyphens, underscores). Invalid values should return `422 Unprocessable Entity`.

## Suggested fix

Add a Pydantic validator using a regex pattern:

```python
from pydantic import field_validator
import re

class CreateWorkspaceRequest(BaseModel):
    workspace_id: str
    display_name: str | None = None

    @field_validator("workspace_id")
    @classmethod
    def validate_workspace_id(cls, v: str) -> str:
        if not v or not re.fullmatch(r"[a-zA-Z0-9_-]+", v):
            raise ValueError("workspace_id must be non-empty and contain only alphanumeric characters, hyphens, or underscores")
        return v
```

## Affected file

`openrag/routers/workspaces.py`, line 24

EnjoyBacon7:
## Summary

Fixes #276.

`CreateWorkspaceRequest` accepted any string for `workspace_id`, including slashes, spaces, and empty strings. A workspace created with a slash in its ID was permanently inaccessible because the slash is interpreted as a URL path separator by the router, making every subsequent API call return 404.

## Change

**`routers/workspaces.py`** — add a Pydantic `field_validator` on `workspace_id` that enforces it must be non-empty and match `[a-zA-Z0-9_-]+`. Invalid values cause Pydantic to return a standard `422 Unprocessable Entity` response before the request reaches any handler.

`display_name` is unchanged — it can contain any characters since it is never used in URL paths.

## Tests

17 new parametrized unit tests in `routers/test_workspace_id_validation.py` covering:
- Valid IDs: alphanumeric, hyphens, underscores, mixed case
- Invalid IDs: slash, space, dot, at-sign, hash, empty string, multi-slash path, newline
- `display_name` remains unrestricted

222 passed, 3 skipped (pre-existing)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for workspace ID validation rules

* **New Features**
  * Workspace ID validation now enforces alphanumeric, hyphen, and underscore characters only; empty workspace IDs are rejected


openrag/routers/workspaces.py
```diff
@@ -17,13 +18,24 @@
 _config = load_config()
 VECTORDB_TIMEOUT = _config.ray.indexer.get("vectordb_timeout", 30)
 
+_WORKSPACE_ID_RE = re.compile(r"[a-zA-Z0-9_-]+")
```
paultranvan, line 21: Why the `_` prefix? Is it an existing convention?

openrag/routers/test_workspace.py
```diff
@@ -0,0 +1,86 @@
+"""Unit tests for workspace_id validation logic.
```
paultranvan, line 1: Could you rename the file into test_workspace.py ? so we have 1 single file for future workspace tests

paultranvan (approved):
Thanks!

openrag/routers/workspaces.py
EnjoyBacon7: Yes, this link about [private names in modules](https://livebook.manning.com/wiki/categories/python/private) explains the convention.

openrag/routers/test_workspace.py
EnjoyBacon7: Done, renamed to `test_workspace.py`.

PR #284, opened 2026-03-13
Step 3/3 Feat/temporal filters in rag pipeline

Ahmath-Gadji:
### Summary

Adds LLM-driven temporal filters to retrieval so users can naturally ask time-bounded questions ("reports from last month", "meeting notes this week") and have them translated into Milvus `created_at` filters before vector search.

### Changes

- **New structured query models** in `openrag/components/pipeline.py`: `TemporalPredicate` (field/op/value on `created_at`) and `Query` (text + AND-combined predicates). `SearchQueries` now returns a list of `Query` objects instead of raw strings.
- **`RetrieverPipeline.retrieve_docs`** takes a `Query` and derives the Milvus filter via `Query.to_milvus_filter()` — no schema change on the collection.
- **Filterless fallback**: when a temporally-filtered search returns zero docs, retry without the filter. Gated by `retriever.allow_filterless_fallback` (env `RETRIEVER_ALLOW_FILTERLESS_FALLBACK`, default `true`) so strict deployments can opt out.
- **Schema-validation retry**: query generation retries once on `ValidationError` / `OutputParserException`, then falls back to the raw user query. `with_structured_output(..., strict=True)`.
- **Prompt template** (`prompts/example1/query_contextualizer_tmpl.txt`, v1) adds query-decomposition rules, temporal-filter rules with explicit lower/upper-bound semantics, and richer `{current_date}` (weekday + time) to disambiguate relative expressions. Key distinction: ingestion date → filter; document topic date → no filter.

### Benchmarks

Offline harness under `benchmarks/prompt_eval/` comparing prompt **v0** (pre-change) vs **v1** (this PR) on the two production-candidate models. Judge: Qwen3-VL-8B-Instruct-FP8. Full results in `benchmarks/prompt_eval/Report_temporal_filter.md` and `benchmarks/prompt_eval/Report_query_decomposition.md`.

**Temporal filter detection** (40 cases, 20 pos / 20 neg):

| Prompt | Model | Accuracy | F1 | Filter correctness (TP) |
|---|---|---:|---:|---:|
| v0 | Mistral-Small-3.1-24B | 75.0% | 66.7% | 8/10 |
| **v1** | **Mistral-Small-3.1-24B** | **100.0%** | **100.0%** | **18/20 (90.0%)** |
| v1 | Qwen3-VL-8B | 94.9% | 95.0% | 17/19 (89.5%) |

**Query decomposition** (80 cases across D1/D2/D3):

| Prompt | Model | count_match | semantic_coverage |
|---|---|---:|---:|
| v0 | Mistral-Small-3.1-24B | 82.5% | 92.5% |
| **v1** | **Mistral-Small-3.1-24B** | **86.2%** | **95.0%** |
| v1 | Qwen3-VL-8B | 82.5% | 88.8% |

Recommended pairing: **v1 + Mistral-Small-3.1-24B** — 100% filter detection, 90% filter-body correctness, and best decomposition numbers. Residual failures are narrow (e.g. spurious upper bound on "since last Monday", over-wide window on "latest") and documented in the reports.

### Backwards compatibility

No collection/schema change. Queries with no temporal reference produce `temporal_filters=None` and hit the same retrieval path as before.

## Summary by CodeRabbit

* **New Features**
  * Temporal (date-based) filter support for generated search queries.
  * Retrieval fallback: retries without temporal filters when filtered results return none.
  * Prompt templates now enforce structured JSON query outputs with explicit temporal predicates.

* **Documentation**
  * New benchmark reports evaluating query decomposition and temporal-filter generation with per-tier, per-prompt analyses and example outputs.

* **Chores**
  * Added benchmark datasets, evaluation tooling, and project manifest to support prompt evaluation.


prompts/example1/query_contextualizer_tmpl.txt
```diff
@@ -33,4 +33,4 @@ From a chat history, reformulate the user's last message into one or more autono
 - User: I'm planning a trip to Italy and I'm interested in historical monuments and local cuisine.
 - Assistant: Italy offers a wealth of history and culinary delights.
 - User: What are the must-see sites?
-Reformulated query: What are the must-see historical monuments and local cuisine restaurants in Italy?
+Reformulated query: What are the must-see historical monuments and local cuisine restaurants in Italy?
```
paultranvan: Did we evaluate this prompt ? How it performs with real-case examples? How it impact the latency ?

utility/data_indexer2.py
```diff
@@ -0,0 +1,121 @@
...
+    logger.info(
+        f"Size (MB) -> min: {min(file_sizes_mb):.2f}, max: {max(file_sizes_mb):.2f}, "
+        f"mean: {statistics.mean(file_sizes_mb):.2f}, std: {statistics.stdev(file_sizes_mb) if len(file_sizes_mb) > 1 else 0:.2f}"
+    )
+else:
+    logger.info("No files were uploaded.")
+
+
+# # How to run this code:
+# uv run python utility/data_indexer2.py \
```
paultranvan: please avoid non numbered naming such as indexer2 🙏 
if we need several data_indexer script, let's name them explicitly with their purpose

Ahmath-Gadji: This was only for my local use. Not meant to be pushed on github.

benchmarks/prompt_eval/.env.example
```diff
@@ -0,0 +1,4 @@
+
+BASE_URLs=https://example/v1/;https://example2/v1/;
+API_KEYs=sk-api_key1;sk-api_key2;
+MODELs=model1;model2;
```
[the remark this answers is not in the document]
paultranvan: I agree with the bot

openrag/components/pipeline.py
```diff
@@ -185,12 +186,20 @@ def __init__(self) -> None:
...
             temperature=config.llm.temperature,
-        ).with_structured_output(SearchQueries, method="function_calling", strict=True)
+        )
+
+        primary = llm.with_structured_output(SearchQueries, method="json_schema", strict=True)
+        fallback = llm.with_structured_output(SearchQueries, method="function_calling", strict=False)
+        self.query_generator = primary.with_fallbacks(
+            [fallback],
+            exceptions_to_handle=(openai.BadRequestError,),
+        )
```
paultranvan, line 220: This commit can be droped since it already exist in https://github.com/linagora/openrag/pull/328
PR must be rebased

paultranvan: and please, no more merge commits 🥲 
let's always rebase

openrag/components/pipeline.py
```diff
@@ -34,19 +37,62 @@ class RAGMODE(Enum):
...
+    query: str = Field(description="A semantically enriched, descriptive query for vector similarity search.")
+    temporal_filters: list[TemporalPredicate] | None = Field(
+        default=None,
+        description="Date predicates on `created_at`, AND-combined. Null when no temporal reference in the query.",
+    )
+
+    def to_milvus_filter(self) -> str | None:
+        if not self.temporal_filters:
+            return None
+        parts = [f'{p.field} {p.operator} ISO "{p.value}"' for p in self.temporal_filters]
```
paultranvan: Do we control the p.value date at some point ? If not, I think this is quite risky. 
We should control that each field has a correct format, otherwise, drop it.

openrag/config/models.py
```diff
@@ -419,6 +419,7 @@ class _BaseRetrieverConfig(ConfigMixin):
     include_ancestors: bool = True
     related_limit: int = 10
     max_ancestor_depth: int = 10
+    allow_filterless_fallback: bool = True
```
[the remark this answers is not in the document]
Ahmath-Gadji: It's fine to default it to true. On instances where temporal filterig is a must, we will deploy openrag disallowing this fallback.

openrag/components/pipeline.py
```diff
@@ -34,19 +37,62 @@ class RAGMODE(Enum):
...
+        return " and ".join(parts)
```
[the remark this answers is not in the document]
paultranvan: Sounds worth checking

openrag/components/pipeline.py
Ahmath-Gadji: I agree.

openrag/components/pipeline.py
Ahmath-Gadji: I'm gonna. It's true that we wanted to check the generated date values.

openrag/components/pipeline.py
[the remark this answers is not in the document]
Ahmath-Gadji: > Sounds worth checking

@paultranvan  A rare instance of False positive from Coderabbit. Milvus 2.6.x+ supports that syntax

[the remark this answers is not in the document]
paultranvan: Thank you, non-hallucinating human :)

openrag/components/pipeline.py
Ahmath-Gadji: Done

tests/api_tests/api_run/mock_vllm.py
```diff
@@ -224,6 +224,46 @@ async def stream_chat_completion(request: ChatCompletionRequest):
...
+    if t == "array":
+        items = _resolve_ref(schema.get("items", {}), root)
+        return [_mock_value(items, root, user_text)]
+    if t == "string":
+        return user_text
+    if t in ("integer", "number"):
+        return 0
+    if t == "boolean":
+        return False
+    return None
```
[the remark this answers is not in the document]
Ahmath-Gadji: The tests are not breaking. We will adapt it when needed

openrag/components/pipeline.py
```diff
@@ -34,19 +37,76 @@ class RAGMODE(Enum):
...
+        return f"Query: {self.query}, Filter: {self.to_milvus_filter()}"
+
 
-    query_list: list[str] = Field(..., description="Search sub-queries to retrieve relevant documents.")
+class SearchQueries(BaseModel):
+    query_list: list[Query] = Field(..., description="Search sub-queries to retrieve relevant documents.")
 
     def __str__(self) -> str:
-        return " -- ".join(f"Query: {q}" for q in self.query_list)
+        return " --- ".join(str(q) for q in self.query_list)
```
paultranvan, line 106: I'm not sure why we need this `---` separator?

openrag/components/pipeline.py
```diff
@@ -34,19 +37,76 @@ class RAGMODE(Enum):
...
+                datetime.fromisoformat(p.value)
+            except (TypeError, ValueError):
+                logger.warning(
+                    "Dropping temporal predicate with non-ISO value",
+                    field=p.field,
+                    operator=p.operator,
+                    value=p.value,
+                )
+                continue
+            parts.append(f'{p.field} {p.operator} ISO "{p.value}"')
```
paultranvan, line 93: I think we should check field and operator as well, with a list of authorized values

openrag/components/pipeline.py
Ahmath-Gadji: This is just a separator for when we log the generated subqueries.

Ahmath-Gadji: Here is when where we log them.
https://github.com/linagora/openrag/blob/6875dc417368c5f8aab4771b6283cdd7894ab60a/openrag/components/pipeline.py#L279-L281

paultranvan: Ok, I'm not sure I like doing this overload in `_str_` which seem dangerous to me, as we could have other situation than logging, but let's keep it for now

openrag/components/pipeline.py
Ahmath-Gadji: For `field` and `operator` they are valided. In fact Pydantic enforces the provided type hints (Literal)

paultranvan: Ah yes, you're right thank you! Maybe a small comment in the function would help to add this precision? Like `The temporal_filters attributes are already checked through the Pydantic types, except for date value that is kept as string, as LLM sometimes give correct but not entirely complete date`

paultranvan (approved):
Thank you for this great and long-expected work!

openrag/components/pipeline.py
Ahmath-Gadji: The field that needs validation is the date, since a string is not necessarily a valid ISO datetime. During our discussion, I considered using the `datetime` type instead of `str` for the `value`.

```diff
from datetime import datetime

class TemporalPredicate(BaseModel):
    """A single constraint on a document's creation date.

    Multiple predicates on the same `Query` are combined with logical AND.
    Use two predicates to express a closed range (e.g. last month):
      [{op: ">=", value: "2026-03-01..."}, {op: "<=", value: "2026-03-31..."}]
    """

    field: Literal["created_at"] = Field(
        default="created_at",
        description="Document metadata field to filter on. Always `created_at` for now.",
    )
    operator: Literal[">", "<", ">=", "<="] = Field(
        description="Comparison operator applied to the date field.",
    )
-    value: str = Field(description='ISO 8601 datetime with timezone, e.g. "2026-03-15T00:00:00+00:00".')
+    value: datetime = Field(description='ISO 8601 datetime with timezone, e.g. "2026-03-15T00:00:00+00:00".')
```

The reason for not using it is that Mistral, in my earlier tests, tended to omit the `+HH:MM:SS` timezone part of the datetime. However, this might be due to task difficulty or unclear prompting. With the newer, simplified version of the prompt, I have observed fewer filtering errors of this kind. Using `datetime` would remove the need for manual validation.

That said, I prefer sticking with `str`, because the generated value might still be correct but not a fully specified datetime. In such cases, we could convert it to a `datetime` afterward.

Ahmath-Gadji: > Ah yes, you're right thank you! Maybe a small comment in the function would help to add this precision? Like `The temporal_filters attributes are already checked through the Pydantic types, except for date value that is kept as string, as LLM sometimes give correct but not entirely complete date`

added

PR #291, opened 2026-03-19
fix(workspaces): eliminate TOCTOU race in add_files_to_workspace

EnjoyBacon7:
## Summary
- `add_files_to_workspace` now bulk-resolves all file_ids → `File.id` in a single query and returns any unresolved IDs, instead of silently skipping them
- The router checks the return value and raises 404 if any files vanished between the pre-check and the insert, closing the TOCTOU race window

Addresses CodeRabbit review comment on PR #281 (comment `2959737769`).


Ahmath-Gadji (approved):
LGTM

PR #313, opened 2026-04-19
fix(auth): detect HTTPS behind TLS-terminating reverse proxy

etiquet:
## Summary

When OpenRAG runs behind a TLS-terminating reverse proxy (Traefik, Nginx, etc.), the OIDC state cookie is set without the `Secure` flag because `request.url.scheme` is `http` (the proxy handles TLS). This causes the browser to not send the cookie back on the HTTPS callback, breaking the OIDC Authorization Code + PKCE flow with:

```
OIDC code exchange failed
```

## Root cause

`_is_request_secure()` only checks `request.url.scheme`, which reflects the internal HTTP connection from the proxy, not the client-facing HTTPS.

## Fix

`_is_request_secure()` now checks three indicators:
1. `PREFERRED_URL_SCHEME=https` env var (explicit config, already used elsewhere in OpenRAG)
2. `X-Forwarded-Proto: https` header (standard reverse proxy header)
3. `request.url.scheme` (existing fallback)

## Test plan

- [x] Tested on Scaleway VM behind Traefik with Let's Encrypt
- [x] OIDC cookie now has `Secure` flag when accessed via HTTPS
- [x] OIDC login flow completes successfully (code exchange + session creation)
- [x] Direct HTTP access still works (Secure=false, correct behavior)


EnjoyBacon7:
Hello @etiquet,

Before putting more time into this PR, would you mind clearing up some qualms I have with the scope of your modifications and the relevance of certain changes?

1. Relevance
- The prompt file [prompts_integration.md](https://github.com/IA-Generative/openrag/blob/ad1ef94b4b80939d609e0ef94e0b8aa11b3174ba/prompts_integration.md)

What is the point of providing this file in the PR? With the new age of AI, there may be some use for this file I don't grasp.

3.  Scope issues
- The [drive connector](https://github.com/IA-Generative/openrag/blob/ad1ef94b4b80939d609e0ef94e0b8aa11b3174ba/openrag/components/connectors/drive.py)
- The notifications components

Unless these changes are core to the auth fix in some way, please resubmit them in other PRs.

PR #328, opened 2026-04-21
feat(rag): use json_schema + function_calling for reliable structured…

Ahmath-Gadji:
# Robust structured output with json_schema + function_calling fallback
 
## Context
 
The query generator uses `with_structured_output` to extract a typed `SearchQueries` object from the LLM response. The reliability of this depends heavily on **which structured output method is used** and **which model is configured**.
 
The pipeline is designed to be model-agnostic — any OpenAI-compatible LLM can be plugged in via configuration. This means we can't assume a specific model's capabilities: some models support strict schema enforcement via `json_schema`, others only support `function_calling`, and some may support neither reliably. The fallback chain handles this transparently without requiring pipeline changes when the model changes.
 
## The three methods compared
 
| Method | How it works | Schema enforced? | Provider support |
|---|---|---|---|
| `json_schema` | Passes JSON schema via `response_format`, uses constrained decoding at token level | ✅ Strictly | Newer models / vLLM |
| `function_calling` | Wraps schema as a fake tool definition, model "calls" it | Loosely | Wide — most models |
| `json_mode` | Instructs model to return valid JSON, no schema passed | ❌ No | Medium |
 
**`json_schema`** is the preferred method on vLLM because structured output is enforced at generation time via token masking — the model is physically prevented from producing output that doesn't match the schema, regardless of fine-tuning. This is more reliable than `function_calling`, which depends on the model having been trained to respect tool call conventions.
 
**`function_calling`** is a widely compatible fallback. It works across most models but schema adherence is softer — failures typically surface as a `null` output or a `BadRequestError` when the model doesn't support the tool calling API at all.
 
## Why `function_calling` works as a structured output mechanism
 
This is a non-obvious design choice worth explaining. `function_calling` was never intended for structured data extraction — it was designed to let models trigger external tools or APIs. However, it became the de-facto structured output hack before native `json_schema` support existed, and LangChain still uses it today as a broad-compatibility fallback.
 
Here is what happens under the hood when you use `method="function_calling"` in LangChain:
 
1. **Your Pydantic schema is converted into a fake tool definition.** LangChain takes your `SearchQueries` class and wraps it as if it were a callable function, with the schema fields mapped to function parameters.
2. **The tool call is forced.** LangChain sets `tool_choice={"type": "function", "name": "SearchQueries"}`, which instructs the model that it *must* call this specific tool — no free-text response is allowed.
3. **The response is intercepted, never executed.** The model returns a tool call payload with the structured arguments. LangChain extracts those arguments and deserializes them back into your Pydantic object. No actual function is ever called.
```python
# What you write
llm.with_structured_output(SearchQueries, method="function_calling")
 
# What LangChain sends to the model
tools = [{
    "type": "function",
    "function": {
        "name": "SearchQueries",        # your class name
        "parameters": { ...schema... }  # your Pydantic fields
    }
}]
tool_choice = {"type": "function", "name": "SearchQueries"}  # forced
 
# The model "calls" the tool → LangChain extracts the arguments → SearchQueries instance
```
 
The reason this works as a fallback is that tool calling is a capability that has been fine-tuned into a wider range of models than native structured output. A model that doesn't understand `response_format: json_schema` will very likely still understand tool call conventions, making `function_calling` a reliable second line of defense.
 
The tradeoff is that schema adherence is softer — the model is guided by its fine-tuning rather than constrained by token masking, so malformed outputs are more likely under edge cases.
 
## Why `json_mode` was not used
 
`json_mode` is the most basic structured output option. It sets `response_format={"type": "json_object"}`, which tells the model to return valid JSON — but nothing more. The schema is **not passed to the model API at all**.
 
This has a critical implication: **you must inject the schema manually into the prompt**, otherwise the model has no idea what fields to produce:
 
```python
# With json_mode, LangChain does NOT pass the schema to the API.
# You are responsible for describing the expected structure in the prompt.
prompt = ChatPromptTemplate.from_template("""
Extract search queries from the user input.
Respond ONLY with a JSON object matching this structure:
{{
  "queries": ["query1", "query2", ...]
}}
 
User input: {input}
""")
 
structured_llm = llm.with_structured_output(SearchQueries, method="json_mode")
chain = prompt | structured_llm
```
 
Without the schema in the prompt, the model will return arbitrary valid JSON — correct syntax, wrong shape. This makes `json_mode` fragile by design: the contract between the prompt and the expected output is implicit and easy to break when either side changes.
 
The other issue is that `json_mode` only guarantees parseable JSON, not schema compliance. Even with a well-crafted prompt, the model can omit fields, use wrong types, or add unexpected keys — and no validation error will be raised at the API level.
 
For these reasons, `json_mode` was ruled out in favor of `json_schema` (enforced at token level) with `function_calling` as a fallback (enforced via fine-tuning). Both methods pass the schema to the model through the API and do not rely on prompt engineering for structural correctness.
 
## Implementation
 
```python
primary = llm.with_structured_output(SearchQueries, method="json_schema")
fallback = llm.with_structured_output(SearchQueries, method="function_calling", fallback_to_raw=True)
 
self.query_generator = primary.with_fallbacks(
    [fallback],
    exceptions_to_handle=(openai.BadRequestError,),
)
```
 
The chain tries `json_schema` first. If the model or vLLM version doesn't support it and raises a `BadRequestError`, LangChain automatically retries with `function_calling`. `fallback_to_raw=True` ensures that if even the fallback fails to parse, the raw response is returned rather than raising, giving the caller a chance to handle degraded output gracefully.
 
## Why this matters
 
Since any OpenAI-compatible model can be swapped in via config, structured output support varies. Using a fallback chain makes the query generator resilient to:
 
- Models that don't support `response_format: json_schema` (e.g. older or smaller models)
- Models not fine-tuned for tool calling (which would cause `function_calling` to fail)
- Different inference backends with varying levels of OpenAI API compatibility (vLLM, Ollama, LiteLLM, etc.)
## References
 
- [Mastering Structured Output in LLMs — Revisiting LangChain](https://medium.com/@docherty/mastering-structured-output-in-llms-revisiting-langchain-and-json-structured-outputs-d95dfc286045)
- [vLLM Structured Outputs docs](https://docs.vllm.ai/en/latest/features/structured_outputs/)
 

## Summary by CodeRabbit

* **Improvements**
  * Enhanced query generation resilience with a fallback mechanism that handles structured output generation more gracefully, reducing failures during query processing.

### DIFF
--- a/openrag/components/pipeline.py
+++ b/openrag/components/pipeline.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Literal
 
+import openai
 import ray
 from components.prompts import (
     QUERY_CONTEXTUALIZER_PROMPT,
@@ -185,12 +186,20 @@ class RagPipeline:
         self.max_context_tokens = config.reranker.top_k * config.chunker.chunk_size
 
         self.llm_client = LLM(config.llm, logger)
-        self.query_generator = ChatOpenAI(
+
+        llm = ChatOpenAI(
             base_url=config.llm.base_url,
             api_key=config.llm.api_key,
             model=config.llm.model,
             temperature=config.llm.temperature,
-        ).with_structured_output(SearchQueries, method="function_calling", strict=True)
+        )
+
+        primary = llm.with_structured_output(SearchQueries, method="json_schema", strict=True)
+        fallback = llm.with_structured_output(SearchQueries, method="function_calling", strict=False)
+        self.query_generator = primary.with_fallbacks(
+            [fallback],
+            exceptions_to_handle=(openai.BadRequestError,),
+        )
 
         self.max_contextualized_query_len = config.rag.max_contextualized_query_len
 


openrag/components/pipeline.py
```diff
@@ -185,12 +186,20 @@ def __init__(self) -> None:
...
+        llm = ChatOpenAI(
             base_url=config.llm.base_url,
             api_key=config.llm.api_key,
             model=config.llm.model,
             temperature=config.llm.temperature,
-        ).with_structured_output(SearchQueries, method="function_calling", strict=True)
+        )
+
+        primary = llm.with_structured_output(SearchQueries, method="json_schema")
+        fallback = llm.with_structured_output(SearchQueries, method="function_calling")
```
paultranvan: no need for `strict=True` like before ?

paultranvan (approved):

paultranvan:
It's great if can rely more on LLM JSON answers, thanks!
Did you test it with the temporal PR ?

paultranvan:
And from what I understood, json_schema will actually leverage vLLM structured_output when available  https://docs.vllm.ai/en/v0.8.2/features/structured_outputs.html , which is good as it is at the logit level

Ahmath-Gadji:
> And from what I understood, json_schema will actually leverage vLLM structured_output when available https://docs.vllm.ai/en/v0.8.2/features/structured_outputs.html , which is good as it is at the logit level

Exactly. On logit level it's less error-prone.

openrag/components/pipeline.py
Ahmath-Gadji: https://docs.langchain.com/oss/python/langchain/structured-output#param-strict

Re-added. Some providers like Openai and Grok can use it.

Ahmath-Gadji:
> It's great if can rely more on LLM JSON answers, thanks! Did you test it with the temporal PR ?

Yes. It works fine.

PR #340, opened 2026-04-24
fix: strip every line-terminal [Sources: N] tag from LLM output

paultranvan:
Avoid having `[Sources: ...]` tag inlined in LLM response

Note it introduces a slight risk of false positive: if the LLM generates a `[Sources: X]` at the end of a sentence in a legitimate way (i.e. not actually citing sources), it will be stripped.

Capture d’écran du 2026-04-24 11-44-27

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of line-terminal citation tags: all terminal tags are stripped appropriately while preserving inline citation markers and accurately merging remaining citations.

* **Chores**
  * Simplified streaming API by removing an unnecessary parameter and refined streaming behavior to prevent in-flight tag leakage and ensure correct final citation emission.
  * Minor initialization tweak for command setup.

* **Tests**
  * Added comprehensive tests covering citation parsing and filtering in both non-streamed and streamed responses.

### DIFF
--- a/openrag/components/utils.py
+++ b/openrag/components/utils.py
@@ -3,7 +3,6 @@ import copy
 import json
 import re
 import threading
-from collections import deque
 from typing import ClassVar
 
 import ray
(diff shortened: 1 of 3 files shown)


openrag/components/utils.py
```diff
@@ -214,91 +225,105 @@ def filter_sources_by_citations(sources: list, citations: set[int] | None) -> li
...
+                choice = data["choices"][0]
+                out = {
+                    **data,
+                    "choices": [
+                        {**choice, "delta": {**choice.get("delta", {}), "content": cleaned[emitted_len:safe_end]}}
+                    ],
+                    "extra": "{}",
+                }
+                yield f"data: {json.dumps(out)}\n\n"
+                emitted_len = safe_end
```
Ahmath-Gadji, line 322: The mechanic here is to accumulate content up to a limit, strip the source tags, and finally stream it.

The streaming here streams the accumulated text with some metadata, in contrast to the previous version where all the metadata was preserved and streaming was done token by token.

- I don't think we have much choice other than doing it this way, as we're trying to get rid of some artifacts mid-streaming. The backend LLM is the problem since it's not following the prompt, so now we have to handle it on our side.
- We're no longer streaming token-by-token, so in terms of OpenAI compatibility we're regressing. For the concatenated streamed text, we can't accurately reconstruct the metadata. For evaluation purposes, some token-level metadata could be interesting — e.g. logprobs, to probe the LLM's certainty in its response — but again, when doing this cleaning, we can't fabricate those for the newly minted text.

I think we can merge this, and in the future we can consider switching back to the old version if we deploy a better LLM.

Ahmath-Gadji: Should we preserve the old version of utils.py in a file like utils_legacy.py ?

Ahmath-Gadji (commented):
LGTM.

Ahmath-Gadji (approved):

openrag/components/utils.py
paultranvan: > Should we preserve the old version of utils.py in a file like utils_legacy.py ?

git history should be enough

PR #341, opened 2026-04-25
feat(auth): optional auto-provisioning of OIDC users on first login

etiquet:
## Why

OpenRAG's OIDC flow today rejects unknown `sub` values with a hard `403 "User not registered"`. That works well for tightly controlled deployments but is painful at scale: every corporate user has to be manually pre-provisioned via `POST /users/` before they can use the platform, even when the IdP itself already enforces who can issue valid tokens.

This PR adds an opt-in flag that lets the callback create a non-admin user on the fly from the ID-token claims — same idea as the existing `OIDC_AUTO_PROVISION` in the bearer-JWT validator (`auth/oidc.py`), now mirrored in the interactive login flow.

**Default is unchanged.** `OIDC_AUTO_PROVISION_LOGIN` defaults to `false` → no behavioral regression for any existing deployment.

## What

When `OIDC_AUTO_PROVISION_LOGIN=true` and the callback receives an unknown `sub`:

| Field | Source |
|---|---|
| `external_user_id` | `sub` (always) |
| `display_name` | `name` → `preferred_username` → `<given_name> <family_name>` → `oidc-<8-char-sub-prefix>` |
| `email` | `email` claim if present, else `null` |
| `is_admin` | **always `false`** — operators must promote afterwards |
| `file_quota` | inherited from server default (same as `POST /users/`) |

A best-effort recovery handles concurrent first-login races (catches the create exception, retries `get_user_by_external_id`; surfaces 500 only if both still fail). An `INFO` log line is emitted on each auto-provisioning so audits can correlate IdP-driven account creations.

## Threat model

The flag is **opt-in** because it shifts the trust boundary: deployments that enable it must be comfortable trusting their IdP's user list as the source of truth for OpenRAG accounts. Operators who need a stricter "manual whitelist" workflow keep the existing 403 behavior by leaving the flag unset.

`is_admin` is hard-coded to `false` for auto-provisioned users — there is no path to grant admin via OIDC claims. Promotion still goes through the existing admin endpoints.

## Tests

- `test_callback_user_not_registered` (existing) — extended to also assert `create_user` is **not** called when the flag is unset.
- `test_callback_auto_provisions_user_when_enabled` — 302 + cookie + `create_user` called with the IdP claims and `is_admin=False`.
- `test_callback_auto_provision_falls_back_to_sub_when_no_name` — `oidc-<sub-prefix>` fallback when no readable name claim.
- `test_callback_auto_provision_disabled_by_default` — explicit `=false` keeps the historical 403.

`_StubVectorDB` grows a `create_user` impl matching the real Ray actor signature (`UserCreate` Pydantic in / dict out).

## Docs

- `.env.example` documents the flag with the exact semantics.
- `CLAUDE.md` updates the OIDC env-var table and the *User Matching & Provisioning* section to describe both flows.

## Files

\`\`\`
.env.example                        |   7 +++
CLAUDE.md                           |   5 +-
openrag/routers/auth.py             |  82 +++++++++++++++++++++++++---
openrag/routers/test_auth_router.py | 105 +++++++++++++++++++++++++++++++++++-
4 files changed, 191 insertions(+), 8 deletions(-)
\`\`\`

## Out of scope (could be follow-ups)

- Group/role sync from claims (already handled by `auth/oidc.py` for bearer JWT — could be unified later).
- Auto-provisioning hooks/callbacks (e.g. emit a webhook when a user is auto-created so external systems can react).

### DIFF
--- a/openrag/routers/auth.py
+++ b/openrag/routers/auth.py
@@ -31,6 +31,7 @@ from components.auth import (
 )
 from fastapi import APIRouter, Form, HTTPException, Request, Response, status
 from fastapi.responses import JSONResponse, RedirectResponse
+from models.user import UserCreate
 from utils.dependencies import get_vectordb
 from utils.logger import get_logger
 
@@ -63,6 +64,40 @@ def _token_encryption_key() -> str:
 
 def _claim_source() -> str:
     return os.getenv("OIDC_CLAIM_SOURCE", "id_token").strip().lower()
+
+
+def _auto_provision_login() -> bool:
+    """Whether to auto-provision a non-admin user on first OIDC login.
+
+    Defaults to ``False`` — keeping the historical "admin pre-creates every
+    user" model. Set ``OIDC_AUTO_PROVISION_LOGIN=true`` to enable: when the
+    callback receives a ``sub`` that isn't yet mapped to an OpenRAG user,
+    a row is created on the fly using the ID-token claims (``name`` /
+    ``preferred_username`` for the display name, ``email`` if present).
+
+    Auto-provisioned users are **never** admin and inherit the default file
+    quota — operators can promote / adjust afterwards via ``/users/``.
+    """
+    return os.getenv("OIDC_AUTO_PROVISION_LOGIN", "false").strip().lower() == "true"
+
+
+def _display_name_from_claims(claims: dict[str, Any], sub: str) -> str:
+    """Pick a sensible display name from the standard OIDC claims.
+
+    Falls back to a short ``sub`` prefix when nothing readable is available
+    so the user row always has something printable for the UI.
+    """
+    for key in ("name", "preferred_username"):
+        value = claims.get(key)
+        if isinstance(value, str) and value.strip():
+            return value.strip()
+    given = claims.get("given_name") or ""
+    family = claims.get("family_name") or ""
+    composed = f"{given} {family}".strip()
+    if composed:
+        return composed
+    # Last-resort fallback — keep enough of the sub to be unique-ish in the UI.
+    return f"oidc-{sub[:8]}"
 
 
 def _claim_mapping() -> dict[str, str]:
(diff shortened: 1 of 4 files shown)


EnjoyBacon7 (commented):
18 passed, 1 skipped.

`is_admin=False` is hard-coded with no override path via claims. Race condition on concurrent first-login handled correctly with catch-and-retry.

Two minor things:
- Email not normalized to lowercase at creation (the stub does `.lower()` but `auth.py` doesn't). Low risk but worth aligning.
- Auto-provisioned users have no partition memberships so they'll see an empty interface after login. By design, but worth noting somewhere.

Tests cover enabled, disabled, sub-prefix fallback, and explicit-false.

openrag/routers/auth.py
```diff
@@ -65,6 +66,40 @@ def _claim_source() -> str:
     return os.getenv("OIDC_CLAIM_SOURCE", "id_token").strip().lower()
 
 
+def _auto_provision_login() -> bool:
+    """Whether to auto-provision a non-admin user on first OIDC login.
+
+    Defaults to ``False`` — keeping the historical "admin pre-creates every
+    user" model. Set ``OIDC_AUTO_PROVISION_LOGIN=true`` to enable: when the
```
Ahmath-Gadji, line 73: It would be nice update the doc with this var.

* https://github.com/linagora/openrag/blob/8468db0b16c3ba365657af319f6fdc86e8494353/docs/oidc.md?plain=1#L102-L113
* https://github.com/linagora/openrag/blob/dev/docs/sso-quickstart.md

openrag/routers/auth.py
```diff
@@ -308,12 +343,47 @@ async def callback(request: Request, code: str | None = None, state: str | None
...
+            )
+
+        # Auto-provision: create a non-admin user from the ID-token claims.
+        # Email is best-effort — populated when the IdP exposes it on the
+        # ``email`` claim (typically via the ``email`` scope, which is in the
+        # default ``OIDC_SCOPES``). Display name falls back to the sub when
+        # the IdP exposes nothing readable.
+        display_name = _display_name_from_claims(bundle.claims, sub)
+        email = bundle.claims.get("email")
+        try:
```
Ahmath-Gadji, line 361: I think it would be nice as well to make the logic **auto-create or update** users from claims. The current logic only handles creation. The code block with `.create_user` will fail if the user already exists.

Ahmath-Gadji (commented):
Thanks a lot for this PR as it's much needed.
Few changes to be made before merge

EnjoyBacon7:
Superseded by #350 (same content from the linagora-side branch).

PR #342, opened 2026-04-25
fix(audio): skip WAV conversion for already-supported formats

etiquet:
## Summary

`OpenAIAudioLoader` was converting **every** non-WAV input to uncompressed WAV before sending it to the transcription endpoint. The conversion inflates a typical mp3 by ~10× — an 11.7 MB mp3 (≈ 30 min audio) becomes 124 MB in WAV, which trips the per-request size cap enforced server-side:

- **Scaleway Generative** (`whisper-large-v3`): 100 MB
- **OpenAI** (`/v1/audio/transcriptions`): 25 MB

The visible failure on Scaleway:

```
BadRequestError: Maximum file size exceeded
(parameter=audio_filesize_mb, value=124.77665328979492)
```

## Why the conversion existed in the first place

The original comment cites vLLM's `soundfile`/`libsndfile` dependency, which only decodes WAV/FLAC/OGG — so mp3 in particular fails with `Format not recognised` against a vLLM-served Whisper.

But the OpenAI `/v1/audio/transcriptions` contract (and every hosted compatible — Scaleway Generative, Groq, Replicate, etc.) explicitly accepts mp3, m4a, mp4, mpeg, mpga, wav, webm, flac, ogg. Forcing every input to WAV is over-conservative for the common case and actively breaks long audio uploads on hosted endpoints.

## The fix

- New module-level whitelist `DIRECT_UPLOAD_SUFFIXES` covering the 9 formats every OpenAI-compatible Whisper accepts.
- `AudioTranscriber.transcribe` now checks the suffix first: if it's in the whitelist, send the compressed file as-is (10× smaller). Otherwise (`.flv`, `.wma`, etc.) keep the WAV-conversion fallback so vLLM-style deployments still work.
- `AudioSegment.from_file` is still called when we keep the original file, because language detection needs to extract a 30-second slice from the decoded audio. ffmpeg-backed pydub handles all formats so this is a no-op for non-WAV inputs.
- Comment block updated: explains the trade-off, mentions the 25 MB OpenAI / 100 MB Scaleway caps explicitly, and notes that the WAV fallback covers the vLLM/libsndfile case.

## Test plan

- [x] 11.7 MB mp3 (30 min réunion) → previously failed mid-pipeline on Scaleway with `Maximum file size exceeded` → now uploads successfully and returns a transcript.
- [x] Pipeline complet validé bout-à-bout en production sur la VM Mirai : transcription Scaleway → 42 chunks → indexation Milvus → recherche RAG → citation MP3 retournée par Chainlit.
- [x] WAV / FLAC inputs continue to bypass conversion (no regression for already-WAV deployments).
- [x] Exotic containers (e.g. `.flv`) still go through the WAV fallback for vLLM/libsndfile compatibility.
- [x] Language detection (`_detect_language`) unaffected — still operates on a decoded 30 s slice via pydub.

## New Features
* Audio transcription now supports direct upload of additional audio formats, improving processing efficiency. Compatible audio formats bypass unnecessary conversion steps, while unsupported formats continue to be automatically converted as needed. Language detection functionality remains consistent across all supported formats.

### DIFF
--- a/openrag/components/indexer/loaders/audio/openai.py
+++ b/openrag/components/indexer/loaders/audio/openai.py
@@ -16,6 +16,17 @@ logger = get_logger()
 # Duration of the audio sample used for language detection
 LANG_DETECT_SAMPLE_MS = 30_000  # 30 s
 
+# Audio container formats that the OpenAI ``/v1/audio/transcriptions`` family
+# accepts directly (Whisper API, vLLM-served Whisper, Scaleway Generative
+# whisper-large-v3, etc.). For these we skip the WAV conversion and send the
+# file as-is — converting an already-compressed input to uncompressed WAV
+# inflates size by ~10×, which trips per-request size limits enforced
+# server-side (Scaleway: 100 MB; OpenAI: 25 MB).
+DIRECT_UPLOAD_SUFFIXES = {
+    ".wav", ".flac", ".ogg", ".mp3", ".mp4", ".m4a",
+    ".webm", ".mpeg", ".mpga",
+}
+
 
 class AudioTranscriber:
     """Transcribes audio in a single request (no chunking).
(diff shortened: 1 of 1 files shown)


Ahmath-Gadji:
Thanks a lot for this PR. It has lead to this one #346.

Previously **vLLM 0.16.0** was used which did not support **MP3/MP4 at the endpoint level**. the server would return a 400 for those formats. As a result, OpenRAG had to convert everything to **WAV** before sending it over.

MP3/MP4 support was introduced in a later vLLM release and the newest vLLM at this moment (v0.19.1) accepts it.

With [#346](https://github.com/linagora/openrag/pull/346) bumping the transcriber to v0.19.1, the endpoint now accepts these formats natively, on par with any other OpenAI-compatible transcription API. The direct-upload approach introduced here (skipping WAV conversion) is therefore now fully supported.

Ahmath-Gadji:
The linting CI is currently failing. To avoid this issue during development, you can configure a pre-commit hook locally that runs linting checks before each commit. Refer to the [LINTING](https://github.com/linagora/openrag/blob/main/LINTING.md?plain=1#L83-L122) guide for setup instructions.

openrag/components/indexer/loaders/audio/openai.py
```diff
@@ -16,6 +16,17 @@
 # Duration of the audio sample used for language detection
 LANG_DETECT_SAMPLE_MS = 30_000  # 30 s
 
+# Audio container formats that the OpenAI ``/v1/audio/transcriptions`` family
+# accepts directly (Whisper API, vLLM-served Whisper, Scaleway Generative
+# whisper-large-v3, etc.). For these we skip the WAV conversion and send the
+# file as-is — converting an already-compressed input to uncompressed WAV
+# inflates size by ~10×, which trips per-request size limits enforced
+# server-side (Scaleway: 100 MB; OpenAI: 25 MB).
+DIRECT_UPLOAD_SUFFIXES = {
```
Ahmath-Gadji, line 25: i suggest having these in conf/config.yaml file and make it configurable with an .env variable DIRECT_UPLOAD_SUFFIXES

```yml
loader:
  ...
  transcriber:
     direct_upload_suffixes: .wav||.ogg||.mp||.mp4 ...
```

EnjoyBacon7 (commented):
Passes all tests. Logic is correct.

`wav_path` is a misleading variable name when the file is mp3/m4a — could be `upload_path`. Pydub still loads the file even when `use_whisper_lang_detector=False`, but that's pre-existing behavior for WAV, not a regression.

The `DIRECT_UPLOAD_SUFFIXES` config issue from the comment below is now addressed.

EnjoyBacon7:
Made `DIRECT_UPLOAD_SUFFIXES` configurable as suggested.

- Added `direct_upload_suffixes` to `conf/config.yaml` under `loader.transcriber` (pipe-separated string, defaults to the existing 9 formats)
- Wired `TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES` env var in `config/loader.py`
- `AudioTranscriber.__init__` now reads from config and falls back to `_DEFAULT_DIRECT_UPLOAD_SUFFIXES` if unset

Example override for a strict vLLM deployment (WAV/FLAC/OGG only):
```
TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES=.wav|.flac|.ogg
```

EnjoyBacon7:
Superseded by #348 (same content plus a follow-up fix on the linagora-side branch).

Issue #344, opened 2026-04-26
OpenAI-compat clients (Open WebUI, LibreChat, …) silently drop the `extra.sources` field — sources never reach the user


etiquet:
## Problem

OpenRAG's `/v1/chat/completions` and `/v1/completions` endpoints return their source list in a **non-standard `extra` field at the top of the response**, encoded as a JSON-string :

```json
{
  "id": "chatcmpl-…",
  "choices": [{"message": {"content": "..."}}],
  "extra": "{\"sources\": [{\"file_url\": \"…\", \"title\": \"…\", \"relevance_score\": 0.71, …}]}"
}
```

This works for **clients written to consume it** (the official `pipe_openrag.py` for Open WebUI, MyRAG bridge, custom programmatic agents).

But **vanilla OpenAI-compat clients silently drop the field**, so the user sees the LLM answer with no source list at all. Verified for :

- **Open WebUI 0.9.2** (the most common deployment target) — `backend/open_webui/utils/middleware.py:non_streaming_chat_response_handler` reads `choices`, `output`, `usage` only ; no reference to `response_data['extra']`. Frontend Svelte/TS does not render any top-level `extra` either.
- **LibreChat** — generic OpenAI proxy, ignores unknown top-level fields.
- **Continue / Cursor / curl / SDK clients** — same story.

For sensitive use cases (legal, compliance, gov), losing the source list is a regression : the user gets an answer without an audit trail, which is unacceptable in our deployment context (French Ministry of the Interior, droit au séjour des étrangers).

## Why not push everyone to use a custom pipe / function ?

The Open WebUI ecosystem has [`pipe_openrag.py`](https://github.com/linagora/openrag/blob/dev/integrations/openwebui-pipe/pipe_openrag.py) (and similar in `mirai-open-webui`) that does the right thing — parses `extra.sources` and emits OWUI native `citation` events. But :

1. It needs to be **manually installed and activated** on every OWUI instance — non-trivial for non-developers
2. **Each new OpenAI-compat client** needs a similar adapter (LibreChat, Continue, etc.) — unbounded maintenance
3. **Programmatic clients** (curl, agents) still see no sources unless they parse `extra`

The OpenAI contract has no notion of citations at the response top-level, and OWUI is not adding support for `extra` (no related PR upstream — I checked `gh pr list --repo open-webui/open-webui --search "extra response field"` returns zero relevant matches).

## Proposed solution

[...]

PR #345, opened 2026-04-26
feat(openai): inline sources in content (opt-in) for clients ignoring `extra`

etiquet:
Companion PR for #344.

## TL;DR

OpenRAG returns its source list in a non-standard `extra` field at the top of `/v1/chat/completions` responses. Vanilla OpenAI-compat clients (Open WebUI 0.9.2, LibreChat, Continue, Cursor, curl, …) silently drop it — the user sees the answer without sources. This PR adds an **opt-in** flag `INLINE_SOURCES_IN_CONTENT` that *also* writes the source list into the assistant message `content` as a markdown block, so any OpenAI-compat client renders it natively. The structured `extra` field is unchanged for clients that already consume it.

Default = `false` → no breaking change. See #344 for full context.

## Changes

| File | What |
|---|---|
| `components/utils.py` | `inline_sources_enabled()` (env-var read) + `format_sources_as_markdown(sources)` helper (dedup on `file_url`, ranked by `relevance_score`, capped by `INLINE_SOURCES_TOP_K`, filtered by `INLINE_SOURCES_MIN_SCORE`). Streaming path injects one extra delta chunk before `finish_reason`. |
| `routers/openai.py` | 3 call sites — append the markdown block right after `extract_and_strip_sources_block`/`filter_sources_by_citations`. |
| `components/test_source_filtering.py` | 5 new tests : on/off behavior, dedup, ranking, top_k, min_score, streaming with the extra delta chunk, no-inline when disabled. |
| `.env.example` | Documents the 3 knobs (`INLINE_SOURCES_IN_CONTENT`, `INLINE_SOURCES_TOP_K`, `INLINE_SOURCES_MIN_SCORE`). |

## Knobs

```bash
INLINE_SOURCES_IN_CONTENT=true     # default false — opt-in
INLINE_SOURCES_TOP_K=5              # default 5 — max sources rendered
INLINE_SOURCES_MIN_SCORE=0.65       # default -inf — filter weak chunks
```

## Sample output (with flag on)

```
…answer from the LLM…

---
**Sources :**

1. [Article L423-1 — Conditions de séjour](https://api.openrag.example.com/static/foo.md) — score 0.71
2. [Décret 2024-… — Procédure CESEDA](https://api.openrag.example.com/static/bar.pdf) — score 0.69
…
```

The block is dedup'd by `file_url` (OpenRAG can return multiple chunks of the same file with different scores ; we show each file once with its best chunk), ranked by relevance score desc, and capped at `INLINE_SOURCES_TOP_K`.

## Validation

Tested in production on a deployment serving the French Ministry of the Interior :

- **Non-streaming `/v1/chat/completions`** : block appears in `choices[0].message.content`, `extra.sources` unchanged.
- **Streaming `/v1/chat/completions`** : extra delta chunk emitted between content stream and `finish_reason` ; OWUI 0.9.2 buffers it correctly and renders the full message with the source block.
- **Non-streaming `/v1/completions`** : block appears in `choices[0].text`.
- **With flag off (default)** : exactly identical to current behavior, no regression.
- Open WebUI 0.9.2 user-side : sources visible as clickable markdown links inside the answer card. No client modification needed. The ecosystem-specific `pipe_openrag.py` (which emits OWUI native `citation` events) keeps working in parallel for operators who want richer UX — both can coexist.
- 5 new unit tests pass (`pytest components/test_source_filtering.py`).

## Why opt-in (not always-on)

- **Zero breaking change** for current deployments using clients that consume `extra` (the existing pipe, MyRAG bridge, programmatic agents).
- The decision belongs to the operator : some workflows want structured citations only (clean for downstream parsing), others want the visible-content fallback for vanilla clients.
- Easy to A/B compare.

## Alternatives considered & rejected

| Option | Rejected because |
|---|---|
| Push OWUI to read `extra` (PR upstream) | No traction (0 related PR on `open-webui/open-webui`), multi-month timeline, doesn't help LibreChat/Continue/curl. |
| Use `choices[0].message.annotations` (OpenAI 2024+) | OWUI doesn't render those either today. |
| Force every operator to install a Pipe Function | Non-trivial UX, doesn't scale to other OpenAI-compat clients. |
| Custom SSE event types | Format undocumented + version-dependent across OWUI versions. |

A trailing markdown block is the **minimum delta** that works in every OpenAI-compat client today, with zero client-side change.

## Test plan

- [x] `pytest components/test_source_filtering.py` — 5 new + existing tests green
- [x] Production VM (Open WebUI 0.9.2 talking to OpenRAG via vanilla `OPENAI_API_BASE_URLS`) : sources visible in chat
- [x] With flag off : output byte-identical to current behavior
- [x] `extra.sources` always present in response (rétrocompat for pipe_openrag.py and MyRAG)

## Release Notes

* **New Features**
  * Optional inline markdown sources now appear directly in assistant responses
  * Sources are ranked, filtered, and deduplicated for improved relevance
  * Configurable behavior to control source inclusion and filtering thresholds

* **Tests**
  * Added unit tests for source formatting and filtering functionality

* **Chores**
  * Extended environment configuration documentation

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,18 @@ EMBEDDER_MODEL_NAME=jinaai/jina-embeddings-v3 # or other embedder from huggingfa
 # RETRIEVER
 # RETRIEVER_TOP_K=20 # number of top documents to retrieve, before reranking (lower (~10) is faster on CPU | on GPU, you can try to increase the value (~40) ).
 
+# INLINE SOURCES IN CONTENT — opt-in, default off
+# When true, OpenRAG appends a markdown source block (`**Sources :**` …) to the
+# assistant `content` of /v1/chat/completions and /v1/completions responses,
+# in addition to the structured `extra.sources` field that already exists.
+# Useful for clients that ignore the non-standard `extra` field at the
+# response top-level (Open WebUI, LibreChat, Continue, Cursor, vanilla curl).
+# Default `false` keeps the existing OpenAI-compat contract intact.
+# See linagora/openrag#344 for context.
+# INLINE_SOURCES_IN_CONTENT=false
+# INLINE_SOURCES_TOP_K=5
+# INLINE_SOURCES_MIN_SCORE=-inf
+
 # RERANKER
 RERANKER_ENABLED=true # deactivate the reranker if your CPU is not powerful enough
 RERANKER_MODEL=Alibaba-NLP/gte-multilingual-reranker-base # or jinaai/jina-reranker-v2-base-multilingual
--- a/openrag/components/test_source_filtering.py
+++ b/openrag/components/test_source_filtering.py
@@ -6,6 +6,7 @@ import pytest
 from components.utils import (
     extract_and_strip_sources_block,
     filter_sources_by_citations,
+    format_sources_as_markdown,
     stream_with_source_filtering,
 )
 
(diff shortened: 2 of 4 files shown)


EnjoyBacon7 (commented):
Two bugs fixed (see comments below): web sources were silently dropped because the dedup key and URL resolution only checked `file_url`/`filename`; and `[`/`]` in titles weren't escaped, breaking markdown links.

`format_sources_as_markdown` logic is otherwise clean. Dedup on best-score chunk per URL is correct. The streaming extra-delta is emitted before `finish_reason` so clients that buffer until finish see it as part of the content.

Minor: page 1 is excluded from the page annotation (`str(page) not in {"0", "1"}`). Likely intentional to avoid noise on single-page docs, but could use a comment.

EnjoyBacon7:
Fixed both CodeRabbit issues:

**Web sources dropped** (`format_sources_as_markdown`): web entries carry `url` not `file_url`/`filename`, so the dedup key resolved to `""` and they were skipped. Added `url` as a fallback in both the key and the link URL resolution. Also added a test fixture with `source_type: "web"` to lock this in.

**`[`/`]` not escaped in label**: only `|` was escaped; a title like `Report [draft]` would break the markdown link. Fixed to also escape `[` and `]`.

Diff:
```python
# dedup key
key = s.get("file_url") or s.get("url") or s.get("filename") or s.get("source") or ""

# link URL
url = s.get("file_url") or s.get("url") or s.get("chunk_url") or ""

# label escaping
label = _label(s).replace("\\", "\\\\").replace("[", "\\[").replace("]", "\\]").replace("|", "\\|")
```

EnjoyBacon7:
Reimplimented in #347 for editing rights

PR #348, opened 2026-04-28
fix(audio): skip WAV conversion for already-supported formats

EnjoyBacon7:
Replaces #342. Same content plus one follow-up fix for review feedback.

`OpenAIAudioLoader` was converting every non-WAV input to uncompressed WAV before sending it to the transcription endpoint. The conversion inflates a typical mp3 by ~10x, so a 30-min mp3 (11.7 MB) becomes 124 MB and trips the per-request size cap on hosted Whisper endpoints (Scaleway 100 MB, OpenAI 25 MB).

Fix: skip the conversion when the suffix is one of the formats every OpenAI-compat Whisper accepts (mp3, m4a, mp4, mpeg, mpga, wav, webm, flac, ogg). Other suffixes (`.flv`, `.wma`, ...) still go through the WAV fallback for vLLM/libsndfile deployments.

## Changes vs #342

- `51711b6` - move the suffix list into config (`loader.transcriber.direct_upload_suffixes`), overridable via `TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES`, addressing CodeRabbit's note about the hard-coded module-level whitelist

## Test plan

- 30-min mp3 that previously failed on Scaleway with `Maximum file size exceeded`: uploads and returns a transcript
- WAV/FLAC inputs unchanged (no conversion)
- `.flv` and other exotic containers still go through the WAV fallback for vLLM
- Default suffix list matches the previous hard-coded set, no regression with default config

## Release Notes

* **New Features**
  * Added support for direct audio file uploads to the transcriber without format conversion for supported file types.
  * New environment variable configuration option to specify which audio/video formats bypass conversion.

* **Documentation**
  * Updated configuration documentation with guidance on audio file format handling and codec compatibility.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,12 @@ EMBEDDER_MODEL_NAME=jinaai/jina-embeddings-v3 # or other embedder from huggingfa
 RERANKER_ENABLED=true # deactivate the reranker if your CPU is not powerful enough
 RERANKER_MODEL=Alibaba-NLP/gte-multilingual-reranker-base # or jinaai/jina-reranker-v2-base-multilingual
 
+# Audio transcription
+# Pipe-separated list of file extensions sent directly to the transcription endpoint
+# without converting to WAV first. Defaults cover all formats the OpenAI Whisper
+# API accepts natively. Override to restrict (e.g. to ".wav" only for vLLM deployments).
+# TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES=.wav|.flac|.ogg|.mp3|.mp4|.m4a|.webm|.mpeg|.mpga
+
 # Prompts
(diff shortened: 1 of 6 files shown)


Ahmath-Gadji (approved):
I've made some changes here in this commit

https://github.com/linagora/openrag/pull/348/changes/0f20e3898b3f56cd58914d0ade305f0d25b231d2

* parse direct_upload_suffixes in TranscriberConfig
* Move the pipe-delimited string → set conversion into a Pydantic
field_validator on TranscriberConfig
  * The audio loader now reads the parsed
set directly, eliminating the duplicated default constant and the
inline split logic.
* Document TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES in
env_vars.md.

PR #392, opened 2026-05-11
fix(audio): bind tmp_wav before try in OpenAI transcribe() (#366)

### LINKED ISSUE #366: Audio loader masks errors with UnboundLocalError on cleanup
`openrag/components/indexer/loaders/audio/openai.py` around line 63 references `tmp_wav` inside a `finally` block. The variable is only bound inside the if/else (set to `None` for `.wav` at line 45, assigned for non-wav at line 50). If `AudioSegment.from_file` at line 48 raises before line 50 executes, the `finally` clause raises `UnboundLocalError` and discards the original exception, leaving operators with no clue what really failed.

Initialize `tmp_wav = None` at the top of the function before the try block. The existing `if tmp_wav:` cleanup will then run safely on every code path and the original exception will reach the caller. Add a regression test that monkey-patches `AudioSegment.from_file` to raise.

SAY-5:
Fixes #366.

`AudioTranscriber.transcribe()` only assigned `tmp_wav` inside the if/else block, so when `AudioSegment.from_file` raised before that point, the `finally` clause hit `UnboundLocalError` and masked the real decode failure.

Move `tmp_wav = None` ahead of the try so the cleanup branch always sees the name; the real exception now propagates to the caller. Added a small regression test that mirrors the buggy / fixed control flow without pulling in the heavy ray / openai dependencies.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression test suite for audio transcription error handling, covering edge cases in control flow where variable initialization timing affects exception propagation. Test cases validate both problematic patterns that inadvertently mask errors and corrected patterns that ensure audio processing failures are properly reported.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/openrag/pull/392)

### DIFF
--- a/openrag/components/indexer/loaders/audio/test_openai.py
+++ b/openrag/components/indexer/loaders/audio/test_openai.py
@@ -276,3 +276,41 @@ class TestAudioChunking:
         # Chunks should be contiguous
         for i in range(len(chunks) - 1):
             assert chunks[i][1] == chunks[i + 1][0]
+
+
+class TestTranscribeFinallyCleanup:
+    """Regression test for the UnboundLocalError in transcribe()'s finally clause.
+
+    The transcribe() method previously bound ``tmp_wav`` only inside the
+    if/else block, so if ``AudioSegment.from_file`` raised at the very top
+    of the try, the finally clause crashed with ``UnboundLocalError`` and
+    masked the real exception. We mirror the buggy / fixed control flow here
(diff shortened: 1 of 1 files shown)


Ahmath-Gadji (commented):
Thank you for your contribution — I approve these changes.
Could you retarget the PR to the dev branch?

Our workflow is as the following: We merge features and bug fixes on dev, and once dev is stable and tested we rebase main on dev then proceed to a new release

PR #403, opened 2026-05-20
fix(partition): require admin for POST /partition/{name} (#358)

### LINKED ISSUE #358: POST /partition/ has no role check, any authenticated user can create
`openrag/routers/partition.py` at line 227 declares `create_partition` with only `request: Request` and `vectordb=Depends(get_vectordb)`. No role dependency is attached, and `app.include_router(partition_router, ...)` in `openrag/api.py` around line 298 does not install one either. Any authenticated non-admin user can `POST /partition/` to create arbitrary partitions and is recorded as the owner via line 233 (`user_id = request.state.user["id"]`), bypassing the RBAC model documented in `CLAUDE.md`.

Add `_=Depends(require_admin)` (or a dedicated `require_partition_creator`) to the route signature, matching the pattern used elsewhere in `routers/utils.py`. If non-admins are intended to create partitions, gate the call behind a per-user partition-count quota and document the policy alongside `file_quota`.

EnjoyBacon7:
## Summary

`routers/partition.py` declared `create_partition` with only `request: Request` and a `PartitionService` dependency. No role dependency was attached, so any authenticated non-admin user could `POST /partition/{name}` to create arbitrary partitions and be recorded as the owner, bypassing the RBAC model documented in `CLAUDE.md`.

This PR adds `_=Depends(require_admin)` to the route signature.

## Test plan

- [ ] Admin user can still create partitions
- [ ] Non-admin user gets 403 when calling POST /partition/{name}

Fixes #358

## Summary by CodeRabbit

* **Bug Fixes**
  * Partition creation is now restricted to administrators only, enforcing stricter authorization controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/openrag/pull/403?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openrag/routers/partition.py
+++ b/openrag/routers/partition.py
@@ -22,6 +22,7 @@ from utils.logger import get_logger
 from .utils import (
     ROLE_HIERARCHY,
     partitions_with_details,
+    require_admin,
     require_partition_owner,
     require_partition_viewer,
 )
@@ -234,6 +235,7 @@ Returns 409 Conflict if partition already exists.
 async def create_partition(
     request: Request,
     partition: str,
+    _=Depends(require_admin),
     service: PartitionService = Depends(get_partition_service),
 ):
     if await service.partition_exists(partition):
--- a/tests/api_tests/test_indexer.py
+++ b/tests/api_tests/test_indexer.py
@@ -385,24 +385,25 @@ class TestUserQuotaEnforcement:
         return response.json()
 
     def _create_partition(self, api_client, partition_name: str, user_token: str):
-        """Helper to create a partition and grant the test user editor access.
+        """Helper to create a partition and promote the test user to owner.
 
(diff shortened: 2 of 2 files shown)


EnjoyBacon7:
Closing: partition creation is self-service by design, not a vulnerability.

The endpoint's own docstring states it 'automatically assigns you as the partition owner', and the base behavior lets any authenticated user create a partition and become its owner. The RBAC model in CLAUDE.md governs membership roles (owner/editor/viewer) on existing partitions; it does not restrict who may create one, so creating-and-owning is consistent with it rather than a bypass.

Requiring admin would break the intended self-service model. The only real concern (unbounded creation) is better addressed by a per-user partition quota mirroring file_quota, which can be tracked separately if needed.

PR #427, opened 2026-05-26
fix(pptx): return placeholder for any chart ValueError, add tests (#388)

### LINKED ISSUE #388: PPTX chart conversion returns None and aborts the file
`openrag/components/indexer/loaders/pptx_loader.py:127-133` only returns from the `except ValueError` branch when the message contains `"unsupported plot type"`. Any other `ValueError` falls through and the function returns `None` implicitly. The caller at line 58 then runs `md_content += self._convert_chart_to_markdown(...)`, which raises `TypeError: can only concatenate str (not "NoneType") to str` and aborts the entire PPTX, even when only a single chart slide is malformed.

Return an empty string (or a placeholder line) instead of `None`, and let the caller decide whether to skip or annotate the chart. Add a unit test that feeds a chart raising a non-matching `ValueError` and asserts the loader still produces markdown for the surrounding slides.

hedhoud:
Closes #388

## Context

`_chart_to_markdown` in `pptx_parser.py` had two identical `ValueError` branches — a leftover from the original `pptx_loader.py` that only returned on `"unsupported plot type"` and fell through (returning `None`) for any other `ValueError`. If that `None` reached the caller it would raise `TypeError: can only concatenate str (not "NoneType") to str`, aborting the entire PPTX.

During the hexagonal refactoring the code moved to `pptx_parser.py` and the else-fall-through was plugged, but the redundant `if "unsupported plot type"` check remained as dead code, with no test coverage proving the fix held.

## Changes

- Collapse the two identical `ValueError` branches into a single `except Exception` clause in `_chart_to_markdown`
- Add `test_pptx_parser.py` with four cases: matching `ValueError`, non-matching `ValueError`, non-`ValueError` exception, and a sweep of error messages — all asserting the method never returns `None`

## Validation

`uv run pytest` — 1010 passed, 4 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Chart conversion now consistently returns a fallback placeholder for any failing chart, preventing conversion failures from interrupting slide rendering.

* **Tests**
  * Added regression tests covering multiple chart failure scenarios and end-to-end slide conversion paths to ensure stable behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/openrag/pull/427?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openrag/core/indexing/parsers/pptx_parser.py
+++ b/openrag/core/indexing/parsers/pptx_parser.py
@@ -187,9 +187,5 @@ class PptxParser(DocumentParser):
             rows = ["| " + " | ".join(map(str, r)) + " |" for r in data]
             separator = "|" + "|".join(["---"] * len(data[0])) + "|"
             return md + "\n".join([rows[0], separator] + rows[1:])
-        except ValueError as exc:
-            if "unsupported plot type" in str(exc):
-                return "\n\n[unsupported chart]\n\n"
-            return "\n\n[unsupported chart]\n\n"
         except Exception:
             return "\n\n[unsupported chart]\n\n"
--- a/openrag/core/indexing/parsers/test_pptx_parser.py
+++ b/openrag/core/indexing/parsers/test_pptx_parser.py
@@ -5,9 +5,7 @@ message contained "unsupported plot type"; any other ValueError fell through
 and the function returned None implicitly, causing TypeError in the caller.
 """
 
-import pytest
-
-from .pptx_parser import PptxParser
+from core.indexing.parsers.pptx_parser import PptxParser
 
 
 class _RaisingCategories:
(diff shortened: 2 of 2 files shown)


hedhoud:
## What was wrong

`PptxParser._chart_to_markdown` had two `except ValueError` branches with identical bodies — both returned the placeholder string, so one was dead code. The original bug (returning `None` instead of a placeholder) had already been fixed during the hexagonal refactoring, but no test verified the behaviour, leaving the dead branch silently in place.

## What was changed

Collapsed both `ValueError` branches into a single `except Exception`, which also covers unexpected chart library errors beyond `ValueError`. Added `test_pptx_parser.py` with four cases: matching `ValueError`, arbitrary `ValueError`, non-`ValueError` exception, and a sweep asserting the return value is never `None`.

## Verified on server

PPTX with a chart (Q3 revenue/costs) indexed successfully, Chainlit answered chart-related questions correctly with source citations. No `None`-return crash observed.

PR #428, opened 2026-05-26
fix(loaders): remove re.sub caption workaround, inject captions verbatim (#389)

### LINKED ISSUE #389: Image captions injected into re.sub raise re.error
`openrag/components/indexer/loaders/pptx_loader.py:151` and `openrag/components/indexer/loaders/docx.py:60` pass VLM-generated caption strings directly as the second argument to `re.sub`. Backslash escapes such as `\1`, `\g<name>`, or any literal numeric backreference produced by the captioning model raise `re.error: invalid group reference`, killing the loader for that document with a stack trace that does not name the offending caption.

Use a callable replacement, for example `re.sub(pattern, lambda _: caption, text)`, so the caption is treated as a literal. Add a regression test with a caption containing `\1` and `\g<x>` and assert the substitution succeeds.

hedhoud:
Closes #389

## Context

`pptx_loader.py` and `docx.py` originally used `re.sub(pattern, caption, text)` to inject VLM captions into the slide/document markdown. Any caption produced by the VLM that contained `\1`, `\g<name>`, or similar backslash sequences was interpreted by `re.sub` as a backreference and raised `re.error`, killing the entire file load.

During the hexagonal refactoring both files were switched from `re.sub` to `str.replace()`, which treats the replacement as a literal string. The root cause of `re.error` was eliminated. However the `caption.replace("\\", "/")` workaround that had been added to paper over the `re.sub` issue was carried over unchanged — it now only serves to mangle captions containing backslashes (e.g. a Windows path `C:\Users\alice` becomes `C:/Users/alice` in the output).

## Changes

- Remove `caption.replace("\\", "/")` from `PPTXLoader.aload_document` and `DocxLoader.aload_document` — captions are now injected verbatim via `str.replace()`
- Add `TestCaptionBackslashInjection` in `test_docx_loader.py` covering `\1`, `\g<name>`, Windows paths, and a sweep of backslash sequences — proving neither an exception is raised nor the content is mangled

## Validation

`uv run pytest` — 1010 passed, 4 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed image caption text handling in Word and PowerPoint document extraction. Captions now preserve backslashes and special characters as literal text instead of normalizing them to forward slashes.

* **Tests**
  * Added comprehensive regression tests to verify caption injection correctly handles special characters, regex patterns, and Windows-style paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/openrag/pull/428?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openrag/components/indexer/loaders/docx.py
+++ b/openrag/components/indexer/loaders/docx.py
@@ -71,7 +71,7 @@ class DocxLoader(BaseLoader):
             for block, caption in zip(processed.images, captions):
                 ref = (block.metadata or {}).get("markdown_ref")
                 if ref:
-                    result = result.replace(ref, caption.replace("\\", "/"))
+                    result = result.replace(ref, caption)
 
             result = await self.replace_markdown_images_with_captions(
                 result,
--- a/openrag/components/indexer/loaders/pptx_loader.py
+++ b/openrag/components/indexer/loaders/pptx_loader.py
@@ -63,7 +63,7 @@ class PPTXLoader(BaseLoader):
             for block, caption in zip(processed.images, captions):
                 ref = (block.metadata or {}).get("markdown_ref")
                 if ref:
-                    md_content = md_content.replace(ref, caption.replace("\\", "/"))
+                    md_content = md_content.replace(ref, caption)
 
         doc = Document(page_content=md_content, metadata=metadata)
         if save_markdown:
--- a/openrag/components/indexer/loaders/test_docx_loader.py
+++ b/openrag/components/indexer/loaders/test_docx_loader.py
@@ -129,6 +129,64 @@ class TestGetImagesFromZip:
         assert images == []
 
 
+class TestCaptionBackslashInjection:
+    """Regression tests for #389.
+
+    The old code passed VLM captions directly to re.sub as the replacement
+    string. Captions containing \\1 or \\g<name> raised re.error. The fix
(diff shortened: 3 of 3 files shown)


hedhoud:
## What was wrong

The original code used `re.sub(pattern, caption, text)` to inject VLM captions into document content. Because `re.sub` interprets backreferences in the replacement string, captions containing `\1`, `\g<name>`, or Windows paths (`C:\Users\...`) raised `re.error` or silently corrupted the text.

During refactoring the call was correctly replaced with `str.replace(ref, caption)`, which treats the replacement as a literal. However, a workaround from the `re.sub` era — `caption.replace("\\", "/")` — was carried over into both `PPTXLoader` and `DocxLoader`. This mangled backslashes in captions even though `str.replace` no longer needed protecting.

## What was changed

Removed `caption.replace("\\", "/")` from both loaders. Captions are now injected verbatim. Added `TestCaptionBackslashInjection` in `test_docx_loader.py` covering `\1`, `\g<name>`, Windows paths, and a full sweep of problematic sequences.

## Verified on server

DOCX with an embedded image indexed to COMPLETED. Server logs confirmed `Captioning embedded images: 100%|██████████| 1/1` with `IMAGE_CAPTIONING=true`, no crash on caption injection.

Issue #383, opened 2026-05-07
Web search fetcher follows redirects and accepts numeric IPs


EnjoyBacon7:
`openrag/components/websearch/content_fetcher.py:50-57` validates only the initial URL with `_is_loopback_url`, then issues `client.get(url, follow_redirects=True)` at lines 67-68 with no per-hop revalidation. The check itself only parses the host string with `ipaddress.ip_address`, so decimal-encoded IPs like `http://2130706433/` (which resolves to 127.0.0.1), octal forms, and DNS names that resolve to internal addresses all bypass the guard, and any benign URL can redirect to `169.254.169.254` to read cloud metadata.

Resolve the host to an IP, reject any address in private, loopback, link-local, or carrier-grade NAT ranges, and re-run that check on every redirect (use httpx event hooks or set `follow_redirects=False` and walk the chain manually). Cap the redirect chain length and the total fetch size, and log structured events when a redirect is rejected.

EnjoyBacon7:
This issue has been fixed and is included in the refactor/hexagonal branch, which will be merged to main shortly.

PR #429, opened 2026-05-26
fix(websearch): harden SSRF guard in ContentFetcher (#383)

hedhoud:
## Why

The web search fetcher fetches URLs returned by an external provider. A compromised or misbehaving provider could return URLs pointing at internal infrastructure. Three vectors were unguarded.

## What was wrong

- Only loopback addresses were blocked — RFC 1918 private ranges, link-local (169.254.x.x), CGNAT (100.64.x.x), and other reserved networks were accepted.
- Decimal-integer-encoded IPv4 addresses (e.g. `2130706433` == `127.0.0.1`) were not caught because `ipaddress.ip_address()` only accepts dotted-decimal notation.
- `follow_redirects=True` meant a legitimate initial URL could silently redirect to a private address after the initial SSRF check passed.

## What changed

- `_is_loopback_url` replaced by `_is_safe_url` (module-level), which checks all private/reserved/non-global IP flags explicitly, handles decimal-integer IPv4 encoding, and rejects non-HTTP(S) schemes.
- Redirect following is now manual (`follow_redirects=False`): every redirect target is validated by `_is_safe_url` before the next request is sent, with a hard cap of 10 hops.

## Tests

39 cases covering: all private/reserved ranges, decimal-integer encoding, non-HTTP schemes, redirect-to-private-IP (redirect target never contacted), safe redirect chains, and redirect limit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security for web search functionality by validating URLs and preventing access to internal/private addresses and unsafe schemes. All redirect paths are now validated to prevent bypass attempts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/openrag/pull/429?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openrag/components/websearch/content_fetcher.py
+++ b/openrag/components/websearch/content_fetcher.py
@@ -1,6 +1,6 @@
 import asyncio
 import ipaddress
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
 
 import httpx
 import lxml.html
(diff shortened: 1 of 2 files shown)


hedhoud:
## What was wrong

`ContentFetcher._fetch_single` had three gaps in its SSRF guard:

1. **Only loopback was blocked.** The old `_is_loopback_url` checked `localhost` and whether `ip_address.is_global` was False for IP literals, but did not explicitly cover RFC 1918 private ranges, link-local (169.254.x.x), CGNAT (100.64.x.x), or other reserved blocks.

2. **Decimal-integer IPv4 encoding slipped through.** A URL like `http://2130706433/` (where `2130706433` is `127.0.0.1` as a 32-bit integer) fails `ipaddress.ip_address("2130706433")` with `ValueError`, so the old guard returned `False` (safe) and let it through. glibc's resolver interprets decimal integers as packed IPv4 addresses, so the actual TCP connection would target `127.0.0.1`.

3. **Redirects were followed without re-checking.** `follow_redirects=True` meant a public URL could return a 302 pointing at an internal address; the redirect target was never validated.

## What was changed

- `_is_loopback_url` replaced by `_is_safe_url` (module-level), which checks `is_loopback | is_private | is_link_local | is_reserved | is_unspecified | is_multicast | not is_global` explicitly and rejects non-HTTP(S) schemes.
- Decimal-integer IPv4 now caught via a second attempt: `ipaddress.ip_address(int(host))` — `ipaddress` interprets the integer as a packed IPv4 address, matching the OS resolver.
- `follow_redirects=False` replaces the blind follow. Redirects are followed manually (up to 10 hops); `_is_safe_url` is called on every redirect target before the next request is sent.

## Tests

39 cases: all private/reserved address ranges, decimal-integer encoding, non-HTTP schemes, redirect-to-private-IP (redirect target never contacted), safe redirect chains, and redirect chain length cap.

## Note on DNS rebinding

Full DNS rebinding protection (where a hostname resolves to a public IP at check time then to a private IP at connect time) requires a custom transport layer and is out of scope here. The per-hop redirect validation and IP-literal checks cover the practical attack vectors from a compromised search provider.

PR #444, opened 2026-06-04
refactor(phase 14): Model Endpoints Registry and Per-Partition Presets

hedhoud:
## Why this PR matters

Phase 14 moves OpenRAG away from one global inference/pipeline setup and gives operators a safer way to manage models and partition behavior. Model endpoints are now named, presets are reusable, and each partition can resolve its own indexation and retrieval pipeline without hardcoding everything in environment variables.

The main goal is operational control: swap or validate model endpoints, update shared presets, and let partitions pick the right pipeline while keeping the existing default behavior for deployments that do not opt into customization.

## What changed

This PR adds the model endpoint registry, the preset registry, and the admin APIs needed to manage them. It also wires those registries into startup, indexing, retrieval, and partition configuration resolution.

A few review fixes were also added before final validation: preset renames are transactional, invalid preset types fail explicitly, retrieval now uses the partition-specific embedder through the correct searcher, and startup/auth error handling is stricter.

## Validation

Automated validation is passing on PR #444:

- CodeRabbit review
- lint
- unit tests
- api-tests
- layer import guard
- Milvus integration

Manual E2E on the server is the remaining check before merge.


Ahmath-Gadji:
@coderabbit

Ahmath-Gadji:
@coderabbit review

PR #446, opened 2026-06-05
fix(test): resolve flaky test_cancel_increments_total_cancelled

### LINKED ISSUE #426: Flaky API test: test_cancel_increments_total_cancelled races the async counter update
## Summary

`tests/api_tests/test_indexer.py::TestTaskCancellation::test_cancel_increments_total_cancelled` is flaky and broke the API Tests workflow on the `chore: bump version to 1.1.11` push to `main` (commit c9d53cc0, run [26233866578](https://github.com/linagora/openrag/actions/runs/26233866578)).

The tag-triggered `Build and Release` workflow itself succeeded — only this test failed.

## Failure

```
FAILED test_indexer.py::TestTaskCancellation::test_cancel_increments_total_cancelled
  assert 1 == (1 + 1)
```

## Root cause

At `tests/api_tests/test_indexer.py:766-791` the test:

1. Reads `cancelled_before = /queue/info → tasks.total_cancelled`
2. Uploads a file and `DELETE /indexer/task/{task_id}`
3. Immediately re-reads `cancelled_after`
4. Asserts `cancelled_after == cancelled_before + 1`

The `DELETE` endpoint only returns *"Cancellation signal sent"*. The `total_cancelled` counter is bumped asynchronously by `TaskStateManager` when the task actually transitions to `CANCELLED`. The test re-reads `/queue/info` before that bump lands, so the counter looks unchanged.

The previous test in the class (`test_cancel_task_sets_cancelled_state`) also cancels a task — that cancellation is what makes `cancelled_before == 1` here, confirming the counter does work, just not synchronously.

## Suggested fix

Poll `/queue/info` (or `/indexer/task/{task_id}` until `task_state == "CANCELLED"`) before asserting the counter, instead of reading immediately after the cancel call returns.

EnjoyBacon7:
## Summary

Fixes issue #426 by eliminating the race condition in the task cancellation test. The test was immediately checking the `total_cancelled` counter after sending a DELETE request, before the TaskStateManager had a chance to asynchronously update the counter.

## Changes

- Added `wait_for_task_cancellation()` helper function in conftest.py that polls the task status endpoint until the task state reaches CANCELLED
- Updated `test_cancel_increments_total_cancelled` to call this helper before asserting the counter increment
- Follows the existing pattern of `wait_for_task()` for consistency

## Why This Fixes It

The DELETE endpoint returns immediately with "Cancellation signal sent", but the counter update happens asynchronously. By polling until the task state becomes CANCELLED, we ensure the counter has been updated before we check it, making the test deterministic.

Fixes #426


EnjoyBacon7:
Closing as duplicate of #432, which already addresses issue #426 with the same solution.

Issue #379, opened 2026-05-07
Partition and file deletes diverge Milvus from PG on partial failure


EnjoyBacon7:
`openrag/components/indexer/vectordb/vectordb.py:687-716` (`delete_file`) and `openrag/components/indexer/vectordb/vectordb.py:1083-1112` (`delete_partition`) both issue the Milvus delete first, then run Postgres cleanup in subsequent steps with no shared transaction or compensation path. If a Postgres call fails after the Milvus delete succeeds, chunks are gone from the vector store but the database still references the file or partition, the per-uploader `file_count` decrement never runs, and operators have no way to detect the divergence other than auditing both stores by hand.

Run the Postgres cleanup first inside a single transaction, then issue the Milvus delete only after the Postgres commit succeeds. If the Milvus delete fails, log a structured error and enqueue a background reconciliation task that retries the orphaned Milvus delete using the captured chunk ids.

EnjoyBacon7:
## Fix Summary

Reordered delete operations to ensure database cleanup precedes vector store deletion:

**dispatcher.delete_file** (for #378):
- Removes file from workspaces and database first
- Then queries and deletes from Milvus
- If Milvus delete fails, logs error with reconciliation task info but doesn't raise

**partition_service.delete_partition** (for #379):
- Deletes partition from database first in transaction
- Then queries and deletes chunks from Milvus
- If Milvus delete fails, logs error with reconciliation task info

This ensures the data model remains consistent:
- If Postgres cleanup succeeds but Milvus delete fails: file/partition is gone from database (authoritative), orphaned chunks logged for reconciliation
- No scenario where database references deleted chunks or vice versa

## Tests Added

Regression tests verify:
1. Correct operation order (database → vector store)
2. Error handling when vector store delete fails
3. Proper structured logging with reconciliation task info

All 27 existing dispatcher and partition service tests still pass.

PR #447, opened 2026-06-05
fix: ensure database cleanup precedes vector store deletion (#378, #379)

### LINKED ISSUE #378: Replace-file path keeps old chunks if PG metadata write fails
`openrag/components/indexer/vectordb/vectordb.py:843-923` runs `add_documents_for_existing_file` in four separate steps: snapshot old chunk ids, insert new chunks, delete old chunks, update Postgres metadata. The docstring acknowledges that a step-3 failure leaves duplicates, but a step-4 failure also leaves the new chunks indexed alongside stale Postgres rows whose `relationship_id` and `file_metadata` no longer match the indexed content. There is no compensation, retry, or background reconciliation, so divergence persists until manual intervention.

Reorder the operations so Postgres metadata is written first inside a transaction that holds row locks on the affected files, then issue Milvus delete and insert under that transaction's scope using a saga pattern with explicit compensation on failure. As a minimum, schedule a follow-up reconciliation task that detects chunks whose `relationship_id` does not match the file's current Postgres row and removes them.

EnjoyBacon7:
## Summary

Fixes data divergence issues where Milvus and Postgres get out of sync during delete operations.

## Changes

**dispatcher.delete_file (#378)**
- Removes file from workspaces/database first
- Then deletes from Milvus
- Logs errors for reconciliation if Milvus delete fails

**partition_service.delete_partition (#379)**  
- Deletes partition from database first
- Then deletes chunks from Milvus
- Logs errors for reconciliation if Milvus delete fails

## Tests

✅ All 27 existing tests pass
✅ 4 new regression tests verify operation order and error handling

## Data Consistency

- **Success**: Both stores cleaned up
- **DB succeeds, Milvus fails**: Authoritative DB is clean, orphaned chunks logged for reconciliation  
- **DB fails**: Transaction rolls back, both stores unchanged

Fixes #378 #379


EnjoyBacon7:
This PR is superseded - the fix has been directly committed to refactor/hexagonal (commit e8a55806). When refactor/hexagonal merges to main, this fix will be included. Closing to avoid duplicate.

Issue #451, opened 2026-06-08
Marker PDF parsing can lose GPU allocation after Phase 11


hedhoud:
Context: Phase 11 moved Marker worker creation behind the Ray parser pool. PDF indexing became much slower because work that should use GPU can stay in SERIALIZING for minutes.\n\nProblem: Marker GPU allocation is decided with CUDA visibility from the pool actor. That actor may not have GPU resources assigned, so the worker can be created without GPU even when the Ray cluster has GPUs available. The result is CPU-bound OCR/layout/table recognition and a large regression versus Phase 10.\n\nExpected behavior: Marker workers should request the configured GPU fraction whenever Ray exposes GPU resources, so PDF serialization uses the intended accelerator and indexing time stays comparable with Phase 10.

EnjoyBacon7:
A fix exists but is not merged, and the targeted code is not on `main`. `a0fba51a` ("fix(workers): preserve Marker GPU allocation from Ray resources") addresses this on the workers/services refactor branch; the `services/workers/parsers` layout it edits does not exist in `main`. Needs merge / re-evaluation against `main`.

Ahmath-Gadji:
Same issue whisper workers. Solved with this PR #496

PR #452, opened 2026-06-08
fix(workers): preserve Marker GPU allocation from Ray resources

hedhoud:
Context: Phase 11 made Marker PDF parsing much slower because GPU allocation can be decided from a Ray actor that does not itself own GPU resources.\n\nThis keeps Marker worker GPU requests based on Ray cluster resources, while still falling back safely when GPU resources are absent. That prevents PDF serialization from silently becoming CPU-bound on GPU deployments.\n\nCloses #451.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU detection for the marker parser: non-positive GPU requests are treated as zero, and GPU honoring now checks cluster-reported resources with a safer fallback when cluster queries fail.
* **Tests**
  * Added a unit test covering GPU-detection behavior and refactored an existing dispatcher test for clearer async mocking.

### DIFF
--- a/openrag/services/workers/parsers/marker_workers.py
+++ b/openrag/services/workers/parsers/marker_workers.py
@@ -25,7 +25,15 @@ logger = get_logger()
 
 
 def _marker_num_gpus(config) -> float:
-    return config.loader.marker_num_gpus if torch.cuda.is_available() else 0
+    """Return the configured Marker GPU request when Ray reports GPU capacity."""
+    requested_gpus = config.loader.marker_num_gpus
+    if requested_gpus <= 0:
+        return 0
+    try:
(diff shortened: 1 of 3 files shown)


Ahmath-Gadji:
@hedhoud, this fix is already included in the `refactor/hexagonal` branch (see commit from PR #444:  [https://github.com/linagora/openrag/pull/444/changes/a0fba51a476cfee5e1d9e5a63a592c606faa31d4](https://github.com/linagora/openrag/pull/444/changes/a0fba51a476cfee5e1d9e5a63a592c606faa31d4)).

I think this PR can be closed since the change has already been implemented.

PR #456, opened 2026-06-10
chore: update template & general improvement

ThibautChoppy:
## Summary by CodeRabbit

* **New Features**
  * Configurable pod/container securityContexts across components
  * Release-scoped resource naming for env, PVCs, RayCluster, services and secrets
  * Flexible secret management: values, ExternalSecret, or VaultStaticSecret with release-scoped secret name helper
  * Health checks and optional runtimeClass for reranker; indexer UI mounts config JSON volume
  * Enhanced ingress routing (regex + /api route) and proxy env support in env ConfigMap
  * Post-install notes now report secret configuration and HF token warnings

* **Chores**
  * Chart bumped to 0.5.0 with metadata/maintainers; images pinned; dependency version ranges tightened


bastiansmn:
+1

charts/openrag-stack/templates/ingress.yaml
```diff
@@ -3,26 +3,31 @@ apiVersion: networking.k8s.io/v1
...
+  {{- if .Values.ingress.className }}
   ingressClassName: {{ .Values.ingress.className | quote }}
+  {{- end }}
   rules:
     - http:
         paths:
           {{- if .Values.ingress.paths.openrag.enabled }}
           - path: /(.*)
-            pathType: ImplementationSpecific
+            pathType: Prefix
```
EnjoyBacon7: **Ingress does not route anything with default values.**

With the default values there are no ingress annotations and pathType is set to `Prefix`, but the path is `/(.*)`. Prefix matching does not understand regex, so it looks for a request path that literally starts with `/(.*)`. Normal requests like `/v1/chat/completions` will not match and traffic goes nowhere. It only works once you add the nginx `use-regex` and `rewrite-target` annotations, and right now only `values-linagora.yaml` has those. Either set `pathType: ImplementationSpecific` back for this regex path, or put the working nginx annotations in the default `values.yaml`.

charts/openrag-stack/values.yaml
```diff
@@ -304,18 +419,16 @@ env:
...
+    EMBEDDER_MODEL_NAME: '{{ .Values.vllm.embedderModelName }}'
     EMBEDDER_BASE_URL: "http://{{ .Release.Name }}-embedder-engine-service/v1"
 
     # Reranker
-    RERANKER_MODEL: *rerankerModel
+    RERANKER_MODEL: '{{ .Values.reranker.rerankerModelName }}'
     RERANKER_ENABLED: "true"
     RERANKER_TOP_K: "4"
-    RERANKER_BASE_URL: "http://{{ .Release.Name }}-reranker:{{ .Values.reranker.servicePort }}"
+    RERANKER_BASE_URL: '{{ if .Values.reranker.externalUrl }}{{ .Values.reranker.externalUrl }}{{ else }}http://{{ .Release.Name }}-reranker:{{ .Values.reranker.servicePort }}{{ end }}'
```
EnjoyBacon7: **Disabling the reranker leaves the app pointing at it.**

When `reranker.enabled` is false and `externalUrl` is empty, the reranker Deployment is not created, but the config still sets `RERANKER_ENABLED` to true and `RERANKER_BASE_URL` to the in-cluster reranker service. The app will keep calling a service that does not exist. Please tie `RERANKER_ENABLED` and `RERANKER_BASE_URL` to `reranker.enabled` and `externalUrl`, so turning the reranker off actually turns it off.

charts/openrag-stack/values.yaml
```diff
@@ -76,12 +94,22 @@ milvus:
...
   whisperModelName: &whisperModel "openai/whisper-large-v3-turbo"
   llmModelName: &llmModel "RedHatAI/Mistral-Small-3.1-24B-Instruct-2503-quantized.w8a8"
+
+  # Single source for the HF_TOKEN secret name passed to all four model specs.
+  # YAML anchor: editing this value propagates to all four hf_token.secretName entries below.
+  # Must match the chart env secret name ({{ .Release.Name }}-env-secrets by default).
+  # values.yaml is not templated, so set this manually when your release is not named "rag":
+  #   --set "vllm.servingEngineSpec.modelSpec[0].hf_token.secretName=<release>-env-secrets"
+  #   (repeat for indices 1, 2, 3 — see NOTES after install for the exact commands)
+  hfTokenSecretName: &vllmHfTokenSecretName "rag-env-secrets"
```
EnjoyBacon7, line 132: **vLLM looks for the wrong secret name unless the release is called "rag".**

`hfTokenSecretName` is fixed to `rag-env-secrets`, but the chart now names its secret `<release>-env-secrets`. So if the release is not called `rag`, the vLLM pods look for a secret that is not there and cannot read `HF_TOKEN`. The same hardcoded `rag-env-secrets` shows up four times in `values-linagora.yaml`. The post-install NOTES warn about this, but it is easy to miss. At minimum the comment should make clear you must change this to match your release name.

charts/openrag-stack/values.yaml
```diff
@@ -11,27 +13,43 @@ networkPolicy:
...
     count: 1
     replicas: 1
     minReplicas: 1
     maxReplicas: 1
-  nodeSelector:
-    gpu-role: serving
+  serviceAccountName: ""
+  podSecurityContext: {}
+  nodeSelector: {}
+  resources: {}  # e.g. limits: {"nvidia.com/gpu": 1}
```
EnjoyBacon7: **The ray block sets `resources` twice.**

Line 40 has `resources: {}` with the GPU example comment, and a few lines down there is the real `resources` with head and worker. YAML keeps only the last one, so this line and its comment do nothing, and stricter linters will fail on the repeated key. Please remove this empty `resources: {}`.

charts/openrag-stack/templates/secrets-env.yaml
```diff
@@ -1,9 +1,75 @@
...
+  refreshAfter: {{ .Values.env.secretsProvider.vaultStaticSecret.refreshAfter | quote }}
+  hmacSecretData: {{ .Values.env.secretsProvider.vaultStaticSecret.hmacSecretData }}
+  destination:
+    create: true
+    name: {{ $secretName }}
+  rolloutRestartTargets:
+    - kind: Deployment
+      name: {{ .Release.Name }}-openrag
+    - kind: Deployment
+      name: {{ .Release.Name }}-reranker
```
EnjoyBacon7: **Vault restart target points at a reranker that may not exist.**

The indexer-ui restart target below is wrapped in `if .Values.indexerUi.enabled`, but this reranker target is not. When `reranker.enabled` is false this still lists `<release>-reranker`, a Deployment that was never created, and the Vault operator will try to restart something that is not there. Please wrap this reranker target in `if .Values.reranker.enabled` the same way.

charts/openrag-stack/values.yaml
```diff
@@ -261,27 +296,107 @@ openrag:
...
+proxy:
+  enabled: false
+  url: "http://192.168.100.100:80"
+  noProxy: "localhost,127.0.0.1,.cluster.local,.svc,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16"
+
 # === Shared env (config + secrets) ===
 env:
+  # Name of an existing K8s Secret to use as-is.
+  # When set, the chart creates no Secret/ExternalSecret/VaultStaticSecret resource.
+  existingSecret: ""
```
EnjoyBacon7, line 433: **App and database passwords can drift apart with external secrets.**

When you use `externalSecret`, `vaultStaticSecret`, or set `existingSecret`, the chart stops creating the env secret, so the app gets `POSTGRES_PASSWORD` from your external source. But the Postgres server is still started from `postgresql.auth.password`. The two values are no longer the same, so the app cannot log in to the database. Please document that in these modes you also have to point Postgres at the same secret (`postgresql.auth.existingSecret`) or set `postgresql.auth.password` to match.

charts/openrag-stack/templates/secrets-env.yaml
```diff
@@ -1,9 +1,75 @@
...
 type: Opaque
 stringData:
 {{- range $key, $value := .Values.env.secrets }}
-  {{ $key }}: "{{ $value }}"
+  {{ $key }}: {{ $value | quote }}
+{{- end }}
+
+{{- else if eq $type "externalSecret" }}
+---
+apiVersion: external-secrets.io/v1
```
EnjoyBacon7, line 22: **External Secrets API version.**

This ExternalSecret uses `external-secrets.io/v1`. That version only exists in newer External Secrets Operator releases (0.14 and up). On older installs only `v1beta1` is served and the apply will fail. Worth noting a minimum operator version in the docs.

EnjoyBacon7 (commented):
Nice. A few issues I picked up:

Ahmath-Gadji (commented):
Did a full pass over the chart changes. The move to release-scoped naming and configurable securityContexts is a good direction. My main concern is that the rename from the hardcoded `rag-` prefix to the `fullname` helper is only half-applied in a few places, which breaks things for any release not named `rag`, plus a couple of ingress/secret wiring issues. Notes inline, grouped roughly by impact.

Likely blockers:
- Ray worker PVC names still use `.Release.Name`, so workers can't find their PVCs and stay Pending
- Ray ingress points at the unauthenticated dashboard port (8265) and is enabled by default
- openrag and ray ingresses are both default-on with the same host and path
- vLLM `hf_token` secret name hardcoded to `rag-env-secrets`
- dead top-level `ingress:` block in values-linagora.yaml
- `reranker.externalUrl` is documented but not implemented
- PVCs lost the `resource-policy: keep` annotation, so uninstall now deletes the data
- GPU resource request dropped from the Ray cluster

Smaller items and nits are inline too. Happy to help with any of these.

charts/openrag-stack/templates/raycluster.yaml
```diff
@@ -127,23 +154,57 @@ spec:
...
               resources:
-                limits:
-                  nvidia.com/gpu: 1
+                {{- toYaml . | nindent 16 }}
+              {{- end }}
           volumes:
             - name: model-weights
               persistentVolumeClaim:
-                claimName: rag-model-weights
+                claimName: {{ $.Release.Name }}-model-weights
```
Ahmath-Gadji: These worker PVC claim names use `{{ $.Release.Name }}`, but the PVCs themselves and the head group just above both use the `fullname` helper. So for a release named `rag`, these resolve to `rag-model-weights` while the actual PVC is `rag-openrag-stack-model-weights`, and the workers can't mount it (they stay stuck Pending). Looks like the rename landed everywhere except these four. Could we switch them to `{{ include "openrag-stack.fullname" $ }}-...` so they match the head group?

charts/openrag-stack/templates/raycluster.yaml
```diff
@@ -127,23 +154,57 @@ spec:
...
+  rules:
+    - http:
+        paths:
+          - path: {{ .Values.ray.ingress.path }}
+            pathType: {{ .Values.ray.ingress.pathType }}
+            backend:
+              service:
+                name: {{ include "openrag-stack.fullname" . }}-raycluster-head-svc
+                port:
+                  number: {{ .Values.env.config.RAY_DASHBOARD_PORT }}
```
Ahmath-Gadji, line 150: Heads up, this points the ingress at the Ray dashboard port (8265), which has no auth and lets anyone submit jobs (effectively RCE on the cluster). The old ingress routed to the Serve port, not the dashboard. Since `ray.ingress.enabled` defaults to true, this would publish the dashboard on any cluster where the NetworkPolicy isn't enforced. Can we default `ray.ingress.enabled` to false and avoid exposing 8265 here?

charts/openrag-stack/values.yaml
```diff
@@ -11,27 +17,58 @@ networkPolicy:
...
+    head:
+      requests:
+        cpu: "4"
+        memory: "8Gi"
+    worker:
+      requests:
+        cpu: "4"
+        memory: "8Gi"
+  ingress:
+    enabled: true
```
Ahmath-Gadji: With both `openrag.ingress` and `ray.ingress` defaulting to enabled, and both using `host: ""` and `path: /`, they'll collide on the same host and path on a single ingress controller. Might be worth defaulting one of them off, or requiring a host before they render.

charts/openrag-stack/values.yaml
```diff
@@ -76,12 +113,22 @@ milvus:
...
   whisperModelName: &whisperModel "openai/whisper-large-v3-turbo"
   llmModelName: &llmModel "RedHatAI/Mistral-Small-3.1-24B-Instruct-2503-quantized.w8a8"
+
+  # Single source for the HF_TOKEN secret name passed to all four model specs.
+  # YAML anchor: editing this value propagates to all four hf_token.secretName entries below.
+  # Must match the chart env secret name ({{ include "openrag-stack.fullname" . }}-env-secrets by default).
+  # values.yaml is not templated, so set this manually when your release is not named "rag":
+  #   --set "vllm.servingEngineSpec.modelSpec[0].hf_token.secretName=<release>-env-secrets"
+  #   (repeat for indices 1, 2, 3 — see NOTES after install for the exact commands)
+  hfTokenSecretName: &vllmHfTokenSecretName "rag-env-secrets"
```
Ahmath-Gadji, line 132: Now that the env secret is named per release (`<release>-env-secrets`), this hardcoded `rag-env-secrets` only matches when the release is literally named `rag`. For any other name the vLLM pods can't find the HF token secret and gated model pulls fail. It used to work for any release name before the secret got renamed. The NOTES output already detects this and prints a workaround, but it would be nicer to wire the real secret name in here so it just works. Could we template it from the release name?

charts/openrag-stack/values-linagora.yaml
```diff
@@ -0,0 +1,133 @@
...
+          maxModelLen: 8192
+          extraArgs: ["--trust-remote-code"]
+        shmSize: "20Gi"
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: gpu-role
+                operator: In
+                values: ["vlm"]
+
+ingress:
```
Ahmath-Gadji: I don't think this top-level `ingress:` block is read anywhere anymore. The chart moved to per-service keys (`openrag.ingress.*` and friends), so the nginx class, the rewrite, and the 500m body-size limit here are silently ignored. Looks like it needs to move under `openrag.ingress.{className,annotations}`. Worth dropping the `/$1` rewrite too, since the routes now use `path: /` with no capture group.

charts/openrag-stack/values.yaml
```diff
@@ -182,25 +218,23 @@ vllm:
...
-                operator: In
-                values: ["vlm"]
+        nodeSelectorTerms: []
   routerSpec:
     enableRouter: false
 
 # === Infinity reranker ===
 reranker:
+  # Set to false to skip deploying the reranker and use an external one instead.
+  # Configure externalUrl to point to the external reranker endpoint.
```
Ahmath-Gadji, line 230: This comment tells operators to set `externalUrl` for an external reranker, but there's no `externalUrl` key used anywhere, and `RERANKER_BASE_URL` still hardcodes the in-cluster service. So setting `reranker.enabled: false` just points the app at a service that no longer exists, and reranking breaks. Could we either implement the override (and make `RERANKER_BASE_URL` conditional on `reranker.enabled`) or update the comment to the real mechanism?

charts/openrag-stack/templates/pvc.yaml
```diff
@@ -2,57 +2,72 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: rag-model-weights
+  name: {{ include "openrag-stack.fullname" . }}-model-weights
+  {{- with .Values.persistence.annotations }}
```
Ahmath-Gadji, line 6: Dropping the `helm.sh/resource-policy: keep` annotation means a `helm uninstall` now deletes these PVCs (model weights, data, logs, venv) and the data with them on most storage classes. Was that intended? It's probably safer to keep it as a default on the data volumes so an uninstall doesn't wipe the model cache.

charts/openrag-stack/templates/raycluster.yaml
```diff
@@ -67,25 +85,26 @@ spec:
                 mountPath: /app/.venv
             envFrom:
               - configMapRef:
-                  name: rag-env
+                  name: {{ include "openrag-stack.fullname" . }}-env
               - secretRef:
-                  name: rag-env-secrets
+                  name: {{ include "openrag-stack.secretName" . }}
+            {{- with .Values.ray.resources.head }}
```
Ahmath-Gadji, line 63: The previous spec hardcoded `nvidia.com/gpu: 1` on the head and workers. Now that resources come from `ray.resources` and only set cpu/memory, the GPU request is gone, and there's no memory limit either. GPU workloads would quietly schedule onto CPU nodes. Should we add a configurable GPU resource (and a memory limit) back?

charts/openrag-stack/templates/openrag.yaml
```diff
@@ -53,33 +76,108 @@ spec:
...
+  ingressClassName: {{ .Values.openrag.ingress.className | quote }}
+  {{- end }}
+  rules:
+    - http:
+        paths:
+          - path: {{ .Values.openrag.ingress.path }}
+            pathType: {{ .Values.openrag.ingress.pathType }}
+            backend:
+              service:
+                name: {{ include "openrag-stack.fullname" . }}-openrag
```
Ahmath-Gadji, line 146: Quick question on this backend. With `ENABLE_RAY_SERVE: true` (the default), the comment in values says the HTTP proxy runs on the Ray head rather than this pod, and the old ingress routed to the head's Serve port for that reason. Does pointing at `-openrag:8080` still reach the API in Serve mode, or should this target the head service?

charts/openrag-stack/templates/_helpers.tpl
```diff
@@ -51,12 +51,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+When env.existingSecret is set, that name is returned directly.
 */}}
-{{- define "openrag-stack.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create }}
-{{- default (include "openrag-stack.fullname" .) .Values.serviceAccount.name }}
-{{- else }}
-{{- default "default" .Values.serviceAccount.name }}
-{{- end }}
+{{- define "openrag-stack.secretName" -}}
+{{- .Values.env.existingSecret | default (printf "%s-env-secrets" .Release.Name) }}
```
Ahmath-Gadji: Small thing: this helper uses `.Release.Name` directly instead of the `fullname` helper the rest of the chart uses, and it skips the `trunc 63` the other names have. It's basically why the secret name diverges from the configmap name, and it ties into the vLLM secret mismatch above. Might be cleaner to base it on `fullname` for consistency.

charts/openrag-stack/templates/secrets-env.yaml
```diff
@@ -1,9 +1,75 @@
...
+  refreshAfter: {{ .Values.env.secretsProvider.vaultStaticSecret.refreshAfter | quote }}
+  hmacSecretData: {{ .Values.env.secretsProvider.vaultStaticSecret.hmacSecretData }}
+  destination:
+    create: true
+    name: {{ $secretName }}
+  rolloutRestartTargets:
+    - kind: Deployment
+      name: {{ include "openrag-stack.fullname" . }}-openrag
+    - kind: Deployment
+      name: {{ include "openrag-stack.fullname" . }}-reranker
```
Ahmath-Gadji: These rollout-restart targets look inverted. The reranker only mounts the ConfigMap (not this secret), while the Ray head and workers do mount the secret via secretRef but aren't listed here. So on a Vault rotation the wrong workloads restart. This reranker entry also isn't gated on `reranker.enabled`, unlike the indexer-ui one just below.

charts/openrag-stack/templates/indexer-ui.yaml
```diff
@@ -2,44 +2,54 @@
...
       labels:
         app.kubernetes.io/name: indexer-ui
-        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/instance: {{ include "openrag-stack.fullname" . }}
     spec:
+      {{- with .Values.indexerUi.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.indexerUi.serviceAccountName }}
```
Ahmath-Gadji: nit: `indexerUi.serviceAccountName` is referenced here but never declared in values.yaml, so it isn't really settable or documented. Worth adding it under the `indexerUi` block.

charts/openrag-stack/templates/configmap-env.yaml
```diff
@@ -1,8 +1,16 @@
...
 kind: ConfigMap
 metadata:
-  name: rag-env
+  name: {{ include "openrag-stack.fullname" . }}-env
 data:
 {{- range $key, $value := .Values.env.config }}
   {{ $key }}: "{{ tpl (printf "%v" $value) $ }}"
 {{- end }}
+{{- if .Values.proxy.enabled }}
+  HTTP_PROXY: {{ .Values.proxy.url | quote }}
```
Ahmath-Gadji, line 19: nit: if the proxy URL ever includes credentials, they'd end up in this plaintext ConfigMap. And if someone also sets `HTTP_PROXY` in `env.config`, you'd get a duplicate key in the same ConfigMap. Fine for the default, just flagging it.

charts/openrag-stack/Chart.yaml
```diff
@@ -1,39 +1,36 @@
...
-# This is the chart version. This version number should be incremented each time you make changes
-# to the chart and its templates, including the app version.
-# Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
-
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Versions are not expected to
-# follow Semantic Versioning. They should reflect the version the application is using.
-# It is recommended to use it with quotes.
+version: 0.5.0
```
Ahmath-Gadji: nit: while bumping the chart version here, it would be a good moment to bump `appVersion` too. It's still `1.1.3` while the default images are on `v1.1.12`.

charts/openrag-stack/templates/NOTES.txt
```diff
@@ -0,0 +1,29 @@
...
+
+⚠  vLLM HF_TOKEN secret mismatch — action required
+---------------------------------------------------
+This release creates the env secret as: {{ $secretName }}
+The vLLM sub-chart values still reference: {{ $defaultVllmSecret }}
+
+vLLM serving pods will fail to mount HF_TOKEN unless you override the four
+hf_token.secretName entries. Re-run the install/upgrade with:
+
+  helm upgrade {{ .Release.Name }} ./openrag-stack --reuse-values \
```
Ahmath-Gadji, line 15: nit: this hardcodes `./openrag-stack` as the chart path, which won't be right for anyone installing from a repo (for example `helm install ... linagora/openrag-stack`).

charts/openrag-stack/templates/extra-objects.yaml
```diff
@@ -0,0 +1,4 @@
+{{- range .Values.extraObjects }}
```
Ahmath-Gadji, line 1: nit: an empty or null entry in `extraObjects` would render as `---` followed by `null` and fail to apply. A `{{- if . }}` guard inside the range would cover it.

charts/openrag-stack/templates/indexer-ui.yaml
```diff
@@ -2,44 +2,54 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Release.Name }}-indexer-ui
+  name: {{ include "openrag-stack.fullname" . }}-indexer-ui
   labels:
     app.kubernetes.io/name: indexer-ui
-    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/instance: {{ include "openrag-stack.fullname" . }}
```
Ahmath-Gadji: nit: `app.kubernetes.io/instance` is set to the fullname here, but the shared `selectorLabels` helper uses `.Release.Name`. Minor label inconsistency across the stack (same pattern in infinity.yaml).

andyne13:
Bonjour @ThibautChoppy 👋

Pour préparer le déploiement K8s/OpenShift de la v2.0, j'aimerais m'appuyer sur ton travail sur le chart Helm. Est-ce que cette PR (#456) correspond bien à ta version de travail la plus récente du chart, ou est-ce que tu as une version plus à jour (sur une autre branche) que tu pourrais partager ?

Merci beaucoup !

charts/openrag-stack/templates/openrag.yaml
ThibautChoppy: Good point, I'm gonna fix that

andyne13:
Merci @ThibautChoppy pour le travail sur le chart 🙏

Pour avancer sur le déploiement K8s/OpenShift de la v2.0 : **les images `v2.0.0-rc.1` sont publiées** (construites via le workflow `build_rc.yml`) et disponibles —

- `ghcr.io/linagora/openrag:v2.0.0-rc.1` (api) — aussi `linagoraai/openrag:v2.0.0-rc.1` sur Docker Hub
- `ghcr.io/linagora/openrag-ray:v2.0.0-rc.1`
- `ghcr.io/linagora/openrag-admin-ui:v2.0.0-rc.1` — aussi `linagoraai/openrag-admin-ui:v2.0.0-rc.1`

Jaimerais quon sappuie sur ta PR #456 comme base pour le chart v2. Le chart actuel est encore en v1.x (`appVersion 1.1.13`, UI `indexer-ui`) ; ce quil reste à couvrir côté v2 :

1. **Tags images → `v2.0.0-rc.1`** : `ray` (registre/repo déjà bons, juste le tag à bumper) et `openrag` (api), + bump `version`/`appVersion` du chart.
2. **Nouvelle UI `admin-ui`** : elle remplace `indexer-ui`. Il faut un Deployment + Service pour `openrag-admin-ui` (limage ci-dessus) — cest une image et une UI différentes d`indexer-ui`. Petit point dattention : `VITE_BASE_PATH=/app/` est *compilé dans limage* au build (Vite), donc limage rc.1 suppose un service sous `/app/` en same-origin ; si on sert lUI ailleurs, il faudra rebuilder avec dautres build-args.
3. **Routage** : tu as retiré l`ingress.yaml` au profit de `extraObjects` (BYO) 👍. Du coup il faut quon décide ensemble la topologie v2 — front-door nginx `admin-ui` qui sert le SPA sous `/app/` + proxy vers `openrag:8080` (façon compose), vs Ingress/Gateway par chemin vers les Services api/admin-ui séparés. Ça impacte `/`, `/app/`, `/v1`, `/auth`, `/chainlit`, le SSE et `X-Forwarded-Proto` (OIDC).

Un détail à garder en tête : ta PR cible `charts/openrag-stack` sur `main`, alors que la v2 (branche `refactor/hexagonal`) a le chart sous `infra/charts/openrag-stack` — il faudra réconcilier les deux à un moment.

Est-ce que #456 est bien ta version de travail la plus à jour du chart ? Si oui, jai un récap écrit des deltas v2 que je peux partager, et on cale la suite ensemble. Merci !

PR #459, opened 2026-06-11
chore: run the OpenRAG container and service stack as non-root users

EnjoyBacon7:
## What

Make the OpenRAG Docker stack run **unprivileged** instead of as root. Two commits:

### 1. App image (`Dockerfile`)
The `openrag` image had no `USER`, so the container ran as root. It only writes under `/app` and binds non-privileged ports, so:
- Add a dedicated `openrag` user (uid/gid `1000`, overridable via `APP_UID`/`APP_GID`) that owns `/app`, with `USER` before the entrypoint.
- Relocate uv's managed Python/cache out of root's `$HOME` so the root-built artifacts stay reachable for the runtime user.

### 2. Compose stack (`docker-compose.yaml`, `vdb/milvus.yaml`, `extern/reranker/infinity.yaml`)
Every other service ran as root too. Now:
- A single one-shot **`init-perms`** (busybox) — the only container that uses root, running no app code — creates the host bind-mount dirs and chowns them to the unprivileged uid/gid. Docker creates bind mounts as root, so this is what lets the non-root services write their volumes. Services wait on it via `depends_on: service_completed_successfully`.
- **postgres** → `999:999` (built-in user); **etcd/minio** → `1000:1000`; **milvus** → `1000:0` (its binary is `root:root 0775`, so it needs group root to exec); **vllm/reranker** → `1000:1000` with `HOME`/`HF_HOME` on a writable mounted cache, plus `USER`/`LOGNAME` so vllm's `getpass.getuser()` doesn't crash on a uid with no `/etc/passwd` entry.

## Verification (all from scratch, non-root)

- **App image:** builds; runs as `uid=1000(openrag)`; uv resolves pinned Python; writes `/app/data`,`/app/logs`,venv. ✅
- **Data layer:** `init-perms → rdb, etcd, minio, milvus` all reach **healthy** (postgres `999`, etcd/minio/milvus `1000`); **every host file is owned by the unprivileged uid — no root-owned data**. ✅
- **Model servers (`--profile cpu`):** **reranker-cpu** and **vllm-cpu** both run as `uid=1000` and serve `/health` **200** — vllm returned a real **1024-dim embedding**; model caches on the host owned by the unprivileged uid. ✅

## Known remaining gap

- **indexer-ui still runs as root** — its entrypoint writes `/app/config.json` into the image's root-owned `/app`, so it can't drop privileges without an image-level fix in the `openrag-admin-ui` repo. Documented inline.

## Notes

- `Dockerfile.ray` (separate Ray-cluster image) is intentionally left as-is.
- uid/gid default to `1000`; deployments whose host user differs override `PUID`/`PGID` (and `APP_UID`/`APP_GID` for the build).

## Summary by CodeRabbit

* **Chores**
  * Services now run unprivileged where possible to improve container security.
  * Added a startup permission-init step so bind mounts, caches and writable dirs get correct ownership before services start.
  * Cache and data mount locations updated so services use writable cache paths at runtime.
  * Service startup ordering adjusted so dependent services wait for permission initialization to complete.
  * Indexer UI still runs as root due to startup constraints.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,17 +56,23 @@ ENV APP_iPORT=${APP_iPORT:-8080}
 
 # --- Run as an unprivileged user -------------------------------------------
 # openrag only ever writes under /app (data, logs, db, the HF model cache and
-# the uv-created venv) and binds non-privileged ports, so it has no need for
-# root. Create a dedicated user that owns those paths. UID/GID are build args
-# so a deployment can match the host user owning the bind-mounted volumes
+# the uv-created venv) plus uv's cache, and binds non-privileged ports, so it
+# has no need for root. Create a dedicated user and give it ownership of *only*
+# those runtime-write paths, plus the /app dir node itself so the runtime
+# `uv run` can create the venv and the editable install's openrag.egg-info at
+# the project root. The application code, config files and uv's managed Python
+# install stay root-owned/read-only (least privilege). UID/GID are build
+# args so a deployment can match the host user owning the bind-mounted volumes
 # (./data, ./logs, ~/.cache/huggingface); the 1000 default fits a typical
 # single-user host.
 ARG APP_UID=1000
 ARG APP_GID=1000
 RUN groupadd --gid ${APP_GID} openrag \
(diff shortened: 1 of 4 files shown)


EnjoyBacon7:
Superceeded by #489

PR #461, opened 2026-06-12
fix: add content when the RAG response is empty

ewan102:
Adding tests in open_ai_chat_completion() and streaming_with_source filtering that verify if the message sent by the llm has any text in it and if not, it had a generic text (I wrote something  but it is just an exemple...) : I do not have the necessary documentation in order to answer your message

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a clear client-facing fallback message when a reply becomes empty after removing source attributions, covering both streaming and non-streaming chat/completions.
  * Improved streaming end-of-response chunking so the final message reliably includes either the remaining content or the fallback, avoiding empty completions.

### DIFF
--- a/openrag/components/utils.py
+++ b/openrag/components/utils.py
@@ -264,21 +264,28 @@ async def stream_with_source_filtering(
             filtered = filter_sources_by_citations(sources, citations)
             filtered_json = json.dumps({"sources": filtered})
 
-            if chunk_template and len(final_clean) > emitted_len:
-                tail_chunk = copy.deepcopy(chunk_template)
-                tail_chunk["choices"][0]["delta"] = {"content": final_clean[emitted_len:]}
-                tail_chunk["extra"] = filtered_json
-                yield f"data: {json.dumps(tail_chunk)}\n\n"
+            def _sse(delta: dict, finish_reason: str | None = None) -> str:
+                """Build an SSE chunk from the template, overriding delta/extra."""
+                chunk = copy.deepcopy(chunk_template)
+                chunk["choices"][0]["delta"] = delta
+                if finish_reason is not None:
+                    chunk["choices"][0]["finish_reason"] = finish_reason
+                chunk["extra"] = filtered_json
+                return f"data: {json.dumps(chunk)}\n\n"
+
+            if chunk_template and not final_clean.strip():
+                # Response was only a [Sources: ...] tag (or empty) — send a default
+                # message so the client never receives an empty completion.
+                logger.warning("LLM and RAG response empty after source stripping, returning fallback message")
+                yield _sse({"content": "I do not have the necessary documentation in order to answer your message"})
+            elif chunk_template and len(final_clean) > emitted_len:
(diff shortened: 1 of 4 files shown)


openrag/components/utils.py
```diff
@@ -264,21 +264,27 @@ async def stream_with_source_filtering(
...
+                chunk["choices"][0]["delta"] = delta
+                if finish_reason is not None:
+                    chunk["choices"][0]["finish_reason"] = finish_reason
+                chunk["extra"] = filtered_json
+                return f"data: {json.dumps(chunk)}\n\n"
+
+            if chunk_template and not final_clean.strip():
+                # Response was only a [Sources: ...] tag (or empty) — send a default
+                # message so the client never receives an empty completion.
+                yield _sse({"content": "I do not have the necessary documentation in order to answer your message"})
```
EnjoyBacon7: This leaves internationalisation pretty tough.

Ahmath-Gadji: Agreed — suggest making this configurable (e.g. a Hydra `llm.empty_response_message` / env var) defaulting to the current string, so deployments can localize. The product already targets non-English users (`WEBSEARCH_LANG` defaults to `fr-FR`).

openrag/routers/openai.py
```diff
@@ -365,6 +366,9 @@ async def stream_response():
         filtered = filter_sources_by_citations(sources, citations)
         chunk["extra"] = json.dumps({"sources": filtered})
         log.debug("Returning non-streaming completion chunk.")
+        if not clean_content.strip():
+            log.warning("LLM and RAG response empty after source stripping, returning fallback message")
+            chunk["choices"][0]["message"]["content"] = EMPTY_RESPONSE_FALLBACK_MESSAGE
```
Ahmath-Gadji, line 371: When the fallback fires, `extra.sources` is still populated (all sources if `citations is None`, or the cited ones if the reply was only a `[Sources: ...]` tag) — so the client gets "I do not have the necessary documentation" alongside a non-empty source list, which is contradictory. Consider clearing the sources here (and in the streaming / `/completions` paths), or softening the message.

openrag/components/utils.py
```diff
@@ -15,6 +15,10 @@
 
 SOURCE_SEPARATOR = "-" * 10 + "\n\n"
 
+# Sent to the client when the LLM produces no usable content (e.g. the whole
+# response was a [Sources: ...] tag), so it never receives an empty completion.
+EMPTY_RESPONSE_FALLBACK_MESSAGE = "I do not have the necessary documentation in order to answer your message"
```
Ahmath-Gadji: The PR description says tests were added for the empty-response handling, but the diff has none. Worth a unit test feeding a stream whose only content is a `[Sources: ...]` tag and asserting the fallback chunk is emitted (and likewise for the two non-streaming paths).

EnjoyBacon7: Not actually in PR.

Ahmath-Gadji:
Solved by 01a13113178.

PR #463, opened 2026-06-12
feat(phase-15): sync Keycloak groups to partition memberships

EnjoyBacon7:
## Phase 15 — OIDC / Keycloak group → partition authorization

Implements the genuinely-missing half of strategy-doc Phase 15: **claim-driven authorization**. On OIDC login, an IdP groups claim is translated into partition memberships and reconciled against the database. SSO login itself already shipped (session-cookie Authorization Code + PKCE), so this phase is **variable uniformization (A)** + **group→partition sync (B)** — *not* a greenfield rebuild.

Full reconciliation plan, intentional deviations from the strategy doc, and design rationale: `docs/refactoring/phase-15.md`.

### Part A — config surface (`e7a3cc04`)
- Four new `OIDC_*` vars on `OIDCConfig`, named to the **shipped** convention (no renames of existing vars, no strategy-doc duplicates):
  - `OIDC_CLAIM_GROUPS` (`""` → feature OFF), `OIDC_GROUP_PREFIX` (`/openrag/`), `OIDC_GROUP_PATTERN` (`(.+)/(owner|editor|viewer)$`), `OIDC_GROUP_SYNC_PRUNE` (`false`).
- `from_env()` validates the pattern compiles + exposes ≥2 capture groups, at startup, regardless of `AUTH_MODE`.

### Part B — group → partition sync (`48be7557`)
- `services/auth/oidc_groups.py`: pure mapper, group → (partition, role) — strip prefix, match pattern, highest role wins per partition.
- `AuthService._sync_oidc_memberships`: called after claim mapping in `handle_oidc_callback`; adds missing, updates changed roles, and — only when `OIDC_GROUP_SYNC_PRUNE` is on — removes stale memberships. Reuses the already-injected membership repo and the verified ID-token claims (no extra IdP round-trip).
- **`is_admin` is never derived from groups** — preserves the documented hard boundary (role→is_admin deferred behind an Open Decision).
- Per-membership `try/except` so an unknown partition (FK violation) is logged + skipped and can never block login.
- **Mass-revocation guard:** prune runs only when the token actually asserts the groups claim. A missing/misconfigured claim is "no signal" and skips prune (an explicit `groups: []` still prunes), so a config typo cannot silently wipe every synced user's access.

### Behavior change
**None unless `OIDC_CLAIM_GROUPS` is set.** Every existing deployment is untouched.

### Tests
Unit coverage for the mapper (prefix/pattern/dedupe/scalar-vs-list/empty short-circuit), the config validator (round-trip, bad-regex, <2 groups, validated in token mode), and the service sync (add/role-change/prune-on-vs-off/unknown-partition isolation + the three guard cases).

### Out of scope (see `docs/refactoring/phase-15.md` → Key considerations)
- Stateless raw-JWT-bearer dual-auth (`jwt_validator`/`_is_jwt`) — redundant; session cookie + `users.token` already serve browser + machine clients. Documented with the implications (no DB lookup ↔ no instant revocation) for if it's ever needed.
- `role → is_admin` mapping — deferred (would breach the `is_admin`-never-claim-writable boundary).

### DIFF
--- a/REFACTORING_DECISION_LOG.md
+++ b/REFACTORING_DECISION_LOG.md
@@ -1399,6 +1399,73 @@ All five CI jobs (Layer guard, Linting, Unit, Integration, API) are
 green on the branch after these fixes.
 
 ---
+## Phase 15 — OIDC group → partition authorization (2026-06-12)
+
+**1. Variable names uniformized onto the shipped `OIDC_*` surface; the
+strategy doc's draft names are not reintroduced.**
+- Why: `REFACTORING_STRATEGY_v1.md` §Phase 15 (and `REFACTORING_DEV_WORKFLOW.md`)
+  sketch a *stateless raw-JWT-bearer* design with its own env vocabulary
+  (`OIDC_ENABLED`, `OIDC_ISSUER_URL`, `OIDC_AUDIENCE`, `OIDC_JWKS_CACHE_TTL`,
+  `VITE_OIDC_*`). The shipped OIDC is the **session-cookie Authorization
+  Code + PKCE** flow (`docs/oidc.md`), which already reads `AUTH_MODE=oidc`,
+  `OIDC_ENDPOINT`, `OIDC_REDIRECT_URI`, … The two never coexisted in code, so
+  Phase 15 adopts the shipped names as canonical and only adds vars for the
+  one genuinely-missing half (group → partition mapping):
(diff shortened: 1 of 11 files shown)


Ahmath-Gadji:
Superseded by https://github.com/linagora/openrag/pull/463 (re-opened under the correct author).

PR #478, opened 2026-06-15
security: address remaining GHSA advisory findings (C3, H2–H8, M2–M13, N3–N13, deps)

EnjoyBacon7:
## Summary

Addresses the remaining open findings across the four draft security advisories (Critical → Low) plus the dependency items. The previously-merged PRs (#466–#476) covered C1, C2, H1, H5, H6, M1, N5 and the websearch SSRF; this PR knocks out the rest, one commit per finding.

## Findings addressed

**Critical / High**
- **C3** — Ray dashboard/Jobs API (unauthenticated, CVE-2023-48022) now binds `127.0.0.1` by default in `api.py`, quick_start and `cluster.yaml` (k8s isolated via the new NetworkPolicy).
- **H2** — SSRF via document image-URL captioning: fetch remote images in-process behind a resolve-then-validate guard (block loopback/RFC1918/link-local/reserved), no redirects, content-type + 20 MB caps, pass only a data URI to the VLM; default `image_captioning_url: false`. Guard extracted to `utils/ssrf.py` and shared with the websearch fetcher.
- **H3** — MinIO creds required via env (`MINIO_ACCESS_KEY`/`MINIO_SECRET_KEY`), no `minioadmin` default; Milvus gets matching creds.
- **H4** — App and Ray images run as non-root (`uid 10001`).
- **H7** — Fail-fast when `CHAINLIT_AUTH_SECRET` is unset (no public-constant fallback; dev opt-in via `ALLOW_NO_AUTH`).
- **H8** — Neutralize `[Source N]` / `[Sources: …]` / separator control tokens in retrieved chunks + web results (prompt-injection / citation forgery).

**Medium** — M2/M3 weak default DB password & AUTH_TOKEN; M6 path-tiered rate limiting (`slowapi`/`limits`); M7 stop leaking stack traces / FS paths; M8 parser-bomb caps (eml/docx/pptx/pdf, incl. a `[None]*max_order` memory bomb); M9 back-channel logout `exp`/`jti` + replay guard; M10 default logs to INFO + drop user queries; M12 enforce token limit in RAG mode + bound `n`/`best_of`; M13 per-user partition cap.

**Low / crypto** — N3 logout CSRF (Fetch-Metadata guard); N6 surrounding-chunk partition scoping; N7 Helm DB password → Secret; N8 no dev bind-mount/`--reload` in prod; N9 Prometheus/exporter exposure + drop `--web.enable-lifecycle`; N10 pin image tags; N11 drop `seccomp:unconfined`; N12 default-deny NetworkPolicy; N13 no token in UI file URLs under OIDC; JWT clock-skew leeway + `nbf`.

**Dependencies** — Starlette → 0.47.3 / FastAPI → 0.116.2 (CVE-2025-54121; transitively bumps Chainlit 2.6.2 → 2.11.1). cairosvg SVG render pinned `unsafe=False` (explicit SSRF/XXE guard).

## Testing
- **Unit suite:** 384 passed, 5 skipped.
- **Live integration suite** (mock-vLLM stack, image built from this branch): 116 passed, 2 skipped. The 1 failure (`test_full_oidc_lifecycle`) is **pre-existing** — reproduces identically on the pre-merge commit (`users.py` calls `task_state_manager.remote()` while the test stubs it `None` under `raise_server_exceptions=True`).
- **Targeted live checks:** M12 `n=100`/`best_of=50` → 422; M6 → 429 + `Retry-After`; M13 non-admin 3rd partition → 403, admin bypass.
- **Docker:** both images build and run as `uid 10001`. **Helm:** chart renders; NetworkPolicy valid, `POSTGRES_PASSWORD` in Secret not ConfigMap.

## Caveats for reviewers
- **Pillow** could NOT be bumped to ≥11.3.0 (CVE-2025-48379): `marker-pdf` caps `pillow<11.0.0` — blocked upstream.
- The **Chainlit 2.6→2.11** bump (forced by the Starlette CVE fix) passes import/unit checks but the **Chainlit UI itself isn't unit-tested** — please smoke-test.
- Helm Postgres password is now a `CHANGE_ME` placeholder that **must** be overridden at install.
- `spire-doc` sandboxing/replacement (untrusted `.doc`) is noted as a follow-up — no safe drop-in.

## Release Notes

* **Security Enhancements**
  * UI session signing now requires a Chainlit auth secret (insecure fallback only when explicitly enabled)
  * Added API rate limiting with per-tier limits and 429 + `Retry-After`
  * Hardened outbound fetching (SSRF blocking) and neutralized untrusted prompt control markers
  * Improved auth hardening (back-channel logout replay/claim validation and safer logout CSRF handling)

* **Reliability & Safety**
  * Added parser-bomb mitigation caps across document ingestion and capped PDF/PPTX/image handling
  * Prevented cross-tenant neighbor leakage and added stricter token-limit enforcement
  * Enforced per-user partition ownership limits

* **Configuration**
  * Pinned app/UI images to `1.1.11`, enabled default-deny NetworkPolicy, and bound dashboards/metrics to localhost
  * Switched to non-root execution; updated env/secrets requirements and default logging to INFO

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,10 @@ VLM_MODEL=
 
 # Set to true, it will mount chainlit chat ui to the fastapi app (Default: true)
 ## WITH_CHAINLIT_UI=true
+# REQUIRED when the Chainlit UI is enabled: secret used to sign UI session
+# cookies. If unset the app refuses to start (unless ALLOW_NO_AUTH=true, dev
+# only). Generate with: python -c "import secrets; print(secrets.token_hex(32))"
+# CHAINLIT_AUTH_SECRET=
 
 # UI language for Chainlit and Indexer UI (e.g. en-US, fr).
 # When unset, the browser language is used, with en-US as the final fallback.
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,12 @@ RERANKER_MODEL=Alibaba-NLP/gte-multilingual-reranker-base # or jinaai/jina-reran
 # API accepts natively. Override to restrict (e.g. to ".wav" only for vLLM deployments).
 # TRANSCRIBER_DIRECT_UPLOAD_SUFFIXES=.wav|.flac|.ogg|.mp3|.mp4|.m4a|.webm|.mpeg|.mpga
 
+# Object storage (MinIO, used by Milvus) — REQUIRED, no default.
+# Generate strong random values, e.g. `openssl rand -hex 16`. These are shared
+# between the minio service and Milvus; both must match.
+MINIO_ACCESS_KEY=
+MINIO_SECRET_KEY=
+
 # Prompts
 PROMPTS_DIR=../prompts/example1
 
--- a/.env.example
+++ b/.env.example
@@ -98,6 +98,12 @@ API_BASE_URL='http://X.X.X.X:APP_PORT'          # Base URL of your FastAPI backe
 # WEBSEARCH_TOP_K=5         # Number of web results to include (default: 5)
 # WEBSEARCH_LANG=fr-FR      # Search language/market (default: fr-FR)
 
+# Rate limiting (per-worker moving window, keyed on user id then client IP)
+# RATE_LIMIT_ENABLED=true
+# RATE_LIMIT_DEFAULT=300/minute   # all paths except those below
+# RATE_LIMIT_AUTH=20/minute       # /auth/* (login/callback/logout)
+# RATE_LIMIT_CHAT=60/minute       # /v1/* (chat completions, tools)
+
 # LOGGING
 LOG_LEVEL=DEBUG # See possible values https://loguru.readthedocs.io/en/stable/api/logger.html
 
--- a/.env.example
+++ b/.env.example
@@ -105,7 +105,8 @@ API_BASE_URL='http://X.X.X.X:APP_PORT'          # Base URL of your FastAPI backe
 # RATE_LIMIT_CHAT=60/minute       # /v1/* (chat completions, tools)
 
 # LOGGING
-LOG_LEVEL=DEBUG # See possible values https://loguru.readthedocs.io/en/stable/api/logger.html
+# INFO by default; DEBUG persists user queries and request data to logs.
+LOG_LEVEL=INFO # See possible values https://loguru.readthedocs.io/en/stable/api/logger.html
 
 # SERVER
 # Set the preferred URL scheme for generated URLs (e.g., task_status_url).
--- a/.env.example
+++ b/.env.example
@@ -27,9 +27,8 @@ VLM_MODEL=
 
 # Set to true, it will mount chainlit chat ui to the fastapi app (Default: true)
 ## WITH_CHAINLIT_UI=true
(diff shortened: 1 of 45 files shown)


EnjoyBacon7:
Thanks for the review — all four findings addressed and folded into their originating commits:

- **`rate_limit.py` (Critical)** — `request.state.user` is a dict, so `getattr(user, "id")` always returned `None` and every request fell back to IP keying. Switched to `user.get("id")` with an `isinstance(user, dict)` guard, and added unit tests asserting `_identity` keys on `user:<id>` (and falls back to IP when unauthenticated).
- **`eml_loader.py` (Major)** — moved the per-email attachment cap check **before** `get_payload(decode=True)`, so overflow attachments are no longer decoded.
- **`docx.py` (Major)** — `order_num` is parsed from the untrusted filename; added a `order_num < 1` guard so a non-positive index can't mis-index `images[pos-1]` or raise.
- **`entrypoint.sh` (Minor, SC2086)** — converted `ENV_ARG`/`RELOAD_ARG` to bash arrays and quoted `${APP_iPORT}` / `${API_NUM_WORKERS}` to avoid argv splitting.

Also fixed the failing `lint` check (`ruff format` on `app_front.py`). Lint, unit suite, and the live integration run all pass.

EnjoyBacon7:
Follow-up: addressed the remaining CodeRabbit findings from the review body (outside-diff + nitpicks), folded into the originating commits:

- **`docs/assets/compose_ollama_cpu.yaml` (Major, C3)** — bound the Ray dashboard port to `127.0.0.1:${RAY_DASHBOARD_PORT:-8265}:8265` for consistency with the main compose.
- **`marker.py` (Major, M8)** — the single-chunk fast path passed `page_range=None`, which bypassed `max_pdf_pages`. It now passes the chunk's explicit range when the page count is capped (and keeps `None` when uncapped).
- **`base.py` (M8/H2 nitpick)** — the remote-image fetch now streams via `resp.aiter_bytes()` and aborts once `MAX_REMOTE_IMAGE_BYTES` is exceeded, instead of buffering the whole body before the size check.
- **`pyproject.toml` (nitpick)** — dropped the unused `slowapi` dependency and declared `limits` directly (the rate limiter imports from `limits`).
- **`app_front.py`** — formatting fixed earlier (`ruff format`); `lint` is green.

Lint, unit suite (303 passed), and imports all pass.

EnjoyBacon7:
Addressed the two findings from the latest review:

- **`partition.py` (Major — TOCTOU race)** — the cap check no longer happens in the router against a request-state snapshot. The owner-count check and the insert now run in a single session inside the `Vectordb` actor (`create_partition(..., max_owned=...)`), which serializes calls, so concurrent creates can't race past the limit. The actor returns `created` / `exists` / `limit_exceeded`, mapped to 201 / 409 / 403 (this also folds the previous separate `partition_exists` check into the atomic path).
- **`entrypoint.sh` (Major)** — `--reload` is incompatible with multiple workers; when `UVICORN_RELOAD=true` the worker count is now forced to `1`.

Folded into the M13 and N8 commits. Lint, unit suite, and the vectordb tests pass.

PR #494, opened 2026-06-16
fix(embedder): avoid vLLM exact-max_model_len hang; tune batch_size default

EnjoyBacon7:
## Problem

When indexing certain files, embedding completes every batch **except one**, which hangs indefinitely — systematically reproducible for the same file, while other files index fine in parallel (the server keeps serving). So it's **content-related**, not load or client.

Root cause: vLLM issue [#29496](https://github.com/vllm-project/vllm/issues/29496) — Qwen3-Embedding (and other vLLM pooling models) **hang forever on an embedding request whose input is exactly `max_model_len` tokens long**. Reducing the input by a single token fixes it; the issue was closed "not planned".

Our embedder sent `truncate_prompt_tokens = max_model_len`, so **any chunk ≥ `max_model_len` was truncated straight onto that hang boundary**, wedging that one batch while the rest of the document (and other files) embedded normally.

## Changes

1. **Truncate one token below the boundary.** `_embed_batch` now sends `truncate_prompt_tokens = max(1, max_model_len - 1)` so a truncated chunk lands at `max_model_len - 1`, never on the hang boundary. Applies to every `VLLMEmbedder` (default and named-endpoint paths). Floored at 1 for safety.

2. **Lower the default embedder `batch_size` 64 → 32** across the three sources of truth (`conf/config.yaml`, `EmbedderConfig`, `VLLMEmbedder` constructor). Smaller batches shrink each request and the blast radius of any single slow/stuck batch, and align the legacy embedder default with the named-endpoint `ModelEndpointConfig` default (also 32). `timeout` (120s) and `embed_concurrency` (4) are unchanged — already reasonable; all three remain overridable via `EMBEDDER_TIMEOUT` / `EMBEDDER_BATCH_SIZE` / `EMBEDDER_CONCURRENCY`.

## Tests

- `test_truncate_prompt_tokens_is_one_below_max_model_len` (8192 → 8191) and a new `test_truncate_prompt_tokens_floors_at_one`.
- Full inference / config / di unit suites pass (280 passed); ruff clean.

## Release Notes

* **Bug Fixes**
  * Fixed prompt truncation logic to avoid truncating at exact model length boundaries, improving embedding stability.

* **Chores**
  * Reduced embedding request batch size from 64 to 32 chunks for optimized memory usage.

* **Tests**
  * Updated truncation validation tests to verify improved boundary handling.

### DIFF
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -47,7 +47,7 @@ embedder:
   api_key: EMPTY
   max_model_len: 8192
   timeout: 120.0       # per-request HTTP timeout (s); raise for slow remote endpoints
-  batch_size: 64       # chunks per embedding request (big docs are split into batches)
+  batch_size: 32       # chunks per embedding request (big docs are split into batches)
   embed_concurrency: 4 # max embedding requests in flight at once
 
 # --- Vector Database (Milvus) ---
--- a/openrag/core/config/endpoints.py
+++ b/openrag/core/config/endpoints.py
@@ -47,7 +47,7 @@ class EmbedderConfig(ConfigMixin):
     timeout: float = Field(default=120.0, gt=0)
     # Big documents are embedded in slices of `batch_size`, at most
     # `embed_concurrency` requests in flight, to stay within the timeout above.
-    batch_size: int = Field(default=64, gt=0)
+    batch_size: int = Field(default=32, gt=0)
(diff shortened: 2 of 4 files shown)


Ahmath-Gadji:
Reopening under the correct author account (Ahmath-Gadji). Superseded by a new PR from the same branch.

PR #496, opened 2026-06-16
fix(workers): preserve Whisper GPU allocation from Ray resources

Ahmath-Gadji:
## The bug

`WhisperPool` silently runs Whisper transcription **on CPU even when the node has a GPU**, making audio/video indexing far slower than it should be.

## Root cause

Same failure mode as the Marker GPU regression (#451 / #452), now in the Whisper path.

`WhisperPool` is created with `num_gpus=0` (it's just a dispatcher — see `bootstrap.py`). Ray therefore clears `CUDA_VISIBLE_DEVICES` inside the pool actor's process, so `torch.cuda.is_available()` returns `False` there *even on a GPU node*.

`_whisper_num_gpus()` asked that question from inside the pool actor before deciding how many GPUs to request for the real `WhisperActor` workers:

```python
return config.loader.local_whisper.whisper_num_gpus if torch.cuda.is_available() else 0
```

Because CUDA is hidden in the pool process, this returned `0`, so the `WhisperActor`s were scheduled with `num_gpus=0` and loaded the model on CPU (`device = "cuda" if torch.cuda.is_available() else "cpu"` → `cpu`). Ray never reserved a GPU. The API still worked, so it looked like a performance issue rather than a misconfiguration.

The wrong question was *"can this pool actor see CUDA?"* — it should be *"does the Ray cluster have GPU capacity Whisper should request?"*

## The fix

`_whisper_num_gpus()` now decides from **Ray cluster resources**, mirroring `_marker_num_gpus()`:

- configured request `<= 0` → stays on CPU (explicit opt-out preserved)
- Ray reports GPU capacity → keep the configured GPU request (e.g. `0.25`)
- Ray lookup fails → log a warning and fall back to the old `torch.cuda.is_available()` check

Adds a regression test covering the hidden-CUDA case plus the config-disabled and Ray-lookup-failure fallbacks.

## Release Notes

* **Bug Fixes**
  * Improved GPU resource detection to increase reliability of GPU allocation.

* **Tests**
  * Added comprehensive test coverage for GPU selection behavior across different system configurations.

### DIFF
--- a/openrag/services/workers/parsers/whisper_workers.py
+++ b/openrag/services/workers/parsers/whisper_workers.py
@@ -20,7 +20,23 @@ logger = get_logger()
 
 
 def _whisper_num_gpus(config) -> float:
-    return config.loader.local_whisper.whisper_num_gpus if torch.cuda.is_available() else 0
+    """Return Whisper's Ray GPU reservation, falling back to CUDA detection.
+
+    Mirrors Marker's GPU selection (see ``_marker_num_gpus``): the
+    ``WhisperPool`` actor runs with ``num_gpus=0``, so Ray hides CUDA in its
+    process and ``torch.cuda.is_available()`` returns False even on a GPU
+    node — which would silently pin the ``WhisperActor``s to CPU. Ask whether
+    the Ray cluster has GPU capacity instead, keeping the CUDA check only as a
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
@coderabbit review

PR #508, opened 2026-06-18
fix(partitions): tolerate missing Milvus collection on delete

### LINKED ISSUE #505: delete_partition returns 500 when the Milvus collection doesn't exist yet (empty partition / fresh stack)
## The bug

Deleting a partition that has never had a document indexed returns **HTTP 500**.

Partitions are created as a pure Postgres operation (`POST /partition/{p}` → `PartitionService.create_partition`), independent of indexing. The shared Milvus collection (`config.vectordb.collection_name`, e.g. `vdb_test`) is created **lazily on the first document insert**. So an empty partition — or any partition on a stack where nothing has ever been indexed — has no backing Milvus collection.

## Root cause

`PartitionService.delete_partition` (`openrag/services/orchestrators/partition_service.py`) deletes the relational rows, then **unconditionally** queries the vector store to find chunks to clean up:

```python
await self._partition_repo.delete_partition(name=partition)
ids = await self._vector_store.query_ids_by_filter(self._collection, {"partition": partition})
```

When the collection doesn't exist, `query_ids_by_filter` → Milvus `query_iterator` → `describe_collection` raises `DescribeCollectionException`. That's unhandled (the existing `try/except` only wraps the subsequent `delete()`), so it surfaces as a 500 — **after** the relational rows have already been removed.

## Reproduction

1. Fresh stack, nothing indexed.
2. `POST /partition/demo` (or create a partition in the admin UI).
3. `DELETE /partition/demo` before uploading any document → **500**.

## Fix

Guard the best-effort vector cleanup with the existing `VectorStore.collection_exists()` port method. If the collection doesn't exist there are no vectors to clean up (DB rows already removed), so log and return early. An explicit existence check — rather than swallowing the exception — avoids masking real Milvus errors (e.g. a connection failure) as "nothing to clean up".

Add a regression test: `delete_partition` with `collection_exists() == False` still removes the relational rows and never queries/raises on the missing collection.

Targets the `refactor/hexagonal` branch.

andyne13:
## What

`delete_partition` returns HTTP 500 when deleting a partition on a stack where nothing has ever been indexed.

Fixes #505.

## Root cause

Partitions are a pure Postgres concept; the shared Milvus collection (`config.vectordb.collection_name`) is created **lazily on the first insert system-wide**. `delete_partition` removed the relational rows and then **unconditionally** queried the vector store for chunks to clean up. When the collection doesn't exist, `query_ids_by_filter` → `query_iterator` → `describe_collection` raises `DescribeCollectionException`, surfacing as a 500 *after* the DB rows were already deleted.

(Note: "partition" is a scalar field on the single shared collection, not a Milvus-native partition — so this only triggers when the collection itself is absent, i.e. nothing indexed yet. With data present, the cleanup query just returns empty.)

## Fix

Guard the best-effort vector cleanup with the existing `VectorStore.collection_exists()` port method. No collection → no chunks to clean up → log and return. An explicit existence check (rather than swallowing the exception) avoids masking real Milvus errors like a connection failure.

## Test

Adds `test_delete_partition_skips_vectors_when_collection_absent`: with `collection_exists() == False`, delete still drops the relational rows and never queries the missing collection. Full suite: 21 passed, ruff clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved partition deletion to gracefully handle cases where the configured vector-store collection was never created, preventing vector-related errors while still removing the relational partition data.

* **Tests**
  * Added a regression unit test ensuring vector querying/deletion are skipped when the vector-store collection is absent, while the relational row deletion still occurs.
  * Added an integration regression test confirming deleting an empty partition succeeds and the partition is removed from listings.

### DIFF
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -178,6 +178,17 @@ class PartitionService:
         """Drop a partition's vectors *and* relational rows (cross-cutting)."""
         await self._ensure_partition(partition)
         await self._partition_repo.delete_partition(name=partition)
+        # The shared Milvus collection is created lazily on the first insert
+        # system-wide, so on a fresh stack (nothing ever indexed) it doesn't
+        # exist; querying it would raise (e.g. DescribeCollectionException) and
+        # fail the whole delete *after* the relational rows are already gone.
+        # No collection means no chunks to clean up — skip the vector cleanup.
+        if not await self._vector_store.collection_exists(self._collection):
+            logger.info(
+                "Partition deleted; no vector collection to clean up",
+                partition=partition,
+            )
+            return
         ids = await self._vector_store.query_ids_by_filter(
             self._collection,
             {"partition": partition},
--- a/tests/integration/api/test_partition.py
+++ b/tests/integration/api/test_partition.py
@@ -41,6 +41,25 @@ class TestPartitionCRUD:
         response = api_client.delete(f"/partition/{test_partition_name}")
         assert response.status_code in [200, 204]
 
+    def test_delete_empty_partition_succeeds(self, api_client, test_partition_name):
+        """Regression for #505: deleting an empty partition must not 500.
+
+        Partitions are a Postgres concept; the shared Milvus collection is
+        created lazily on the first insert system-wide. An empty partition has
+        no chunks to clean up, so the delete must drop the relational rows and
+        return success instead of failing on the (possibly absent) collection.
+        """
+        create = api_client.post(f"/partition/{test_partition_name}")
+        assert create.status_code in [200, 201], create.text
+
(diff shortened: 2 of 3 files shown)


Ahmath-Gadji:
Addressed the CodeRabbit nitpick on `tests/unit/services/orchestrators/test_partition_service.py` (strengthen the "no vector calls" assertion) in commit https://github.com/linagora/openrag/commit/41639c24.

`ExplodingVectorStore` now also overrides `delete()` to flip a `delete_called` flag (and raise), and the test asserts `vstore.delete_called is False`. This closes the gap where a stray `delete([])` would have left `deleted_ids == []` and slipped past the original assertion. Local suite: 21 passed, ruff clean.

Ahmath-Gadji (approved):

Ahmath-Gadji (approved):

PR #511, opened 2026-06-18
fix(indexer): keep pool actor alive for queued batches

hedhoud:
## Context
Large indexing benchmarks can submit more work than the indexer can process immediately. In that case, queued tasks must remain attached to a live worker actor until their turn starts.

## Problem
The indexer pool actor could be cleaned up by Ray while queued work still existed. Those tasks stayed visible as `QUEUED`, but nothing was left to consume them. A full dataset run could then stop after the first window of files.

Concurrent startup also let multiple tasks initialize the catalog store at the same time, which put unnecessary pressure on Postgres.

## Expected behavior
The indexer pool remains alive for queued batches, respects the configured worker concurrency, and initializes its catalog connection once under concurrent load.

## Validation
- `uv run pytest tests/unit/services/workers/test_indexer_pool.py -q`
- `uv run pytest tests/unit/services/workers -q`
- `uv run ruff format --check openrag/services/workers/indexer_pool.py tests/unit/services/workers/test_indexer_pool.py`
- `uv run ruff check openrag/services/workers/indexer_pool.py tests/unit/services/workers/test_indexer_pool.py`

## Summary by CodeRabbit

* **Improvements**
  * Improved indexing pool initialization to be safe under concurrent startup, ensuring the catalog is initialized only once.
  * Updated worker pool creation to respect configured concurrency settings and run the actor as a detached Ray worker for more reliable background operation.
* **Tests**
  * Added unit coverage for single-flight catalog initialization under concurrent calls.
  * Added unit coverage verifying detached actor creation and correct concurrency configuration.

### DIFF
--- a/openrag/services/workers/indexer_pool.py
+++ b/openrag/services/workers/indexer_pool.py
@@ -53,6 +53,7 @@ class IndexerPool:
         rdb_cfg = cfg.rdb.model_copy(update={"database": f"partitions_for_collection_{cfg.vectordb.collection_name}"})
         self._catalog_store = PostgresStore(rdb_cfg, run_migrations=False)
         self._catalog_initialized = False
+        self._catalog_init_lock = asyncio.Lock()
         self._worker = IndexerWorker(
             pipeline=pipeline,
             task_state_manager=task_state_manager,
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji (approved):
This solves the issue we had with batch indexing.

PR #516, opened 2026-06-18
perf(caption): concurrent image captioning under the VLM semaphore

Ahmath-Gadji:
## What

Replace the sequential per-image VLM loop in `caption_stage` with a **concurrent fan-out**, bounded by the cluster-wide VLM semaphore (`get_vlm_semaphore`) so parallelism is shared across every indexer actor — the same headroom the embedder already uses for its batched fan-out. This restores the concurrency the legacy `BaseLoader.caption_images` had before the hexagonal parser migration moved captioning into a dedicated stage.

## Behavior

- **Fan-out**: all images for a document are captioned concurrently; the cluster-wide semaphore (default `VLM_SEMAPHORE=10`) bounds total in-flight VLM calls across actors.
- **Deterministic assembly**: captions are zipped back in image order, so markdown-ref replacement and caption append ordering are unchanged from the sequential version.
- **Cancellation scoped to real cancel signals**: one image failing propagates the error but leaves the other in-flight captions to drain (semaphore-bounded). Only a `CancelledError` (task cancel on shutdown/reload) tears the siblings down — explicitly, because `tqdm.gather` runs on `as_completed`, which won't propagate cancellation to children on its own.
- Progress bar preserved (`tqdm.gather` with per-document `desc`).

## Tests

`test_caption_stage_captions_images_concurrently_and_in_order` asserts all three VLM calls are in flight simultaneously (`max_in_flight == 3`) and captions still land on the correct image even when completion order is the reverse of submission order. An autouse fixture stubs the cluster-wide semaphore so the unit tests don't boot Ray.

## Notes

Builds on the hexagonal pipeline-stage refactor; targets `refactor/hexagonal`. The semaphore infrastructure (`runtime.get_vlm_semaphore`, `DistributedSemaphore`, bootstrap init) already exists on the base branch — this PR only wires the caption stage to use it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in image captioning: when a caption operation fails, related tasks are now properly cancelled and cleaned up to prevent resource leaks.

* **Refactor**
  * Image captioning now runs concurrently with intelligent resource budgeting, improving performance while maintaining deterministic output order.

### DIFF
--- a/openrag/services/workers/stages/caption.py
+++ b/openrag/services/workers/stages/caption.py
@@ -61,18 +61,18 @@ async def _caption_document(
         async with get_vlm_semaphore():
             return await vlm.caption_image(image.image_bytes, prompt=prompt)
 
-    # One image failing propagates but leaves the other in-flight captions to drain
-    # (the VLM semaphore bounds them). Only a cancellation signal — task cancel on
-    # shutdown/reload — tears the siblings down, which tqdm.gather (built on
-    # as_completed) won't do on its own, so we cancel them explicitly.
+    # tqdm.gather is built on as_completed, so a failing child does not clean up
+    # sibling tasks for us. Drain them explicitly so failed documents do not keep
+    # using the shared VLM semaphore in the background.
     label = processed_document.metadata.get("filename") or processed_document.document_id
     tasks = [asyncio.ensure_future(caption_one(image)) for image in images]
     try:
         captions = await tqdm.gather(*tasks, desc=f"Captioning images of *{label}*")
     except asyncio.CancelledError:
-        for task in tasks:
-            task.cancel()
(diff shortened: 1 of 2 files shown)


hedhoud:
I found and fixed one failure-path issue in the concurrent captioning change.

When one image caption failed, the stage raised correctly, but sibling caption tasks could keep running in the background and continue consuming the shared VLM semaphore. The pushed fix now cancels and drains pending sibling tasks for normal VLM failures as well as shutdown cancellation.

I added a regression test that starts three image captions, forces one to fail, and verifies the two slow siblings are cancelled instead of completing after the document has already failed.

hedhoud (approved):

PR #519, opened 2026-06-18
feat(presets): expose parsing strategies via /presets/options

### LINKED ISSUE #518: Expose parsing strategies via /presets/options (UI hardcodes an incomplete list)
## Problem

The admin preset editor's **parsing strategy** dropdown is hardcoded in the UI to `marker` and `pymupdf`, but `IndexationPipelineConfig.parsing_strategy` is `Literal["pymupdf", "marker", "docling"]` — so **`docling` is missing** and the list silently drifts from the backend.

Unlike chunking / retrieval / reranker (which the UI pulls from `GET /presets/options`), parsing strategies aren't exposed there, so the UI has no choice but to hardcode them.

## Fix

Expose `parsing_strategies` on `GET /presets/options`, derived from the validated `Literal` via `typing.get_args(...)` so the option list can never drift from what the config actually accepts. The UI then reads it like the other dropdowns.

Targets `refactor/hexagonal`. (Frontend switch is a separate change on the admin-UI branch.)

andyne13:
## What

Exposes `parsing_strategies` on `GET /presets/options` so the admin preset editor can drive its parsing-strategy dropdown from the backend instead of a hardcoded list.

Closes #518.

## Why

The UI hardcoded `marker` + `pymupdf` and had drifted from `IndexationPipelineConfig.parsing_strategy = Literal["pymupdf", "marker", "docling"]` — `docling` was missing. Chunking/retrieval/reranker options already come from this endpoint; parsing didn't.

## Change

- `PresetOptionsResponse` gains `parsing_strategies: list[str]`.
- The `/options` handler populates it from `get_args(IndexationPipelineConfig.model_fields["parsing_strategy"].annotation)` — derived from the validated `Literal`, so the option list can never drift from what the config accepts.

## Test

`test_preset_options_crud_and_rename` now asserts `parsing_strategies == {pymupdf, marker, docling}`. Ruff clean; response model verified to build with the derived list.

## Follow-up

Frontend switch (read `parsing_strategies` from options instead of the hardcoded `<SelectItem>`s) lands separately on the admin-UI branch.

## Summary by CodeRabbit

* **New Features**
  * The preset options endpoint now exposes available parsing strategies (pymupdf, marker, and docling) in its response.

* **Tests**
  * Updated integration tests to validate parsing strategies in the preset options endpoint response.

### DIFF
--- a/openrag/api/routers/admin/presets.py
+++ b/openrag/api/routers/admin/presets.py
@@ -1,5 +1,7 @@
 """Admin routes for the Phase 14 pipeline preset registry."""
 
+from typing import get_args
+
 from api.dependencies.auth import require_admin
 from api.schemas.admin.preset_schemas import (
     CreatePresetRequest,
@@ -9,6 +11,7 @@ from api.schemas.admin.preset_schemas import (
     UpdatePresetRequest,
 )
 from core.chunking import chunking_registry
+from core.config.indexation_pipeline import IndexationPipelineConfig
 from core.rerankers.registry import reranker_registry
 from core.retrieval import retriever_registry
 from di.providers import get_preset_service
@@ -17,6 +20,10 @@ from fastapi import APIRouter, Depends, Response, status
 router = APIRouter(dependencies=[Depends(require_admin)])
 
 _DEFAULT_RERANKER_PROVIDERS = ["infinity", "openai"]
+
+# Derived from the validated Literal so the exposed options can never drift from
+# what IndexationPipelineConfig actually accepts.
+_PARSING_STRATEGIES = list(get_args(IndexationPipelineConfig.model_fields["parsing_strategy"].annotation))
 
 
 def _registered_or_default(registered: list[str], defaults: list[str]) -> list[str]:
@@ -29,6 +36,7 @@ async def get_preset_options():
     """Return available preset strategy choices."""
     return PresetOptionsResponse(
         chunking_strategies=chunking_registry.list_registered(),
+        parsing_strategies=_PARSING_STRATEGIES,
         retrieval_types=retriever_registry.list_registered(),
         reranker_providers=_registered_or_default(
             reranker_registry.list_registered(),
--- a/openrag/api/schemas/admin/preset_schemas.py
+++ b/openrag/api/schemas/admin/preset_schemas.py
@@ -83,6 +83,7 @@ class PresetOptionsResponse(BaseModel):
     """Response body listing allowed preset option values."""
 
     chunking_strategies: list[str]
+    parsing_strategies: list[str]
     retrieval_types: list[str]
     reranker_providers: list[str]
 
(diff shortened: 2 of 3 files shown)


Ahmath-Gadji (approved):
LGTM — clean and correctly scoped. Verified the derivation resolves to `["pymupdf", "marker", "docling"]` and that deriving from `IndexationPipelineConfig`'s `Literal` (rather than `parser_registry.list_registered()`) is the right source of truth, since the registry also holds non-PDF/auto-dispatched parsers (`docx`, `image`, `eml`, audio, …). CI is green and the new test covers it.

### Verified live end-to-end

Against a running instance I exercised the full create → attach → resolve flow:

1. `GET /presets/options` → `parsing_strategies: ["pymupdf", "marker", "docling"]` ✅
2. `POST /presets/` with `{"name":"pymupdf-fast","preset_type":"indexation","config":{"parsing_strategy":"pymupdf"}}` → `201`
3. `PATCH /partition/{p}` with `{"indexation_preset":"pymupdf-fast"}` → `200`
4. `GET /partition/{p}/config` → `indexation_preset: pymupdf-fast`, resolved `indexation_pipeline.parsing_strategy: pymupdf` ✅

So the option flows correctly through preset creation, partition assignment, and resolved config. (Step 4 stops at the *resolved config* on purpose — see the gap below for why an actual parse wouldn't yet reflect it.)

### One known gap (out of scope for this PR) — described against this branch (`refactor/hexagonal`)

The `parsing_strategy` this endpoint now exposes is **not yet honored at parse time on this branch**. The per-partition `indexation_config` is plumbed all the way into the pipeline, and the **chunker and embedder already respect it** (their factories are wired). The parser is the exception:

- `IndexerPool` builds a single `parser = DocSerializerBridgeParser(config=cfg)` and calls `build_indexing_pipeline(parser=..., chunker_factory=..., embedder_factory=...)` **without** a `parser_factory` (`openrag/services/workers/indexer_pool.py`).
- Because no `parser_factory` is wired, `_select_parser(config)` (`openrag/services/workers/pipeline_builder.py`) never reaches its `config.parsing_strategy` branch and just returns that single bridge parser. That `config.parsing_strategy` line is, in fact, the *only* place the field is read in the whole parse path on this branch — and it's currently dead.
- The bridge then selects the concrete loader **by file extension / mimetype** from `get_loader_classes(config)`, which reads the **global** `config.loader.file_loaders` map. The per-file override (`_legacy_loader_config`) only toggles image captioning — it never touches the PDF loader.

Net effect: a partition set to `pymupdf` still parses PDFs with whatever the **global** `file_loaders.pdf` is. Exposing the option here is still correct and useful — it just isn't wired into actual parsing yet.

### Why the fix waits for #513

#513 replaces `DocSerializerBridgeParser` with the new `ParserDispatcher`, but keeps the same shape: `IndexerPool` still builds one global dispatcher and passes no `parser_factory`, and the dispatcher resolves the PDF backend from the global `config.loader.file_loaders.pdf`. So the gap carries straight over to the consolidated parser path. Fixing it on this branch would mean wiring the now-legacy bridge, only to rewrite it once #513 lands.

A separate PR will therefore wire the partition's `parsing_strategy` into the parse path **after #513 is merged**, using the existing (currently-unused) `parser_factory` / `_select_parser` seam with the global loader as fallback. It must also address the operational side: marker/docling require provisioned GPU pools, so the follow-up should validate the selected strategy against available backends (or document the provisioning requirement).

Approving so this can merge; the wiring is tracked as the follow-up above.

PR #524, opened 2026-06-19
fix(indexer): wire contextualization into hexagonal pool

### LINKED ISSUE #522: Wire contextualization into the hexagonal indexer (implemented but not connected)
## Summary

Re-wire **contextualization (contextual retrieval)** into the refactored indexing pipeline. The implementation exists but isn't connected, so it silently does nothing on `refactor/hexagonal` (whereas it works pre-refactor on `main`).

## Current state

The pieces are all present:
- `core/indexing/contextualize.py::ChunkContextualizer` (LLM prepends a 1–2 sentence context block to each chunk) + prompt template `chunk_contextualizer_tmpl.txt`.
- `services/workers/stages/contextualize.py` stage, called by `pipeline_builder.py` (`contextualize_stage`).
- `build_indexing_pipeline(...)` accepts `contextualizer` / `contextualizer_factory`, and `_select_contextualizer()` gates on `enable_contextualization`.

But the production indexer **doesn't wire it**: `services/workers/indexer_pool.py` builds the pipeline with `chunker_factory` and `embedder_factory` but **no `contextualizer` / `contextualizer_factory`**, so `_select_contextualizer()` always returns `None` and the stage no-ops. Net effect: toggling `enable_contextualization` (or a partition/preset asking for it) has no effect.

## Proposed

- [ ] Build a `contextualizer_factory` (resolve the `contextualization_llm` endpoint → `ChunkContextualizer`), mirroring the existing `embedder_factory`, and pass it into `build_indexing_pipeline` in `indexer_pool.py`.
- [ ] Decide and document the gate: preset `enable_contextualization` vs the chunker-level `contextual_retrieval` flag (the one that drove it pre-refactor) — align with the benchmarking workflow.
- [ ] Confirm `contextualization_mode` ("simple"/"structured"): only "simple" has an implementation today; either implement "structured" or keep mode out of scope. (The admin UI's simple/structured dropdown was removed as unbacked.)

## Acceptance criteria
- With contextualization enabled, indexed chunks carry the `[CONTEXT]` block on `refactor/hexagonal` (parity with `main`).
- A with/without-contextualization indexing run shows the difference (supports the team's benchmarks).

## Context
Found while auditing the admin-UI indexation preset editor. Being handled by the team; filing to track the gap.

hedhoud:
Context
Contextual retrieval was implemented in the refactored pipeline, but the production Ray indexer never provided the contextualizer dependency. That made contextualization silently no-op on refactor/hexagonal even when a partition preset enabled it.

Problem
This restores the missing composition link so the indexer can build contextualizers from named LLM endpoints, with a fallback to the legacy/global LLM config. It also handles OpenAI-compatible chat response payloads correctly before prepending context to chunks.

Validation
- uv run ruff format --check openrag/ tests/
- uv run ruff check openrag/ tests/
- uv run pytest tests/unit/services/workers/test_pipeline_builder.py tests/unit/services/workers/test_indexer_worker.py tests/unit/services/orchestrators/test_indexing_service.py tests/unit/core/indexing/test_contextualize.py tests/unit/services/workers/test_indexer_pool.py
- API_KEY=llm-key EMBEDDER_API_KEY=embed-key VLM_API_KEY=vlm-key RERANKER_API_KEY=rerank-key uv run --no-env-file pytest tests/unit

Closes #522

## Release Notes

* **New Features**
  * Enhanced support for multiple LLM response formats during contextualization, with robust fallback handling for malformed responses.
  * Improved concurrency management for LLM operations with configurable batch sizing.

* **Configuration**
  * Contextualization settings now intelligently derive from global configuration when preset-specific values are not defined.

### DIFF
--- a/openrag/services/orchestrators/preset_service.py
+++ b/openrag/services/orchestrators/preset_service.py
@@ -28,12 +28,16 @@ logger = get_logger()
 _VALID_PRESET_TYPES = frozenset({"indexation", "retrieval"})
 
 _DEFAULT_SEEDS: dict[str, dict[str, dict[str, Any]]] = {
+    # ``enable_contextualization`` is intentionally omitted from the ``default``
+    # indexation preset below — it is injected at seed time from the global
+    # ``chunker.contextual_retrieval`` (``CONTEXTUAL_RETRIEVAL``) toggle in
+    # _finalize_seed, so the env flag still drives contextualization on fresh
+    # deployments. Named presets (legal/finance) keep their explicit choice.
     "indexation": {
         "default": {
             "chunking": {"name": "recursive_splitter", "chunk_size": 512, "chunk_overlap_rate": 0.2},
             "parsing_strategy": "marker",
             "enable_image_captioning": True,
-            "enable_contextualization": False,
             "enable_entity_extraction": True,
             "enable_topic_tagging": True,
         },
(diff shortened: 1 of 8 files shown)


tests/unit/services/workers/test_indexer_pool.py
```diff
@@ -98,3 +98,147 @@ def fake_options(**kwargs):
...
+        instances = []
+
+        def __init__(self, **kwargs):
+            self.kwargs = kwargs
+            self.instances.append(self)
+
+        async def chat(self, messages, **kwargs):
+            return {"choices": [{"message": {"content": "document context"}}]}
+
+    llm_registry.register("test-contextualizer-llm")(FakeLLM)
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/37649d74. The `FakeLLM` registration is now removed from the shared `llm_registry` in a `finally` block, and the unused `instances` class attribute was dropped.

Ahmath-Gadji:
Pushed two follow-up commits on top of the contextualization wiring:

**`37649d74` — feat(indexer): rate-limit chunk contextualization via the distributed LLM semaphore**
The hexagonal contextualizer was calling the LLM without sharing the cluster-wide `llmSemaphore`, so a large indexing job could flood the vLLM endpoint and run uncapped against query-time calls. `_llm.chat` is now wrapped by an optional `llm_semaphore` gate injected from the services-layer factory (`DistributedSemaphore("llmSemaphore")`), keeping `core` free of any Ray/services import (layer guard stays green).

**`8aa7249c` — fix(presets): drive default contextualization from CONTEXTUAL_RETRIEVAL**
Contextualization was gated only by the per-partition preset's `enable_contextualization` (default `false`); the global `CONTEXTUAL_RETRIEVAL` (`chunker.contextual_retrieval`) flag was orphaned and never took effect. Now (a) the `default` indexation preset inherits the flag at seed time (mirrors the `reranker.enabled` kill-switch), and (b) `PresetService.sync_env_toggles` re-applies it to the `default` preset on every boot, so existing deployments honour a changed env value after restart. Named presets (legal/finance) keep their explicit values.

Verified end-to-end on a live stack: forcing the preset to `false` + restart → boot re-sync flips it back to `true`; indexing then runs contextualization. Full unit suite: 1268 passed, 1 skipped; ruff + layer-import-guard green.

Ahmath-Gadji (approved):

openrag/services/orchestrators/preset_service.py
```diff
@@ -105,21 +109,53 @@ async def seed_defaults(self) -> None:
...
+        startup sequence, so the refreshed value reaches the in-memory caches.
+        """
+        row = await self._repo.get("default", "indexation")
+        if row is None:
+            return
+        desired = self._config.chunker.contextual_retrieval
+        config = row["config"]
+        if config.get("enable_contextualization") == desired:
+            return
+        await self._repo.upsert("default", "indexation", {**config, "enable_contextualization": desired})
```
hedhoud: I’m worried about this one because it changes the ownership model of the `default` preset. After an admin updates `default` from `/admin/presets`, a restart can silently put `enable_contextualization` back to whatever `CONTEXTUAL_RETRIEVAL` says. That makes the DB preset look persistent from the UI, but not actually authoritative. Could we keep the env value for first-time seeding only, or make this an explicit one-shot migration/backfill instead of syncing it on every boot?

hedhoud (commented):
Left one concern about the default preset ownership model.

openrag/services/orchestrators/preset_service.py
hedhoud: Fixed in https://github.com/linagora/openrag/commit/77769ab4e73036ba3fd82558c7149c1f00ea3a05. The startup re-sync was removed, so `CONTEXTUAL_RETRIEVAL` now only drives the initial seed for fresh installs. Once the default preset exists in the DB, admin changes stay authoritative across restarts.

Ahmath-Gadji (approved):
LGTM. Tested end to end

PR #526, opened 2026-06-19
feat(api): connect to an external Ray cluster via RAY_ADDRESS

Ahmath-Gadji:
## What

Make `ray.init()` in `openrag/api.py` branch on `RAY_ADDRESS`:

- **`RAY_ADDRESS` set** → `ray.init(address=...)` to **attach to an external Ray cluster** (e.g. a dedicated ray-head container / KubeRay). No embedded dashboard is started — the head node owns it.
- **unset** → embedded mode as before, binding the unauthenticated dashboard to `127.0.0.1` by default ([CVE-2023-48022](https://nvd.nist.gov/vuln/detail/CVE-2023-48022)), overridable via `RAY_DASHBOARD_HOST`.

```python
_ray_address = os.environ.get("RAY_ADDRESS")
if _ray_address:
    # Connect to an external Ray cluster (e.g. a dedicated ray-head container).
    ray.init(address=_ray_address)
else:
    # Embedded mode: start a local Ray cluster inside this process.
    ray.init(dashboard_host=os.environ.get("RAY_DASHBOARD_HOST", "127.0.0.1"))
```

## Why

`RAY_ADDRESS` was already referenced in the cluster-deployment docs and relied on Ray implicitly reading the env var, while the code unconditionally passed `dashboard_host` (meaningless when attaching to an existing cluster). This makes the two modes explicit and keeps the secure-by-default dashboard binding for embedded runs.

## Docs / env

- `.env.example`, `docs/assets/env_example.env`, `docs/assets/env_linux_gpu.env` — document `RAY_ADDRESS` and `RAY_DASHBOARD_HOST` (both commented out).
- `docs/content/docs/documentation/env_vars.md` — new rows for `RAY_ADDRESS` and `RAY_DASHBOARD_HOST`.
- `docs/content/docs/documentation/deploy_ray_cluster.md` — note clarifying that in attach mode the head node owns the dashboard.

## Notes

- No `docker-compose` changes: the dashboard-host default on `main` is already `127.0.0.1`; this PR is scoped to the `RAY_ADDRESS` attach path + docs.
- `ruff check openrag/api.py` passes.

## Summary by CodeRabbit

* **New Features**
  * Added support for connecting to an external Ray cluster using `RAY_ADDRESS`, with embedded mode used only when it’s unset.
  * Improved embedded Ray dashboard configuration via `RAY_DASHBOARD_HOST` (defaulting to loopback).

* **Documentation**
  * Updated environment examples and Ray documentation to reflect external cluster attachment behavior and dashboard ownership.
  * Added security guidance around the embedded Ray dashboard/job-submission API and recommended exposure controls.

### DIFF
--- a/docs/content/docs/documentation/deploy_ray_cluster.md
+++ b/docs/content/docs/documentation/deploy_ray_cluster.md
@@ -136,6 +136,10 @@ docker compose up -d
 
 Once running, **OpenRAG will auto-connect** to the Ray cluster using `RAY_ADDRESS` from `.env`.
 
+:::note
+When `RAY_ADDRESS` is set, the app **attaches** to the external cluster and does **not** start its own embedded Ray dashboard — the head node owns it (started above via `--dashboard-host 0.0.0.0 --dashboard-port ${RAY_DASHBOARD_PORT:-8265}`). The app-side `RAY_DASHBOARD_HOST` setting is only used in embedded (single-node) mode, where it defaults to `127.0.0.1` because the dashboard API is unauthenticated ([CVE-2023-48022](https://nvd.nist.gov/vuln/detail/CVE-2023-48022)).
+:::
+
 ---
 
 With this setup, your app is now fully distributed and ready to handle concurrent tasks across your Ray cluster.
--- a/.env.example
+++ b/.env.example
@@ -74,6 +74,14 @@ RAY_task_retry_delay_ms=3000
 RAY_ENABLE_UV_RUN_RUNTIME_ENV=0 # critical with the newest version of UV
 # # To disable worker killing
 # RAY_memory_monitor_refresh_ms=0
+# Connect to an external Ray cluster instead of starting an embedded one.
+# When set, the app attaches to this cluster and does NOT start a local dashboard
+# (the head node owns it). See docs/documentation/deploy_ray_cluster.
+# RAY_ADDRESS=ray://X.X.X.X:10001
+# Interface the embedded Ray dashboard binds to. Defaults to 127.0.0.1 (loopback)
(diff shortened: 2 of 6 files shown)


hedhoud (commented):
One small docs concern: this note makes the app-side behavior clear, but the security warning should also cover the Ray head dashboard. In cluster mode, that head dashboard is the one users will actually have running, and the example starts it on all interfaces. If the port is not kept private, firewalled, or placed behind auth, users can still end up exposing Ray's unauthenticated dashboard/job API.

docs/content/docs/documentation/deploy_ray_cluster.md
```diff
@@ -136,6 +136,10 @@ docker compose up -d
 
 Once running, **OpenRAG will auto-connect** to the Ray cluster using `RAY_ADDRESS` from `.env`.
 
+:::note
+When `RAY_ADDRESS` is set, the app **attaches** to the external cluster and does **not** start its own embedded Ray dashboard — the head node owns it (started above via `--dashboard-host 0.0.0.0 --dashboard-port ${RAY_DASHBOARD_PORT:-8265}`). The app-side `RAY_DASHBOARD_HOST` setting is only used in embedded (single-node) mode, where it defaults to `127.0.0.1` because the dashboard API is unauthenticated ([CVE-2023-48022](https://nvd.nist.gov/vuln/detail/CVE-2023-48022)).
```
hedhoud, line 140: This should probably call out the Ray head dashboard too. In cluster mode the app does not start its own dashboard, but the example above starts the head dashboard on all interfaces. If that port is not private, firewalled, or behind auth, users can still expose Ray's unauthenticated dashboard/job API.

PR #527, opened 2026-06-19
fix(quota): honor per-user file quota when global default is unlimited

andyne13:
## Problem
With `DEFAULT_FILE_QUOTA` negative (the default, `-1` = unlimited), an admin-set per-user `file_quota` was **silently ignored** — the user showed as "Unlimited" and uploads were never blocked. Verified on a live stack: a user with `file_quota=200` and 200 indexed files was still allowed to upload.

## Cause
All three quota code paths short-circuited to "unlimited" on `default < 0` **before** reading the per-user `file_quota`:
- `user_service.get_current_user_info` (the `/users/info` display)
- `dependencies/auth.check_user_file_quota` (the upload gate)
- `auth_service.validate_file_quota` (the pure check)

The per-user value was only consulted in the branch *after* that early return, which never ran.

## Fix
Drop the early `default < 0` return at each site. The existing `None → default`, then `< 0 → unlimited` resolution already handles every case correctly:
- per-user `None` + default `-1` → still unlimited (default users unchanged)
- per-user `100` + default `-1` → **now enforced** (the bug)
- per-user `-1` → unlimited · admin → bypass

The only behaviour that changes is the broken case; default-everyone-unlimited deployments are unaffected.

## Tests
- New `validate_file_quota` regression (per-user limit enforced even when global default is unlimited; a user with no per-user quota still inherits the unlimited default).
- Corrected `test_current_user_info_negative_default_is_unlimited`, which had **codified the bug** (asserted a per-user quota of 3 resolved to -1) — split into per-user-override-honored and no-override-unlimited cases.

Verified live (in-container): `200/200 @ default -1` now raises; `5/200` and no-override+default-1 do not.

## Release Notes

* **Bug Fixes**
  * Fixed file quota enforcement so per-user `file_quota` limits are correctly applied even when the global default quota is set to unlimited.
  * Refined unlimited quota behavior so negative quota values act as “unlimited” only after applying per-user overrides, while admins remain unlimited.
  * Improved quota checks to avoid unnecessary pending-task counting when the user effectively has unlimited quota.
* **Tests**
  * Expanded unit tests to cover per-user quota override scenarios and the new pending-count behavior.

### DIFF
--- a/openrag/api/dependencies/auth.py
+++ b/openrag/api/dependencies/auth.py
@@ -239,11 +239,9 @@ async def check_user_file_quota(
     default_file_quota = config.rdb.default_file_quota
     if user.get("is_admin", False):
         return user
-    # A per-user quota takes precedence; the global default is only the
-    # fallback for users with no override (resolved in validate_file_quota).
-    # An explicit per-user ``< 0`` means unlimited.
     user_quota = user.get("file_quota")
-    if user_quota is not None and user_quota < 0:
+    effective_quota = default_file_quota if user_quota is None else user_quota
+    if effective_quota < 0:
         return user
 
     user_id = user.get("id")
--- a/openrag/api/dependencies/auth.py
+++ b/openrag/api/dependencies/auth.py
@@ -239,8 +239,9 @@ async def check_user_file_quota(
     default_file_quota = config.rdb.default_file_quota
     if user.get("is_admin", False):
         return user
-    if default_file_quota < 0:
-        return user
+    # A per-user quota takes precedence; the global default is only the
+    # fallback for users with no override (resolved in validate_file_quota).
+    # An explicit per-user ``< 0`` means unlimited.
     user_quota = user.get("file_quota")
     if user_quota is not None and user_quota < 0:
         return user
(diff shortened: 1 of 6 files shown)


hedhoud (commented):
I found one small issue worth fixing before merge.

The new precedence rule is right, but `check_user_file_quota()` still calls the pending-task service when the resolved quota is unlimited through the global default.

Example: `file_quota=None` and `DEFAULT_FILE_QUOTA=-1`. That user is unlimited, so upload quota checking should return before calling `job_service.get_user_pending_task_count()`. Right now, if the task manager is degraded, an unlimited user can still fail on quota checking even though no quota needs to be evaluated.

I would resolve the effective quota first: per-user value when present, otherwise the global default. If that resolved value is `< 0`, return immediately. Then call the pending-task service only for users with an actual limit.

The rest of the PR looks good to me, and the service-level behavior matches the intended precedence.

hedhoud:
Small clarification on my previous review note: the `file_quota=None` behavior is correct. It should use `DEFAULT_FILE_QUOTA`.

The fix I pushed only changes the order: resolve that effective quota first, then return immediately when it is unlimited (`-1`). That avoids calling the pending-task service for users who do not need quota enforcement, while keeping the same precedence rule for limited users.

Ahmath-Gadji (approved):
LGTM.
Works and i've tested it manually

hedhoud (approved):

hedhoud (approved):
Looks ready to merge. CodeRabbit is green, all GitHub checks pass, and the quota behavior is now covered for both per-user override and unlimited-default cases.

PR #529, opened 2026-06-19
feat(files): persist + expose indexation timestamp (files.created_at)

andyne13:
## Problem
The admin UI documents list has an "Indexed" column, but it's always empty. The list is built from the Postgres `files` catalog (`SELECT * FROM files` → `_row_to_dict`), and that table has **no timestamp column** — the index time exists only in the Milvus chunk metadata (which is why the file *detail* view shows `created_at`/`indexed_at` but the *list* can't).

## Fix
- Add a `files.created_at` column — `DateTime(timezone=True)`, `server_default now()`, `NOT NULL` — to the model (`schema.py`).
- Surface it in `_row_to_dict` (placed after the `**metadata` spread so the column value always wins).
- Idempotent Alembic migration (`column_exists` guard, both directions), consistent with the repo's create_all-then-migrate policy.

The file INSERT doesn't list `created_at`, so the server default sets it at insert time = the indexation time. No UI change needed — the existing "Indexed" column reads `indexed_at ?? created_at`.

## Caveat
Existing files have no recorded index time, so the `server_default` backfills them to the **migration run time**; newly indexed files get their true insert time.

## Tests
Persistence + partition-service unit suites green (66 passed); ruff clean. Both `_row_to_dict` callers use `SELECT *`, so the new column is always present.

## Summary by CodeRabbit

* **New Features**
  * Added tracking for an indexing timestamp across document catalog writes and vector-store chunk upserts, with the ability to pin it to a caller-provided time (otherwise it uses the current time).

* **Chores**
  * Added a database migration to introduce a new non-null `indexed_at` column with a default for existing rows.
  * Updated unit test doubles and assertions to include the new timestamp parameter in indexing flows.

### DIFF
--- a/openrag/services/persistence/document_repo.py
+++ b/openrag/services/persistence/document_repo.py
@@ -29,6 +29,8 @@ from core.models.catalog import DocumentRecord, DocumentStatus
 from core.ports.document_repo import DocumentRepository
 
 if TYPE_CHECKING:
+    from datetime import datetime
+
     import asyncpg
 
 # Note on JSON: ``ConnectionManager.initialize`` registers a json/jsonb codec
@@ -271,11 +273,15 @@ class PgDocumentRepository(DocumentRepository):
         relationship_id: str | None = None,
         parent_id: str | None = None,
         indexation_config: dict | None = None,
+        indexed_at: datetime | None = None,
     ) -> bool:
         """TODO(phase-9): remove. Mirror of legacy ``add_file_to_partition``.
 
         Creates the partition row on first use (legacy behaviour). Returns
         ``False`` if a row with the same (file_id, partition) already exists.
+
+        ``indexed_at`` pins the indexation timestamp so it matches the Milvus
+        chunks; when ``None`` the ``files.indexed_at`` server default applies.
         """
         async with self.pool.acquire() as conn:
             async with conn.transaction():
@@ -311,12 +317,16 @@ class PgDocumentRepository(DocumentRepository):
                         user_id,
                     )
 
-                await conn.execute(
-                    """
-                    INSERT INTO files (file_id, partition_name, file_metadata,
-                                       indexation_config, created_by, relationship_id, parent_id)
-                    VALUES ($1, $2, $3::json, $4::jsonb, $5, $6, $7)
-                    """,
+                columns = [
+                    "file_id",
+                    "partition_name",
+                    "file_metadata",
+                    "indexation_config",
+                    "created_by",
+                    "relationship_id",
+                    "parent_id",
+                ]
+                values: list[Any] = [
                     file_id,
                     partition,
                     file_metadata or {},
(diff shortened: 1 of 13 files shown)


openrag/services/persistence/document_repo.py
```diff
@@ -550,12 +550,16 @@ def _row_to_dict(row: asyncpg.Record) -> dict:
...
+        created_at = row["created_at"]
         return {
             "partition": row["partition_name"],
             "file_id": row["file_id"],
             "relationship_id": row["relationship_id"],
             "parent_id": row["parent_id"],
             **metadata,
+            # Indexation timestamp lives on the row, not in file_metadata;
+            # placed after the spread so the column value always wins.
+            "created_at": created_at.isoformat() if created_at else None,
```
Ahmath-Gadji: **Non-blocking design observation — `created_at` key overload**

This override is correct for the goal (the admin list's "Indexed" column now populates), but it reuses the `created_at` key for a *second* meaning. The indexing API documents `created_at` in `file_metadata` as a user-supplied **temporal field** (`openrag/api/routers/admin/indexing.py:100,239` — "Optional temporal field … for time-based queries and filtering"). Placing the column after the spread means a user-provided document date is silently shadowed by the indexation/insert time in this flattened dict.

It's harmless in practice — the underlying `file_metadata` JSON (and the Milvus chunk metadata used for filtering) still hold the user value, and the file-detail view reads `metadata.created_at` separately — but the list surface now overloads one key with two concepts.

Worth considering: expose the column as **`indexed_at`** instead. The UI already reads `file.indexed_at ?? file.created_at` (`extern/indexer-ui/.../partition/[partition]/+page.svelte:252,298`), so `indexed_at` is the codebase's intended key for indexation time. That would fill the "Indexed" column just the same while leaving the user's temporal `created_at` intact (and `created_at` sorting at `+page.svelte:70` would then sort by document date rather than insert time — pick whichever is intended).

`openrag/services/persistence/document_repo.py:562`

Ahmath-Gadji (changes requested):
### Request: expose the indexation timestamp as `indexed_at`, not `created_at`

Thanks for tracking down why the "Indexed" column was empty — the diagnosis and the idempotent migration are spot-on. One blocking concern with the key name, though.

`created_at` is a **reserved, client-supplied temporal field**, not a free key we can repurpose. Per the API docs ([API.mdx#L93-L102](https://github.com/linagora/openrag/blob/8f875685bbb81b1bb5828f0601b0814080fb556b/docs/content/docs/documentation/API.mdx?plain=1#L93-L102)):

> `created_at`: ISO 8601 format date of when the file was created
> `created_at` is provided by the client in the metadata of the file during upload.

It's the user's *document date* (used for temporal-aware search/filtering), not the system's insert time. Placing the new column after the `**metadata` spread in `_row_to_dict` (`openrag/services/persistence/document_repo.py:562`) silently **shadows** that client value with the indexation time in the file-listing surface.

The two concepts should stay separate. Please expose the indexation timestamp as **`indexed_at`** instead:

- It's already the codebase's intended key for this — the UI reads `file.indexed_at ?? file.created_at` (`extern/indexer-ui/src/routes/indexer/partition/[partition]/+page.svelte:252,298`), so `indexed_at` fills the "Indexed" column directly.
- It leaves the reserved `created_at` (client temporal field) intact, so `created_at` sorting (`+page.svelte:70`) and temporal filtering keep working on the document date.

Concretely: rename the new column `files.created_at` → `files.indexed_at` across the migration, `schema.py`, and the `_row_to_dict` key. The rest of the change (server default, idempotent guard, INSERTs relying on the default) carries over unchanged.

Everything else in the PR looks good and CI is green — just this naming.

openrag/services/persistence/document_repo.py
Ahmath-Gadji: Done — the column and its `_row_to_dict` surface are renamed to `indexed_at`, so `created_at` stays reserved for the client-supplied temporal field.

Fixed in https://github.com/linagora/openrag/commit/9ee5c5ab (rename) with a follow-up in https://github.com/linagora/openrag/commit/0f0c9252 that mints a single `indexed_at` shared by the Milvus chunks and the `files` row.

Ahmath-Gadji (approved):
Requested change implemented: the indexation timestamp is now `indexed_at` (commit 9ee5c5ab), with a follow-up so Milvus chunks and the `files` row share one timestamp (commit 0f0c9252). `created_at` stays reserved for the client temporal field. CI re-running.

PR #530, opened 2026-06-22
fix(postgres): support managed database provisioning

### LINKED ISSUE #520: Harden PostgreSQL provisioning for managed/OpenShift deploys (no CREATEDB, migrations-only schema)
## Summary

Harden OpenRag's PostgreSQL provisioning so it deploys cleanly on managed / least-privilege Postgres (OpenShift, cloud RDS/Crunchy, etc.). Today the app **creates its own database and schema at startup**, which requires elevated privileges and fights the way managed Postgres is operated.

## Current behaviour (refactor/hexagonal)

1. **App auto-creates the database.** `connection.py::_ensure_database_exists()` calls `sqlalchemy_utils.create_database()` if the DB is absent → the connecting role needs the **`CREATEDB`** privilege, which managed Postgres typically doesn't grant to app roles.
2. **Schema created at runtime AND via migrations.** `Base.metadata.create_all()` runs at startup (`services/persistence/partition_repo.py`) *and* Alembic runs (`PostgresStore.initialize`, `create_catalog_store(run_migrations=...)`). The two overlap, which is why migrations must be hand-kept idempotent. Schema should come from migrations only.
3. **DB name derived from the Milvus collection** (`partitions_for_collection_<collection>`), now overridable via `POSTGRES_DATABASE` / `rdb.database` (good — keep this).

Net effect on OpenShift: the app fails at boot unless the role has `CREATEDB`, or the DB is pre-created — and even then `create_all()` needs schema-creation rights at runtime.

## Proposed hardening

- [ ] **Make DB auto-creation opt-out.** Add a flag (e.g. `POSTGRES_AUTO_CREATE_DB`, default `true` for the self-contained compose dev stack) so production can point at a **pre-provisioned** database and skip `create_database()` entirely — no `CREATEDB` needed.
- [ ] **Schema from migrations only in prod.** Make the startup `Base.metadata.create_all()` opt-out (or gate it behind the same dev flag) so schema is owned by Alembic; provide a dedicated **migration entrypoint** (e.g. `alembic upgrade head`) runnable as an init-container / Job before the app starts.
- [ ] **Helm chart support** (`infra/charts/openrag-stack`): a pre-install/pre-upgrade **migration Job** (or init-container), plus values for external/managed Postgres (host/port/db/secret) and a toggle for "DB already provisioned, role has no CREATEDB".
- [ ] **Docs**: least-privilege runbook — pre-create the DB (default name `partitions_for_collection_<VDB_COLLECTION_NAME>` or explicit `POSTGRES_DATABASE`), `OWNER` = app role, grant only `CONNECT`/`USAGE`/`CREATE on schema`; no `CREATEDB`/superuser.

## Acceptance criteria

- OpenRag boots against a Postgres where the app role **lacks `CREATEDB`** and the database is pre-created, with schema applied by migrations.
- The self-contained `docker compose` dev stack still works out-of-the-box (auto-create remains the default there).
- No runtime `create_all()` required in the production path; Alembic is the single source of schema truth.

## Context

Surfaced during an OpenShift deployment: operators asked why the vector layer "needs Postgres and creates a DB". Root cause is the app-provisions-its-own-DB pattern (and historically the Vectordb actor co-owning the catalog — already untangled by the hexagonal ports/adapters split). This issue is the remaining ops-hardening piece.

hedhoud:
Closes #520

## Why
Managed PostgreSQL deployments usually do not allow the application role to create databases. OpenRAG should be able to start against a pre-provisioned database, with schema changes applied as an explicit deployment step.

## What changed
- Added Postgres provisioning flags so local compose keeps its current behavior while managed deployments can skip database creation and app-start migrations.
- Added a standalone migration runner and an optional Helm migration hook for install/upgrade flows.
- Updated the Kubernetes and migration docs to describe the least-privilege path.

## Validation
- `uv run --no-env-file ruff format --check openrag tests`
- `uv run --no-env-file ruff check openrag tests`
- `uv run --no-env-file pytest tests/unit/services/persistence/test_connection.py tests/unit/services/persistence/test_migration_entrypoint.py tests/unit/di/test_container.py::TestRepositoriesFactory tests/unit/core/config/test_rdb_env.py`
- Full `tests/unit` was also run locally: 1315 passed, 1 existing model-endpoint test failed because the local repo `.env` injects API keys into that test.

Note: I could not render the Helm chart locally because `helm` is not installed in this environment.

## Summary by CodeRabbit

* **New Features**
  * Added configuration options to control automatic database creation and migration execution independently
  * Enabled migration jobs for Kubernetes deployments as separate pre-install/pre-upgrade hooks
  * New environment variables: `POSTGRES_AUTO_CREATE_DB`, `POSTGRES_RUN_MIGRATIONS`, `POSTGRES_POOL_MIN_SIZE`, `POSTGRES_POOL_MAX_SIZE`, `POSTGRES_COMMAND_TIMEOUT`, and `POSTGRES_DATABASE`

* **Documentation**
  * Updated guides for managed PostgreSQL database setup, migration strategies, and Kubernetes deployments
  * Enhanced environment variable documentation with new PostgreSQL configuration options

* **Tests**
  * Added test coverage for new configuration options and migration job behavior

### DIFF
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -63,13 +63,26 @@ vectordb:
   schema_version: 1 # Increment when the collection schema changes and a migration is required
 
 # --- Relational Database (PostgreSQL) ---
-# Env: POSTGRES_HOST, POSTGRES_PORT, POSTGRES_USER, POSTGRES_PASSWORD, DEFAULT_FILE_QUOTA
+# Env: POSTGRES_HOST, POSTGRES_PORT, POSTGRES_USER, POSTGRES_PASSWORD,
+#      POSTGRES_DATABASE, POSTGRES_AUTO_CREATE_DB, POSTGRES_RUN_MIGRATIONS,
+#      DEFAULT_FILE_QUOTA, POSTGRES_POOL_MIN_SIZE, POSTGRES_POOL_MAX_SIZE,
+#      POSTGRES_COMMAND_TIMEOUT
 rdb:
   host: rdb
   port: 5432
   user: root
   password: "root_password"
   default_file_quota: -1
+  # Leave unset to derive partitions_for_collection_<VDB_COLLECTION_NAME>.
+  database: null
+  # Compose/dev convenience. Managed Postgres deployments should pre-create
+  # the database and set POSTGRES_AUTO_CREATE_DB=false.
+  auto_create_database: true
+  # Disable in the app when migrations run as a separate Job/init step.
+  run_migrations: true
+  pool_min_size: 5
+  pool_max_size: 20
+  command_timeout: 30
 
 # --- Reranker ---
 # Env: RERANKER_PROVIDER, RERANKER_ENABLED, RERANKER_MODEL, RERANKER_TOP_K, RERANKER_BASE_URL, RERANKER_API_KEY, RERANKER_TIMEOUT, RERANKER_SEMAPHORE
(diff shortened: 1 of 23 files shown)


Ahmath-Gadji (approved):
LGTM. I've made some small changes

PR #531, opened 2026-06-22
fix(compose): use named volumes for stateful data

### LINKED ISSUE #521: Use named volumes / PVCs for stateful data instead of repo-tree bind mounts
## Summary

Move the stack's stateful storage off repo-tree **host bind mounts** to **named volumes** (compose) and **PVCs** (Helm), so deployments are portable, k8s/OpenShift-friendly, and have predictable data lifecycle.

## Current behaviour (`infra/compose/docker-compose.yaml`)

Stateful data defaults to bind mounts under the repo tree:
- `${DB_VOLUME:-../../db}:/var/lib/postgresql/data` — Postgres data
- `${DATA_VOLUME:-../../data}:/app/data` — uploaded files / app data
- `../../logs:/app/logs` — logs
- Milvus deps (etcd / MinIO) similarly bind-mount under the repo
- `../../openrag:/app/openrag` — **dev-mode source mount** (should not exist in a prod image run)

Consequences:
- Not portable to Kubernetes/OpenShift (needs PVCs, not host paths).
- `docker compose down -v` does **not** clear bind-mounted data (e.g. Milvus/Postgres survive), which surprises operators and complicates clean redeploys.
- Host-path perms (the `init-perms` chown step) only exist because of bind mounts.

## Proposed

- [ ] **Compose**: default the stateful paths to **named volumes** (`pgdata`, `appdata`, `milvus`, `etcd`, `minio`, `logs`), keeping the `*_VOLUME` env overrides for users who want host paths. Drop/relocate the dev-only `../../openrag` source mount to a dev override file.
- [ ] **Helm** (`infra/charts/openrag-stack`): ensure Postgres, Milvus, etcd, MinIO and app `data` use **PVCs** with configurable `storageClass`/size; no host paths.
- [ ] Make `init-perms` unnecessary when volumes/PVCs are used.
- [ ] Docs: data lifecycle + backup/restore note (named volumes / PVCs vs the old bind-mount behaviour).

## Context
Deferred deployment-hardening item; relevant to the OpenShift work alongside #520 (Postgres provisioning).

hedhoud:
Closes #521

## Why
Compose was still writing runtime state into repo-tree bind mounts. That makes local cleanup surprising, creates host-permission issues, and does not match how the stack is expected to run on Kubernetes or OpenShift.

## What changed
- Compose now defaults stateful storage to named volumes for app data, logs, model cache, PostgreSQL, Milvus, etcd, and MinIO.
- The live source-code bind mount moved to a dev-only compose override.
- Docs now explain the data lifecycle for named volumes, host-path overrides, and PVC-backed deployments.
- Added tests that keep the compose defaults and Helm storage expectations from drifting back.

## Validation
- `uv run --no-env-file ruff format --check tests/unit/infra/test_compose_storage.py`
- `uv run --no-env-file ruff check tests/unit/infra/test_compose_storage.py`
- `uv run --no-env-file pytest tests/unit/infra/test_compose_storage.py`
- `SHARED_ENV=.env docker compose config`
- `SHARED_ENV=.env docker compose -f docker-compose.yaml -f docker-compose.dev.yaml config`

## Summary by CodeRabbit

* **New Features**
  * Added an optional Docker Compose named-volume storage profile for volume-based persistence
  * Expanded storage configuration via environment variables across app data, logs, caches, PostgreSQL, and Milvus/etcd/MinIO
* **Documentation**
  * Improved OpenShift deployment, migration, and backup/restore guidance (including a fixed code snippet)
  * Added “Compose Storage Volumes” and updated environment variable references
* **Tests**
  * Added unit tests validating Compose/Helm persistence behavior across storage backends
* **Chores**
  * Made Milvus compose inclusion and log volume source configurable; ensured Milvus starts after MinIO; updated reranker CPU profile label

### DIFF
--- a/docs/content/docs/documentation/backup_restore.md
+++ b/docs/content/docs/documentation/backup_restore.md
@@ -4,17 +4,7 @@ title: How to backup or restore OpenRag partition/data ?
 
 # How to backup partitions?
 
-## Storage lifecycle
-
-Docker Compose keeps the historical host-path storage layout by default: app data, logs, model cache, PostgreSQL, and Milvus state stay in the same local paths as before. This avoids moving existing local deployments unexpectedly.
-
-If you want Docker named volumes instead, copy the values from `infra/compose/.env.named-volumes.example` into your `.env`. That profile is useful when you want a storage layout closer to Kubernetes/OpenShift PVC behavior.
-
-No migration is required unless you choose to switch an existing deployment from host paths to named volumes. In that case, stop the stack, copy the existing data into the new volumes, then restart the stack and verify partitions before removing the old host directories.
-
-Use `docker compose down` to stop the stack while keeping data. Use `docker compose down -v` only when you intentionally want to remove Docker named volumes and start from a clean state.
-
-For Kubernetes and OpenShift, the Helm chart stores state in PVCs. Treat PVC snapshots or storage-provider backups as infrastructure backups, and use the OpenRAG backup commands below when you need a portable partition-level export.
+> Storage layout (host paths vs. named volumes) and migrating data between storage modes are documented in [Deploying OpenRAG on OpenShift](/openrag/documentation/openshift_deployment/).
 
 ## Backup one partition
 
--- a/docs/content/docs/documentation/backup_restore.md
+++ b/docs/content/docs/documentation/backup_restore.md
@@ -3,6 +3,17 @@ title: How to backup or restore OpenRag partition/data ?
 ---
 
 # How to backup partitions?
+
+## Storage lifecycle
+
+Docker Compose uses named volumes for OpenRAG state by default: app data, logs, model cache, PostgreSQL, Milvus, etcd, and MinIO. This makes local deployments closer to Kubernetes/PVC behavior and avoids writing runtime data into the repository tree.
+
+Use `docker compose down` to stop the stack while keeping data. Use `docker compose down -v` only when you intentionally want to remove the named volumes and start from a clean state.
+
+If you need host-mounted storage, set the matching volume variable in `.env` (`DATA_VOLUME`, `DB_VOLUME`, `MILVUS_VOLUME`, `ETCD_VOLUME`, `MINIO_VOLUME`, `LOG_VOLUME`, or `MODEL_WEIGHTS_VOLUME`). In that case, lifecycle and permissions are managed by the host path owner.
+
+For Kubernetes and OpenShift, the Helm chart stores state in PVCs. Treat PVC snapshots or storage-provider backups as infrastructure backups, and use the OpenRAG backup commands below when you need a portable partition-level export.
+
 ## Backup one partition
 
 :::caution
--- a/docs/content/docs/documentation/backup_restore.md
+++ b/docs/content/docs/documentation/backup_restore.md
@@ -6,11 +6,13 @@ title: How to backup or restore OpenRag partition/data ?
 
 ## Storage lifecycle
 
-Docker Compose uses named volumes for OpenRAG state by default: app data, logs, model cache, PostgreSQL, Milvus, etcd, and MinIO. This makes local deployments closer to Kubernetes/PVC behavior and avoids writing runtime data into the repository tree.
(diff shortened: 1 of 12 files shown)


docs/content/docs/documentation/backup_restore.md
```diff
@@ -3,6 +3,17 @@ title: How to backup or restore OpenRag partition/data ?
...
+## Storage lifecycle
+
+Docker Compose uses named volumes for OpenRAG state by default: app data, logs, model cache, PostgreSQL, Milvus, etcd, and MinIO. This makes local deployments closer to Kubernetes/PVC behavior and avoids writing runtime data into the repository tree.
+
+Use `docker compose down` to stop the stack while keeping data. Use `docker compose down -v` only when you intentionally want to remove the named volumes and start from a clean state.
+
+If you need host-mounted storage, set the matching volume variable in `.env` (`DATA_VOLUME`, `DB_VOLUME`, `MILVUS_VOLUME`, `ETCD_VOLUME`, `MINIO_VOLUME`, `LOG_VOLUME`, or `MODEL_WEIGHTS_VOLUME`). In that case, lifecycle and permissions are managed by the host path owner.
+
+For Kubernetes and OpenShift, the Helm chart stores state in PVCs. Treat PVC snapshots or storage-provider backups as infrastructure backups, and use the OpenRAG backup commands below when you need a portable partition-level export.
+
```
Ahmath-Gadji, line 8: This section seems to me to be k8s / OpenShift-deployment-specific.
So I think it's better to create a new doc file `openshift_deployment.md` in `docs/content/docs/documentation`.

docs/content/docs/documentation/env_vars.md
```diff
@@ -195,6 +195,23 @@ The PostgreSQL database is configured using the following environment variables:
...
+|----------|---------|-------------|
+| `DATA_VOLUME` | `appdata` | OpenRAG uploaded files and app data mounted at `/app/data`. |
+| `LOG_VOLUME` | `logs` | OpenRAG logs mounted at `/app/logs`. |
+| `MODEL_WEIGHTS_VOLUME` | `modelweights` | Model cache mounted at `/app/model_weights`. |
+| `VLLM_CACHE` | `modelweights` | Hugging Face cache used by vLLM, reranker, and transcriber services. Set this to a host path only when you intentionally want to reuse a local model cache. |
+| `DB_VOLUME` | `pgdata` | PostgreSQL data mounted at `/var/lib/postgresql/data`. |
+| `ETCD_VOLUME` | `etcd` | Milvus etcd data mounted at `/etcd`. |
+| `MINIO_VOLUME` | `minio` | Milvus object storage data mounted at `/minio_data`. |
+| `MILVUS_VOLUME` | `milvus` | Milvus data mounted at `/var/lib/milvus`. |
+
```
Ahmath-Gadji, line 216: Switching from path-based volumes to named volumes means existing deployments start with empty storage on the next `up`. Data under `../../data`, `../../db`, and `infra/compose/milvus/volumes` is silently ignored (not lost on disk, just unused). This is a breaking change for anyone switching to this configuration.

The question is how to preserve existing data when switching: we'd need a migration script from path-based to named volumes, precisely because this is a breaking change.

Also, given that we don't use the OpenShift/k8s deployments much and that existing instances aren't k8s-based, I suggest keeping the previous values for `VLLM_CACHE`, `DATA_VOLUME`, etc. as the default behaviour, and making the OpenShift version opt-in with the new (named-volume) values to set — e.g. `VLLM_CACHE=modelweights`.

Ahmath-Gadji (commented):
These changes contain breaking changes and shall provide migration script if needed. Also previous configuration could be preserved as default and the openshift one as opt-in

hedhoud:
Addressed the compatibility concern from the review.

The default Docker Compose stack now keeps the existing host-path storage layout, so current local deployments do not move data unexpectedly. The named-volume layout is now explicit opt-in via `infra/compose/.env.named-volumes.example`, with a separate Milvus named-volume include.

I also updated the storage docs to say that no migration is required unless someone chooses to switch an existing deployment to the named-volume profile. In that case, the migration is an intentional host-path to named-volume move, not a default behavior change.

Ahmath-Gadji (approved):

PR #532, opened 2026-06-22
fix: sanitize messages before LLM call

ewan102:
Drop empty assistant messages (defensive guard against upstream that should never happen, logged as error to trace the  cause of the blank) and merge consecutive same-role messages that result from the drop, to avoid an error from the LLM 

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat requests now sanitize outbound LLM message payloads in both streaming and non-streaming modes. Assistant messages with missing, empty, or whitespace-only `content` are replaced with a standard placeholder (excluding tool/function-call assistant turns).

* **Tests**
  * Added unit tests covering sanitization behavior, including preservation of system/user messages, handling multiple patched assistant messages while keeping ordering, treating empty list `content` as empty, and ensuring non-empty multimodal content is unchanged.


ewan102:
waiting for Paul's review... (maybe the solution will change)

ewan102:
Going through all the messages once instead of twice is possible but is less clear for an outside reader so I let the function sanitize like this for now. I can eventually switch it if needed

openrag/services/orchestrators/query_service.py
```diff
@@ -325,6 +325,43 @@ async def _prepare_completions(self, partition: list[str], payload: dict):
...
+        """Sanitize the message list before forwarding to the LLM.
+
+        Replaces ``role="assistant"`` messages whose content is absent or
+        whitespace-only with the placeholder ``"NO_CONTENT"``.  This preserves
+        role alternation (no consecutive same-role messages are created) while
+        acting as a defensive guard against upstream bugs (broken streaming
+        response or client-side history persistence).  Logged at ERROR level so
+        the root cause is traceable.
+
+        Non-string content (e.g. multimodal list) is never replaced.
```
paultranvan: Why not dealing with empty list? We might have the same issue

openrag/services/orchestrators/query_service.py
```diff
@@ -325,6 +325,43 @@ async def _prepare_completions(self, partition: list[str], payload: dict):
...
+            content = msg.get("content")
+            if msg.get("role") == "assistant" and (
+                content is None or (isinstance(content, str) and not content.strip())
+            ):
+                result.append({**msg, "content": "NO_CONTENT"})
+                patched_count += 1
+            else:
+                result.append(msg)
+        if patched_count:
+            logger.error(
```
paultranvan: Actually I think we should have a warning rather than an error. This is not an openrag error, but rather a client mistake

openrag/services/orchestrators/query_service.py
```diff
@@ -325,6 +325,43 @@ async def _prepare_completions(self, partition: list[str], payload: dict):
...
+    @staticmethod
+    def _sanitize_messages(messages: list[dict]) -> list[dict]:
+        """Sanitize the message list before forwarding to the LLM.
+
+        Replaces ``role="assistant"`` messages whose content is absent or
+        whitespace-only with the placeholder ``"NO_CONTENT"``.  This preserves
+        role alternation (no consecutive same-role messages are created) while
+        acting as a defensive guard against upstream bugs (broken streaming
+        response or client-side history persistence).  Logged at ERROR level so
+        the root cause is traceable.
```
paultranvan: This description is too focused on the "what" rather than the "why".
The "what" is not good, because the implem should be self explanatory, and it can change.
I would prefer something like this: 

```
Replace ``role="assistant"`` messages with empty content, as they are
rejected by some LLM. This preserves role alternation (no consecutive 
same-role messages are created). 
Note this should normally never happen, but we do it as a safeguard 
``

File: openrag/services/orchestrators/query_service.py [hunk 4]
```diff
@@ -325,6 +325,43 @@ async def _prepare_completions(self, partition: list[str], payload: dict):
...
+
+        Non-string content (e.g. multimodal list) is never replaced.
+        """
+        result = []
+        patched_count = 0
+        for msg in messages:
+            content = msg.get("content")
+            if msg.get("role") == "assistant" and (
+                content is None or (isinstance(content, str) and not content.strip())
+            ):
```
paultranvan: I forgot that it's actually ok to have empty assistant messages when using tools or function calling. This is not supported by openRAG ATM, but it will be eventually, so we should add it now: 

```
should_replace_msg = (
              msg.get("role") == "assistant"
              and not msg.get("tool_calls")
              and not msg.get("function_call")
              and (content is None or (isinstance(content, str) and not content.strip()))
          )
          if should_replace_msg:
...
```

tests/unit/services/orchestrators/test_query_service.py
```diff
@@ -240,3 +240,141 @@ def test_dedupe_web_preserves_first_seen():
...
+
+    def debug(self, *a, **kw) -> None:
+        pass
+
+
+@pytest.fixture()
+def rec_logger(monkeypatch):
+    rec = _RecordingLogger()
+    monkeypatch.setattr(qs, "logger", rec)
+    return rec
```
paultranvan: I don't think it's worth testing the log ; it's an implementation detail, we should not bother about it.
I think you can remove this part and all `rec_logger` asserts

openrag/services/orchestrators/query_service.py
EnjoyBacon7: Addressed in commit 158332341. `_is_empty_content` now treats an empty list as empty (`if isinstance(content, list): return not content`), so an assistant turn with `content=[]` is replaced too. Covered by `test_sanitize_empty_list_content_replaced`.

openrag/services/orchestrators/query_service.py
EnjoyBacon7: Addressed in commit 158332341 — switched to `logger.warning(...)`. An empty assistant turn is a client-side history issue rather than an OpenRag error, so a warning is the right level.

openrag/services/orchestrators/query_service.py
EnjoyBacon7: Addressed in commit 158332341 — the docstring now leads with the *why* ("Some LLMs reject an assistant turn with no content. Substituting a placeholder preserves role alternation…") and drops the step-by-step description of the implementation.

openrag/services/orchestrators/query_service.py
EnjoyBacon7: Addressed in commit 158332341 — the guard now exempts tool/function-calling turns: `not msg.get("tool_calls") and not msg.get("function_call")`, so a legitimately content-free assistant turn is left untouched. Covered by `test_sanitize_assistant_with_tool_calls_untouched` and `test_sanitize_assistant_with_function_call_untouched`.

tests/unit/services/orchestrators/test_query_service.py
EnjoyBacon7: Addressed in commit 158332341 — removed the log assertions; the tests now assert only the sanitized output, not the logging side effect.

EnjoyBacon7 (approved):
Reviewed. The message sanitization is correct: `_is_empty_content` handles `None` / empty-string / whitespace / empty-list content, the guard exempts `tool_calls`/`function_call` turns, it warns rather than errors, and it runs before both the `chat` and `chat_stream` LLM calls. 12 unit tests cover the cases. All review feedback addressed and CI is green. LGTM.

hedhoud (approved):
Looks good to me. The fix is focused and easy to follow: empty assistant messages are handled before we call the LLM, and the important cases are covered by tests. I don’t see anything blocking here. The more general invalid-history cleanup can be handled later if needed.

PR #536, opened 2026-06-23
fix(query): apply per-partition chat_history_depth

### LINKED ISSUE #534: Per-partition chat_history_depth is stored but never applied (QueryService uses global config)
## Summary

A partition's `chat_history_depth` (settable via the admin UI / `PATCH /partition/{name}`) is **stored and round-trips correctly through the data layer, but is never applied to chat**. Every chat completion truncates conversation history using the **global** `config.rag.chat_history_depth` (default `4`), regardless of the partition's configured value. Changing it per-partition has zero effect.

This is the same class of gap as #522 (contextualization implemented but not wired): present end-to-end in the data layer, ignored by the consumer.

## Wired correctly (data layer) ✅
- API accepts it: `openrag/api/schemas/admin/partition_schemas.py:36`
- DB stores it: `openrag/services/persistence/schema.py:108`, `partition_repo.py`
- Resolved per-partition config carries it: `openrag/core/models/preset.py:55` (`PartitionConfig.chat_history_depth`), cached in `Settings.partitions` (`core/config/root.py:71`)

## Not applied (the bug) ❌
- `QueryService` is a **singleton** built once at startup with the global `config`: `openrag/di/container.py:538`
- `__init__` snapshots the global value: `self._chat_history_depth = config.rag.chat_history_depth` — `openrag/services/orchestrators/query_service.py:117` (global default `4`, `core/config/retrieval.py:94`)
- Applied at `query_service.py:222`: `messages = payload["messages"][-self._chat_history_depth:]`
- `_prepare_chat(self, partition, payload)` receives the partition but **never reads** `config.partitions[partition].chat_history_depth`. The value is never re-resolved per request.

## Secondary footgun (and default mismatch)
- The slice is `messages[-depth:]`; with `depth == 0` that is `messages[-0:]` → `messages[0:]` → **all** history, not none.
- The partition column / `PartitionConfig` default is `0`, while the effective global default is `4` — the defaults disagree. Any fix must define `0` semantics (proposal: `0` = inherit the global default, so `0` never reaches the slice).

## Proposed fix
- Hold a reference to the live `config` in `QueryService` and resolve the effective depth per request from the partition's resolved config, falling back to the global default when the partition value is `0`/absent.
- Multi-partition (`openrag-all`): no single owning partition → fall back to the global default (or `max` of explicit per-partition values).
- Add unit coverage for the resolution (single partition explicit, `0`→inherit, multi-partition, unknown partition).

## Affected
`refactor/hexagonal` (fix target). Likely also present on `main` — worth a follow-up check.

andyne13:
Closes #534.

## Problem
A partition's `chat_history_depth` (set via the admin UI / `PATCH /partition/{name}`) was stored and round-tripped correctly, but **never applied to chat**. `QueryService` is a startup singleton that snapshotted the global `config.rag.chat_history_depth` (default `4`) once and used it for every completion (`query_service.py:117` → slice at `:222`), regardless of partition. Changing the per-partition value had no effect — same class of gap as #522.

## Fix
Resolve the depth **per request** instead of at startup:
- `QueryService` holds a live `config` reference and reads the partition's resolved `chat_history_depth` from `config.partitions` (refreshed on every preset change via `clear()+update()`).
- A partition value of `0` means **inherit the global default**, so it never reaches the `messages[-depth:]` slice — where `0` would otherwise select the *entire* history (`[-0:]` == `[0:]`).
- Multi-partition queries (`openrag-all`) take the largest explicit value, falling back to the global default when none is set.
- Purely additive: behavior is unchanged when no partition value is configured.

## Tests
+4 unit tests in `test_query_service.py` (explicit value used; `0` → inherit default; `None`/unknown → default; multi-partition → max explicit).

## Verification
- `uv run pytest tests/unit/` → **1289 passed**
- `uv run ruff check` → clean

## Note
The same gap likely exists on `main` — worth a separate follow-up check.

## Release Notes

* **New Features**
  * Chat history depth is now resolved dynamically per request, with support for per-partition overrides and inheritance from global defaults when a partition sets `chat_history_depth` to `0`.
  * The `"all"`/`openrag-all` partition sentinel now uses the global default, and multi-partition requests use the maximum explicitly set depth.

* **Tests**
  * Added unit coverage for chat-history depth resolution across single partitions, unknown partitions, the `"all"` sentinel, and multi-partition inputs.

### DIFF
--- a/openrag/services/orchestrators/query_service.py
+++ b/openrag/services/orchestrators/query_service.py
@@ -138,17 +138,23 @@ class QueryService:
         Honors a partition's configured ``chat_history_depth`` (set via the
         admin API) over the global default. A partition value of ``0`` means
         "inherit the global default" — it never reaches the ``messages[-depth:]``
-        slice, where ``0`` would otherwise select the *entire* history. For
-        multi-partition queries (e.g. ``openrag-all``) there is no single owning
-        partition, so the largest explicit value wins, falling back to the
-        global default when none is set.
+        slice, where ``0`` would otherwise select the *entire* history.
+
+        The ``"all"`` sentinel (``openrag-all``) reaches this layer un-expanded
+        (retrieval resolves it to concrete partitions downstream) and is a
+        cross-partition query with no single owning partition, so it uses the
+        global default. A request scoped to one or more named partitions takes
+        the largest explicit (>0) value among them.
         """
-        explicit = [
-            cfg.chat_history_depth
-            for name in (partition or [])
-            if (cfg := self._config.partitions.get(name)) is not None and cfg.chat_history_depth > 0
-        ]
-        return max(explicit) if explicit else self._default_chat_history_depth
+        if partition and "all" not in partition:
+            explicit = [
+                cfg.chat_history_depth
+                for name in partition
+                if (cfg := self._config.partitions.get(name)) is not None and cfg.chat_history_depth > 0
+            ]
+            if explicit:
+                return max(explicit)
+        return self._default_chat_history_depth
 
     # ------------------------------------------------------------------
     # Query generation (was RagPipeline.generate_query — no LangChain)
--- a/openrag/services/orchestrators/query_service.py
+++ b/openrag/services/orchestrators/query_service.py
@@ -113,8 +113,12 @@ class QueryService:
         self._web = web_search_service
         self._workspace = workspace_service
 
+        # Keep a live reference so per-partition config (resolved into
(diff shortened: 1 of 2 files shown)


EnjoyBacon7:
all good on my end

hedhoud (approved):
Looks good to me. The change is scoped, the per-partition depth behavior is covered by tests, and the latest handling for `openrag-all` is clearly documented in the code. CI is green, so this looks mergeable from my side.

Ahmath-Gadji (approved):
LGTM

PR #537, opened 2026-06-23
fix(admin-ui): OIDC logout, same-origin API base, link encoding + numeric guards (with tests)

EnjoyBacon7:
## Summary

Security/correctness fixes surfaced by an audit of the admin UI, each with a regression test. Branches off the current `feat/admin-ui` tip and contains **only these changes** (17 files, +524/−15).

## Bug fixes

- **Revoke the OIDC session on logout.** `logout()` only cleared the localStorage bearer token; in OIDC mode identity comes from the httpOnly `openrag_session` cookie, so the server/IdP session was never ended and `/users/info` re-authenticated on reload. `logout()` now reports whether it was an OIDC/cookie session and the header full-navigates to `GET /auth/logout` (token-mode logout unchanged — which also avoids the 400 `/auth/logout` returns when `AUTH_MODE != oidc`).
- **Default the API base to same-origin.** `API_BASE` fell back to `http://localhost:8000` when `VITE_API_BASE_URL` was undefined, which would send the bearer token cross-origin on a misconfigured build. Now defaults to `""` (relative, same-origin) — fails closed.
- **Encode partition names in detail links.** Raw interpolation broke routing for names containing `/ # ? %`; extracted a `partitionDetailPath()` helper that encodes (single source of truth).
- **Guard numeric form fields against `NaN`.** `parseInt("")`→`NaN`→`null` for `chat_history_depth` / `batch_size` / `timeout`; added `intOr`/`numOr` helpers that fall back to a sane default.

## Tests added

- Wired up a `test` script (`vitest run`) — the suite had no runner before.
- Frontend: logout contract, same-origin/bearer/401/empty-body + `ApiError` envelope parsing, route encoding, `intOr`/`numOr`, the `usePermissions` capability matrix, and `pickDefaultEndpoint`/`resolveEmbedderName`.
- Backend: `GET /partition/` list authorization (no cross-partition leak; super-admin `all`) and the draft `POST /model-endpoints/validate` route.

## Verification

- Frontend: **47 tests / 7 files pass**, `tsc -b` clean, `eslint` 0 warnings.
- Backend: the two router test files pass (15 tests), `ruff` clean.

> Note: no CI job runs `npm test` yet — happy to add a workflow step (`cd ui && npm ci && npm test`) if wanted.

## Summary by CodeRabbit

* **New Features**
  * Enhanced logout flow to properly handle OIDC/SSO sessions with appropriate redirects.
  * Added support for special characters in partition names via proper URL encoding.
  * Improved numeric input validation with sensible fallback defaults.

* **Bug Fixes**
  * Fixed API client to use relative URLs by default for enhanced security.
  * Improved bearer token handling and cleared tokens on authentication failures.

* **Tests**
  * Added comprehensive test coverage for admin routes, API client, authentication, permissions, and utilities.

* **Chores**
  * Added npm test scripts to the UI package.

### DIFF
--- a/tests/unit/api/routers/admin/test_partition_list_route.py
+++ b/tests/unit/api/routers/admin/test_partition_list_route.py
@@ -1,0 +1,103 @@
+"""Authorization + shaping tests for ``GET /partition/`` (list_existant_partitions).
+
+The list route was reworked to return per-partition stored config +
+``document_count`` (via ``PartitionService.list_partition_summaries``). These
+tests pin the security-relevant branch: a regular user must see summaries for
+*only* their own partitions, while the super-admin ``"all"`` sentinel lists
+every partition.
+"""
+
+from __future__ import annotations
+
+import pytest
+from api.dependencies.auth import partitions_with_details
+from api.routers.admin import partitions
+from di.providers import get_partition_service
+from fastapi import FastAPI
+
+
+def _summary(name: str, **overrides) -> dict:
+    base = {
+        "partition": name,
+        "description": "",
+        "embedder": "default",
+        "indexation_preset": "default",
+        "retrieval_preset": "default",
+        "dimension": 1024,
+        "chat_history_depth": 0,
+        "chat_llm": None,
+        "created_at": "2026-01-01T00:00:00+00:00",
+        "document_count": 0,
+    }
+    base.update(overrides)
+    return base
+
+
+class _FakeService:
+    """Minimal partition service exposing only what the list route calls."""
+
+    def __init__(self, summaries: dict[str, dict]) -> None:
+        self._summaries = summaries
+
+    async def list_partition_summaries(self) -> dict[str, dict]:
+        return self._summaries
+
+
+def _build_app(service: _FakeService, principal_partitions: list[dict]) -> FastAPI:
+    app = FastAPI()
+    app.include_router(partitions.router, prefix="/partition")
+    app.dependency_overrides[partitions_with_details] = lambda: principal_partitions
+    app.dependency_overrides[get_partition_service] = lambda: service
+    return app
+
+
+@pytest.mark.asyncio
+async def test_list_returns_only_the_callers_partitions(async_client_factory):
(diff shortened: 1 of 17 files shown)


andyne13 (approved):
LGTM — reviewed against the diff (not just the description). Approving.

Verified:
- **OIDC logout** (`auth.tsx` + `header.tsx`): correct — `wasOidcSession` is read *before* `removeItem` (order matters), and the side-effect lives in the component: header does `window.location.assign("/auth/logout")` for OIDC vs `navigate("/login")` for token mode (which also avoids the 400 `/auth/logout` returns when `AUTH_MODE != oidc`). Both contracts covered by tests. Clean separation — `logout()` stays a pure state mutation that returns a signal.
- **Same-origin API base** (`client.ts` → `?? ""`): fail-closed, good.
- **`intOr`/`numOr`** (`utils.ts`): correct — `0` is preserved; only empty/non-numeric falls back.
- **`partitionDetailPath`** (`routes.ts`): proper `encodeURIComponent`, nice catch.
- All 7 CI checks green; mergeable/clean.

Two **non-blocking** notes (be-aware / possible follow-up, not change requests):

**1. `chat_history_depth: intOr(chatHistoryDepth, 0)` resets to default on edit.**
Clearing the field on an existing partition sends `0` (overwrites the stored value) rather than omitting it (keeping it). That's a defensible "clear = reset to default" choice, and it's coherent once #536 lands — there `chat_history_depth = 0` means "inherit the global default," so clearing → `0` → "use the default" reads correctly. Just flagging the semantic + the mild ordering dependency on #536.

**2. `encodeURIComponent` → `%2F` for names containing `/`.**
The encoding is the right fix and client-side react-router handles it fine. The residual edge: a **hard load** of such a URL (refresh / shared link) hits nginx, and `%2F` in a path segment is rejected/normalized by default in many proxies (`AllowEncodedSlashes off`, `merge_slashes on`) → 404 before the SPA loads. Only reachable if partition names with `/` are actually allowed — so the real fix is upstream name validation (same root as the `all`-sentinel collision). Worth a follow-up, not this PR.

PR #549, opened 2026-06-23
fix(partitions): reserve the `all` partition name to prevent a metadata leak

### LINKED ISSUE #548: Partition named `all` leaks every partition's config to a regular user
## Summary

A regular (non-admin) user whose **sole partition membership is a partition literally named `all`** receives **every partition's** config + document counts from `GET /partition/`, not just their own.

## Root cause

`all` is overloaded as the cross-partition sentinel. `partitions_with_details` (`openrag/api/dependencies/auth.py:27`) returns `[{"partition": "all", ...}]` for admins under `SUPER_ADMIN_MODE`, and the list endpoint expands that to every partition:

```python
# openrag/api/routers/admin/partitions.py:66-67
if len(partitions) == 1 and partitions[0]["partition"] == "all":
    result = list(summaries.values())   # ← every partition's config + counts
```

The branch keys on the **name** `"all"`, not on `is_admin`. Two gaps let a regular user reach it:

1. **No reserved-name guard on create.** `create_partition` (`partitions.py:268`) takes the name from the **URL path** (`partition: str`) and never validates it against reserved sentinels — the `CreatePartitionRequest` body validator doesn't apply to this path-based route. Regular users can create partitions, so a user can create one named `all`, becoming its owner.
2. With that partition as their only membership, their `partitions` list is `[{"partition": "all", role: "owner"}]` → the branch fires → full leak.

## Fix

1. **Reserved-name guard** in the partition-creation path (service layer, so it covers all callers): reject `all` (case-insensitive) with a 4xx. This makes the `"all"` sentinel unambiguous everywhere (also de-risks the `%2F`-encoding and `chat_history_depth` resolver that lean on it).
2. **Defense-in-depth**: gate the all-expansion branch on the user actually being an admin, so even a pre-existing `all` partition can't trigger it.
3. Tests: reserved-name rejection on create (`all`/`ALL`/`All`); the list branch only expands for admins.

## Severity

Medium (security-flavored): cross-tenant metadata disclosure (partition names, configs, document counts) to a regular user. No document *content* is exposed via this path, but partition topology/config is.

andyne13:
Closes #548.

## Problem
A regular (non-admin) user whose **sole partition membership is a partition literally named `all`** received **every partition's** config + document counts from `GET /partition/`.

`all` is overloaded as the cross-partition sentinel: `partitions_with_details` returns `[{"partition": "all"}]` for admins under `SUPER_ADMIN_MODE`, and the list route expanded that to every partition (`partitions.py:66`). The branch keyed on the **name** `"all"`, not on `is_admin` — and `create_partition` takes the name from the URL path with **no reserved-name guard**, so a regular user (who can create partitions) could create one named `all` and trigger the expansion.

## Fix
1. **Reserved-name guard (root fix)** — `PartitionService.create_partition` now rejects `all` (case-insensitive, trimmed) with `400 RESERVED_PARTITION_NAME`, at the service layer so it covers every caller. No new partition can collide with the sentinel.
2. **Defense-in-depth** — the all-expansion branch in the list route now also requires `is_admin`, so even a *legacy* `all` partition owned by a regular user can't leak every partition.

## Tests (+6)
- `create_partition` rejects `all` / `ALL` / `All` / `"  all  "` (never written)
- list route **expands** `all` for an admin
- list route does **not** expand `all` for a non-admin

`uv run pytest tests/unit/` → **1311 passed** · `ruff` clean.

## Notes
- Only **create** is guarded — there is no partition-rename endpoint.
- The reserved set is currently `{"all"}`, structured to extend if other sentinels appear.
- This also de-risks the `all` sentinel that the per-partition `chat_history_depth` resolver (#536) and partition-link encoding lean on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted the special `all` partition behavior to admins, so non-admin users no longer trigger a full partition listing by using that value.
  * Prevented creation of partitions named `all` (including case-insensitive or whitespace variants), returning a validation error instead.
* **Tests**
  * Added unit and API route tests covering the admin vs non-admin `all` expansion behavior and rejection of the reserved partition name.

### DIFF
--- a/openrag/api/routers/admin/partitions.py
+++ b/openrag/api/routers/admin/partitions.py
@@ -59,11 +59,18 @@ Returns a list of partitions you have access to, including:
 """,
 )
 async def list_existant_partitions(
+    request: Request,
     partitions=Depends(partitions_with_details),
     service=Depends(get_partition_service),
 ):
     """List partitions visible to the current user."""
-    if len(partitions) == 1 and partitions[0]["partition"] == "all":
+    # The ``all`` entry is the admin/SUPER_ADMIN_MODE sentinel from
+    # partitions_with_details. Gate the all-expansion on the caller actually being
+    # an admin, so a (legacy) partition literally named ``all`` owned by a regular
+    # user cannot leak every partition. New ``all`` partitions are already rejected
+    # at creation (_RESERVED_PARTITION_NAMES).
+    is_admin = bool(request.state.user.get("is_admin"))
+    if is_admin and len(partitions) == 1 and partitions[0]["partition"] == "all":
         partitions = await service.list_partitions()
     logger.debug("Returned list of existing partitions.", partition_count=len(partitions))
     return JSONResponse(status_code=status.HTTP_200_OK, content={"partitions": partitions})
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -50,6 +50,12 @@ if TYPE_CHECKING:
 
 logger = get_logger()
 
+# Names reserved as cross-partition sentinels (e.g. ``openrag-all`` /
+# ``?partitions=all``). A real partition named ``all`` collides with the sentinel
+# and the admin partition-list route would expand it to *every* partition — see
+# ``list_existant_partitions``. Matched case-insensitively.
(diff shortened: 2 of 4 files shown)


Ahmath-Gadji (approved):
LGTM.
I've tested this with admin and regular user and the fix works.

PR #553, opened 2026-06-23
docs(oidc): clarify OIDC_REDIRECT_URI is the UI front door, not the API port

EnjoyBacon7:
## Why

`OIDC_REDIRECT_URI` has a silent footgun once the **admin-ui (nginx)** is the front door.

The rule people assume — *"it's the FastAPI backend's base URL"* — is only true when the backend also serves the UI you land on (e.g. Chainlit on `APP_PORT`). With the bundled admin-ui, the UI is served by **nginx** (which also proxies `/auth/*` to the backend), so the front door is the **admin-ui port, not `APP_PORT`**.

If you point the redirect at the bare API port anyway:
- login **completes** (the API does host `/auth/callback`), but
- the post-login redirect is **relative** (`/app/`), so the browser lands back on the API port, which **doesn't serve `/app/`** → a blank/JSON page.

**No error is shown.** That's the silent failure this PR documents.

## What changed

Docs-only — every place the variable is mentioned now states that `OIDC_REDIRECT_URI` is the **front door that serves the UI *and* reaches `/auth/callback`**, and names the silent failure mode:

| File | Change |
|------|--------|
| `docs/.../oidc.md` | Rewrote the **"Choosing `OIDC_REDIRECT_URI`"** section into three deployment shapes (admin-ui / backend / reverse proxy) with the two must-be-true conditions; updated the env-table row |
| `docs/.../sso-quickstart.md` | Distinguish a **reverse-proxy front (valid — the admin-ui)** from a **static SPA front (infinite loop)**; flag the bare-backend-port pitfall |
| `infra/compose/.env.example` | Expanded the `OIDC_REDIRECT_URI` comment with the 3 cases; documented `ADMIN_UI_PORT` |
| `CLAUDE.md` | Tightened the env-table description |
| `openrag/core/config/auth.py` | Comment on the `redirect_uri` config field pointing at the docs |

The prior docs said *"always the backend, never the front"* — correct only for a **dumb SPA** front. The admin-ui is a **reverse-proxy** front (serves `/app/` *and* proxies `/auth/callback`), so it's the **recommended** redirect target, not a forbidden one. The docs now say that explicitly.

## Notes
- No code/behavior change; the validator and runtime are untouched.
- Base is `feat/admin-ui` (same as the other admin-ui PRs, e.g. #537).

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to set the OIDC redirect URI to the UI “front door” that routes to the backend `/auth/callback`, with guidance for bundled admin UI, reverse proxy, and backend-direct setups.
  * Expanded troubleshooting and failure-mode explanations for mispointed redirect URIs (e.g., blank/404/JSON pages and login loops).
  * Updated the SSO quickstart and configuration examples (including environment template comments) to reflect the corrected front-door vs backend routing expectations.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -385,7 +385,7 @@ OpenRag supports two authentication modes, controlled by the `AUTH_MODE` environ
 | `OIDC_ENDPOINT` | Issuer URL for auto-discovery | `https://idp.example.com/realms/openrag` |
 | `OIDC_CLIENT_ID` | Client registered at IdP | `openrag` |
 | `OIDC_CLIENT_SECRET` | Client secret | (provided by IdP) |
-| `OIDC_REDIRECT_URI` | Callback URL (must match IdP config) | `https://openrag.example.com/auth/callback` |
+| `OIDC_REDIRECT_URI` | Callback URL — the **front door** that serves the UI *and* reaches the backend `/auth/callback` (the admin-ui / proxy port, **not necessarily** `APP_PORT`); must match IdP config | `https://openrag.example.com/auth/callback` |
 | `OIDC_TOKEN_ENCRYPTION_KEY` | Fernet key for token encryption | (generate via: `python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"`) |
 
 **Optional Env Variables**:
--- a/docs/content/docs/documentation/oidc.md
+++ b/docs/content/docs/documentation/oidc.md
@@ -129,19 +129,19 @@ Pick the value by deployment shape:
 
 - **Bundled admin-ui (nginx) is the front door** — it serves the React SPA at `/app/` and proxies `/auth`, `/v1`, … to the backend. Use the **admin-ui port** (`ADMIN_UI_PORT`), **not** `APP_PORT`. Both ports host a working `/auth/callback`, but only the admin-ui port *also* serves `/app/`, so only it lands you back on the UI:
 
-  ```
+  ```bash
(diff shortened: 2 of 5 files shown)


hedhoud:
The new “front door” wording matches the admin-ui setup much better and should prevent the silent post-login landing issue we saw when the redirect points at the bare API port.

I only see two small doc tweaks worth making before merge:

- For the back-channel logout URL, I’d clarify that it is called by the IdP server, not by the user’s browser. Using the same admin-ui/proxy origin is fine when that origin forwards `/auth/backchannel-logout` and is reachable from the IdP, but the important requirement is server-side reachability to the backend logout endpoint.
- In the quickstart, the backend is described as the container running `openrag/api.py`. That path is stale on this branch, so I’d make it layout-neutral, for example “the OpenRAG API backend container”.

Nothing blocking from my side. The main redirect guidance looks correct, especially for the bundled admin-ui/nginx deployment.

Ahmath-Gadji:
@hedhoud thanks for the review — both points addressed in 995b2500:

- **Stale `openrag/api.py` path** → made layout-neutral ("the OpenRag API backend container"). The file doesn't exist on this branch anymore (`openrag/api/` package), so the literal path was misleading.
- **Back-channel logout** → clarified it's a **server-to-server call from the IdP**, not a browser redirect, so it has no UI front-door requirement — only server-side reachability to the backend's logout endpoint. Updated both the warning block and the table row (which previously said "same origin as above").

Also folded in CodeRabbit's nitpick (language specifiers on the three `OIDC_REDIRECT_URI` fenced blocks) and added a troubleshooting entry for the silent blank-page-after-login symptom this PR documents.

PR #555, opened 2026-06-24
fix(auth): revoke OIDC sessions on user access changes

### LINKED ISSUE #486: [security] OIDC sessions not revoked on user token regeneration / de-admin
**Severity:** Low

Rotating a user's `users.token` (`regenerate_user_token`) or reducing privileges via PATCH does not touch the `oidc_sessions` table. De-admin does take effect on the next request (the middleware re-reads the current user row), but there is no way to forcibly kill an active `openrag_session` cookie short of deleting the user or waiting for `session_expires_at`.

This is an incident-response gap ("revoke this user's access now").

**Fix:** add `revoke_oidc_sessions_by_user_id` and call it from `regenerate_user_token`; optionally expose an admin "revoke sessions" action.

hedhoud:
## Why

#486 is an incident-response gap: changing a user's access should not leave their existing browser session alive until it naturally expires.

The refactor branch already had the persistence method to revoke OIDC sessions by user, but the user-management flow was not calling it.

## What changed

This wires the revocation into the two access-change paths that matter here:

- rotating a user's API token
- demoting an admin user back to a regular user

Regular profile edits keep the current behavior and do not revoke sessions.

Closes #486.

## Summary by CodeRabbit

* **New Features**
  * Regenerating a user token now revokes the user’s active OIDC sessions.
  * Updating a user from admin to non-admin now revokes that user’s active OIDC sessions.

* **Bug Fixes**
  * Improved missing-user handling during user updates to prevent silent failures.
  * Logging now includes the number of revoked OIDC sessions for clearer auditing.

* **Tests**
  * Added and extended unit tests to verify session revocation delegation and behavior.

### DIFF
--- a/openrag/services/orchestrators/auth_service.py
+++ b/openrag/services/orchestrators/auth_service.py
@@ -364,6 +364,12 @@ class AuthService:
     async def revoke_oidc_session_by_id_for_request(self, session_id: int) -> None:
         await self._oidc_session_repo.revoke_session(session_id)
 
+    async def revoke_oidc_sessions_by_user(self, user_id: int) -> int:
+        """Revoke every active OIDC browser session for a user."""
+        count = await self._oidc_session_repo.revoke_by_user(user_id)
+        logger.info("Revoked OIDC sessions by user", user_id=user_id, count=count)
+        return count
+
     async def refresh_session_if_needed(self, *, session: dict[str, Any], enc_key: str) -> dict[str, Any] | None:
         """Refresh the IdP access token when it is near expiry.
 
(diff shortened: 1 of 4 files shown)


hedhoud:
Addressed the valid consistency issue.

The revocation now happens before the access mutation, so a revocation failure no longer leaves a rotated token or admin demotion committed while old OIDC browser sessions remain active. This keeps the flow fail-closed without adding a larger cross-repository unit-of-work in this PR.

I also added ordering coverage around token regeneration and admin demotion.

Ahmath-Gadji (approved):
LGTM.
Revocation works. 
I haven't tested it though with the new admin-ui to see the consequences on front

PR #557, opened 2026-06-24
fix(helm): harden chart security defaults

### LINKED ISSUE #482: [security] Helm chart hardening: placeholder secrets, no securityContext, no Ingress TLS
**Severity:** Medium

`charts/openrag-stack/`:

- **Placeholder secrets rendered unconditionally** into the Secret: `values.yaml:350 AUTH_TOKEN: "sk-xxxx"`, `values.yaml:42 POSTGRES_PASSWORD: "CHANGE_ME_STRONG_PASSWORD"`, plus `API_KEY`/`VLM_API_KEY: "sk-xxxx"` and `HF_TOKEN: "hf_xxxx"`. A `helm install` with no overrides brings up the API with a publicly-known master bearer token and DB password. Gate `AUTH_TOKEN` and `POSTGRES_PASSWORD` behind `{{ required }}`/`fail`.
- **No `securityContext` on any workload** (Deployments + RayCluster). The image runs as UID 10001 but the chart does not enforce `runAsNonRoot` / `allowPrivilegeEscalation: false` / `drop: [ALL]` / `seccompProfile`. Fails the restricted Pod Security Standard. A `serviceAccountName` helper exists but is unused; `automountServiceAccountToken: false` is absent.
- **No Ingress TLS.** `ingress.yaml` has no `tls:` block and `values.yaml:268 host: ""` matches any Host — the bearer token, uploads, and RAG responses traverse the network in plaintext.

**Fix:** `required`-gate the secrets, add a restricted `securityContext` to all workloads, and add a `tls:` block to the Ingress.

hedhoud:
## Summary

This hardens the OpenRAG Helm chart so a default install can no longer accidentally expose known placeholder credentials or deploy chart-owned workloads without basic restricted security settings.

The chart now fails fast when required secrets are missing or still use common placeholder values. The PostgreSQL password is no longer rendered through the shared ConfigMap, ingress is opt-in instead of wildcard-by-default, and TLS can be configured when ingress is enabled.

The chart-owned workloads also get restricted pod/container security defaults, including non-root execution, no service account token mount by default, no privilege escalation, dropped capabilities, and a runtime seccomp profile. The vLLM dependency is overridden through values so its engine pods follow the same non-root default.

Closes #482

## Validation

- ............                                                             [100%]
12 passed in 0.09s
- All checks passed!
- 3 files already formatted
- 
- Dockerized Helm lint with required secrets and ingress values
- Dockerized Helm render with required secrets and ingress TLS values

### DIFF
--- a/infra/charts/openrag-stack/Chart.yaml
+++ b/infra/charts/openrag-stack/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
+version: 0.4.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/infra/charts/openrag-stack/templates/indexer-ui.yaml
+++ b/infra/charts/openrag-stack/templates/indexer-ui.yaml
@@ -18,10 +18,15 @@ spec:
         app.kubernetes.io/name: indexer-ui
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      automountServiceAccountToken: {{ .Values.security.automountServiceAccountToken }}
+      securityContext:
+        {{- toYaml .Values.security.podSecurityContext | nindent 8 }}
       containers:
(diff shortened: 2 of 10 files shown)


hedhoud:
Closing this fork-based PR so I can reopen the same fix from the origin branch, matching the team workflow.

Issue #554, opened 2026-06-23
Indexer can't resolve admin-UI-registered LLM endpoints; contextualization hard-fails indexing


andyne13:
## Problem
Indexing a file fails hard when the indexation preset enables **contextualization** with an LLM endpoint that was registered via the **admin UI**:

```
KeyError: "Unknown llm '<registered-endpoint-name>'. Available: []"
  indexer_pool.py:223 in factory  (_build_contextualizer_factory)
```

## Root cause (two issues)
1. **Admin-UI-registered endpoints never reach the indexer.** `ModelEndpointService.load_all()` hydrates `Settings.models.llm` from the DB, but it only runs in the **API process** (`di/container.py:188`). The **Indexer is a separate Ray actor** whose `__init__` does `cfg = load_config()` (fresh from `conf/config.yaml`, no DB endpoints) and builds `_build_contextualizer_factory(cfg)` from that empty `cfg.models.llm`. So a contextualization LLM that exists in the DB (and works for query/chat) is invisible to the indexer — `Available: []`. A restart does not fix it.
2. **No graceful degradation.** `_select_contextualizer` calls `contextualizer_factory(name)` which raises, and the error propagates up to fail the entire file. Contextualization is an *enhancement* — a missing/unresolvable LLM should not block ingestion.

## Fix
- **Graceful degrade (this PR):** if the contextualization LLM can't be resolved, log a warning and skip contextualization for that run instead of hard-failing the file. (Same applies to topic-tagging.)
- **Hydrate indexer endpoints (follow-up):** make admin-UI-registered endpoints available to the indexer actor (load DB endpoints into its `cfg.models`, or resolve via the shared model-endpoint service), so a named contextualization endpoint actually works — not just `default`.

## Impact
Backend only; the admin UI is correct (the endpoint is registered and works for query). Surfaced on a deploy after contextualization was wired in (#522/#524).

andyne13:
Fixed by #560 (targets `refactor/hexagonal`).

The indexer Ray actor now hydrates the DB-backed model-endpoint registry, so contextualization / topic tagging configured with an LLM endpoint registered via the admin UI resolves correctly instead of failing with `KeyError: Unknown llm '<name>'`. A missing/unresolvable enhancement LLM now degrades gracefully (skip + warn) rather than failing the file.

Closing manually since #560 targets `refactor/hexagonal` rather than the default branch, so merge won't auto-close.

PR #560, opened 2026-06-24
fix(indexer): hydrate model-endpoint registry so admin-UI-registered LLM endpoints resolve

andyne13:
## Problem
Contextualization or topic tagging configured with an LLM endpoint **registered via the admin UI** (i.e. stored in the DB) made indexing fail with `KeyError: Unknown llm '<name>'`. The global/default LLM path worked; only DB-registered named endpoints failed.

Root cause: the indexer Ray actor builds its enhancement-LLM factories from `Settings.models`, but that registry is hydrated from the DB **only in the API process at startup** — never in the indexer process. So a named endpoint never resolved there, and a restart didn't help.

## Fix
- Hydrate `cfg.models` from the DB inside the indexer actor, lazily: once on first use, on a 60s TTL refresh (run in the background, off the file's critical path), and on a cache miss (rate-limited so an unknown/typo'd name can't storm the DB). Reloads are single-flight.
- The contextualizer/topic-tagger factories read the live registry and cache one client per endpoint. The cache key is a **hash of the full endpoint config** — endpoint URL, model, *and* every `extra` key (`api_key`, `temperature`, …) — so any edit, including an api-key rotation that keeps the same URL/model, yields a new identity and rebuilds the client on the next reload (≤60s), without leaking the old one. This matches the API process's invalidate-on-any-change behaviour.
- A missing/unresolvable enhancement LLM degrades gracefully (warn + skip) rather than failing the whole file.
- The actor logger is an instance attribute, not a module global, so Ray's by-value pickling of the actor class doesn't pull in loguru's `enqueue=True` `SimpleQueue`.

## Freshness semantics
- **New endpoint, used immediately** → resolved on first use via reload-on-miss (not gated by the TTL).
- **Edit / delete / change-default of an existing endpoint** → propagates within ≤60s (TTL); the client rebuilds because the cache key covers the full config.
- Instant (sub-60s) propagation via an API→actor reload signal is possible but intentionally deferred (it would couple the orchestrator to Ray).

## Testing
- Unit tests: registry-reload decisions (initial / hit / ttl / miss + rate-limit), single-flight and background TTL reload, live-registry resolution, edit-aware client cache (URL/model change **and** api-key-only rotation), full-config identity hashing, and graceful-degrade skip.
- Verified on a running stack: named-endpoint contextualization and topic tagging index to completion, and indexing is performance-neutral vs. the pre-fix path (identical per-stage timings).

Refs #554

## Summary by CodeRabbit

* **New Features**
  * Indexing now resolves required LLM endpoint settings on demand, allowing processing to begin even if full LLM configuration becomes available later.
  * Contextualization and topic tagging now initialize lazily and refresh automatically when endpoint settings change, including API key updates.

* **Bug Fixes**
  * Improved LLM registry refresh with TTL-based background updates, rate-limited retries for missing endpoints, and single-flight reloads to avoid spikes.
  * If an LLM can’t be resolved, contextualization/topic tagging are skipped with warnings; other contextualization errors still surface instead of being suppressed.

### DIFF
--- a/openrag/services/workers/indexer_pool.py
+++ b/openrag/services/workers/indexer_pool.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import threading
+import time
 from types import SimpleNamespace
 from typing import Any
 
@@ -9,6 +10,11 @@ import ray
 from services.workers.indexer_actor import IndexerWorker
 
 from openrag.core.config.root import Settings
+
+# The indexer reloads the DB-backed model-endpoint registry at most once per
+# this window (and on a miss), bounding both staleness and DB load regardless
+# of indexing throughput.
+_MODEL_REGISTRY_TTL_SECONDS = 60.0
 
 
 @ray.remote
@@ -20,6 +26,7 @@ class IndexerPool:
         import services.inference.vllm_client  # noqa: F401
         from core.config import load_config
         from core.embeddings import embedder_registry
+        from core.utils.logging import get_logger
         from services.storage.milvus_store import MilvusVectorStore
         from services.storage.postgres_store import PostgresStore
         from services.workers.parsers.parser_dispatcher import build_caption_vlm, build_parser_dispatcher
@@ -63,6 +70,27 @@ class IndexerPool:
         self._catalog_store = PostgresStore(rdb_cfg, run_migrations=False)
         self._catalog_initialized = False
         self._catalog_init_lock = asyncio.Lock()
+        # Model-endpoint registry hydration. Unlike the API process, the indexer
+        # never runs ModelEndpointService at startup, so cfg.models starts empty
+        # and named endpoints (registered via the admin UI) can't resolve here.
+        # We hydrate cfg.models from the DB lazily on first use — see
+        # _ensure_registry_fresh. The factories above hold a live reference to
+        # cfg.models.* so in-place hydration becomes visible to them.
+        # Instance-level logger: kept off the module so Ray's by-value pickling
+        # of this actor class never drags in loguru's enqueue SimpleQueue (a
+        # module-global logger referenced by a method makes the class
+        # unpicklable). Created here, in the actor process, it is never pickled.
+        self._logger = get_logger()
+        self._cfg = cfg
+        # Whether "default" resolves via the global cfg.llm fallback (static —
+        # cfg.llm doesn't change). Lets _reload_decision avoid a perpetual
+        # reload-on-miss for "default" when no is_default row exists in the DB.
+        self._has_default_fallback = _global_llm_endpoint_config(cfg) is not None
+        self._model_endpoint_service: Any = None
+        self._registry_loaded_at: float | None = None
+        self._last_miss_reload_at: float | None = None
+        self._registry_lock = asyncio.Lock()
+        self._registry_reload_task: asyncio.Task[None] | None = None
         self._worker = IndexerWorker(
             pipeline=pipeline,
             task_state_manager=task_state_manager,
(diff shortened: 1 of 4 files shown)


andyne13:
Addressed CodeRabbit's note: narrowed both enhancement-skip blocks from `except Exception` to `except KeyError` — that's the error the factory raises for an unresolvable endpoint name, the only case we want to skip. Other factory errors (missing prompt template, bad client/implementation config, …) now surface instead of being silently masked. Added a regression test (`test_pipeline_propagates_non_keyerror_contextualizer_factory_error`) asserting a non-KeyError factory error propagates and fails the file.

andyne13:
Addressed: `_registry_reload_decision` now checks `miss` before `ttl`. A missing required endpoint blocks once to resolve (so the triggering file gets the LLM stage) even when the registry is also stale, instead of falling through to a non-blocking `ttl` refresh and skipping. Rate-limiting is preserved — a rate-limited miss that's also stale still does a background `ttl` refresh. Added decision-table cases for both (missing+stale → `miss`; rate-limited-miss+stale → `ttl`).

hedhoud (commented):
I did an impact pass beyond the happy path. This looks safe for the issue it targets: the indexer can now hydrate DB-backed LLM endpoints, so contextualization and topic tagging can resolve endpoints created from the admin UI. The graceful-degrade behavior also makes sense here: a missing enhancement LLM should not fail the whole file.

I also checked the adjacent paths around API startup, the model-endpoint service, the DI container, dispatcher, and indexer worker. I did not see a regression there, and the focused local tests plus the GitHub checks are green.

One scope note: this fixes the LLM-backed enhancement path. Named embedders and VLMs registered through the admin UI still look like a separate follow-up if we expect per-preset embedder/VLM selection to work fully inside the indexer. I would not block this PR on that, but I would track it separately.

andyne13:
Ran an independent multi-lens review pass over the diff. Fixed three items it surfaced:

1. **Perpetual reload-on-miss for `default` (real, non-converging).** `_reload_decision` computed `missing` purely from `cfg.models.llm`, but the factory also resolves `default` via the global `cfg.llm` fallback. In a deployment with a global LLM but no `is_default` DB row (so no `default` alias), topic tagging worked via fallback yet the registry was deemed perpetually stale → a blocking reload every 60s that never converged. Now `_reload_decision` mirrors the fallback (`_has_default_fallback`, computed once).
2. **Atomic lazy build** of the contextualizer's shared prompt+semaphore, so a partial init can't leave a half-populated dict that would later raise a stray `KeyError` (which the `except KeyError` skip would misread as an unresolvable endpoint).
3. **Docstring correctness**: the cache identity is the full endpoint config (incl. `extra`/api_key), and the superseded client is dropped but not explicitly closed (rare; reclaimed on actor exit) — corrected the earlier 'without leaking' claim.

Added regression tests for the fallback-resolvability case. 28 unit tests pass; actor still cloud-pickles with the enqueue sink active.

Noted but intentionally **not** in scope here: a dedicated `UnknownEndpointError` instead of the builtin `KeyError` sentinel (durable hardening; current catch is verified safe), and extending reload-on-miss to embedder/VLM names (bounded to the 60s TTL today).

hedhoud (commented):
Re-reviewed the updated head `56446fda` after the extra fallback/reload hardening.

The new `default` fallback handling looks correct: the reload decision now matches the factory resolution path, so a deployment that only has the global LLM fallback should not keep doing a blocking reload every TTL window. The lazy prompt/semaphore build is also safer now because it avoids publishing half-built shared state.

I reran the focused worker/indexer suites and the adjacent model-endpoint/container/dispatcher tests locally, plus ruff, diff whitespace, and a Ray cloudpickle smoke for the actor class. I do not see a blocker from this update.

Same scope note as before: embedder/VLM endpoint hydration for indexation presets still looks like a separate follow-up, not something I would block this LLM-contextualization fix on.

hedhoud (approved):
Approving this from my side.

I tested the current head with an isolated stack and exercised the Admin API flow that the UI depends on: create a model endpoint, create a preset with image captioning + contextual retrieval + topic tagging enabled, assign it to a partition, then upload PDFs through the indexer.

The E2E check passed with 10 PDFs submitted concurrently:

- all uploads returned successfully
- all 10 tasks reached COMPLETED
- 0 failed tasks
- indexer-ui showed COMPLETED (10), FAILED (0)
- contextualization was visible in the indexed chunks through the stored context and [CONTEXT] prefix
- topic tags were persisted for all 10 files

GitHub checks are green as well. The remaining comments look like follow-up/nitpick-level hardening, not blockers for this fix.

PR #561, opened 2026-06-24
fix: make chat-template thinking configurable

### LINKED ISSUE #559: chat_template_kwargs={"enable_thinking": False} is hardcoded and breaks Mistral-tokenizer models, comment-out is not a good fix
The LLM/VLM calls pass extra_body={"chat_template_kwargs": {"enable_thinking": False}}, a Qwen-specific chat-template kwarg. When the configured model uses a Mistral tokenizer (e.g. Mistral-Small-3.2-24B-Instruct-2506-FP8 served via vLLM/litellm), this raises:

litellm.BadRequestError: OpenAIException -
chat_template is not supported for Mistral tokenizers.
Received Model Group=Mistral-Small-3.2-24B-Instruct-2506-FP8
→ the gateway returns 400 → OpenRAG returns 500 on the chat-completion → every RAG query fails (no answer, empty generation/retrieval).

**Affected versions**
≤ v1.1.9: the line ships active (e.g. openrag/components/pipeline.py:164 in v1.1.9) → any deployment against a Mistral-tokenizer model is fully broken (all queries 500).

current main / v1.1.11: the line is commented out at openrag/components/pipeline.py:251 and openrag/components/indexer/loaders/base.py:50.

**Why commenting it out is insufficient**
**Regression for the intended models:** enable_thinking: False exists to suppress reasoning traces on Qwen3-style models. Commenting it out means those models now emit <think>…</think> content into answers/contextualized queries/captions.
**Not configurable:** the behavior is a hardcoded literal , operators can't opt in (Qwen) or out (Mistral) per deployment.
**Scattered:** the same literal appears at multiple call sites (generation pipeline + loader/VLM), so it's easy to leave one active (which is how ≤v1.1.9 breaks).
**Silent:** there's no config surface or doc noting that the generation model must (not) be a Mistral tokenizer.

** A proposed fix**

Add an opt-in flag on the LLM/VLM config (openrag/config/models.py — LLMParamsConfig, line 59) rather than a magic literal, e.g.:

```
class LLMParamsConfig(ConfigMixin):
    temperature: float = 0.1
    timeout: int = 60
    max_retries: int = 2
    logprobs: bool = True
    enable_thinking: bool | None = None   # None → don't send chat_template_kwargs at all
```

Then at each call site, inject only when set:

```
extra_body = {}
if self.config.llm.enable_thinking is not None:
    extra_body["chat_template_kwargs"] = {"enable_thinking": self.config.llm.enable_thinking}
# pass extra_body only if non-empty
```

This way Qwen deployments set enable_thinking: false, Mistral deployments leave it unset (default), and there's a single source of truth instead of scattered commented literals.

hedhoud:
## Why

Mistral-tokenizer deployments fail when OpenRAG sends Qwen-specific chat-template kwargs by default. Removing the option entirely is also a regression for deployments that rely on suppressing Qwen-style reasoning traces.

## Change

This makes the thinking control opt-in across the affected LLM and VLM paths. Mistral deployments omit it by default, while Qwen-style deployments can enable the setting explicitly.

Fixes #559.

## Verification

- `uv run --no-env-file pytest tests/unit -q`
- `AUTH_TOKEN=test-admin-token OPENRAG_API_URL=http://localhost:8080 uv run --no-env-file pytest tests/integration/api -q --tb=short`
- `uv run --no-env-file ruff check ...`
- `uv run --no-env-file ruff format --check ...`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added optional “thinking” controls for LLM, VLM, and the OpenAI-compatible OCR loader.
  * These settings are now respected across chat, streaming, and image captioning request payloads.
* **Bug Fixes**
  * Token counting and endpoint seeding no longer pass unsupported thinking parameters.
  * Environment overrides for the new options are consistently applied.
* **Documentation**
  * Updated environment-variable docs and the example compose config with the new thinking toggles.
* **Tests**
  * Added/expanded unit tests to verify request shaping and override behavior.

### DIFF
--- a/openrag/services/workers/parsers/parser_dispatcher.py
+++ b/openrag/services/workers/parsers/parser_dispatcher.py
@@ -174,7 +174,7 @@ class ParserDispatcher(DocumentParser):
             ocfg.model,
             ocfg.api_key,
             ocfg.timeout,
-            getattr(ocfg, "enable_thinking", None),
+            ocfg.enable_thinking,
         )
         client = DotsOCRPdfClient(vlm, concurrency_limit=ocfg.concurrency_limit)
         return _create("core.indexing.parsers.pdf.client_based", "pdf_client", client=client)
@@ -262,7 +262,7 @@ def build_caption_vlm(config: Any) -> Any | None:
         vlm_cfg.model,
         vlm_cfg.api_key,
         vlm_cfg.timeout,
-        getattr(vlm_cfg, "enable_thinking", None),
+        vlm_cfg.enable_thinking,
     )
 
 
--- a/openrag/services/workers/parsers/parser_dispatcher.py
+++ b/openrag/services/workers/parsers/parser_dispatcher.py
@@ -169,7 +169,13 @@ class ParserDispatcher(DocumentParser):
         from services.inference.parsers.dotsocr import DotsOCRPdfClient
 
         ocfg = self._config.loader.openai
-        vlm = _build_vlm(ocfg.base_url, ocfg.model, ocfg.api_key, ocfg.timeout)
+        vlm = _build_vlm(
+            ocfg.base_url,
+            ocfg.model,
+            ocfg.api_key,
+            ocfg.timeout,
+            getattr(ocfg, "enable_thinking", None),
+        )
         client = DotsOCRPdfClient(vlm, concurrency_limit=ocfg.concurrency_limit)
         return _create("core.indexing.parsers.pdf.client_based", "pdf_client", client=client)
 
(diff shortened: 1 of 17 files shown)


openrag/services/workers/parsers/parser_dispatcher.py
```diff
@@ -169,7 +169,13 @@ def _build_pdf_client(self) -> DocumentParser:
         from services.inference.parsers.dotsocr import DotsOCRPdfClient
 
         ocfg = self._config.loader.openai
-        vlm = _build_vlm(ocfg.base_url, ocfg.model, ocfg.api_key, ocfg.timeout)
+        vlm = _build_vlm(
+            ocfg.base_url,
+            ocfg.model,
+            ocfg.api_key,
+            ocfg.timeout,
+            getattr(ocfg, "enable_thinking", None),
```
Ahmath-Gadji: i think we can use ocfg.enable_thinking directly

Ahmath-Gadji (approved):
LGTM.
I've tested it. Nothing is broken on my side

PR #564, opened 2026-06-24
fix(indexer): resolve named embedder and VLM endpoints

### LINKED ISSUE #562: fix(indexer): resolve named embedder/VLM endpoints from the hydrated registry (fail loud on embedder mismatch)
## Background
#560 fixed the indexer Ray actor to hydrate the DB-backed model-endpoint registry and resolve **named LLM** endpoints (contextualization / topic tagging). The same root cause still affects other model types at index time, because the indexer's factories are built in `IndexerPool.__init__` **before** the lazy registry hydration.

## Problem
- **Embedder — high severity (silent data corruption).** `_build_embedder_factory(cfg)` returns `None` when `cfg.models.embedder` is empty at `__init__` (it is, pre-hydration), so the factory is captured as `None`. `_select_embedder` then silently falls back to the global default embedder (`cfg.embedder` from env). A partition configured with a **named** embedder is silently ignored at index time → documents embedded with the wrong model → vectors incompatible with the query-side embedder → **silently broken retrieval**, with no error raised.
- **VLM — lower severity.** `build_indexer_pool` passes no `vlm_factory` (only a default VLM built from env `cfg.vlm`), so a per-partition named VLM is ignored for index-time image captioning, and admin-UI changes to the VLM endpoint never reach the indexer.

**Not affected:** the **reranker** (retrieval-only, runs in the API process) and the **query/API process** in general (it hydrates at startup and invalidates client caches on every endpoint CRUD).

## Fix
- Give the embedder factory (and a new VLM factory) the same treatment the LLM factories got in #560: read the **live** `cfg.models.*` dict instead of capturing an empty snapshot / early-returning `None` at build time, and cache one client per endpoint **identity** (full-config hash, like `_endpoint_identity`) so endpoint edits propagate after a reload.
- **Embedder must fail loud, never silent-fallback:** an enabled named embedder that cannot be resolved must raise and fail the file — silently embedding with a different model (wrong vectors) is worse than a failed file. (This differs from the LLM enhancement path, which intentionally skips gracefully.)
- Wire a `vlm_factory` into the indexer pipeline for per-partition VLM selection.

## References
Follows #560 / #554. Part of the v2.0 (`refactor/hexagonal`) release prep.

hedhoud:
## Context

This fixes the remaining model-endpoint registry gap in the refactored indexer. After #560, named LLM endpoints were resolved correctly for contextualization and topic tagging, but named embedders and VLMs could still be missed at index time.

## Why this matters

The embedder case is especially risky because indexing can appear successful while using the wrong vector model. That leaves retrieval silently degraded or incompatible with the partition configuration. VLM selection had a similar Admin UI mismatch for image captioning.

## What changed

The indexer now refreshes the hydrated model registry for embedder and VLM requirements too. Named embedders and named VLMs are resolved dynamically from the live registry, cached by full endpoint identity, and rebuilt when endpoint config changes. Explicit named endpoint misses fail loudly instead of falling back to another model.

Default env/config fallback remains available for deployments that have not configured named endpoints. Seeded default endpoints now preserve the configured timeout and embedder batch size so the DB-backed default stays faithful to the running environment.

## Validation

- Added an in-process E2E-style regression covering factories built before registry hydration, then hydrated named embedder/VLM endpoints used by the real indexing pipeline.
- `uv run --no-env-file pytest tests/unit -q`
- `uv run --no-env-file ruff check openrag/services/orchestrators/model_endpoint_service.py openrag/services/workers/indexer_pool.py openrag/services/workers/pipeline_builder.py tests/unit/services/orchestrators/test_model_endpoint_service.py tests/unit/services/workers/test_indexer_pool.py tests/unit/services/workers/test_pipeline_builder.py`
- `git diff --check`

Closes #562

## Summary by CodeRabbit

* **New Features**
  * Expanded indexing to resolve embedder and image captioning models via the model endpoint registry, including generalized refresh logic and per-endpoint fallback support.
  * Added identity-based client caching so endpoint changes (e.g., credentials or config) trigger rebuilt clients.

* **Bug Fixes**
  * Default endpoint seeding now preserves configured `batch_size` and `timeout` values for embedder/LLM/VLM/reranker.
  * More defensive VLM selection: avoids errors when the default entry is missing and falls back to the provided VLM.

* **Tests**
  * Added/updated unit and e2e coverage for embedder/VLM registry resolution, rebuild behavior, and timeout preservation.

### DIFF
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -87,6 +87,8 @@ class ModelEndpointService:
                 model_type=model_type,
                 endpoint=endpoint,
                 model_name=model_name or None,
+                batch_size=data.get("batch_size", 32),
+                timeout=data.get("timeout", 30.0),
                 extra=data.get("extra", {}),
                 is_default=True,
                 created_at=now,
@@ -102,6 +104,8 @@ class ModelEndpointService:
             "embedder": {
                 "endpoint": os.getenv("EMBEDDER_ENDPOINT", s.embedder.base_url),
                 "model_name": os.getenv("EMBEDDING_MODEL", s.embedder.model_name),
+                "batch_size": s.embedder.batch_size,
+                "timeout": s.embedder.timeout,
                 "extra": _with_api_key(
                     {"implementation": "vllm"},
                     os.getenv("EMBEDDER_API_KEY", s.embedder.api_key),
@@ -110,6 +114,7 @@ class ModelEndpointService:
             "llm": {
                 "endpoint": os.getenv("LLM_ENDPOINT", s.llm.base_url),
                 "model_name": os.getenv("LLM_MODEL", s.llm.model),
+                "timeout": s.llm.timeout,
                 "extra": _with_enable_thinking(
                     _with_api_key(
                         {"implementation": "vllm"},
@@ -121,6 +126,7 @@ class ModelEndpointService:
             "vlm": {
                 "endpoint": os.getenv("VLM_ENDPOINT", s.vlm.base_url),
                 "model_name": os.getenv("VLM_MODEL", s.vlm.model),
+                "timeout": s.vlm.timeout,
                 "extra": _with_enable_thinking(
                     _with_api_key(
                         {"implementation": "vllm"},
@@ -138,6 +144,7 @@ class ModelEndpointService:
                 # remains available for per-partition opt-in.
                 "endpoint": os.getenv("RERANKER_ENDPOINT", s.reranker.base_url),
                 "model_name": os.getenv("RERANKER_MODEL", s.reranker.model_name),
+                "timeout": s.reranker.timeout,
                 "extra": _with_api_key(
                     {"implementation": s.reranker.provider},
                     os.getenv("RERANKER_API_KEY", s.reranker.api_key),
--- a/openrag/services/workers/indexer_pool.py
+++ b/openrag/services/workers/indexer_pool.py
@@ -96,6 +96,7 @@ class IndexerPool:
         self._model_endpoint_service: Any = None
         self._registry_loaded_at: float | None = None
         self._last_miss_reload_at: float | None = None
+        self._last_miss_reload_key: tuple[tuple[str, tuple[str, ...]], ...] | None = None
         self._registry_lock = asyncio.Lock()
         self._registry_reload_task: asyncio.Task[None] | None = None
         self._worker = IndexerWorker(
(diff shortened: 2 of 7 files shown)


Ahmath-Gadji (commented):
## Manual end-to-end verification

I exercised this branch against a live deployment built from the PR (real OpenAI-compatible inference backend) to confirm named embedder/VLM resolution behaves as intended at index time. Endpoints/models are anonymized below as placeholders.

### Setup
Registered endpoints in the DB-backed model-endpoint registry, then created three partitions, each pointed at a different embedder:

1. **`default`** — partition uses the default embedder.
2. **`embedder-A`** — partition pointed at a *valid* named endpoint (distinct registry entry).
3. **`embedder-bogus`** — partition pointed at a named endpoint whose URL is intentionally unreachable (`http://127.0.0.1:1/v1`), everything else valid.

For each partition I indexed a small document and checked which endpoint the embedding request actually targeted, per file.

### Results
- **Named resolution works.** Partitions (1) and (2) resolved their configured embedder from the registry and indexed successfully. This also confirms the indexer lazily hydrates `cfg.models` from the DB — the registry starts empty in the indexer process, which is precisely the gap this PR closes.
- **Routing is correct (decisive check).** Partition (3)'s embedding request went to its *own* configured (unreachable) endpoint and failed there — it did **not** silently fall back to the working default. Under the pre-fix behavior (inert embedder factory → silent default fallback), this partition would have wrongly succeeded against the default endpoint. It didn't, which is the proof the fix routes per-partition correctly.
- **Fail-loud on misconfiguration.** A partition pointed at an *unregistered* embedder name fails the file loudly instead of indexing with the wrong model; a named endpoint with a bad credential fails at the embedding call (auth error) rather than silently substituting another model. This matters because a wrong embedding model corrupts a partition's vectors invisibly.
- **VLM path.** Image captioning resolved the configured VLM endpoint and made a real call **only when a document actually contained an image**; text-only files correctly made no VLM call.

### Principle of the test
Because a correctly-routed call and a silently-misrouted call look identical when the named endpoint and the default point at the same backend, the routing check is only meaningful when they **diverge**. So the key partition deliberately uses an endpoint that *must* fail if (and only if) it is actually used — confirming the indexer contacts each partition's own configured endpoint, not the default.

> Note: OpenRAG uses one shared Milvus collection with a single embedding dimension (partition = `partition_key` row tag), so a named embedder must match the collection's dimension. Worth documenting, but independent of this PR.

**Net:** named embedder/VLM endpoint resolution works as intended — each partition indexes through exactly its configured endpoint (hydrated from the registry), and fails loud on a bad/missing endpoint instead of silently using the default.

Ahmath-Gadji (approved):
This is good for merge

PR #567, opened 2026-06-25
fix(llm): drop logprobs on streamed chat to avoid LiteLLM serialization crash

### LINKED ISSUE #563: Investigate LiteLLM streaming serialization failure in chat
## Context

A chat request can fail during streamed completion with a LiteLLM/Pydantic serialization error. This matters because the user only sees a generic chat failure, while the backend request itself may already have reached the downstream model service.

## Problem

The observed failure happens inside LiteLLM while serializing streamed chunks: `MockValSer` cannot be converted to `SchemaSerializer`. The stack trace points to LiteLLM streaming response serialization, not directly to OpenRAG business logic.

PR #535 may make this easier to trigger if extra OpenAI-compatible parameters are forwarded to LiteLLM, but the default Chainlit chat path does not appear to send the new `response_format` fields itself. The likely area to verify is streamed chat behavior with downstream LiteLLM, especially when optional parameters such as logprobs or provider-specific extras are present.

## Expected behavior

Chat streaming should either complete normally or fail with a clear provider error. OpenRAG should avoid sending parameters that are known to break the configured downstream provider, or handle this incompatibility explicitly so the UI does not fail with an opaque serialization error.

Ahmath-Gadji:
## Problem

Closes #563. A streamed chat completion fails with an opaque error from the downstream LiteLLM proxy:

> `'MockValSer' object cannot be converted to 'SchemaSerializer'`

The user only sees a generic chat failure even though the request reached the model service.

## Root cause

OpenRAG doesn't use LiteLLM as a library — `VLLMClient` POSTs raw OpenAI-style requests via `httpx` to the configured OpenAI-compatible endpoint (here, a LiteLLM proxy). The client is built from `settings.llm.model_dump()`, so two server-side config knobs leaked into `self._defaults` and were sent as request-body fields on **every** call:

- **`logprobs: true`** (config default) — yet OpenRAG **never reads logprobs back** anywhere. Per [BerriAI/litellm#24357](https://github.com/BerriAI/litellm/issues/24357), **streaming + logprobs** is exactly the break: the final stream chunk's `ChoiceLogprobs` still carries a `MockValSer` serializer, so LiteLLM crashes serializing it.
- **`max_retries: 2`** — not an OpenAI body field at all; retries are handled by `@with_retry` (which hardcodes attempts and ignores it). Dead weight on the wire that strict downstream validators can reject.

## Fix

At the single wire-building choke point (`VLLMClient`, covering RAG chat, direct chat, query-gen, map-reduce, completions, vision captioning, indexer contextualization):

1. **Strip `logprobs` from streamed requests** — regardless of source (config default or client-supplied). Non-streaming keeps it (that path doesn't hit the bug), so behavior is preserved for direct-vLLM users.
2. **Filter `max_retries` out of `_defaults` at construction** so it never reaches any payload.

## Testing

- 3 new unit tests in `test_vllm_client.py`: streaming omits `logprobs`; non-streaming keeps it; `max_retries` never sent (chat + completions).
- Full inference / query-service / DI / config unit suites green (324 passed); ruff lint + format clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming chat requests now avoid sending unsupported `logprobs` options, preventing request failures.
  * Retry settings are no longer included in outgoing request payloads, reducing incompatible request data.
* **Tests**
  * Added coverage for streaming and non-streaming chat behavior, including payload handling for `logprobs` and retry settings.

### DIFF
--- a/openrag/core/config/endpoints.py
+++ b/openrag/core/config/endpoints.py
@@ -13,6 +13,10 @@ class LLMParamsConfig(ConfigMixin):
     temperature: float = 0.1
     timeout: int = 60
     max_retries: int = 2
+    # Forwarded on non-streaming calls only. OpenRAG never reads logprobs back,
+    # and streaming them through a LiteLLM proxy triggers a pydantic
+    # serialization crash, so the client strips logprobs from streamed requests
+    # (see VLLMClient / linagora/openrag#563).
     logprobs: bool = True
     enable_thinking: bool | None = None
 
--- a/openrag/services/inference/vllm_client.py
+++ b/openrag/services/inference/vllm_client.py
@@ -47,6 +47,13 @@ def _parse_response(resp: httpx.Response) -> dict:
 # LLM
 # ---------------------------------------------------------------------------
 
+# Knobs that travel inside ``settings.llm`` (and therefore the client
+# ``_defaults``) but are NOT OpenAI request-body fields. ``max_retries`` is a
+# client-side retry policy — retries are handled by ``@with_retry``, so
+# forwarding it on the wire does nothing useful and only risks tripping strict
+# downstream validators (e.g. a LiteLLM proxy).
+_NON_WIRE_DEFAULTS = frozenset({"max_retries"})
+
 
 @llm_registry.register("vllm")
 class VLLMClient(LLM):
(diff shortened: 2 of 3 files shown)


Ahmath-Gadji:
Closing: dropping `logprobs` is the wrong trade-off here. logprobs is a metric we intend to consume, so suppressing it (even only on streamed requests) loses data we want. Reworking toward keeping logprobs available while handling the LiteLLM streaming-serialization incompatibility explicitly instead.

PR #577, opened 2026-06-25
fix(indexing): return 403 (not 500) when a create-race loser lacks access

Ahmath-Gadji:
## What

Follow-up to a review finding on #568 (F11). The partition auto-create race guard (`_ensure_editor_access_after_create_race` in `indexing_service.py`) correctly **fails closed** when a concurrent create lost the race and the loser turns out not to be a member of the partition — but it raised a builtin `PermissionError`.

## Why it matters

The global exception handlers (`api/error_handlers.py`) map `OpenRAGError` subclasses to their declared status and everything else to the catch-all **500**. So a builtin `PermissionError` from this guard surfaced a legitimate authorization denial as an *Internal Server Error* instead of a `403 Forbidden`.

This only triggers on the narrow cross-user race (two different users creating the same brand-new partition name simultaneously, where the loser has no membership). It already fails closed — this just corrects the status code. The common case (same user, concurrent uploads — the scenario the race guard was added for) is unaffected: the user owns the partition and the upload continues.

## Change

- `indexing_service.py`: raise `AuthError` (status 403) instead of `PermissionError`.
- `test_indexing_service.py`: the cross-user race test now expects `AuthError` and asserts `status_code == 403`.

## Verification

- `uv run pytest tests/unit/services/orchestrators/test_indexing_service.py` → 18 passed
- `ruff check` clean

### DIFF
--- a/openrag/services/orchestrators/indexing_service.py
+++ b/openrag/services/orchestrators/indexing_service.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
-from core.utils.exceptions import PartitionNotFoundError, ValidationError
+from core.utils.exceptions import AuthError, PartitionNotFoundError, ValidationError
 from core.utils.filename import extract_temporal_fields
 from core.utils.logging import get_logger
 from core.utils.partition_limits import max_partitions_for_user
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
Folded directly into #568 as commit e4314d64 on `forward-port/main-to-hexagonal` (this PR's base), since it's a follow-up to #568's F11 finding and belongs in that branch. Closing — no separate PR needed.

PR #578, opened 2026-06-25
fix(chainlit): serve source downloads from standalone Chainlit app (Ray Serve mode)

Ahmath-Gadji:
## What

Follow-up to a review finding on #568 (F9). In **Ray Serve mode**, source-document previews (PDFs/images/audio) fail to load in the Chainlit chat UI.

## Why it's broken

In Ray Serve mode the API and Chainlit run on **separate ports**:
- the API (with `download_router` mounted) via `serve.run(OpenRagAPI.bind(), route_prefix="/")` on `ray.serve.port`,
- Chainlit via its own `uvicorn.run(chainlit_app, port=ray.serve.chainlit_port)` — where `chainlit_app` is `chainlit_api.app`.

`_format_sources` (`app_front.py`) rewrites each file URL from the internal base URL to the **browser origin** (`get_external_url()`, derived from the `Referer`). In separate-port mode that origin is the **Chainlit host**, but `chainlit_api.py` only mounted the Chainlit UI — so `/static/{extract_id}` didn't exist there and every source preview 404'd.

The default mounted `/chainlit` deployment is fine because the UI and API share one origin (the API already serves the route).

## Change

- `openrag/chainlit_api.py`: mount `download_router` on the standalone Chainlit FastAPI app.
  - Auth is unchanged: the same `AuthMiddleware` runs here. In OIDC mode the `openrag_session` cookie is sent to the Chainlit port too (cookies aren't port-scoped), and in token mode the preview link carries `?token=`. The route resolves its services from the container the middleware initializes on `app.state.container`.
- `tests/unit/test_chainlit_api_download.py`: regression test asserting the standalone app exposes `download_source` at `/static/{extract_id}` (would fail before this change).

## Verification

- `uv run pytest tests/unit/test_chainlit_api_download.py tests/unit/test_chainlit_api_auth.py tests/unit/api/routers/user/test_download.py` → all pass
- `ruff check` clean; layer-import guard OK

### DIFF
--- a/openrag/chainlit_api.py
+++ b/openrag/chainlit_api.py
@@ -1,4 +1,5 @@
 from api.middleware.auth import AuthMiddleware
+from api.routers.user.download import router as download_router
 from chainlit.utils import mount_chainlit
 from fastapi import FastAPI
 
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
Folded directly into #568 as commit 18050a23 on `forward-port/main-to-hexagonal` (this PR's base), since it's a follow-up to #568's F9 finding and belongs in that branch. Closing — no separate PR needed.

Issue #575, opened 2026-06-25
Preset docling parser fails: DoclingPool actor never created when global PDF backend isn't docling


andyne13:
### Summary
Indexing a PDF with a preset whose `parsing_strategy = docling` fails immediately:

```
File "services/workers/parsers/docling_workers.py", line 123, in DoclingLoader.__init__
    self.worker = ray.get_actor("DoclingPool", namespace="openrag")
ValueError: Failed to look up actor with name 'DoclingPool'.
```

### Root cause
Bootstrap provisions the PDF pool from the **global** `loader.file_loaders.pdf`:

```python
# bootstrap.get_marker_pool()
match pdf_loader:
    case "DoclingLoader": get_or_create_actor("DoclingPool", ...)
    case "MarkerLoader":  get_or_create_actor("MarkerPool", ...)
```

With the default global (`MarkerLoader`), only `MarkerPool` is created; `DoclingPool` never is. But `DoclingLoader.__init__` does a **get-only** `ray.get_actor("DoclingPool")`, so it fails the moment a *preset* selects docling. (pymupdf is unaffected — it has no Ray pool.)

This is the pool-bootstrap analog of #569: pools are provisioned per global config, but #569 lets any preset pick any backend at runtime. `MarkerLoader.__init__` has the same latent bug (would fail if the global default were docling and a preset used marker).

### Suggested fix
Create the pool lazily in the loader — use `get_or_create_actor(...)` (idempotent / `get_if_exists`) instead of `ray.get_actor(...)` in both `DoclingLoader` and `MarkerLoader`, so the pool for whichever backend a preset selects is created on first use. Depends on #569.

Issue #571, opened 2026-06-25
Indexer parse stage has no timeout — a slow/wedged parse can hang indexing


andyne13:
### Summary
The indexing pipeline runs the parse stage with **no timeout**, so a slow or wedged parse can stall indexing indefinitely.

### Root cause
`services/workers/indexer_pool.py` calls `build_indexing_pipeline(...)` **without `timeouts=`**, so `PipelineTimeouts.parse` defaults to `None` and `parse_stage(..., timeout=None)` never bounds the parse.

- marker is self-protected (its pool has its own `MARKER_TIMEOUT` + retry), so it survives a bad PDF.
- pymupdf/docling have no parse timeout. pymupdf additionally serializes all work on a single shared worker thread (`_PYMUPDF_EXECUTOR`, `max_workers=1`), so one wedged parse blocks **every** subsequent pymupdf parse — the whole backend appears frozen.

### Suggested fix
Wire a parse-stage timeout so a pathological PDF fails *that file* (and is reported) instead of stalling the pipeline. Consider per-backend defaults.

Area: backend (`refactor/hexagonal`).

andyne13:
Fixed by #582 (merged into refactor/hexagonal; ships in v2.0).

Issue #570, opened 2026-06-25
pymupdf markdown mode inlines base64 images into chunk text → oversized chunks & Milvus insert failure


andyne13:
### Summary
pymupdf's default `markdown` mode renders images and embeds each one as a base64 `data:image/...` URI **inside the page text**. That base64 is never stripped from the text, so it flows into chunks → embeddings → Milvus. Result: bloated chunks, meaningless embeddings, and insert failures.

### Evidence
Indexing a PDF with pymupdf fails at the store stage:
```
grpc: received message larger than max (~147 MB vs 64 MB)  →  StatusCode.RESOURCE_EXHAUSTED
```
Likely the same root cause as #61 (a single chunk's varchar text exceeding Milvus's 65535-char field limit) — both stem from base64 image data living in the text.

### Root cause
`core/indexing/parsers/pdf/pymupdf.py::_extract_markdown` calls `pymupdf4llm.to_markdown(..., embed_images=True, dpi=300)` and appends the full page text (base64 and all). `image_preprocessor.extract_data_uri_image_blocks` only **copies** the data URIs into `ImageBlock`s — it does not remove them from the text. The base64 is removed only if the caption stage runs (substituting `markdown_ref → caption`); with captioning off it persists.

### Suggested fix
- Make pymupdf the lightweight backend: build in `text` mode (`page.get_text()`, no image rendering) — fast, no base64. Image-aware parsing is marker/docling's job.
- And/or: in markdown mode, strip base64 from text (compact placeholder) regardless of whether captioning runs, so raw base64 never reaches chunks.

Related: #61, #183, #453. Area: backend (`refactor/hexagonal`).

andyne13:
Fixed by #582 (merged into refactor/hexagonal; ships in v2.0).

Issue #569, opened 2026-06-25
Indexer ignores a preset's parsing_strategy (pymupdf/docling unreachable, always marker)


andyne13:
### Summary
A partition/preset's `parsing_strategy` (pymupdf / docling / marker) is silently ignored for indexing. Every PDF is parsed with the **global** default loader (`loader.file_loaders.pdf`, normally marker), regardless of the preset.

### Evidence
Setting a preset's `parsing_strategy` to `pymupdf` and indexing a PDF logs:
```
DEBUG | parser_dispatcher:_build - Building parser backend: marker
```

### Root cause
`services/workers/indexer_pool.py` builds the pipeline with `chunker_factory`, `embedder_factory`, `contextualizer_factory`, `topic_tagger_factory` — but **no `parser_factory`**. So `pipeline_builder.IndexingPipeline._select_parser` (which would call `parser_factory(config.parsing_strategy)`) always falls back to `self.parser`, the single global `ParserDispatcher` that resolves the PDF backend from `loader.file_loaders.pdf` and never reads `parsing_strategy`.

Because of this, the pymupdf and docling backends have effectively never been exercised through the indexer.

### Suggested fix
Wire a `parser_factory` that honors `parsing_strategy` per preset (reusing the dispatcher's shared backend cache so no duplicate marker/docling pool is created).

Area: backend (`refactor/hexagonal`).

andyne13:
Fixed by #582 (merged into refactor/hexagonal; ships in v2.0).

PR #582, opened 2026-06-26
fix(indexer): honor preset parsing_strategy, lazy pools, pymupdf markdown, parse timeout

andyne13:
Makes the indexing pipeline honor a partition preset's `parsing_strategy` and hardens the PDF parse stage. Four related parser fixes (entangled in shared files, so one PR; #575 is a separate commit).

- **#569** honor a preset's `parsing_strategy`: `ParserDispatcher.for_pdf_strategy()` + a parser_factory wired into the pipeline, so pymupdf/docling are reachable per preset instead of always using the global default.
- **#575** create the marker/docling Ray pools **lazily** (`get_or_create_actor` + `get_if_exists`) so a per-preset backend works even when it isn't the global default and bootstrap didn't pre-warm it.
- **#570** build pymupdf in markdown mode with `embed_images=False`: structured text for the chunker, no base64 bloat / Milvus gRPC overflow.
- **#571** bound the parse stage with `loader.parse_timeout` (`PARSE_TIMEOUT`, default 3600s) so a wedged parse fails that file instead of hanging indexing.

Tests: strategy dispatch, lazy pool creation, pymupdf markdown mode, parse-timeout error. Refs #569 #570 #571 #575.

## Summary by CodeRabbit

* **New Features / Configuration**
  * Added `parse_timeout` (default 3600s) to bound per-file parsing, with `PARSE_TIMEOUT` environment override support.
  * Preset PDF parsing strategies can now be selected per document, with defaults able to inherit the deployment-wide choice.
* **Bug Fixes**
  * Parsing timeouts now report clearer timeout errors (including filename) and consistently mark parse failures.
  * Improved robustness of worker pool startup by lazily creating missing pools/actors.
  * PDF markdown parsing no longer extracts or inlines embedded images.
* **Tests**
  * Added/expanded unit coverage for strategy routing, lazy pool creation, and parse timeout behavior.

### DIFF
--- a/openrag/core/indexing/parsers/pdf/pymupdf.py
+++ b/openrag/core/indexing/parsers/pdf/pymupdf.py
@@ -31,7 +31,6 @@ import pymupdf
 import pymupdf4llm
 
 from ....models.document import Document, DocumentType, ImageBlock, ProcessedDocument, TextBlock
-from ...image_preprocessor import extract_data_uri_image_blocks
 from ..document_parser import DocumentParser
 from ..registry import parser_registry
 
@@ -48,29 +47,25 @@ def _extract_text(raw: bytes) -> tuple[list[str], list[ImageBlock]]:
 
 
 def _extract_markdown(raw: bytes) -> tuple[list[str], list[ImageBlock]]:
-    """Return Markdown per page + ``ImageBlock``s built from embedded data URIs.
+    """Return structured Markdown per page (no images).
 
-    ``embed_images=True`` makes ``pymupdf4llm`` write images as base64
-    data URIs in-line. We decode each ref into an ``ImageBlock`` and
-    leave the ref in the page text untouched so the caption stage can
-    substitute later via ``ImageBlock.metadata['markdown_ref']``.
+    pymupdf is the lightweight, no-VLM backend. ``pymupdf4llm`` preserves
+    document structure (headings, lists, tables) — which the markdown-aware
+    chunker needs to cut on real boundaries instead of mid-sentence — while
+    ``embed_images=False`` keeps base64 image data out of the text. That keeps
+    chunks small (no Milvus gRPC overflow) and skips image rendering entirely
+    (fast). Image-aware parsing is marker/docling's job, so no ``ImageBlock``s
+    are produced here.
     """
     with pymupdf.open(stream=raw, filetype="pdf") as doc:
         chunks = pymupdf4llm.to_markdown(
             doc,
             page_chunks=True,
-            embed_images=True,
+            embed_images=False,
             write_images=False,
-            dpi=300,
         )
-    pages: list[str] = []
-    images: list[ImageBlock] = []
-    for i, chunk in enumerate(chunks, start=1):
-        text = (chunk.get("text") or "").strip()
-        pages.append(text)
-        if text:
-            images.extend(extract_data_uri_image_blocks(text, page_number=i))
-    return pages, images
+    pages = [(chunk.get("text") or "").strip() for chunk in chunks]
+    return pages, []
 
 
 @parser_registry.register("pymupdf")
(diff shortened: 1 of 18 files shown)


andyne13:
### Testing independently
**Unit:** `uv run pytest tests/unit/services/workers/parsers/ tests/unit/services/workers/stages/test_parse.py tests/unit/services/workers/test_indexer_pool.py`

**Functional (API/logs, no admin UI needed):**
- Create indexation presets with `parsing_strategy` = pymupdf / docling / marker (`POST /presets/`), assign to partitions (`PATCH /partition/{p}` → `indexation_preset`), index a PDF into each → logs show the matching backend per preset (#569).
- **#575:** after a fresh restart with the global default = marker, index into the docling preset → `Docling pool: …` appears at upload (lazy), no `Failed to look up actor`.
- **#570:** index an image-heavy PDF via pymupdf → job completes (no Milvus 64MB overflow); chunks contain no `data:image/base64` blobs.
- **#571:** set `PARSE_TIMEOUT=2`, index a large PDF → task FAILED with `parse timed out after 2s for '<file>'`; worker survives the next upload. Restore `PARSE_TIMEOUT=3600`.

openrag/core/indexing/parsers/pdf/pymupdf.py
```diff
@@ -48,29 +47,25 @@ def _extract_text(raw: bytes) -> tuple[list[str], list[ImageBlock]]:
...
-    pages: list[str] = []
-    images: list[ImageBlock] = []
-    for i, chunk in enumerate(chunks, start=1):
-        text = (chunk.get("text") or "").strip()
-        pages.append(text)
-        if text:
-            images.extend(extract_data_uri_image_blocks(text, page_number=i))
-    return pages, images
+    pages = [(chunk.get("text") or "").strip() for chunk in chunks]
+    return pages, []
```
Ahmath-Gadji, line 68: The issue with this implementation is that we're dropping embedded images, even though this parser is capable of handling them. Do we really want to do that?
We can avoid embedding images inside the markdown, but preserve ImageBlocks.

Ahmath-Gadji:
**Found while testing this PR: docling parses run CPU-only on a GPU host**

Exercising the new per-preset `parsing_strategy` routing (this PR makes docling reachable per preset, #569), I indexed a PDF into a `docling`-preset partition on a GPU host — and the parse ran on **CPU**. Symptom: a flood of

```
[W…] NNPACK.cpp:57 Could not initialize NNPACK! Reason: Unsupported hardware.
```

and slow parses. `nvidia-smi` confirmed it — `ray::MarkerWorker` was on the GPU, but there was **no `ray::DoclingWorker`** process on the device.

**Root cause:** `_docling_num_gpus` gated the worker's GPU reservation on `torch.cuda.is_available()`, which is evaluated inside the GPU-less `DoclingPool` actor (Ray sets `CUDA_VISIBLE_DEVICES=""` for a 0-GPU actor) → always `False` → `DoclingWorker` got `num_gpus=0` → CUDA hidden → docling's `AcceleratorDevice.AUTO` fell back to CPU. Marker avoids this by gating on `ray.cluster_resources()["GPU"]` (cluster-wide, visible from a GPU-less pool).

It's a pre-existing docling bug that this PR *surfaces* (docling was rarely reached before per-preset routing). Fixed separately in #584, which makes `_docling_num_gpus` mirror `_marker_num_gpus`.

openrag/core/indexing/parsers/pdf/pymupdf.py
Ahmath-Gadji: Since pymupdf image cropping quality has not been validated and compared to other advanced parsers like (docling & marker), we've decided to remove the embedded images until that validation.

Ahmath-Gadji (approved):
LGTM.
I've tested it and it works

Issue #580, opened 2026-06-25
Docling parses on CPU — GPU reservation uses a process-local CUDA check, unlike marker


andyne13:
### Summary
The Docling PDF backend silently runs entirely on **CPU**, making it ~5–10× slower than marker. `nvidia-smi` during indexing shows `ray::MarkerWorker` on the GPU but **no `DoclingWorker`** process at all.

### Root cause
`_docling_num_gpus()` decides the worker's GPU reservation with a **process-local** check:
```python
def _docling_num_gpus(config):
    return config.loader.docling_num_gpus if torch.cuda.is_available() else 0
```
It runs inside the `DoclingPool` actor, which Ray schedules with **no GPU** — so `CUDA_VISIBLE_DEVICES=""` there and `torch.cuda.is_available()` is **False**. It therefore reserves **0 GPUs** for the `DoclingWorker`, Ray hides the GPU from the worker too, and Docling's `AcceleratorDevice.AUTO` resolves to **CPU**.

Marker does **not** have this bug — `_marker_num_gpus()` queries the Ray **cluster** (`ray.cluster_resources().get("GPU")`) with local CUDA only as a fallback. That fix (`a0fba51a` "preserve Marker GPU allocation from Ray resources") was never applied to Docling.

### Evidence (per-stage timings, same 10 PDFs)
- Docling (CPU): first/least-queued doc `ms_parse ≈ 123000` (123 s for a 13-chunk PDF); parse is 95–99% of total.
- Marker (GPU, page-parallel): comparable docs `ms_parse ≈ 14000–26000`.

### Fix
Make `_docling_num_gpus()` mirror `_marker_num_gpus()` (use `ray.cluster_resources()`, fall back to `torch.cuda`). Unit tests cover: cluster-has-GPU + CUDA-hidden → reserves; no cluster GPU → 0; not requested → 0.

### Follow-up (separate)
Even on GPU, Docling uses the slowest options (`TableFormerMode.ACCURATE`, full `do_ocr`, `images_scale=2`) and lacks marker's intra-document page parallelism — tunable later. Area: backend.

PR #584, opened 2026-06-26
fix(docling): reserve the GPU from Ray cluster resources, not local CUDA (#580)

andyne13:
Docling parsed entirely on **CPU** (~5-10x slower than marker; no `DoclingWorker` on `nvidia-smi`). `_docling_num_gpus()` ran inside the `DoclingPool` actor — scheduled with no GPU, so `torch.cuda.is_available()` was False there and it reserved **0 GPUs** for the worker, leaving Docling's `AcceleratorDevice.AUTO` on CPU.

Mirror `_marker_num_gpus()`: query the Ray **cluster** for GPU capacity, with local CUDA only as a startup fallback. (Marker got this fix earlier; Docling never did.)

Validated: with the fix, `DoclingWorker` shows on the GPU and per-doc parse dropped from ~123s to ~10s (on par with marker). Tests cover cluster-has-GPU/CUDA-hidden → reserves, no cluster GPU → 0, not requested → 0. Refs #580.

### DIFF
--- a/openrag/services/workers/parsers/docling_workers.py
+++ b/openrag/services/workers/parsers/docling_workers.py
@@ -37,7 +37,22 @@ logger = get_logger()
 
 
 def _docling_num_gpus(config) -> float:
-    return config.loader.docling_num_gpus if torch.cuda.is_available() else 0
+    """Return Docling's Ray GPU reservation, falling back to CUDA detection.
+
+    Must mirror Marker: ``_docling_num_gpus`` runs inside the ``DoclingPool``
+    actor, which Ray schedules with no GPU — so ``torch.cuda.is_available()`` is
+    False there (``CUDA_VISIBLE_DEVICES=""``) and a naive local check reserves 0
+    GPUs for the worker, leaving Docling stuck on CPU. Query the Ray *cluster*
+    for GPU capacity instead, with local CUDA only as a startup fallback.
+    """
+    requested_gpus = config.loader.docling_num_gpus
+    if requested_gpus <= 0:
+        return 0
+    try:
+        return requested_gpus if ray.cluster_resources().get("GPU", 0) > 0 else 0
+    except Exception as exc:
+        logger.warning("Failed to query Ray cluster GPU resources; falling back to CUDA check", error=str(exc))
+        return requested_gpus if torch.cuda.is_available() else 0
 
 
 @ray.remote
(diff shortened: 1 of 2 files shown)


andyne13:
### Testing independently (no other PR needed)
**Unit:** `uv run pytest tests/unit/services/workers/parsers/test_docling_workers.py`

**Functional standalone** — set docling as the **global** backend (`PDFLoader=DoclingLoader`), so you don't need #582:
- Index a PDF; during the parse run `nvidia-smi` → a `ray::DoclingWorker` process now holds VRAM (it was **absent**/CPU-only before this fix).
- Per-doc `ms_parse` drops sharply (observed ~123s → ~10s, on par with marker).
Before-fix signature: no DoclingWorker on `nvidia-smi`, docling parse 5–10× slower than marker.

openrag/services/workers/parsers/docling_workers.py
```diff
@@ -37,7 +37,22 @@
...
+    actor, which Ray schedules with no GPU — so ``torch.cuda.is_available()`` is
+    False there (``CUDA_VISIBLE_DEVICES=""``) and a naive local check reserves 0
+    GPUs for the worker, leaving Docling stuck on CPU. Query the Ray *cluster*
+    for GPU capacity instead, with local CUDA only as a startup fallback.
+    """
+    requested_gpus = config.loader.docling_num_gpus
+    if requested_gpus <= 0:
+        return 0
+    try:
+        return requested_gpus if ray.cluster_resources().get("GPU", 0) > 0 else 0
```
hedhoud, line 52: This fixes fresh `DoclingPool` creation, but it may not take effect during a live upgrade. Startup uses `get_or_create_actor("DoclingPool", ...)`, so if the old detached actor is still alive, it will keep its existing workers that were created with `num_gpus=0`. The deployment would still parse on CPU until someone manually restarts `DoclingPool`.

For this PR, it would be good to either add a small migration/restart guard for the existing actor, or call out clearly that `DoclingPool` must be restarted after deploying the fix.

hedhoud: I re-checked this. The code fix is already enough for any newly created DoclingPool. The only remaining case is a live upgrade with an old detached pool still running, and we can handle that by restarting DoclingPool after deployment. So I don’t think we need an extra code change here.

hedhoud (approved):

PR #585, opened 2026-06-26
feat(indexer): per-stage indexing timings

andyne13:
Times each pipeline stage and logs one structured line per file (`ms_parse`, `ms_chunk`, `ms_embed`, `ms_store`, …, `ms_total`) so a slow backend or stage is easy to spot (e.g. quantifying docling vs marker parse cost). Fields are bound via `logger.bind()` so they render, and emitted from a `finally` block so a failed stage still reports how far it got.

Docstring notes the timings are **wall-clock** (include scheduling/queue contention under concurrent batch indexing) — clean per-stage numbers need low concurrency. Test asserts the line is logged with per-stage keys.

## Summary by CodeRabbit

- **New Features**
  - Indexing now carries the job/task identifier through the processing flow, improving traceability of file processing.

- **Bug Fixes**
  - Added more reliable timing and logging for indexing stages, so performance data is recorded even if a step fails.
  - Logging now includes clearer per-stage and total duration details to help diagnose slow or incomplete indexing runs.

- **Tests**
  - Added coverage for task identifier propagation and indexing timing logs.

### DIFF
--- a/openrag/services/workers/pipeline_builder.py
+++ b/openrag/services/workers/pipeline_builder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import time
 from collections.abc import Callable, MutableMapping
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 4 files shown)


andyne13:
### Testing independently
**Unit:** `uv run pytest tests/unit/services/workers/test_pipeline_builder.py` (asserts the timing line is logged with per-stage keys).

**Functional:** index any file, then `docker compose logs openrag | grep 'stage timings'` → one structured line per file: `ms_parse / ms_chunk / ms_embed / ms_store / … / ms_total` (+ `ms_caption`/`ms_contextualize` when active). Note: these are **wall-clock** — under a concurrent batch they include scheduling/queue contention, so index **one file at a time** for clean per-stage numbers.

openrag/services/workers/pipeline_builder.py
```diff
@@ -67,54 +68,102 @@ class IndexingPipeline:
...
-        )
-        return row
+            return row
+        finally:
+            logger.bind(
+                filename=row.get("filename", ""),
+                n_chunks=len(row.get("chunks") or []),
+                **{f"ms_{name}": round(value) for name, value in timings.items()},
+                ms_total=round(sum(timings.values())),
+            ).info("indexing stage timings (ms)")
```
hedhoud, line 167: The timing log is useful, but it isn't associated with the indexing task. The task log endpoints only return records that include `extra.task_id`, while this log currently only includes fields such as the filename, chunk count, and timing information.

As a result, the timing log will appear in the container logs, but it won't be available through `/task/{task_id}/logs` or `get_task_logs` via MCP. This makes it harder to debug a specific indexing task using the existing task log view.

hedhoud:
@codex review

hedhoud (approved):

Issue #453, opened 2026-06-08
Image captioning can still run when disabled during indexing


hedhoud:
## Context

PDF indexing performance depends heavily on whether image captioning runs. When captioning is not needed, disabling it should reliably skip the VLM work across every indexing path.

## Problem

One indexing path still goes through the legacy parser bridge. That bridge keeps using the global captioning configuration instead of the per-file indexing configuration, so captioning may run even when it was disabled for the current indexing request. This makes PDF ingestion unexpectedly slow and makes performance comparisons misleading.

## Expected behavior

When image captioning is disabled for an indexing request, every parser path should skip captioning for that file, and no internal configuration details should leak into stored file metadata.

EnjoyBacon7:
A fix exists but is not merged, and the targeted code is not on `main`. `ee55e44b` ("fix(indexing): honor disabled image captioning in legacy parser") addresses this on the phase-14 refactor branch. Both the architecture this issue describes (the legacy parser bridge) and its fix live only on that branch; on `main`, captioning is driven by a single global toggle. Needs merge / re-evaluation against `main`.

hedhoud:
Closing this now that #454 has been merged and the image-captioning config path is fixed.

PR #586, opened 2026-06-26
feat(admin-ui): vendored admin UI (partitions, documents, jobs, presets, models, users) on the OpenRag API

andyne13:
Brings the new admin UI into the hexagonal refactor: a static SPA (served via nginx, same-origin API proxy) wired to the OpenRag API for partitions, documents, jobs, presets, models, users and system ops, with token + OIDC auth and a role-aware capability layer.

**Highlights**
- Partitions / documents / jobs / presets / models / users / system pages on the real API (poll-based jobs + document list).
- Indexation & retrieval preset editors driven by `/presets/options`.
- Recent fixes: keep paginated lists on the current page after a delete; sortable columns + bulk delete in the documents list; image-captioning toggle defaults to the backend's value (#453); poll the documents list so newly-indexed files appear; persist the selected partition across navigation; topic-tagging toggle grayed as 'coming soon'.

**Merge status:** no conflicts with `refactor/hexagonal` (disjoint divergence from #561). Recommend a merge commit to preserve history.

Note: a few additive backend endpoints support the UI (partition document_count / chat fields, model-endpoint validation, etc.).

## Summary by CodeRabbit

* **New Features**
  * Added a revamped Admin UI with partitions, documents, jobs, users, presets, model endpoints, system status, and an overview.
  * Added admin-only “model endpoint validate (draft)” to probe endpoint settings.
* **Bug Fixes**
  * Improved OIDC redirect URI guidance to reduce blank/404/JSON landings and infinite login loops.
  * Enhanced partition listing to return consistent partition summaries with document counts and safer missing-partition behavior.
* **Documentation**
  * Updated OIDC and SSO quickstart guides with “front door” vs backend/proxy redirect URI examples and troubleshooting.


ui/src/pages/admin/jobs/list.tsx
```diff
@@ -0,0 +1,86 @@
...
+const STATUS_TABS = ["ALL", "ACTIVE", "COMPLETED", "FAILED", "CANCELLED"] as const;
+
+const str = (v: unknown) => (v == null ? "" : String(v));
+
+const columns: ColumnDef<TaskListItem, unknown>[] = [
+  {
+    accessorKey: "task_id",
+    header: "Task",
+    cell: ({ row }) => (
+      <Link to={`/jobs/${row.original.task_id}`} className="text-primary hover:underline font-mono text-sm">
```
hedhoud, line 23: This link can put admins into a dead end. The jobs list shows admins every task, but the task detail, error, logs, and cancel endpoints still only allow the original task owner. So an admin can see another user's task here, click it, and immediately get a 403.
I would either make those task endpoints accept admin-or-owner access, or avoid showing links/actions for tasks the current user cannot actually open.

ui/src/pages/admin/documents/detail.tsx
```diff
@@ -0,0 +1,330 @@
...
+            <span className="text-sm">{partition}</span>
+            {mimetype && (
+              <>
+                <span>·</span>
+                <span className="text-sm">{mimetype}</span>
+              </>
+            )}
+          </div>
+        </div>
+        <ConfirmDialog
```
hedhoud: This page brings write actions back for users who may only have viewer access. The documents list already hides upload/delete when the partition role is read-only, but the detail view always shows delete, replace, and copy. The API still rejects those calls for non-editors, so the UI offers actions that are guaranteed to fail.
I would reuse the same role check here and only show write controls for editor/owner access. For copy, the target list should also only include partitions where the user can write.

ui/src/pages/admin/partitions/detail.tsx
```diff
@@ -0,0 +1,623 @@
...
+  const [chatHistoryDepth, setChatHistoryDepth] = useState(String(partition.chat_history_depth));
+  const [indexationPreset, setIndexationPreset] = useState(partition.indexation_preset);
+  const [retrievalPreset, setRetrievalPreset] = useState(partition.retrieval_preset);
+  const [chatLlm, setChatLlm] = useState(partition.chat_llm ?? "__default__");
+  const [llmValidated, setLlmValidated] = useState<boolean | null>(
+    partition.chat_llm ? null : true,
+  );
+  const [llmValidating, setLlmValidating] = useState(false);
+
+  const { data: presetsData } = useQuery({
```
hedhoud, line 80: This tab is available from a shared partition route, but it loads admin-only preset/model data and still renders editable settings for users who may only be viewers or editors. Owners also have an awkward path here: the backend lets owners update partition settings, but these option lists are guarded as admin-only.
I would make the page role-aware, like the members tab already is: read-only for non-owners, and avoid calling admin-only registries unless the current user can actually access them.

ui/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "clsx": "^2.1.1",
+    "cmdk": "^1.1.1",
+    "lucide-react": "^0.563.0",
+    "next-themes": "^0.4.6",
+    "radix-ui": "^1.4.3",
+    "react": "^19.2.0",
+    "react-dom": "^19.2.0",
+    "react-hook-form": "^7.71.1",
+    "react-markdown": "^10.1.0",
+    "react-router-dom": "^7.13.0",
```
hedhoud: The new UI currently pins react-router-dom to a version that npm audit --omit=dev reports as vulnerable through react-router. Some of the advisories may matter less for this static SPA than for React Router framework usage, but this is a new direct dependency and there are patched versions available.
I would bump this before merging so the vendored UI does not land with known audit findings on day one.

ui/src/pages/admin/presets.tsx
```diff
@@ -0,0 +1,906 @@
...
+    queryKey: ["prompts-for-presets"],
+    queryFn: () => listPrompts({ limit: 200 }),
+    enabled: open && presetType === "indexation",
+  });
+  const allPrompts = promptData?.prompts ?? [];
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (editing) {
+      onUpdate(editing.preset_type, editing.name, name, config);
```
hedhoud, line 855: Renaming a preset from this UI can leave partitions pointing at a preset name that no longer exists. Delete already blocks presets that are still used by partitions, but rename does not apply the same protection or migrate the partition references to the new name.

ui/src/pages/admin/models.tsx
```diff
@@ -0,0 +1,452 @@
...
+        endpoint,
+        model_name: modelName || undefined,
+        batch_size: intOr(batchSize, 32),
+        timeout: numOr(timeout, 30),
+        extra,
+      };
+      if (name !== editing.name) {
+        updateData.name = name;
+      }
+      onUpdate(editing.model_type, editing.name, updateData);
```
hedhoud, line 357: Editing the current default endpoint from this dialog can leave the cached "default" client stale. Callers that already resolved "default" may keep using the old endpoint after the update, because the backend invalidates the endpoint's real name but not the default alias.

ui/src/pages/admin/models.tsx
```diff
@@ -0,0 +1,452 @@
...
+                                <Star className="mr-1 h-3 w-3" /> Set Default
+                              </Button>
+                            )}
+                            <Button size="sm" variant="outline" onClick={() => handleOpenEdit(ep)}>
+                              <Pencil className="mr-1 h-3 w-3" /> Edit
+                            </Button>
+                            <ConfirmDialog
+                              title="Delete endpoint?"
+                              description={`This will delete "${ep.name}". Partitions referencing it will break.`}
+                              onConfirm={() => deleteMut.mutate({ type: ep.model_type, name: ep.name })}
```
hedhoud, line 192: Deleting the current default endpoint is allowed as long as another endpoint of the same type exists, but nothing promotes a replacement. After the registry reloads, that model type can lose its "default" alias even though a fallback endpoint still exists.

ui/src/pages/admin/users/detail.tsx
```diff
@@ -0,0 +1,551 @@
...
+    enabled: !!id,
+  });
+
+  const [displayName, setDisplayName] = useState("");
+  const [email, setEmail] = useState("");
+  const [externalId, setExternalId] = useState("");
+  const [isAdmin, setIsAdmin] = useState(false);
+  const [formLoaded, setFormLoaded] = useState(false);
+
+  if (user && !formLoaded) {
```
hedhoud, line 82: The profile form is initialized by setting state during render, and formLoaded never resets when the route moves from one user id to another. If the component is reused while navigating between users, the next user's form can keep the previous user's values and Save will submit those stale values to the new user.

ui/src/pages/admin/users/detail.tsx
```diff
@@ -0,0 +1,551 @@
...
+    setEmail(user.email ?? "");
+    setExternalId(user.external_user_id ?? "");
+    setIsAdmin(user.is_admin);
+    setFormLoaded(true);
+  }
+
+  const updateMut = useMutation({
+    mutationFn: () =>
+      updateUser(id!, {
+        display_name: displayName.trim() || undefined,
```
hedhoud: These fields cannot be cleared from the UI. Empty values are converted to undefined, so PATCH omits the field instead of sending a clear value. That matters for external_user_id in particular, because the page describes an empty external id as the token-only state.

ui/src/lib/permissions.ts
```diff
@@ -0,0 +1,53 @@
...
+  return {
+    isAdmin,
+    canViewPlatform: isAdmin,
+    canViewSystem: isAdmin,
+    canManageUsers: isAdmin,
+    canManageModels: isAdmin,
+    canManagePresets: isAdmin,
+    canManagePartitions: isAdmin,
+    canCreatePartition: true,
+    canRead: (role) => isAdmin || !!role,
```
hedhoud: The frontend treats every is_admin user as having partition read/write/member-management access. The backend only grants that partition bypass when SUPER_ADMIN_MODE=true; otherwise admins still need the partition role. This makes several UI paths look available when the API will reject them.

ui/src/pages/admin/jobs/list.tsx
andyne13: Addressed in #589 (backend): the four task endpoints (status/error/logs/cancel) now allow admin access through a central `AuthService.authorize` decision point, matching the admin jobs list that already shows every user's tasks. This UI link needs no change once #589 lands.

ui/src/pages/admin/documents/detail.tsx
andyne13: Addressed in 59c0f824: the detail view now gates delete/replace on the caller's partition role (parity with the documents list), and the copy-target list is filtered to partitions the user can write.

ui/src/pages/admin/partitions/detail.tsx
andyne13: Addressed in 59c0f824: the General settings tab is now role-aware — read-only for non-owners (banner + disabled inputs + no Save) — and the admin-only preset/model registries are only fetched for admins, so a member's view no longer fires guaranteed-403 requests.

ui/src/pages/admin/presets.tsx
andyne13: Addressed in #588 (backend): `PgPresetRepository.rename` now migrates every referencing partition to the new name atomically, and renaming onto an existing name is rejected. The reserved `default` preset is also protected from rename/delete.

ui/src/pages/admin/models.tsx
andyne13: Addressed in #588 (backend): updating the current default endpoint now also evicts the `default` alias client cache (mirroring `set_default`), so callers that resolved `default` no longer keep a stale client.

ui/src/pages/admin/models.tsx
andyne13: Addressed in #588 (backend): deleting the current default now promotes a surviving endpoint to default, so the `default` alias keeps resolving instead of disappearing.

ui/src/pages/admin/users/detail.tsx
andyne13: Addressed in 4f7c75da: the detail view is remounted via a route key when the user id changes, so the profile form re-seeds for each user instead of leaking the previous user's values.

ui/src/lib/permissions.ts
andyne13: Addressed in 59c0f824: `usePermissions` now grants the admin partition bypass only when `SUPER_ADMIN_MODE` is on (read from /config); otherwise admins are treated by their actual partition role, matching the backend. (The flag is exposed by #589.)

ui/src/components/ui/form.tsx
```diff
@@ -0,0 +1,165 @@
+import * as React from "react"
+import type { Label as LabelPrimitive } from "radix-ui"
```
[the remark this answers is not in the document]
andyne13: Verified against current code: claim 1 is a false positive — `tsc --noEmit` compiles clean, since `typeof LabelPrimitive.Root` is a type-position query that accepts the type-only namespace import. Claim 2 (the `if (!fieldContext)` guard) is the upstream shadcn pattern, and this component is currently unused in the admin UI. No functional issue — resolving.

hedhoud (approved):
LGTM, approved! 
I rechecked the Admin UI changes and the issues I previously reported all look fixed now.
The remaining backend improvements are already being handled in PR [#588](https://github.com/linagora/openrag/pull/588), so I don't see them as blockers for this Admin UI PR. Nice work!

PR #587, opened 2026-06-26
fix(embedder): plumb max_model_len into model-endpoint embedders (default 2047)

Ahmath-Gadji:
## Problem

The indexer (and API) embedders are built from the DB-backed **model-endpoint** config, whose `extra` carries only `implementation` + `api_key`. `max_model_len` is dropped, so `VLLMEmbedder` is constructed with `max_model_len=None`. That disables the `truncate_prompt_tokens` guard, and pooling models such as **Qwen3-Embedding** then **400 / hang** on context-boundary inputs ([vllm-project/vllm#29496](https://github.com/vllm-project/vllm/issues/29496)).

Setting `MAX_MODEL_LEN` did **not** help: that env var only feeds the *static* `embedder` settings, not the named model-endpoint embedders the indexer/API actually use. The indexer also has its **own** factory (`_build_embedder_factory`), separate from the API container's `make_component_factory`, so each path dropped the value independently.

Observed in the wild: a litellm-proxied endpoint returned `ContextWindowExceededError: This model's maximum context length is 2048 tokens. However, you requested 1000003 tokens`. Sending `truncate_prompt_tokens` fixes it.

## Changes

- **Backfill** `max_model_len` / `embed_concurrency` from the static `embedder` settings in **both** factories — `di/container.py` (`_embedder_extra_kwargs`) and `services/workers/indexer_pool.py` (`_build_embedder_factory`). An explicit per-endpoint `extra` value still wins, so it stays **presetable per endpoint**.
- **Default** `max_model_len` `8192 → 2047` (one below the common 2048 boundary) so a fresh deployment is safe out of the box.
- **Minimal logging**: one DEBUG line at embedder construction (shows `max_model_len`), a WARNING when `max_model_len` is unset, and a WARNING when a batched embed fails (with how many batches completed). No per-batch / INFO noise.

## Tests

- `test_container.py`: factory backfills the two kwargs; new `test_embedder_extra_overrides_backfilled_max_model_len` (explicit `extra` wins).
- `test_indexer_pool.py`: new `test_embedder_factory_backfills_max_model_len_from_settings` (backfill + explicit-wins).
- Verified end-to-end: a partition pointed at an embedder endpoint with `extra.max_model_len=1000` indexes with `truncate_prompt_tokens=999`.

Base branch: `refactor/hexagonal`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated embedder default `max_model_len` to better align with prompt truncation behavior.
  * When endpoint embedder settings omit `max_model_len` or `embed_concurrency`, the service now backfills safe defaults from global embedder settings (endpoint overrides still take precedence).
  * Improved embedder logging on startup readiness and on embedding failures, including batch completion progress to aid diagnosis.

* **Tests**
  * Expanded unit tests to verify default backfilling and that per-endpoint overrides override global values.

### DIFF
--- a/tests/unit/di/test_container.py
+++ b/tests/unit/di/test_container.py
@@ -548,6 +548,8 @@ class TestPhase14NamedComponentFactories:
         llm = c.llm_factory("chat-a")
         vlm = c.vlm_factory("vision-a")
 
+        # max_model_len / embed_concurrency aren't carried by the endpoint config,
+        # so the embedder factory backfills them from the static ``embedder`` settings.
         assert embedder.kwargs == {
             "endpoint": "http://embedder:8000/v1",
             "model_name": "embed-model",
@@ -555,6 +557,8 @@ class TestPhase14NamedComponentFactories:
             "timeout": 12.5,
             "api_key": "embed-key",
             "dimension": 384,
+            "max_model_len": 2047,
+            "embed_concurrency": 4,
         }
         assert reranker.kwargs["endpoint"] == "http://reranker:8000"
         assert reranker.kwargs["model_name"] == "rank-model"
@@ -563,6 +567,14 @@ class TestPhase14NamedComponentFactories:
         assert llm.kwargs["temperature"] == 0.2
         assert vlm.kwargs["endpoint"] == "http://vlm:8000/v1"
         assert vlm.kwargs["max_tokens"] == 256
+
+    def test_embedder_extra_overrides_backfilled_max_model_len(self):
+        """A per-endpoint ``extra.max_model_len`` wins over the settings default."""
+        settings = _settings_with_named_models()
+        settings.models.embedder["embed-a"].extra["max_model_len"] = 4096
+        c = ServiceContainer(settings)
+
+        assert c.embedder_factory("embed-a").kwargs["max_model_len"] == 4096
 
     def test_named_factories_cache_by_endpoint_name(self):
         """Repeated factory calls for the same endpoint return one client."""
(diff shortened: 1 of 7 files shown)


hedhoud:
@codex review

tests/unit/di/test_container.py
```diff
@@ -564,6 +568,14 @@ def test_container_wires_named_factories_from_models_config(self):
         assert vlm.kwargs["endpoint"] == "http://vlm:8000/v1"
         assert vlm.kwargs["max_tokens"] == 256
 
+    def test_embedder_extra_overrides_backfilled_max_model_len(self):
+        """A per-endpoint ``extra.max_model_len`` wins over the settings default."""
+        settings = _settings_with_named_models()
+        settings.models.embedder["embed-a"].extra["max_model_len"] = 4096
```
hedhoud, line 574: Small fix: add coverage for `embed_concurrency` override too, for example:
  `settings.models.embedder["embed-a"].extra["embed_concurrency"] = 4`

tests/unit/di/test_container.py
```diff
@@ -564,6 +568,14 @@ def test_container_wires_named_factories_from_models_config(self):
...
+        c = ServiceContainer(settings)
+
+        assert c.embedder_factory("embed-a").kwargs["max_model_len"] == 4096
```
hedhoud: Also assert the built embedder keeps the explicit `embed_concurrency` value, e.g. `9`, so we know it is not overwritten by the default `4`. like this egs below: 
```
embedder = c.embedder_factory("embed-a")
assert embedder.kwargs["max_model_len"] == 4096
assert embedder.kwargs["embed_concurrency"] == 9
```
instead of : 
```
assert c.embedder_factory("embed-a").kwargs["max_model_len"] == 4096
```

hedhoud (approved):

PR #588, opened 2026-06-26
fix(registry): preset/model-endpoint mutation integrity (rename, default, delete)

andyne13:
Backend integrity fixes surfaced while reviewing the admin UI: several admin registry mutations did not maintain referential / default-alias integrity. Each is server-side and independent of the UI.

## Fixes

- **Preset rename orphaned partition references.** Partitions reference a preset by its plain name string (`partitions.indexation_preset` / `retrieval_preset`, no FK). A rename did `DELETE old + INSERT new`, leaving every referencing partition pointing at a name that no longer existed. `PgPresetRepository.rename` now migrates references atomically (insert new → repoint partitions → drop old, one transaction), and the service rejects renaming onto an existing name.

- **Reserved `default` preset.** The `default` preset is the system fallback (schema `server_default`, `create_partition`, and `resolve_partition_row` all key off the literal name; `seed_defaults` only re-seeds when a type has zero rows). Renaming it away or deleting it would dangle that fallback for every new/unset partition. Both are now rejected with 422; editing its config stays allowed.

- **Stale `default` model-endpoint client on update.** Updating the current default endpoint only evicted the named cache entry, not the `default` alias — callers that resolved `default` kept a stale client. It now evicts `default` too (mirrors `set_default`).

- **Lost `default` model-endpoint alias on delete.** Deleting the current default promoted no replacement, so `load_all()` dropped the `default` alias entirely and a later `factory("default")` raised `KeyError`. Delete now promotes a surviving endpoint to default and evicts the stale `default` client.

## Tests / verification

- Unit suite: 1476 passing (added targeted tests for every case above).
- The one new piece of raw SQL — the rename reference-migration — was verified against a real Postgres: referencing partitions repoint correctly, `default`-referencing partitions are untouched, and the old/new preset rows swap as expected.

## Summary by CodeRabbit

* **New Features**
  * Added atomic “delete default with promotion” for model endpoints: deleting the current default automatically selects a surviving endpoint as the new default.
* **Bug Fixes**
  * Enforced a strict single-default invariant per model type across create/update/set-default/delete flows, including correct cache invalidation for the default alias.
  * Prevented destructive operations on the reserved `default` preset and made preset renames update references atomically.
* **Tests**
  * Expanded unit and integration coverage for default promotion, single-default guarantees, cache eviction, preset rename/delete safeguards.

### DIFF
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -233,6 +233,12 @@ class ModelEndpointService:
             raise NotFoundError(f"Endpoint '{name}' of type '{model_type}' not found.")
 
         new_name: str | None = fields.pop("new_name", None)  # type: ignore[assignment]
+        # 'is_default' is not a plain column edit: flipping it must clear the previous
+        # default in the same transaction, so route a truthy value through set_default
+        # (atomic clear-then-set) instead of letting the repo write a second
+        # is_default=true row. A false/None value is a no-op here — you switch the
+        # default by promoting another endpoint, never by leaving the type with none.
+        promote_to_default = bool(fields.pop("is_default", None))
 
         if fields:
             updated = await self._repo.update(name, model_type, **fields)
@@ -246,7 +252,12 @@ class ModelEndpointService:
             self._invalidate_client_cache(model_type, name)
 
         self._invalidate_client_cache(model_type, effective_name)
-        if existing.is_default:
+        if promote_to_default:
+            # Clears any prior default and sets this row in one transaction, then
+            # re-points the 'default' alias to it — never leaves two defaults.
+            await self._repo.set_default(model_type, effective_name)
+            self._invalidate_client_cache(model_type, "default")
+        elif existing.is_default:
             # The 'default' alias resolves to this endpoint, so a client built under
             # the 'default' cache key is now stale too. Evict it alongside the named
             # entry (set_default already does this; update must match).
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -268,16 +268,19 @@ class ModelEndpointService:
         if len(all_of_type) <= 1:
             raise ValidationError(f"Cannot delete the last '{model_type}' endpoint. Register a replacement first.")
 
-        await self._repo.delete(name, model_type)
+        # load_all only adds the 'default' alias for an is_default row, so deleting
+        # the default would drop it entirely and make factory("default") raise.
+        # Promote a surviving endpoint (deterministic: first by name) so 'default'
+        # keeps resolving. Delete + promote run in a single repo transaction, so a
+        # failure can never leave the type with no default.
+        promote_to = None
+        if existing.is_default:
+            survivor = next((e for e in sorted(all_of_type, key=lambda e: e.name) if e.name != name), None)
+            promote_to = survivor.name if survivor is not None else None
+
+        await self._repo.delete_and_promote_default(name, model_type, promote_to)
         self._invalidate_client_cache(model_type, name)
         if existing.is_default:
-            # load_all only adds the 'default' alias for an is_default row, so
-            # deleting the default would drop it entirely and make factory("default")
-            # raise. Promote a surviving endpoint (deterministic: first by name) so
-            # 'default' keeps resolving, and evict its now-stale cached client.
-            survivor = next((e for e in sorted(all_of_type, key=lambda e: e.name) if e.name != name), None)
-            if survivor is not None:
-                await self._repo.set_default(model_type, survivor.name)
             self._invalidate_client_cache(model_type, "default")
         await self.load_all()
 
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -246,23 +246,28 @@ class ModelEndpointService:
             updated = existing
 
         effective_name = name
+        renamed_from: str | None = None
         if new_name and new_name != name:
             await self._repo.rename(name, model_type, new_name)
             effective_name = new_name
-            self._invalidate_client_cache(model_type, name)
+            renamed_from = name
 
-        self._invalidate_client_cache(model_type, effective_name)
         if promote_to_default:
             # Clears any prior default and sets this row in one transaction, then
             # re-points the 'default' alias to it — never leaves two defaults.
             await self._repo.set_default(model_type, effective_name)
-            self._invalidate_client_cache(model_type, "default")
-        elif existing.is_default:
-            # The 'default' alias resolves to this endpoint, so a client built under
-            # the 'default' cache key is now stale too. Evict it alongside the named
-            # entry (set_default already does this; update must match).
-            self._invalidate_client_cache(model_type, "default")
+        # The 'default' alias client is stale when this endpoint becomes the default
+        # OR was already the default (its config just changed).
+        evict_default = bool(promote_to_default or existing.is_default)
+
+        # Reload the in-memory config FIRST, then evict, so a client rebuilt during
+        # the reload window from the old config cannot survive in the cache.
         await self.load_all()
+        if renamed_from is not None:
+            self._invalidate_client_cache(model_type, renamed_from)
+        self._invalidate_client_cache(model_type, effective_name)
+        if evict_default:
+            self._invalidate_client_cache(model_type, "default")
         return await self._repo.get(effective_name, model_type) or (updated or existing)
 
     async def delete_model_endpoint(self, name: str, model_type: str) -> None:
@@ -282,11 +287,15 @@ class ModelEndpointService:
         if status == "last":
             raise ValidationError(f"Cannot delete the last '{model_type}' endpoint. Register a replacement first.")
 
+        # Reload the in-memory config FIRST, then evict: evicting before load_all
+        # leaves a window where a concurrent request rebuilds a client from the old
+        # config and re-caches it; evicting after the reload drops any such stale
+        # client so the next request rebuilds from the fresh config.
+        await self.load_all()
         self._invalidate_client_cache(model_type, name)
         if promoted is not None:
             # The deleted endpoint was the default; its 'default' alias client is now stale.
             self._invalidate_client_cache(model_type, "default")
-        await self.load_all()
 
     async def set_default(self, model_type: str, name: str) -> None:
         """Promote ``name`` to the default endpoint for ``model_type``."""
@@ -294,8 +303,9 @@ class ModelEndpointService:
         if existing is None:
             raise NotFoundError(f"Endpoint '{name}' of type '{model_type}' not found.")
         await self._repo.set_default(model_type, name)
-        self._invalidate_client_cache(model_type, "default")
+        # Reload before evicting so a client rebuilt during the window can't survive.
         await self.load_all()
+        self._invalidate_client_cache(model_type, "default")
 
     # ------------------------------------------------------------------
     # Endpoint validation
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -260,27 +260,20 @@ class ModelEndpointService:
         Raises 404 if not found, 422 if it is the last endpoint of its type
         (would leave components with no fallback).
         """
-        existing = await self._repo.get(name, model_type)
-        if existing is None:
+        # The last-endpoint guard and the survivor/default choice are made INSIDE
+        # the repo's locked transaction (not from a stale snapshot here), so
+        # concurrent deletes of the same type can't both pass the count check or
+        # promote an already-deleted survivor — which would leave the type with no
+        # endpoint / no default. The repo reports what happened.
+        status, promoted = await self._repo.delete_and_promote_default(name, model_type)
+        if status == "not_found":
             raise NotFoundError(f"Endpoint '{name}' of type '{model_type}' not found.")
-
-        all_of_type = await self._repo.list_all(model_type=model_type)
-        if len(all_of_type) <= 1:
+        if status == "last":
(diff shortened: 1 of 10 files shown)


hedhoud:
@codex review

openrag/services/persistence/preset_repo.py
```diff
@@ -79,19 +79,29 @@ async def upsert(self, name: str, preset_type: str, config: dict) -> dict:
...
+                await conn.execute(
+                    f"UPDATE partitions SET {col} = $1 WHERE {col} = $2",
+                    new_name,
+                    old_name,
+                )
+                await conn.execute(
+                    "DELETE FROM pipeline_presets WHERE name = $1 AND preset_type = $2",
+                    old_name,
+                    preset_type,
+                )
```
hedhoud: I think we should avoid using the upsert here. The service already checks for duplicate names, but the repo could still be called directly, or another request could create the target name between the check and the update. For example, if we rename `legal` to `finance` and `finance` already exists, the upsert could update the existing `finance` preset instead of failing. Would it be safer to use a simple `UPDATE` on the existing row and let the database's unique constraint handle duplicate names?

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -265,6 +270,15 @@ async def delete_model_endpoint(self, name: str, model_type: str) -> None:
...
+        if existing.is_default:
+            # load_all only adds the 'default' alias for an is_default row, so
+            # deleting the default would drop it entirely and make factory("default")
+            # raise. Promote a surviving endpoint (deterministic: first by name) so
+            # 'default' keeps resolving, and evict its now-stale cached client.
+            survivor = next((e for e in sorted(all_of_type, key=lambda e: e.name) if e.name != name), None)
+            if survivor is not None:
+                await self._repo.set_default(model_type, survivor.name)
+            self._invalidate_client_cache(model_type, "default")
         await self.load_all()
```
hedhoud: This fixes the missing default alias, but the delete and default promotion still happen in two separate repository calls. If the delete succeeds but `set_default() ` fails (for example because of a DB error), we could end up with no default endpoint for that model type. Would it make sense to perform both operations in a single transaction? That way they're all-or-nothing and we can't leave the database in a partial state.

openrag/services/persistence/preset_repo.py
andyne13: Done in a9c2ca40: rename now uses a plain UPDATE (with the config) instead of an upsert, so the (name, preset_type) unique constraint rejects a collision with an existing target name atomically — no silent overwrite, even on a direct call or a check-then-act race — and the row keeps its created_at. Follow-up ffb89b30 also raises NotFoundError if old_name is concurrently deleted mid-rename.

openrag/services/orchestrators/model_endpoint_service.py
andyne13: Done in a9c2ca40 and hardened in ffb89b30: delete + survivor-promotion now run in a single repo transaction, and (per CodeRabbit's follow-up on the same code) the last-endpoint guard and survivor choice are made INSIDE that transaction under SELECT ... FOR UPDATE on the model type. Concurrent deletes of the same type serialize, so they can't both pass a stale count check or promote an already-deleted survivor — the type can never be left with no endpoint or no default.

openrag/services/orchestrators/model_endpoint_service.py
hedhoud, line 276: I think there's still a small stale-cache race here.
We evict the client cache before reloading `Settings.models`. If another request comes in during that window, it can rebuild a client using the old configuration and cache it again. Since we don't evict after `load_all()`, that stale client could stick around.
Would it make sense to either move the eviction after `load_all()`, or evict again after the reload (including the `default` alias when applicable)?

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -233,6 +233,12 @@ async def update_model_endpoint(self, name: str, model_type: str, **fields: obje
             raise NotFoundError(f"Endpoint '{name}' of type '{model_type}' not found.")
 
         new_name: str | None = fields.pop("new_name", None)  # type: ignore[assignment]
+        # 'is_default' is not a plain column edit: flipping it must clear the previous
+        # default in the same transaction, so route a truthy value through set_default
+        # (atomic clear-then-set) instead of letting the repo write a second
+        # is_default=true row. A false/None value is a no-op here — you switch the
+        # default by promoting another endpoint, never by leaving the type with none.
+        promote_to_default = bool(fields.pop("is_default", None))
```
Ahmath-Gadji, line 246: **Bug found in review (fixed in this PR): `update` could create two defaults for one model type**

`is_default` was in the repo's `_ALLOWED_UPDATE_FIELDS`, so `PUT /model-endpoints/{type}/{name}` with `{"is_default": true}` ran a bare `UPDATE ... SET is_default = true` (`model_endpoint_repo.py`) that never cleared the previous default. Consequences:

- Two `is_default=true` rows for one `model_type`.
- `load_all()` resolves the `"default"` alias to whichever endpoint sorts **last by name** — not the one the caller just promoted.
- The eviction added in this PR is gated on `existing.is_default` (the *old* value), so promoting a previously-non-default endpoint also left the `"default"` client cache stale.

**Verified against the live admin API** (anonymised): with two endpoints of one type and one set as default, a single `PUT {"is_default": true}` on the *other* endpoint produced **two** defaults, and the `"default"` alias then resolved to the wrong one. `POST …/set-default` (the dedicated path) correctly kept exactly one.

Pre-existing (the field was already updatable on `refactor/hexagonal`) but squarely on this PR's default-integrity theme, so fixed here: `is_default` dropped from the bare-update allowlist, and a truthy `is_default` routed through `set_default` (atomic clear-then-set + `"default"` eviction). `false`/`None` is a no-op — the default is switched by promoting another endpoint, never by leaving the type with none.

Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/61258c3786413043b82c9e1a50903fdf571f4000.

- `model_endpoint_repo.py`: `is_default` removed from `_ALLOWED_UPDATE_FIELDS` (with a comment noting the live-API repro), so a bare `UPDATE` can never write a second default.
- `model_endpoint_service.py:update_model_endpoint`: a truthy `is_default` is routed through `set_default` (atomic clear-then-set, evicts the `"default"` alias); `false`/`None` is a no-op.
- Tests: unit (`test_update_is_default_promotes_atomically_single_default`, `test_update_is_default_false_does_not_demote_or_touch_default_cache`) + an HTTP-level integration regression (`test_update_is_default_keeps_single_default`) asserting exactly one default after the update.

Validated locally: `ruff check`/`format` clean, 396 unit tests pass. The integration test runs against the live server in the `api-tests` CI job.

openrag/services/persistence/preset_repo.py
```diff
@@ -96,6 +97,11 @@ async def rename(self, old_name: str, new_name: str, preset_type: str, config: d
                     config,
                     preset_type,
                 )
```
hedhoud, line 106: The rename looks much safer now, but I think there's still one small consistency edge case.
This transaction updates existing partition references from `old_name` to `new_name`, but partition preset references are stored as plain strings. If another request creates or updates a partition while the rename is in progress, it could still save `old_name`, leaving behind a stale reference.
Would it make sense to serialize preset renames with partition writes, for example by using a shared advisory lock or another locking mechanism?

openrag/services/persistence/model_endpoint_repo.py
```diff
@@ -132,28 +132,46 @@ async def set_default(self, model_type: str, name: str) -> None:
...
-        # no default endpoint. ``promote_to is None`` => a plain delete.
+    async def delete_and_promote_default(self, name: str, model_type: str) -> tuple[str, str | None]:
+        # Lock every row of this model_type (FOR UPDATE) so concurrent deletes of
+        # the same type serialize, then make the last-endpoint guard and survivor
+        # choice from the LOCKED, current state — not a stale snapshot. Otherwise two
+        # concurrent deletes could each pass a snapshot count check and remove the
+        # last row, or promote a survivor that another tx just deleted, leaving the
+        # type with no endpoint / no default. Returns (status, promoted_name):
+        # status is "not_found" | "last" | "ok"; promoted_name is set only when the
+        # deleted row was the default and a survivor was promoted.
```
hedhoud, line 191: This method now handles the most important part of the endpoint deletion logic, so I think it would be worth adding a few repository-level tests for it.
The service tests already cover the expected behavior, but they use a fake repository. It would be nice to have a few focused tests against the real repository as well, covering cases like deleting a missing endpoint, deleting the last endpoint, deleting a non-default endpoint, deleting the default endpoint, and making sure the transaction behaves correctly.
That would help prevent this logic from accidentally breaking in the future.

openrag/services/orchestrators/model_endpoint_service.py
andyne13: Good catch — fixed in da9269c3: eviction now runs AFTER load_all() in update/delete/set_default, so a client rebuilt during the reload window can't survive stale in the cache (the 'default' alias is re-evicted there too).

openrag/services/persistence/model_endpoint_repo.py
andyne13: Added in da9269c3: PgModelEndpointRepository tests for delete_and_promote_default covering missing, last, non-default, and default-promotes-survivor (asserting the FOR UPDATE lock + DELETE + promotion SQL).

openrag/services/persistence/preset_repo.py
andyne13: Valid edge — a partition write committing old_name concurrently with the rename can still leave a stale reference. The robust fix (a shared advisory lock spanning preset rename AND the partition write path) is cross-cutting and touches the partition-write hot path, so we're deferring it to a focused follow-up after the pre-release rather than expanding this PR. Leaving this thread open to track it.

hedhoud:
I think one default-endpoint path is still worth closing or tracking explicitly. The PR fixes update/delete integrity, but create and set-default can still affect the same invariant: for each model type, we should have one valid default endpoint.

A create request marked as default can still add another default without demoting the current one, and set-default checks the target before the transaction that clears/promotes defaults. If the target disappears in that gap, the model type can end up with no default.

I know this may pre-exist, but because this PR is about registry integrity, it is easy to read it as covering the whole invariant. Could we either route these paths through the same transactional default-promotion logic, or explicitly track them as follow-up items separate from the preset rename / partition-write race?

openrag/services/persistence/model_endpoint_repo.py
```diff
@@ -119,8 +128,25 @@ async def delete(self, name: str, model_type: str) -> bool:
         return result == "DELETE 1"
 
     async def set_default(self, model_type: str, name: str) -> None:
```
EnjoyBacon7, line 142: **Bug (found in review, fixed here): `set_default` can leave a model type with NO default under a concurrent delete**

Same "never leave a type defaultless" invariant that `delete_and_promote_default` was just hardened for — but `set_default` was left with two bare `UPDATE`s, **no `FOR UPDATE` and no existence check**:

```
UPDATE … SET is_default=false WHERE model_type=$1   -- clears the old default
UPDATE … SET is_default=true  WHERE name=$1 …       -- 0 rows if name was just deleted
```

Concretely — type `llm` with `primary` (default) + `backup`; two concurrent admin requests, `POST …/backup/set-default` and `DELETE …/backup`:
1. set-default: service `get("backup")` → exists ✓
2. delete: `delete_and_promote_default("backup")` removes it (non-default, not last) and **commits** → only `primary` left, still default
3. set-default txn: `UPDATE … is_default=false WHERE model_type='llm'` → clears `primary`
4. set-default txn: `UPDATE … is_default=true WHERE name='backup'` → **0 rows** (gone)
5. **commit → `llm` has a healthy endpoint but no default**; `load_all()` registers no `"default"` alias, so the default LLM stops resolving until an admin re-runs set-default.

Low probability (admin-only path, ms-scale window) but high, silent impact.

**Fix:** `set_default` now `SELECT … FOR UPDATE`s the type's rows, raises `NotFoundError` if the target is gone, then clears-then-sets inside the transaction — so a concurrent delete serializes (either set-default wins, or it sees `backup` gone and 404s) and the type is never left defaultless. Mirrors `delete_and_promote_default`.

Ahmath-Gadji:
> I think one default-endpoint path is still worth closing or tracking explicitly. The PR fixes update/delete integrity, but create and set-default can still affect the same invariant: for each model type, we should have one valid default endpoint.
> 
> A create request marked as default can still add another default without demoting the current one, and set-default checks the target before the transaction that clears/promotes defaults. If the target disappears in that gap, the model type can end up with no default.
> 
> I know this may pre-exist, but because this PR is about registry integrity, it is easy to read it as covering the whole invariant. Could we either route these paths through the same transactional default-promotion logic, or explicitly track them as follow-up items separate from the preset rename / partition-write race?

Solved with:
https://github.com/linagora/openrag/commit/61258c3786413043b82c9e1a50903fdf571f4000
https://github.com/linagora/openrag/pull/588/changes/7e693c15cf84057893b700403568778bd347a256
https://github.com/linagora/openrag/pull/588/changes/1477003a14fc2f32d040101ce46ec6ecce7b8675

openrag/services/persistence/model_endpoint_repo.py
EnjoyBacon7: Fixed in commit https://github.com/linagora/openrag/commit/7e693c15cf84057893b700403568778bd347a256.

- `set_default` now selects the type's rows `FOR UPDATE`, raises `NotFoundError` if the target endpoint is absent, then clears-then-sets inside the same transaction.
- Repo tests: `test_set_default_locks_rows_then_runs_two_updates` (asserts `FOR UPDATE` + both `is_default` updates) and `test_set_default_raises_not_found_without_clearing_when_target_missing` (asserts it aborts **before** clearing the old default).

Validated locally: `ruff` clean, 44 unit tests pass (repo + service). Note this also covers the `update_model_endpoint(is_default=true)` path, which routes through `set_default`.

hedhoud (approved):

PR #589, opened 2026-06-29
fix(auth): admins can access any task; expose super_admin_mode; introduce authorize() PDP

andyne13:
First slice of the admin-UI permission work (review item #1), done through a small, extensible authorization seam rather than an inline bypass.

## Problem
Admins see every user's task in the admin jobs list, but the task status/error/logs/cancel endpoints (all guarded by `require_task_owner`) were owner-only — no admin path. So an admin clicking another user's job in the admin UI got a guaranteed **403 dead-end**.

## Change
- **Single authorization decision point:** `AuthService.authorize(user, action, resource)` — deny-by-default, policy in one place instead of inlined at call sites (the fragmentation that produced these bugs). Today it handles `task:access` (owner **or** admin); unknown actions raise rather than silently allow. It's the seam other checks migrate onto, with a `TODO(org)` for org-scoped admins.
- `require_task_owner` now delegates to it (covers all four task endpoints).
- **Expose `super_admin_mode`** via the admin-only `/config` so the admin UI permission layer can mirror the backend rule (admins only bypass partition checks when `SUPER_ADMIN_MODE` is on) instead of assuming every admin has partition access.

## Scope / non-goals
- No org model — extensible seam only (see `docs/refactoring/AUTHORIZATION_ARCHITECTURE.md`, Phase 0).
- The MCP path has its own divergent task-auth; **intentionally not touched** here — it needs a separate senior review of the MCP auth model.

## Verification
- Unit: 1471 passing (added direct `authorize()` tests + `require_task_owner` admin-allow / non-owner-deny).
- Running stack: admin → another user's task **200** (was 403); owner → own task **200**; non-owner non-admin **403**; `/config` returns `super_admin_mode` with `rdb` intact.

## Summary by CodeRabbit

* **New Features**
  * Updated the admin `/config` response to return settings in a client-friendly serialized format and added a `super_admin_mode` field.
  * Introduced centralized authorization policy support for action-based decisions.

* **Bug Fixes**
  * Improved task access authorization so admins can access tasks regardless of ownership, while non-admins remain restricted; unknown actions are rejected.

* **Tests**
  * Added unit tests covering admin vs non-admin task access and validation for unknown authorization actions.

### DIFF
--- a/openrag/api/dependencies/auth.py
+++ b/openrag/api/dependencies/auth.py
@@ -5,7 +5,6 @@ from core.utils.exceptions import OpenRAGError
 from core.utils.logging import get_logger
 from di.providers import get_auth_service, get_config, get_job_service, get_partition_service
 from fastapi import Depends, HTTPException, Request, status
-from services.orchestrators.auth_service import AuthService
 
 logger = get_logger()
 
@@ -193,6 +192,7 @@ async def require_task_owner(
     task_id=Depends(request_task_id),
     user=Depends(current_user),
     job_service=Depends(get_job_service),
+    auth_service=Depends(get_auth_service),
 ):
     task_details = await job_service.get_task_details(task_id)
     if not task_details:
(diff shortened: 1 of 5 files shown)


hedhoud (approved):
Approved!
I tested the task access flow locally: owner access works, admin access to another user’s task works, and another non-admin user is still denied. `/config` also exposes `super_admin_mode` as expected.

Ahmath-Gadji (approved):
I've tested it and it works

PR #593, opened 2026-06-30
ci: add build_rc.yml on main so the RC workflow is dispatchable

andyne13:
`workflow_dispatch` workflows are only runnable when the file exists on the **default branch** (`main`). The temporary RC image build (PR #592) targets `refactor/hexagonal`, so without this it would not be dispatchable at all.

This adds the **identical** `build_rc.yml` to `main` so the workflow is exposed/dispatchable. Its in-workflow `guard` restricts actual runs to `github.ref == refs/heads/refactor/hexagonal` (and to `v2.0.0-rc.N` tags), so the file is **inert on `main`** — dispatching it against `main` is refused.

Flow: merge #592 (adds it to `refactor/hexagonal`) + this PR (adds it to `main`) → dispatch `Build RC` selecting the `refactor/hexagonal` ref → builds api/ray/admin-ui RC images.

**Temporary** — delete alongside the `refactor/hexagonal` copy after the v2.0 cutover to `main`.

Found via codex review on #592 (P1: workflow not reachable from the default branch).

## Summary by CodeRabbit

* **Chores**
  * Updated the GitHub Actions workflow to publish v2.0.0 RC container images in two phases: build/push run-scoped staging tags first, then promote to the final RC tags only after all images succeed.
  * Added workflow guardrails: runs only on the intended branch, validates `version` format, and prevents overwriting already-published RC images unless forced.
  * Promotion retags images without rebuilding and updates both GHCR and Docker Hub for the API, Ray, and Admin UI images.

### DIFF
--- a/.github/workflows/build_rc.yml
+++ b/.github/workflows/build_rc.yml
@@ -9,6 +9,11 @@
 # pipeline) builds only api + ray — it needs the admin-ui job added before GA.
 # Never tags `latest`.
 #
+# Publishing is ATOMIC: the three image jobs build and push a run-scoped staging
+# tag; the final `v2.0.0-rc.N` tags are created only by the `promote` job, which
+# runs after all three succeed — so a partial failure never leaves a half-
+# published RC. (Run-scoped `rc-staging-<run_id>` tags are left in ghcr.)
+#
 # DELETE this whole file once `refactor/hexagonal` is merged to `main` and
 # releases build from `main` again.
 name: Build RC (refactor/hexagonal)
@@ -17,13 +22,20 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Image tag to publish (e.g. v2.0.0-rc.1). Must NOT be 'latest'."
+        description: "Image tag to publish (e.g. v2.0.0-rc.1). Must match v2.0.0-rc.N."
         required: true
         default: "v2.0.0-rc.1"
+
+# Serialize runs that target the same version so two dispatches can't race to
+# publish the same tags.
+concurrency:
+  group: build-rc-${{ inputs.version }}
+  cancel-in-progress: false
 
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
+  STAGING_TAG: rc-staging-${{ github.run_id }}
 
 jobs:
   guard:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_rc.yml
```diff
@@ -0,0 +1,195 @@
...
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+            linagoraai/openrag
+          tags: |
+            type=raw,value=${{ inputs.version }}
+
+      - name: Build and push Docker image
```
hedhoud: I think we should avoid pushing the final RC tag from each job directly.

Right now, if one image builds successfully and another one fails, we can still end up with a visible `v2.0.0-rc.N` tag for only part of the release. That would make the RC look published even though the full image set is not ready.

Could we first push temporary/run-specific tags, then publish the final RC tags only in a last step after all three image jobs pass?

Same concern applies to the Ray and Admin UI image jobs below.

.github/workflows/build_rc.yml
```diff
@@ -0,0 +1,195 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Image tag to publish (e.g. v2.0.0-rc.1). Must NOT be 'latest'."
+        required: true
+        default: "v2.0.0-rc.1"
+
+env:
```
hedhoud, line 43: Can we add a concurrency guard for the RC version?

Since this workflow is manual, it’s possible to start two runs with the same `version`. If they overlap, they can both push the same tags, and we may not know which run actually produced the final images.

Something keyed by the version input, with `cancel-in-progress: false`, should be enough.

.github/workflows/build_rc.yml
```diff
@@ -0,0 +1,195 @@
...
+          fi
+
+  build-and-push-image:
+    needs: guard
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
```
hedhoud, line 105: Small hardening suggestion: can we set `persist-credentials: false` on the checkout steps?

These jobs don’t need git credentials after checkout, so we don’t need to keep the token around during the Docker build.

Same applies to the other checkout steps in this workflow.

hedhoud (commented):
I left a few comments on the RC publishing flow.

.github/workflows/build_rc.yml
Ahmath-Gadji: Added in 622c0834 — workflow-level `concurrency` keyed by the version input (`group: build-rc-${{ inputs.version }}`, `cancel-in-progress: false`), so two dispatches of the same `version` serialize instead of racing to push the same tags.

.github/workflows/build_rc.yml
Ahmath-Gadji: Done in 622c0834 — `persist-credentials: false` is now set on all three `actions/checkout` steps (the jobs only build/push images and need no git access after checkout).

Ahmath-Gadji (approved):
LGTM

hedhoud (commented):
I left a few comments on the RC workflow pairing and publish safety.

.github/workflows/build_rc.yml
```diff
@@ -0,0 +1,195 @@
...
+          context: .
+          file: infra/docker/ui.Dockerfile
+          platforms: linux/amd64
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-admin-ui:${{ env.STAGING_TAG }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+  promote:
+    # Atomic publish: only after all three image builds succeed, retag the
```
andyne13: Fixed in de08ee9c (identical to #595 131e4261) — softened the wording instead of promising full atomicity. Header now reads "Publish safety (NOT fully atomic)" and both the header and the promote-job comment state the three retags are sequential and not transactional across images, with the force:true recovery path documented inline.

.github/workflows/build_rc.yml
```diff
@@ -0,0 +1,195 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Image tag to publish (e.g. v2.0.0-rc.1). Must match v2.0.0-rc.N."
+        required: true
+        default: "v2.0.0-rc.1"
+
+# Serialize runs that target the same version so two dispatches can't race to
+# publish the same tags.
+concurrency:
```
andyne13, line 39: Added in de08ee9c (identical to #595 131e4261). New `force` boolean input + a `guard` preflight that fails when the target `v2.0.0-rc.N` already exists unless force=true, so a later re-dispatch can no longer silently overwrite a published RC. force=true is the explicit override for a deliberate re-cut or partial-promote recovery.

andyne13:
Pushed de08ee9c (byte-identical to #595 131e4261) addressing the review:

- **"atomic" wording:** softened to "Publish safety (NOT fully atomic)"; header + promote-job comment now state the retags are sequential and not transactional across images, with the force:true recovery path inline.
- **re-dispatch overwrite:** added a `force` boolean input + a `guard` preflight that fails when the target `v2.0.0-rc.N` already exists unless `force=true`.
- **PR pairing:** confirmed — kept byte-identical to #595 and **merged together**; this copy makes the workflow dispatchable from the default branch, #595 is what runs.

Re-requesting review 🙏

hedhoud (approved):
Looks good after the latest updates. The previous concerns are handled.

PR #604, opened 2026-07-01
fix(admin-ui): guide upload action by partition capability

### LINKED ISSUE #598: Admin UI overview should not advertise upload actions to read-only users
Context
In the live Admin UI test, a viewer user correctly had no Upload button inside the assigned partition documents view. However, the overview still showed an Upload Documents quick action.

Problem
This creates an inconsistent experience: the overview suggests the user can upload, but the actual partition view prevents it. For viewer-only users this looks like a broken or misleading shortcut.

Expected behavior
Only show upload quick actions when the user has write access to at least one partition, or change the action text so it does not imply write permissions.

hedhoud:
## Summary

Fixes #598.

The Overview upload quick action now checks whether the current user has a writable partition.

- Users with an `owner` or `editor` partition still get **Upload Documents**, pre-filtered to that writable partition.
- Users with no writable partition are guided to create a partition first, instead of seeing an upload action that later disappears on the Documents page.

This keeps the UI aligned with the partition-scoped role model: a viewer on one partition may still create and own another partition.

## Validation

- `npm run test -- overview.test.tsx`
- `npm run lint`
- `npm run test`
- `npm run build`

## Summary by CodeRabbit

* **New Features**
  * The admin overview’s “Upload Documents” quick action now adapts to partition write access, routing uploads to the best available partition.
  * Remembers a previously used partition in the session to improve upload destination selection.
  * When no writable partition is available, it prompts users to create one via an in-page dialog and a “create partition” link.

* **Bug Fixes**
  * Hides the upload action when partitions are still loading or fail to load, and when partition creation isn’t available.

* **Tests**
  * Added coverage for the overview quick-action rendering scenarios.

### DIFF
--- a/ui/src/pages/admin/overview.test.tsx
+++ b/ui/src/pages/admin/overview.test.tsx
@@ -95,11 +95,11 @@ describe("OverviewPage quick actions", () => {
 
     await userEvent.click(screen.getByRole("button", { name: /upload documents/i }));
 
-    expect(screen.getByRole("heading", { name: /create a partition before uploading/i })).toBeTruthy();
+    expect(await screen.findByRole("heading", { name: /create a partition before uploading/i })).toBeTruthy();
     expect(
-      screen.getByText("You don't have a partition you can upload to yet. Create one first?"),
+      await screen.findByText("You don't have a partition you can upload to yet. Create one first?"),
     ).toBeTruthy();
-    expect(screen.getByRole("link", { name: /^create partition$/i }).getAttribute("href")).toBe(
+    expect((await screen.findByRole("link", { name: /^create partition$/i })).getAttribute("href")).toBe(
       "/partitions?create=1",
     );
   });
--- a/ui/src/pages/admin/overview.test.tsx
+++ b/ui/src/pages/admin/overview.test.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { MemoryRouter } from "react-router-dom";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
@@ -8,12 +9,17 @@ const queryData = vi.hoisted(() => ({
   partitions: [
     {
       partition: "shared-docs",
-      name: "shared-docs",
+      name: "Shared docs display name",
       role: "viewer",
       document_count: 3,
       created_at: "2026-01-01T00:00:00Z",
     },
   ],
+  isError: false,
+  isLoading: false,
+}));
+const permissions = vi.hoisted(() => ({
+  canCreatePartition: true,
 }));
 
 vi.mock("@tanstack/react-query", () => ({
@@ -22,7 +28,8 @@ vi.mock("@tanstack/react-query", () => ({
     if (key === "partitions") {
       return {
         data: { partitions: queryData.partitions },
-        isLoading: false,
+        isError: queryData.isError,
+        isLoading: queryData.isLoading,
       };
     }
     if (key === "tasks") {
@@ -55,17 +62,21 @@ vi.mock("@/lib/permissions", () => ({
   usePermissions: () => ({
     canManageUsers: false,
     canViewSystem: false,
-    canCreatePartition: true,
+    canCreatePartition: permissions.canCreatePartition,
     canWrite: (role: string | null | undefined) => role === "editor" || role === "owner",
   }),
 }));
 
 describe("OverviewPage quick actions", () => {
   beforeEach(() => {
+    sessionStorage.clear();
+    permissions.canCreatePartition = true;
+    queryData.isError = false;
+    queryData.isLoading = false;
     queryData.partitions = [
       {
         partition: "shared-docs",
-        name: "shared-docs",
+        name: "Shared docs display name",
         role: "viewer",
         document_count: 3,
         created_at: "2026-01-01T00:00:00Z",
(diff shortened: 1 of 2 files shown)


hedhoud:
@coderabbitai review

andyne13:
Reviewed + smoke-tested this on a running stack (both the super-admin and a regular-user perspective). The three-state model is implemented cleanly and it correctly fixes the #598 dead-end. Three refinements worth considering before merge — one behavior fix, one cheap polish, one known limitation:

**1. The upload quick action overrides the Documents page's remembered partition (behavior regression).**
`overview.tsx` links to `/documents?partition=${writablePartition.name}`, where `writablePartition` is the *first* writable partition. But the Documents page resolves selection as `searchParams.get("partition") || remembered || partitions[0]` (`documents/list.tsx:60`) — so the explicit `?partition=` **wins over the remembered (sessionStorage) partition**. Result: every other entry into Documents (sidebar, back button, Jobs→Documents) lands on your last-viewed partition, but the Overview "Upload Documents" action jumps to the first writable one instead. For an admin in `SUPER_ADMIN_MODE` (everything writable) that's an arbitrary `partitions[0]`, discarding the remembered choice. Pre-PR the action was a bare `/documents`, which respected `remembered`.
*Suggested fix:* prefer the remembered partition when the user can write to it, else fall back to the first writable one — preserves continuity in the common case while still guaranteeing the upload button is present.

**2. The action is hidden while loading, but not on error.**
`uploadQuickAction` is gated on `partitionsQuery.isLoading` only. If the partitions fetch **errors**, `partitions` falls back to `[]` → no writable partition found → the UI shows "Create Partition to Upload" even to a user who actually has writable partitions. One-liner: also neutralize on `partitionsQuery.isError` (e.g. `isLoading || isError ? null : …`).

**3. `canCreatePartition` is hardcoded `true`, so the "create" guidance can dead-end in capped deployments (known limitation).**
`permissions.ts` sets `canCreatePartition: true` unconditionally, and the backend never surfaces `MAX_PARTITIONS_PER_USER`/owned-count. In a deployment with `MAX_PARTITIONS_PER_USER=0` (or a user at the cap), a no-writable-partition user is steered to "Create Partition to Upload" → the create dialog → a `403 PARTITION_LIMIT_EXCEEDED` (it does degrade gracefully to a toast, not a crash). The intended `: null` "hide" branch is effectively dead code until `canCreatePartition` becomes dynamic. Not a regression and fine for the default config (cap 100) — but making it fully correct needs the backend to expose a partition-create capability (e.g. on `/users/info`). Fine to accept as a known limitation with a follow-up.

None of these are crashes or security issues; #1 is the one I'd fix here, #2 is cheap, #3 can be a follow-up.

andyne13:
Follow-up on the state-2 design (no writable partition), after testing as a fresh non-admin user — I think the current approach should be reconsidered.

**The problem:** when the user has no writable partition, the Quick Actions now shows *two* buttons that are effectively identical:
- **"Create Partition"** (the pre-existing action, gated on `canCreatePartition`) → `/partitions?create=1`, Plus icon.
- **"Create Partition to Upload"** (this PR's state-2 variant) → **also `/partitions?create=1`**, **also Plus icon**.

Same icon, same destination, both worded "Create Partition…" — so state-2 duplicates an action that's already on the page. It reads as inconsistent/confusing rather than helpful.

**Suggested approach:** keep a **single, consistent "Upload Documents"** action (don't mutate its label into a near-duplicate of "Create Partition"). When the user has no writable partition, clicking it opens a short **info dialog** instead of navigating straight to the documents view:

> *"You don't have a partition you can upload to yet. Create one first?"* → **Create Partition** button → `/partitions?create=1`.

Why this is nicer:
- **Consistent entry point** — "Upload Documents" is always present and always means "I want to upload," regardless of the user's partition state (matches what every other user sees).
- **No duplicate button** — the general "Create Partition" quick action stays the only "create" button; "Upload Documents" stays the only upload button.
- **Educational** — the dialog explains *why* uploading isn't available yet (you need a partition you own/can edit), which a bare second "Create Partition" button doesn't convey. It also avoids a silent surprise-redirect from a button labelled "Upload".

So the state-2 branch becomes "open an info dialog that routes to partition creation" rather than "swap the button to a second Create-Partition link." States 1 (has writable → Upload Documents) and 3 (unchanged) stay as-is.

(This is a distinct **state-2** finding — separate from my earlier review, whose three points were all state-1/other: the remembered-partition override, the missing error-state guard, and the `canCreatePartition` cap limitation. Those all still stand.)

andyne13:
Re-reviewed the reworked version (`3e80de0f`) — the earlier points are addressed, nicely done:

- **Remembered-partition override** → fixed: `rememberedWritablePartition … ?? writablePartitions[0]` prefers the remembered partition when it is writable. ✅ (covered by the "prefers the remembered partition" test)
- **Error state** → fixed: the action is now gated on `isLoading || isError`, so a failed partitions fetch no longer mislabels the action. ✅
- **State-2 redesign** → done: a single **"Upload Documents"** action that, with no writable partition, opens a "Create a partition before uploading" dialog → **Create Partition**, instead of a second "Create Partition to Upload" button. This removes the duplication with the existing Create-Partition quick action. ✅
- Super-admins also get remembered-if-writable rather than an arbitrary first partition. ✅

Tests cover link / dialog / error / hidden / remembered-preference — good coverage.

Two residual, non-blocking notes:
1. **`canCreatePartition` still hardcoded** — in a `MAX_PARTITIONS_PER_USER=0` (or cap-reached) deployment, the dialogs Create-Partition CTA would still 403. That needs the backend to surface a partition-create capability (e.g. on `/users/info`); fine as a separate follow-up.
2. **Sequencing with #607** — both PRs edit `overview.tsx` and each adds a new `overview.test.tsx`, so they conflict (content + add/add). Whichever merges first, the other will need a rebase + a merge of the two `overview.test.tsx` files.

LGTM otherwise.

andyne13 (approved):
LGTM ✅ — re-reviewed the reworked version (`3e80de0f`) carefully and validated it on a running stack.

The rework cleanly addresses the earlier review points:
- **Remembered-partition preference** — `rememberedWritablePartition ?? writablePartitions[0]` prefers the last-used partition when it is writable.
- **Error/loading gating** — `partitionsQuery.isLoading || partitionsQuery.isError ? null : …`, so a failed fetch no longer mislabels the action.
- **Single "Upload Documents" action** with a guidance dialog for the no-writable-partition case, replacing the duplicate "Create Partition to Upload".

Traced the logic across states (writable / remembered / none / super-admin) — sound; the `type: "link" | "dialog"` discriminated union is type-safe; tests cover link/dialog/error/hidden/remembered-preference. CI green.

Two trivial, optional nits (non-blocking):
- `overview.tsx` and `documents/list.tsx` both hardcode the `"documents.partition"` sessionStorage key — a shared constant would avoid silent drift if it ever changes.
- `partition.partition === rememberedPartition || partition.name === rememberedPartition` is redundant (`_toRow` sets both to the same value).

Known follow-ups (not this PR): `canCreatePartition` is still hardcoded, so the dialog CTA would 403 in a `MAX_PARTITIONS_PER_USER=0` deployment (needs a backend capability signal); and this needs sequencing with #607 (they conflict on `overview.tsx` + `overview.test.tsx`).

Approving.

Issue #601, opened 2026-06-30
Backend should audit admin access to user partitions and documents


hedhoud:
Context
Super-admin access is useful for operations and support, but it also means admins can inspect user-owned partitions and documents. That access should be traceable.

Problem
Without an audit trail, it is hard to answer who accessed which user content and when. This weakens accountability for sensitive document access.

Expected behavior
Record admin access to user partitions and documents with enough context to support review and incident investigation, without logging document contents or secrets.

PR #605, opened 2026-07-01
fix(admin-ui): show super admin deployment indicator

### LINKED ISSUE #600: Admin UI should make super-admin mode visually explicit
Context
When SUPER_ADMIN_MODE is enabled, admins can inspect partitions and documents beyond ordinary membership boundaries. The current badge says Admin, but it does not clearly communicate that elevated mode.

Problem
This makes sensitive access feel like normal browsing. Operators need a stronger visual cue when they are acting with platform-level access, especially around user documents.

Expected behavior
Show a clear super-admin indicator when elevated access is active, and consider a lightweight confirmation before opening another user's documents.

hedhoud:
Closes #600.

## Summary

This PR adds a persistent indicator in the Admin UI when the current user is an admin and `SUPER_ADMIN_MODE` is enabled for the deployment. The messaging is intentionally framed as a **deployment-wide configuration**, not as a special per-user role.

The startup banner now also displays **Super Admin Mode: enabled** when the same deployment flag is active, making it immediately visible to operators in the container logs.

This change does **not** modify any roles, permissions, or authorization behavior. It simply improves operator awareness while continuing to rely on the existing backend authorization rules.

## Deferred work

This PR does not implement a confirmation dialog before opening another user's documents. The frontend currently cannot distinguish between partitions owned by the current admin and partitions that are only visible because of the super-admin bypass. Supporting that workflow requires additional backend metadata (for example, an ownership or `via_super_admin` indicator) and is better handled in a follow-up change.

## Validation

* `npm run test -- super-admin-mode-indicator.test.tsx`
* `npm run lint`
* `npm run test`
* `npm run build`
* `uv run --no-env-file pytest tests/unit/core/utils/test_banner.py`
* `uv run ruff check openrag/core/utils/banner.py tests/unit/core/utils/test_banner.py`
* `uv run ruff format --check openrag/core/utils/banner.py tests/unit/core/utils/test_banner.py`
* Browser smoke test: verified that the indicator is displayed for admin users when `SUPER_ADMIN_MODE` is enabled and is not shown to non-admin users.

### DIFF
--- a/openrag/core/utils/banner.py
+++ b/openrag/core/utils/banner.py
@@ -45,6 +45,11 @@ _BOLD = "\033[1m"
 _NORMAL = "\033[22m"
 
 
+def _super_admin_mode_enabled() -> bool:
+    """Mirror the backend SUPER_ADMIN_MODE flag used by /config and auth."""
+    return os.environ.get("SUPER_ADMIN_MODE", "false").lower() == "true"
+
+
 def _status_box(version: str, docs_url: str, *, color: bool) -> str:
     """Render the boxed version / docs-URL / status panel shown under the logo.
 
@@ -56,6 +61,8 @@ def _status_box(version: str, docs_url: str, *, color: bool) -> str:
         f"API docs: {docs_url}",
         "Status: ready",
     ]
+    if _super_admin_mode_enabled():
+        plain_lines.insert(-1, "Super Admin Mode: enabled")
     width = max(len(line) for line in plain_lines) + 2
 
     if not color:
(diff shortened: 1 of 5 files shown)


andyne13 (approved):
Reviewed and smoke-tested on a running stack (as a super-admin). LGTM ✅

- The indicator is correctly gated: `enabled: isAdmin` so non-admins never hit the admin-only `/config`, strict `super_admin_mode === true`, and it is hidden for non-admins and when the flag is off. Verified the persistent banner shows across all admin pages for a super-admin, and is absent for a regular user.
- `banner.py` startup line is correct — the plain/visible line lists stay aligned with and without the flag, and the tests cover both.
- Framing is right: deployment-wide awareness, not a per-user role, and no authorization behaviour changes.
- Deferring the "confirm before opening another user's documents" dialog is the correct call — it needs backend metadata (an ownership / `via_super_admin` marker), tracked in #601.

Only trivial, non-blocking nits (not requesting changes): `role="status"` could be `role="region"` for a persistent banner, and the component re-derives `super_admin_mode` instead of reusing `usePermissions().superAdmin` (harmless — shares the query key, no extra fetch).

Approving.

PR #606, opened 2026-07-01
fix(chainlit): render source PDF previews in the chat side panel

Ahmath-Gadji:
## Summary

Source **PDF previews in the Chainlit chat side panel were broken** — clicking a source rendered **"Failed to load PDF file"** even though the document was indexed and retrievable. This turned out to be **two independent bugs**; fixing only one is not enough.

| Symptom | Root cause |
| --- | --- |
| The PDF panel shows *"Failed to load PDF file"* and the PDF is **never even fetched** | react-pdf's **pdf.js worker** can't load — it's requested at the origin-root `/assets/…` which isn't auth-bypassed → `403 application/json` → the browser blocks the module worker on a forbidden MIME type |
| Even with the worker loading, the byte stream would download instead of display | `/static/{chunk_id}` served the file as `Content-Disposition: attachment` (a download), not `inline` |

The browser console made the primary cause explicit:

> Loading the worker at `http://<host>/assets/pdf.worker.min-*.mjs` was blocked due to a forbidden MIME type (**"application/json"**).

---

## Root cause

Chainlit is sub-mounted at `/chainlit`, and its HTML asset references are rewritten to `/chainlit/assets/*`. But react-pdf computes its worker URL **at runtime in JS** as the **origin-root** `/assets/pdf.worker*.mjs` — the HTML root-path rewrite never sees it, so it misses the `/chainlit` prefix and lands on a path that isn't auth-bypassed.

### Before — the worker request is rejected

```mermaid
sequenceDiagram
    participant B as Browser<br/>(Chainlit UI @ /chainlit)
    participant A as OpenRag API

    Note over B: react-pdf boots its pdf.js worker
    B->>A: GET /assets/pdf.worker.min-*.mjs
    Note over A: /assets is NOT under the<br/>/chainlit auth bypass,<br/>and nothing serves it
    A-->>B: 403 {"detail":"Missing token"}<br/>Content-Type: application/json
    Note over B: module worker blocked —<br/>forbidden MIME "application/json"
    Note over B: ❌ "Failed to load PDF file"<br/>(the PDF /static request never happens)
```

### After — the worker (and the PDF) load

```mermaid
sequenceDiagram
    participant B as Browser<br/>(Chainlit UI @ /chainlit)
    participant A as OpenRag API

    B->>A: GET /assets/pdf.worker.min-*.mjs
    Note over A: /assets/* bypassed in auth +<br/>StaticFiles mount serves the file
    A-->>B: 200 Content-Type: application/javascript
    Note over B: pdf.js worker starts ✅
    B->>A: GET /static/{chunk_id}?token=… (Range)
    Note over A: FileResponse, Content-Disposition: inline
    A-->>B: 206 Partial Content · application/pdf
    Note over B: ✅ PDF renders in the side panel
```

---

## Changes

**`openrag/api/main.py`** — serve Chainlit's frontend asset dir at the origin-root `/assets` too (in addition to `/chainlit/assets`), so the runtime-computed worker URL resolves with a JavaScript MIME type.

**`openrag/api/middleware/auth.py`** — bypass `/assets/*` in `AuthMiddleware`. These are public frontend bundles — the same files already served (and already bypassed) under `/chainlit/assets` — so the module worker gets `application/javascript` instead of a `403 application/json`.

**`openrag/api/routers/user/download.py`** — serve `/static/{chunk_id}` **inline** for safe previewable media types (PDF / image / audio / video) so the viewers render in place. `FileResponse` defaults to `attachment` when `filename` is set. HTML / SVG deliberately stay `attachment` so a crafted indexed file can't turn this same-origin route into a stored-XSS vector.

> Note: the two `/assets` changes are a pair — the `main.py` mount **serves** the file (turns `404`→`200`), the `auth.py` bypass **lets the request reach** it (turns `403`→ reaches the mount). Removing either regresses the preview.

---

## Verification

Fresh request (no browser cache), against a running stack:

| Request | Before | After |
| --- | --- | --- |
| `GET /assets/pdf.worker.min-*.mjs` | `403 application/json` (worker blocked) | **`200 application/javascript`** |
| `GET /static/{chunk_id}?token=…` | `200 … attachment` (download) | **`inline · application/pdf`** (renders); range fetches return `206` |
| Path traversal on the new mount (`/assets/../…`) | — | blocked (`403` / `404`) |

Confirmed end-to-end in a **private/incognito tab** (no cache): the source PDF now renders in the Chainlit side panel.

- `ruff check` clean on all three files.
- `pytest tests/unit/api/routers/user/test_download.py` — passing.

## Summary by CodeRabbit

* **New Features**
  * Improved Chainlit UI asset delivery by serving required frontend/static files from the root `/assets` path (including the pdf.js worker).

* **Bug Fixes**
  * Prevented public asset requests from being blocked by expanding the authentication bypass to include `/assets/`.
  * Fixed standalone source downloads by restoring the correct request-token context.

* **Security / Improvements**
  * Updated static file downloads to be more browser-safe: only previewable MIME types render inline; others download as attachments, and responses now include `nosniff`.

### DIFF
--- a/openrag/api/main.py
+++ b/openrag/api/main.py
@@ -362,9 +362,26 @@ if WITH_OPENAI_API or WITH_CHAINLIT_UI:
     app.include_router(openai_router, prefix="/v1", tags=[Tags.OPENAI])
 
 if WITH_CHAINLIT_UI:
+    from pathlib import Path as _Path
+
+    import chainlit as _chainlit
     from chainlit.utils import mount_chainlit
+    from starlette.staticfiles import StaticFiles
 
     mount_chainlit(app, "./app_front.py", path="/chainlit")
+
+    # Chainlit is submounted at /chainlit, so its HTML asset references are
+    # rewritten to /chainlit/assets/*. Its bundled pdf.js worker URL, however,
+    # is computed at runtime in JS as the root-absolute /assets/pdf.worker*.mjs
+    # (no mount prefix), so react-pdf fetches the worker from the origin root.
+    # That path is not under the /chainlit auth bypass, so it returns a 403 JSON
+    # body and the browser blocks the module worker on a bad MIME type — source
+    # PDF previews then fail with "Failed to load PDF file". Serve the same asset
+    # files at /assets too (AuthMiddleware bypasses /assets/*) so the worker
+    # loads with a JavaScript MIME type.
+    _chainlit_assets = _Path(_chainlit.__file__).parent / "frontend" / "dist" / "assets"
+    if _chainlit_assets.is_dir():
+        app.mount("/assets", StaticFiles(directory=str(_chainlit_assets)), name="chainlit_root_assets")
 
 
 if __name__ == "__main__":
--- a/openrag/api/main.py
+++ b/openrag/api/main.py
@@ -362,26 +362,14 @@ if WITH_OPENAI_API or WITH_CHAINLIT_UI:
     app.include_router(openai_router, prefix="/v1", tags=[Tags.OPENAI])
 
 if WITH_CHAINLIT_UI:
-    from pathlib import Path as _Path
-
-    import chainlit as _chainlit
+    from api.chainlit_assets import mount_chainlit_root_assets
     from chainlit.utils import mount_chainlit
-    from starlette.staticfiles import StaticFiles
 
     mount_chainlit(app, "./app_front.py", path="/chainlit")
 
-    # Chainlit is submounted at /chainlit, so its HTML asset references are
-    # rewritten to /chainlit/assets/*. Its bundled pdf.js worker URL, however,
-    # is computed at runtime in JS as the root-absolute /assets/pdf.worker*.mjs
-    # (no mount prefix), so react-pdf fetches the worker from the origin root.
-    # That path is not under the /chainlit auth bypass, so it returns a 403 JSON
-    # body and the browser blocks the module worker on a bad MIME type — source
-    # PDF previews then fail with "Failed to load PDF file". Serve the same asset
-    # files at /assets too (AuthMiddleware bypasses /assets/*) so the worker
-    # loads with a JavaScript MIME type.
-    _chainlit_assets = _Path(_chainlit.__file__).parent / "frontend" / "dist" / "assets"
-    if _chainlit_assets.is_dir():
-        app.mount("/assets", StaticFiles(directory=str(_chainlit_assets)), name="chainlit_root_assets")
+    # Also serve Chainlit's bundled pdf.js worker at the origin root so source
+    # PDF previews load (see mount_chainlit_root_assets for the full rationale).
+    mount_chainlit_root_assets(app)
 
 
 if __name__ == "__main__":
--- a/openrag/api/routers/user/download.py
+++ b/openrag/api/routers/user/download.py
@@ -8,6 +8,7 @@ stored ``source`` path and confines it to ``DATA_DIR``. The URL lives under
 ``/static`` so the middleware's browser ``?token=`` access works the same way.
 """
 
+import mimetypes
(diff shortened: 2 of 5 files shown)


Ahmath-Gadji:
@coderabbits revuew this PR.

hedhoud:
@codex review

openrag/api/main.py
```diff
@@ -362,10 +362,27 @@ def get_config():
...
+    # is computed at runtime in JS as the root-absolute /assets/pdf.worker*.mjs
+    # (no mount prefix), so react-pdf fetches the worker from the origin root.
+    # That path is not under the /chainlit auth bypass, so it returns a 403 JSON
+    # body and the browser blocks the module worker on a bad MIME type — source
+    # PDF previews then fail with "Failed to load PDF file". Serve the same asset
+    # files at /assets too (AuthMiddleware bypasses /assets/*) so the worker
+    # loads with a JavaScript MIME type.
+    _chainlit_assets = _Path(_chainlit.__file__).parent / "frontend" / "dist" / "assets"
+    if _chainlit_assets.is_dir():
+        app.mount("/assets", StaticFiles(directory=str(_chainlit_assets)), name="chainlit_root_assets")
```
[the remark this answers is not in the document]
Ahmath-Gadji: Good catch — fixed in commit https://github.com/linagora/openrag/commit/655f7030efdd4414912191af1bc2ae8ddba1afb0.

The `/assets` worker mount is now factored into a shared `mount_chainlit_root_assets(app)` helper (`openrag/api/chainlit_assets.py`) and called from **both** `api.main` (mounted deployment) and `chainlit_api` (the standalone Ray Serve Chainlit origin), so `GET /assets/pdf.worker*.mjs` now resolves in Ray Serve deployments too. This mirrors how `chainlit_api` already replicates the `/static` download route for the same separate-origin reason.

hedhoud (commented):
I left one small hardening comment on the inline preview response.

openrag/api/routers/user/download.py
```diff
@@ -62,4 +63,27 @@ async def download_source(
...
+    # disposition. Restrict inline rendering to media types that cannot execute
+    # script in the app's origin: HTML / SVG / etc. stay ``attachment`` so a crafted
+    # indexed file can't turn this same-origin route into a stored-XSS vector.
+    media_type, _ = mimetypes.guess_type(file_path.name)
+    inline_ok = bool(media_type) and (
+        media_type == "application/pdf"
+        or media_type.startswith(("audio/", "video/"))
+        or (media_type.startswith("image/") and media_type != "image/svg+xml")
+    )
+    return FileResponse(
```
hedhoud, line 91: The fix makes sense to me and the Chainlit preview flow is much clearer now: the pdf.js worker can load from `/assets`, and preview-safe files like PDFs can render inline instead of being downloaded.

One small security hardening point before approval: since this route now serves some files inline and the MIME type is inferred from the filename, could we also add `X-Content-Type-Options: nosniff` on this response? That helps make sure the browser does not reinterpret a previewed file as something executable.

After that, I think this is good to go.

Ahmath-Gadji: Good call — added in ed28fd65. The `/static` response now sends `X-Content-Type-Options: nosniff` (via `FileResponse(headers=...)`), so the browser won't sniff a mislabeled file into an executable type. This stacks on top of the existing `attachment` fallback for HTML/SVG, keeping inline PDF/image/audio/video previews working since we already declare the correct Content-Type. Thanks for the review!

hedhoud (approved):
Tested the preview path locally against the latest head. The Chainlit worker is now served from /assets with a JavaScript content type, PDF sources render inline, and HTML/SVG stay as attachments with nosniff set. CI is green as well. Looks good to merge from my side.

PR #607, opened 2026-07-01
fix(admin-ui): clarify self-service partition creation

### LINKED ISSUE #599: Admin UI should align Create Partition visibility with the intended user policy
Context
Normal users currently see Create Partition in the Admin UI. This may be intentional because the backend allows users to create their own partitions, but the permission model is not obvious from the interface.

Problem
If partition creation is meant to be admin-only, the current UI exposes the wrong action. If it is meant to be user-facing, the UI should make that policy clear so it does not look like an admin control leak.

Expected behavior
Decide the intended policy and align the UI with it: either hide the action for users who should not create partitions, or label it clearly as a permitted self-service action.

hedhoud:
## Summary

This clarifies that partition creation is an intended self-service action for regular users, not an admin-only control leaking into their UI.

Normal users now see personal-partition wording in the quick action and create dialog. Admins keep the broader partition-management wording.

Closes #599

## Validation

- UI tests passed
- UI build passed
- Manual Playwright smoke check passed for a normal user and an admin

Known note: the existing local dev warning about the parent Astro tsconfig still appears, but it does not block the UI build or tests.

## Summary by CodeRabbit

* **New Features**
  * Updated admin quick actions and “create partition” UI copy to be permission-aware, showing the correct titles/descriptions for admins vs non-admins and matching the create dialog content accordingly.

* **Tests**
  * Expanded end-to-end UI test coverage for the admin overview and partitions list to verify permission-specific quick action/button/dialog wording, including upload-guidance visibility and availability scenarios.

### DIFF
--- a/ui/src/pages/admin/overview.tsx
+++ b/ui/src/pages/admin/overview.tsx
@@ -132,7 +132,13 @@ function RecentTasks({ isLoading, tasks }: { isLoading: boolean; tasks: TaskList
 }
 
 export default function OverviewPage() {
-  const { canManageUsers, canViewSystem, canCreatePartition, canWrite } = usePermissions();
+  const {
+    canManageUsers,
+    canViewSystem,
+    canManagePartitions,
+    canCreatePartition,
+    canWrite,
+  } = usePermissions();
   const [uploadHelpOpen, setUploadHelpOpen] = useState(false);
   const [rememberedPartition] = useState(() => sessionStorage.getItem("documents.partition") || "");
 
@@ -178,6 +184,10 @@ export default function OverviewPage() {
   const quotaIndexed = me?.file_count ?? 0;
   const quotaEff: number | "unlimited" =
     me?.file_quota == null || me.file_quota < 0 ? "unlimited" : me.file_quota;
+  const createPartitionLabel = canManagePartitions ? "Create Partition" : "Create Personal Partition";
+  const createPartitionDescription = canManagePartitions
+    ? "Set up a new document partition"
+    : "Create a personal partition for your documents";
   const writablePartitions = partitions.filter((partition) => canWrite(partition.role));
   const rememberedWritablePartition = writablePartitions.find(
     (partition) => partition.partition === rememberedPartition || partition.name === rememberedPartition,
@@ -300,9 +310,9 @@ export default function OverviewPage() {
                   <Link to="/partitions?create=1">
                     <Plus className="h-4 w-4 mr-2" />
                     <div className="text-left">
-                      <div className="font-medium">Create Partition</div>
+                      <div className="font-medium">{createPartitionLabel}</div>
                       <div className="text-xs text-muted-foreground">
-                        Set up a new document partition
+                        {createPartitionDescription}
                       </div>
                     </div>
                     <ArrowRight className="h-4 w-4 ml-auto" />
--- a/ui/src/pages/admin/overview.test.tsx
+++ b/ui/src/pages/admin/overview.test.tsx
@@ -5,21 +5,36 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import OverviewPage from "./overview";
 
-const queryData = vi.hoisted(() => ({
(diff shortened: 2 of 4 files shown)


andyne13:
Reviewed — the copy change is clean, correctly resolves #599 (labels partition creation as a permitted self-service action rather than hiding it), and the tests are solid. Two notes:

**1. Merge conflict with #604.** Both PRs edit `ui/src/pages/admin/overview.tsx` **and** each adds a new `ui/src/pages/admin/overview.test.tsx`, so they conflict on both files — a content conflict on the page and an add/add conflict on the test (verified against #604's current head). Since both are yours, worth sequencing: merge one, then rebase the other and reconcile `overview.tsx` + merge the two `overview.test.tsx` files.

**2. Minor consistency nit (non-blocking).** `overview.tsx` gates the label on `me?.is_admin` (the raw flag), while `partitions/list.tsx` uses `canManagePartitions` from `usePermissions`. Same result today, but `permissions.ts` is meant to be the single place pages read capabilities from — and `overview.tsx` already imports `usePermissions` — so reading an `isAdmin`/capability from there instead would keep it consistent (and org-ready for when the platform/org-admin split lands). Also a small heads-up: during load `me?.is_admin` is `undefined`, so an admin momentarily sees the "personal" wording before it flips — cosmetic.

ui/src/pages/admin/overview.tsx
```diff
@@ -178,27 +168,10 @@ export default function OverviewPage() {
...
-          type: "link" as const,
-        }
-      : canCreatePartition
-        ? {
-            title: "Upload Documents",
-            description: "Create a partition first, then add documents",
-            type: "dialog" as const,
-          }
-        : null;
+  const createPartitionLabel = me?.is_admin ? "Create Partition" : "Create Personal Partition";
```
Ahmath-Gadji: Consider gating this on `canManagePartitions` from `usePermissions()` rather than reading `me?.is_admin` directly.

The sibling page (`partitions/list.tsx`) decides the same label via `canManagePartitions`. They are equivalent today, but `permissions.ts` explicitly documents the intent that pages ask `usePermissions()` rather than read `is_admin` directly, so that only that file changes when the planned org-admin layer lands. Under that layer an org-admin would be `canManagePartitions === true` but `is_admin === false`, and this quick-action would show "Create Personal Partition" while the list page it links to shows "Create Partition".

`overview.tsx` already calls `usePermissions()` here, so this is a one-line swap:

```ts
const createPartitionLabel = canManagePartitions ? "Create Partition" : "Create Personal Partition";
const createPartitionDescription = canManagePartitions
  ? "Set up a new document partition"
  : "Create a personal partition for your documents";
```

Non-blocking — behavior is correct as-is today.

ui/src/pages/admin/overview.tsx
```diff
@@ -178,6 +178,10 @@ export default function OverviewPage() {
   const quotaIndexed = me?.file_count ?? 0;
   const quotaEff: number | "unlimited" =
     me?.file_quota == null || me.file_quota < 0 ? "unlimited" : me.file_quota;
+  const createPartitionLabel = isAdmin ? "Create Partition" : "Create Personal Partition";
```
Ahmath-Gadji: Thanks for routing this through `usePermissions()` 👍

One small consistency nit remaining: this uses `isAdmin`, while the sibling `partitions/list.tsx` decides the same label via `canManagePartitions`. They are identical today (`canManagePartitions === isAdmin`), but `permissions.ts` documents them as semantically different under the planned org layer — `canManagePartitions` becomes org-admin-scoped while `isAdmin` stays platform-operator-only. At that point an org-admin would see "Create Personal Partition" here but "Create Partition" on the list page this links to.

Since `canManagePartitions` is already available from the same `usePermissions()` call, swapping `isAdmin` → `canManagePartitions` on lines 181–182 keeps the two pages in lockstep. Non-blocking.

Ahmath-Gadji (approved):
LGTM.

PR #610, opened 2026-07-01
fix(admin-ui): let partition owners delete their own partition

### LINKED ISSUE #608: Admin UI: a partition owner (non-admin) cannot delete their own partition
Found while testing the admin UI as a regular (non-admin) user.

## Problem
A regular user who is `owner` of a partition has **no way to delete it** in the admin UI, even though the backend permits it.

## Root cause (frontend over-gating)
The backend allows it: `DELETE /partition/{name}` → `require_partition_owner` → `ensure_partition_role(..., "owner")`, and an owner has `role: "owner"` in their memberships → passes (`openrag/api/dependencies/auth.py`).

But the UI only exposes delete to admins. In `ui/src/pages/admin/partitions/list.tsx` the delete button lives in the "Actions" column, gated on `canManagePartitions` (= `isAdmin`):
- `{canManagePartitions && <TableHead>Actions</TableHead>}` and the delete block only render for admins.
- Non-admins get a role-only row with **no actions**, and `partitions/detail.tsx` has no partition-delete (its only `Trash2` removes *members*).

So a non-admin owner has no delete affordance anywhere.

## Expected
An `owner` (non-admin) can delete their own partition from the UI, with the existing confirm dialog.

## Fix direction
Gate the delete affordance on ownership — `canConfigurePartition(p.role)` (= `superAdmin || role === "owner"`) — and surface it in the non-admin row view (or the detail page), not only the admin-only Actions column.

andyne13:
Closes #608.

## Problem
A non-admin **owner** of a partition had no way to delete it in the admin UI, even though the backend allows it (`DELETE /partition/{name}` → `require_partition_owner`, which a partition owner satisfies). The delete action was gated on `canManagePartitions` (= `isAdmin`), so non-admins never saw an Actions column.

## Change
- Gate the delete affordance on `canConfigurePartition(role)` (= `superAdmin || role === "owner"`) — mirrors the backend `require_partition_owner`, so a non-admin owner can delete their own partition while a viewer/editor still cannot.
- Render the Actions column whenever the caller can act on at least one row (`showActions`), not only for admins.
- Edit (pencil → detail) stays admin-only; owners already reach the detail page via the partition name link.

## Validation
- `npx tsc -b`, `npx eslint`, `npx vitest run` — all clean (55/55).
- Verified on a dev server: a non-admin owner now sees + can use the delete button; a viewer sees none.

## Summary by CodeRabbit

* **New Features**
  * The partitions list now shows edit and delete actions only when you have the appropriate permissions.
  * The Actions column appears dynamically, so the table adapts to what you’re allowed to do.
* **Bug Fixes**
  * Improved the empty-state layout so it displays correctly whether or not the Actions column is shown.

### DIFF
--- a/ui/src/pages/admin/partitions/list.tsx
+++ b/ui/src/pages/admin/partitions/list.tsx
@@ -52,7 +52,15 @@ import { partitionDetailPath } from "@/lib/routes";
 
 type SortDir = "asc" | "desc" | null;
 
-function RowActions({ partition }: { partition: PartitionResponse }) {
+function RowActions({
+  partition,
+  showEdit,
+  showDelete,
+}: {
+  partition: PartitionResponse;
+  showEdit: boolean;
+  showDelete: boolean;
+}) {
   const queryClient = useQueryClient();
 
   const deleteMutation = useMutation({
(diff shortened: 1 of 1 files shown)


Ahmath-Gadji (approved):
## Approve — audited, correct and low-risk

Reviewed the full change against the backend contract. LGTM.

**Correctness**
- The UI gate mirrors the server exactly: `delete_partition` → `require_partition_owner` → `ensure_partition_role("owner")` allows *either* `SUPER_ADMIN_MODE && is_admin` *or* an `owner` membership (`openrag/api/dependencies/auth.py`). `canConfigurePartition(role) = superAdmin || role === "owner"` is a precise mirror, so no affordance is shown that the API would reject.
- `colSpan={(canManagePartitions ? 7 : 4) + (showActions ? 1 : 0)}` is right in all cases — 8 (admin), 5 (non-admin owner), 4 (viewer). The old `? 8 : 4` would have under-counted the new non-admin-owner row, so the update was required.
- Header and empty-state cell both key off `showActions`, so they stay consistent (including when search empties the list).

**Security**
- Client-side gating only; the server remains authoritative. The change *surfaces* an existing backend capability (owner-delete) — it grants no new privilege and weakens no control.

**Non-blocking notes (optional, not requesting changes)**
1. Behavior change beyond the stated goal: an admin *without* `SUPER_ADMIN_MODE` who isn't an owner no longer sees a delete button on that partition. This is correct (the old button 403'd), just worth noting.
2. This repo's CI runs Python only — the frontend `tsc`/`eslint`/`vitest` aren't gated here. A `list.test.tsx` covering admin/owner/viewer/mixed rows would lock in the colSpan + delete-gating for the future.

PR #612, opened 2026-07-01
feat(rate-limit): exempt admins and raise default limits

Ahmath-Gadji:
## Why

Admin users already bypass the per-user **file quota** (`check_user_file_quota` in `openrag/api/dependencies/auth.py`), but they were still subject to request **rate limiting**. Trusted operators — the admin UI polling job status, bulk indexing scripts — shouldn't be throttled. This extends the same admin exemption to `RateLimitMiddleware`.

While there, the default limits were low for real usage, so this raises them.

## Changes

**Admin bypass** (`openrag/api/middleware/rate_limit.py`)
- New `RateLimitMiddleware._is_admin(request)` reads `request.state.user["is_admin"]` — the same dict `AuthMiddleware` populates.
- `dispatch()` short-circuits for admins **before** touching the limiter, mirroring the file-quota bypass. `RateLimitMiddleware` runs *after* `AuthMiddleware` (registration is reverse of execution, see `api/main.py`), so `is_admin` is available. Unauthenticated `/auth/*` paths have no user on `request.state`, so this only ever exempts an authenticated admin.

**Raised default limits** (per-identity, per-worker moving window)
| Env | Old | New | Rationale |
|-----|-----|-----|-----------|
| `RATE_LIMIT_AUTH` | 20/min | **60/min** | Keyed on **client IP** (callers are unauthenticated during login). 20/min ≈ 10 logins/min for the *whole* IP — a shared corporate/NAT egress IP throttles a legitimate morning login rush. |
| `RATE_LIMIT_CHAT` | 60/min | **120/min** | 1 req/s is tight for streaming / agentic clients doing rapid tool-call follow-ups. |
| `RATE_LIMIT_DEFAULT` | 300/min | **600/min** | A UI polling job status while loading lists can approach 5 req/s. |

`.env.example` updated to match, with a note about the IP-keyed `/auth/*` tier.

## Not changed

The IP-keyed **auth-failure** limiter in `AuthMiddleware` is deliberately left as-is: it runs *before* authentication and only counts *failed* attempts, so a successful admin never trips it — and the caller's identity is unknown at that point, so an admin bypass isn't possible (or desirable, since it's brute-force protection).

## Tests

Added to `tests/unit/api/middleware/test_rate_limit.py`:
- `_is_admin` → true for admin dict, false for non-admin and unauthenticated.
- End-to-end: an admin sails past a 1/min limit (5 requests, all 200).

`ruff check` / `ruff format` clean; new tests pass locally.

> Note: two **pre-existing** tests in this file (`test_blocks_over_limit_with_retry_after`, `test_tiers_have_independent_budgets`) fail on `refactor/hexagonal` as well — a TestClient/`limits.aio` async-storage artifact (the memory store binds to a per-request event loop under the sync test client). Verified the limiter blocks correctly at runtime (`hit` → `[True, True, False, False]` for a 2/min limit), so it's a test-harness quirk, not a functional regression, and out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Admin users now bypass request rate limits.
  * Default rate-limit settings have been increased for general, authentication, and API chat requests.

* **Bug Fixes**
  * Improved rate-limit handling so authenticated admins are no longer blocked by normal throttling.
  * Added clearer guidance in the sample configuration for rate-limit behavior and recommended limits.

### DIFF
--- a/infra/compose/.env.example
+++ b/infra/compose/.env.example
@@ -130,11 +130,14 @@ API_BASE_URL='http://X.X.X.X:APP_PORT'          # Base URL of your FastAPI backe
 # Max partitions a non-admin user may own (-1 = unlimited; admins bypass).
 # MAX_PARTITIONS_PER_USER=100
 
-# Rate limiting (per-worker moving window, keyed on user id then client IP)
+# Rate limiting (per-worker moving window, keyed on user id then client IP).
+# Admin users bypass these limits entirely. The /auth/* tier is keyed on client
+# IP (callers are unauthenticated there): keep it high enough that a shared
+# corporate/NAT egress IP does not throttle a legitimate login rush.
 # RATE_LIMIT_ENABLED=true
-# RATE_LIMIT_DEFAULT=300/minute   # all paths except those below
-# RATE_LIMIT_AUTH=20/minute       # /auth/* (login/callback/logout)
-# RATE_LIMIT_CHAT=60/minute       # /v1/* (chat completions, tools)
+# RATE_LIMIT_DEFAULT=600/minute   # all paths except those below
+# RATE_LIMIT_AUTH=60/minute       # /auth/* (login/callback/logout)
+# RATE_LIMIT_CHAT=120/minute      # /v1/* (chat completions, tools)
 
 # MCP SERVER
 # Standalone Model Context Protocol server (openrag/api/mcp/server.py).
--- a/openrag/api/middleware/rate_limit.py
+++ b/openrag/api/middleware/rate_limit.py
@@ -5,8 +5,12 @@ Runs after AuthMiddleware (registered before it in ``api.main`` so it executes
 after auth populates ``request.state.user``). Limits are per-worker; use a Redis
 storage to share them across workers.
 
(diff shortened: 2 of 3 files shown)


hedhoud (approved):
Looks good to me. I only noticed one small test-only nit: the admin bypass test comment says the same-tier non-admin still gets 429, but the test itself only exercises the admin path. Either trimming that sentence or adding the non-admin assertion would make the coverage clearer. Not blocking from my side.

Ahmath-Gadji:
Good catch @hedhoud — the comment was over-claiming. Fixed in 4a2b3361: reworded it to describe what the test actually asserts (admin path only) and noted that the non-admin side is covered by `test_is_admin_false_for_non_admin_and_unauthenticated`.

I deliberately didn't add a live non-admin 429 assertion: exhausting the limit across `TestClient` requests is unreliable under `limits.aio`'s per-event-loop memory storage — it's the same artifact that makes the two pre-existing sibling tests (`test_blocks_over_limit_with_retry_after`, `test_tiers_have_independent_budgets`) pass in CI but fail on a local `uv run pytest`. So the discrimination is verified at the unit level instead.

PR #615, opened 2026-07-02
fix(admin): redact saved model endpoint secrets

### LINKED ISSUE #614: Redact secrets from admin-facing config and model endpoint responses
## Context

The Admin UI needs to show and manage runtime configuration and model endpoints, but it should not receive raw secret values after they have been saved.

This is slightly sensitive because model endpoint API keys are not just display data. They are needed server-side for saved endpoint validation and for the indexer/model registry to call private endpoints correctly. So the fix must redact read responses without removing or weakening the backend's stored secret handling.

## Problem

Some admin-facing read responses can expose sensitive values such as API keys, tokens, or passwords. In particular, model endpoint configuration may include an API key in its extra configuration, and the backend can return that value back to the browser.

This is a data exposure issue rather than a functionality bug. The platform may work correctly, but it reveals more sensitive information than the UI needs.

Even for trusted admins, browser-visible secrets are harder to control. They can appear in network tabs, screenshots, copied JSON, browser extensions, local logs, or support/debug sessions.

There is also a regression risk if this is fixed too aggressively: if the API simply stops returning the key and the edit form sends the redacted object back, an existing endpoint could accidentally lose its stored API key. That would break validation or indexing for private model endpoints.

## Expected behavior

Secrets should stay server-side after they are saved.

Admin-facing read responses should redact secret fields and only expose whether a secret exists. For example, the UI should be able to display that an endpoint has an API key configured, without receiving the real key.

A safer behavior would be:

- write/create/update requests may accept secret values;
- read/list/detail responses never return raw secret values;
- responses expose a simple indicator such as `has_api_key` when useful;
- updating an endpoint without a new key preserves the existing stored key;
- sending a new key rotates the stored key;
- backend services continue using the real secret internally.

## Notes

This should cover both the general configuration response and model endpoint responses, because both can be consumed by the Admin UI.

A local E2E check on `refactor/hexagonal` confirmed that a temporary model endpoint created with a fake API key returned that fake key in create, list, and detail responses. The fix should keep backend validation/indexing behavior working while preventing that browser-side exposure.

hedhoud:
## Summary

Fixes #614 by redacting saved secrets from admin-facing read responses while keeping the real values available server-side.

Model endpoint responses now hide secret values from `extra` and expose `has_api_key` so the Admin UI can show that a key exists without receiving it. The service also preserves existing stored secrets when an endpoint update omits the key, which prevents the edit form from accidentally clearing private endpoint credentials.

The `/config` response now redacts known secret fields before returning the settings object to the Admin UI.

## Why

The Admin UI needs to manage model endpoints and runtime config, but raw API keys, tokens, and passwords should not be sent back to the browser after they are saved. At the same time, saved model endpoint keys are still required by backend validation and indexing, so this keeps storage/internal use intact.

## Validation

- `uv run --no-env-file pytest tests/unit/api/routers/admin/test_phase14_admin_routers.py tests/unit/services/orchestrators/test_model_endpoint_service.py tests/unit/api/test_secret_redaction.py tests/unit/api/test_main_proxy_headers.py -q`
- `uv run --no-env-file ruff check openrag/core/utils/redaction.py openrag/api/schemas/admin/model_endpoint_schemas.py openrag/api/main.py openrag/services/orchestrators/model_endpoint_service.py tests/unit/api/test_secret_redaction.py tests/unit/api/routers/admin/test_phase14_admin_routers.py tests/unit/services/orchestrators/test_model_endpoint_service.py tests/integration/api/test_model_endpoints.py`
- `git diff --check`
- `cd ui && npm run lint && npm run test && npm run build`
- Local API smoke test: create/list/detail/update model endpoint with a fake API key; no read response contained the fake key, and Postgres still preserved the stored key after update without `api_key`.
- Playwright check against local Admin UI: edit dialog hides the key, shows the preservation hint, keeps Update enabled for an unchanged saved endpoint, and System config does not expose the fake key.

## Notes

I also ran `tests/integration/api/test_model_endpoints.py` against the local CPU stack. Three tests passed; the CRUD validation test failed on the existing local setup because it expects a mock vLLM service at `http://vllm:8000/v1`, which this CPU stack does not start.

TestSprite preflight passed, but I did not run a TestSprite suite because this change is only deployed on localhost and TestSprite requires a publicly reachable target URL.

## Summary by CodeRabbit

* **New Features**
  * Added admin support to reveal a server-stored endpoint API key on demand.
  * Enhanced the model endpoint dialog to reuse an existing API key or enter a new one, including show/hide and copy actions.
  * Endpoint details now indicate whether an API key exists without exposing its full value.
* **Bug Fixes**
  * Sensitive configuration values and endpoint `extra` secret data are now redacted in API responses.
  * Updating endpoint details preserves previously saved API keys when not changed.
  * Draft endpoint validation can reuse stored credentials and now rejects invalid URLs more reliably.

### DIFF
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -1,0 +1,92 @@
+"""Helpers for shaping public data without exposing secrets."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any
+
+REDACTED_SECRET = "<redacted>"
+
+SECRET_FIELD_NAMES = frozenset(
+    {
+        "api_key",
+        "api_token",
+        "auth_token",
+        "chainlit_auth_secret",
+        "client_secret",
+        "hf_token",
+        "oidc_client_secret",
+        "oidc_token_encryption_key",
+        "password",
+        "secret",
+        "secret_key",
+        "token",
+        "token_encryption_key",
+    }
+)
+
+
+def is_secret_field(key: str) -> bool:
+    """Return true only for known secret field names, not fuzzy token matches."""
+    return key.lower() in SECRET_FIELD_NAMES
+
+
+def redact_secrets(value: Any) -> Any:
+    """Recursively redact values for known secret fields without mutating input."""
+    if isinstance(value, Mapping):
+        return {
+            key: REDACTED_SECRET if is_secret_field(str(key)) else redact_secrets(item) for key, item in value.items()
+        }
+    if isinstance(value, list):
+        return [redact_secrets(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(redact_secrets(item) for item in value)
+    return value
+
+
+def redact_secret_mapping(extra: Mapping[str, Any] | None) -> dict[str, Any]:
+    """Drop secret fields from endpoint extras while preserving non-secret metadata."""
+    return {key: redact_secrets(item) for key, item in dict(extra or {}).items() if not is_secret_field(str(key))}
+
+
+def preserve_existing_secrets(existing: Mapping[str, Any] | None, incoming: Mapping[str, Any]) -> dict[str, Any]:
+    """Keep stored secrets when an update payload omits or echoes a redacted value."""
+    return _preserve_existing_secrets(existing or {}, incoming)
+
+
+def _preserve_existing_secrets(existing: Mapping[str, Any], incoming: Mapping[str, Any]) -> dict[str, Any]:
+    merged = dict(incoming)
+    for key, value in existing.items():
+        incoming_value = merged.get(key)
+        if is_secret_field(str(key)):
+            if key not in merged or incoming_value == REDACTED_SECRET:
+                merged[key] = value
+            continue
+        if isinstance(value, Mapping) and isinstance(incoming_value, Mapping):
+            merged[key] = _preserve_existing_secrets(value, incoming_value)
+        elif isinstance(value, list) and isinstance(incoming_value, list):
+            merged[key] = _preserve_existing_secret_lists(value, incoming_value)
+    return merged
+
+
+def _preserve_existing_secret_lists(existing: list[Any], incoming: list[Any]) -> list[Any]:
+    merged = list(incoming)
+    for index, existing_item in enumerate(existing):
+        if index >= len(merged):
+            break
+        incoming_item = merged[index]
+        if isinstance(existing_item, Mapping) and isinstance(incoming_item, Mapping):
+            merged[index] = _preserve_existing_secrets(existing_item, incoming_item)
+        elif isinstance(existing_item, list) and isinstance(incoming_item, list):
+            merged[index] = _preserve_existing_secret_lists(existing_item, incoming_item)
+    return merged
+
+
+__all__ = [
+    "REDACTED_SECRET",
+    "SECRET_FIELD_NAMES",
+    "is_secret_field",
+    "preserve_existing_secrets",
+    "redact_secret_mapping",
+    "redact_secrets",
+]
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -71,8 +71,7 @@ def redact_secrets(value: Any) -> Any:
     """Recursively redact values for known secret fields without mutating input."""
     if isinstance(value, Mapping):
         return {
-            key: REDACTED_SECRET if is_secret_field(str(key)) else redact_secrets(item)
-            for key, item in value.items()
+            key: REDACTED_SECRET if is_secret_field(str(key)) else redact_secrets(item) for key, item in value.items()
         }
     if isinstance(value, list):
         return [redact_secrets(item) for item in value]
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -71,7 +71,7 @@ def redact_secrets(value: Any) -> Any:
     """Recursively redact values for known secret fields without mutating input."""
     if isinstance(value, Mapping):
         return {
-            key: mask_secret_value(item) if is_secret_field(str(key)) else redact_secrets(item)
+            key: REDACTED_SECRET if is_secret_field(str(key)) else redact_secrets(item)
             for key, item in value.items()
         }
     if isinstance(value, list):
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -82,13 +82,19 @@ def redact_secrets(value: Any) -> Any:
 
 def redact_secret_mapping(extra: Mapping[str, Any] | None) -> dict[str, Any]:
     """Return the public model-endpoint extra shape shown in Admin UI."""
-    source = dict(extra or {})
-    redacted: dict[str, Any] = {}
-    if "api_key" in source:
-        redacted["api_key"] = mask_secret_value(source.get("api_key"))
-    if "implementation" in source:
-        redacted["implementation"] = source["implementation"]
-    return redacted
+    return _redact_endpoint_extra(dict(extra or {}))
+
+
+def _redact_endpoint_extra(value: Any, key: str | None = None) -> Any:
+    if key is not None and is_secret_field(key):
+        return mask_secret_value(value)
+    if isinstance(value, Mapping):
+        return {entry_key: _redact_endpoint_extra(item, str(entry_key)) for entry_key, item in value.items()}
+    if isinstance(value, list):
+        return [_redact_endpoint_extra(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_redact_endpoint_extra(item) for item in value)
+    return value
 
 
 def preserve_existing_secrets(existing: Mapping[str, Any] | None, incoming: Mapping[str, Any]) -> dict[str, Any]:
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -67,6 +67,11 @@ def is_masked_secret_value(value: Any) -> bool:
     )
 
 
+def is_clear_secret_value(value: Any) -> bool:
+    """Return true when an update explicitly clears a stored secret."""
+    return value is None or value == ""
+
+
 def redact_secrets(value: Any) -> Any:
     """Recursively redact values for known secret fields without mutating input."""
     if isinstance(value, Mapping):
@@ -107,6 +112,9 @@ def _preserve_existing_secrets(existing: Mapping[str, Any], incoming: Mapping[st
     for key, value in existing.items():
         incoming_value = merged.get(key)
         if is_secret_field(str(key)):
+            if key in merged and is_clear_secret_value(incoming_value):
+                merged.pop(key, None)
+                continue
             if key not in merged or incoming_value == REDACTED_SECRET or is_masked_secret_value(incoming_value):
                 merged[key] = value
             continue
@@ -114,6 +122,9 @@ def _preserve_existing_secrets(existing: Mapping[str, Any], incoming: Mapping[st
             merged[key] = _preserve_existing_secrets(value, incoming_value)
         elif isinstance(value, list) and isinstance(incoming_value, list):
             merged[key] = _preserve_existing_secret_lists(value, incoming_value)
+    for key, value in list(merged.items()):
+        if is_secret_field(str(key)) and is_clear_secret_value(value):
+            merged.pop(key, None)
     return merged
 
 
@@ -135,6 +146,7 @@ __all__ = [
     "SECRET_FIELD_NAMES",
     "is_masked_secret_value",
     "is_secret_field",
+    "is_clear_secret_value",
     "mask_secret_value",
     "preserve_existing_secrets",
     "redact_secret_mapping",
--- a/openrag/core/utils/redaction.py
+++ b/openrag/core/utils/redaction.py
@@ -72,6 +72,11 @@ def is_clear_secret_value(value: Any) -> bool:
     return value is None or value == ""
 
 
+def is_secret_placeholder_value(value: Any) -> bool:
+    """Return true for redacted values that should never be stored as secrets."""
+    return value == REDACTED_SECRET or is_masked_secret_value(value)
+
+
 def redact_secrets(value: Any) -> Any:
     """Recursively redact values for known secret fields without mutating input."""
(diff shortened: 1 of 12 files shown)


andyne13 (commented):
Thanks for this — the write-only handling is clean and extending redaction to `/config` is a good catch. A few points from reviewing the diff:

## Blocker — draft validation can send a stored key to an arbitrary URL

`validate_endpoint_draft` (`api/routers/admin/model_endpoints.py`) now reuses a stored endpoint's `api_key` (via `stored_api_key_name` / `stored_api_key_model_type`) while probing `body.endpoint`, which is fully caller-controlled. `validate_endpoint` (`services/orchestrators/model_endpoint_service.py`) then sends `Authorization: Bearer <stored key>` to `{url}/models` with no binding between `url` and the stored endpoint and no URL safety check.

So a caller can set `endpoint` to a server they control, reference any saved endpoint by name/type, and read the bearer token off the wire — recovering the key this PR otherwise makes write-only. The same unbound URL also lets the probe reach internal addresses (SSRF).

Suggestion: when a stored key is reused, bind the probe to that endpoint's saved URL (or require `body.endpoint`'s host to match it); if the URL differs, require the key to be passed explicitly rather than reused. An `is_safe_url`-style guard on the probe URL with `follow_redirects=False` would close the SSRF angle too.

## Should-fix — `/config` redaction is a denylist, not safe-by-default

`redact_secrets` redacts by exact field name against `SECRET_FIELD_NAMES`. It covers everything sensitive in the settings today, but because it's exact-match, a future secret field with an unlisted name (e.g. `access_key`, `private_key`, `signing_key`) would be returned in clear by default.

Two options: build the `/config` payload from an explicit allowlist of safe fields (new fields default to hidden), or keep the denylist but add a guard test that fails when a new `str` field is added to the settings model without being classified — so it can't silently rot.

## Minor

- `token` and `secret` as exact names will redact any field literally named that, secret or not. Fine for the current settings; just flagging in case a benign field named `token` ever appears.

The redaction helper, `has_api_key`, and preserve-on-update all look correct otherwise. 👍

andyne13 (commented):
Thanks for the quick turnaround — re-reviewed the latest commits.

**Resolved ✅**
- **Draft-validate key reuse** is now bound to the saved URL via `_same_endpoint_url` (400 otherwise). Checked it for bypasses — the compare is strict (case differences only make it stricter/fail-closed), so a stored key can't be redirected to an arbitrary host. `follow_redirects=False` on the probe is a nice addition too.
- **Denylist robustness** — `SECRET_FIELD_SUFFIXES` now catches compound names (`*_api_key`, `*_token`, `*_password`, `*_secret`, `*_access_key`, …), so things like `minio_secret_key` / `smtp_password` are covered, and the `max_tokens` false-positive test is a good guard. 👍

**🔴 New regression to fix before merge — `/config` now leaks the first 3 chars of every secret**

`redact_secrets` (used by `/config`, `main.py:333`) no longer fully redacts — it now calls `mask_secret_value`, which for any secret ≥ 8 chars returns `value[:3] + "********"` (`redaction.py:58-60`). Your own test confirms it: `redact_secrets(...)["websearch"]["api_token"] == "sea********"`.

So `/config` returns the first 3 characters of `rdb.password`, the OIDC `token_encryption_key` (session-encryption Fernet key), `oidc.client_secret`, every `api_key`, and `websearch.api_token`. That's a partial-disclosure regression from the full `<redacted>` this PR started with.

The prefix hint makes sense for a user-supplied endpoint API key (to identify which key) — but `mask_secret_value` is only ever reached via `/config` (the endpoint path drops the key through `redact_secret_mapping` + `has_api_key`). So the masking applies exclusively to infrastructure secrets, where it has no UI purpose: revealing the first 3 chars of a human-chosen DB password aids guessing, and exposing any bytes of the session-encryption key is best avoided.

Suggestion: keep `/config` on full `<redacted>` (or omit) and don't route it through `mask_secret_value` — e.g. a full-redact variant for the config path, leaving masking (if you want it) for endpoint-key display only.

**Minor (unchanged):** the validate probe still has no `is_safe_url` guard, so it can reach internal addresses with an admin-supplied key. Admin-gated and `follow_redirects=False` covers the redirect pivot, so a should-fix rather than a blocker.

openrag/core/utils/redaction.py
```diff
@@ -0,0 +1,137 @@
...
+    if isinstance(value, tuple):
+        return tuple(redact_secrets(item) for item in value)
+    return value
+
+
+def redact_secret_mapping(extra: Mapping[str, Any] | None) -> dict[str, Any]:
+    """Return the public model-endpoint extra shape shown in Admin UI."""
+    source = dict(extra or {})
+    redacted: dict[str, Any] = {}
+    if "api_key" in source:
```
Ahmath-Gadji: **Data-loss: `redact_secret_mapping` whitelists `extra` down to `api_key` + `implementation`, so Admin-UI edits silently drop all other endpoint config.**

`extra` is an open-ended kwargs bag — `di/factories.py` and `services/workers/indexer_pool.py` splat every non-`implementation` key into the client constructor, and e.g. `VLLMClient.__init__` accepts `enable_thinking` plus arbitrary `**kwargs` (temperature, top_p, …). This function keeps only `api_key` and `implementation`, so any other (non-secret) field vanishes from the read response.

That becomes permanent storage loss on any UI edit:

1. `GET` an endpoint with `extra: {implementation, api_key, enable_thinking, temperature}` → the response `extra` is reduced to `{api_key: "sk-***", implementation}`.
2. The edit dialog builds its `extra` textarea from that redacted response (`ui/src/pages/admin/models.tsx`, `handleSubmit`), so it can only echo back the two surviving keys.
3. `update_model_endpoint` replaces `extra` wholesale; `preserve_existing_secrets` re-adds only *secret* keys, so `enable_thinking`/`temperature` (non-secret, now absent from the payload) are dropped from the DB.

Net effect: editing *anything* on such an endpoint (even just the timeout) strips its inference config. Worth noting too that the JS nested-secret handling (`displayModelEndpointExtra` for `auth.token`, `headers[].api_key`) and the backend's recursive `preserve_existing_secrets` are both built for a *mask-in-place* shape that this whitelist never actually produces — so that machinery is currently unreachable through the UI round-trip.

**Suggestion:** redact `extra` in place — mask known secret fields, keep everything else (i.e. route it through the generic `redact_secrets` here) instead of whitelisting two keys. That removes the data loss and makes the nested-preservation logic on both sides coherent. `has_api_key` can still be derived from the raw `extra`.

`openrag/core/utils/redaction.py:88`

ui/src/pages/admin/models.tsx
```diff
@@ -289,18 +418,28 @@ function EndpointDialog({
...
     }
     setValidating(true);
     setValidationMsg(null);
     try {
-      const res = await validateModelEndpoint({
-        endpoint,
-        model_name: modelName || undefined,
-        api_key: apiKey,
-      });
+      const canUseStoredSecret =
```
Ahmath-Gadji, line 451: **Minor UX: a URL-only edit can't be saved without re-entering the stored key.**

Save is gated on `validated === true`. When an admin changes only the endpoint URL and leaves the stored key untouched, `handleValidate` takes the stored-key reuse path, and the backend correctly rejects it (`_same_endpoint_url` mismatch → 400) — that SSRF binding is the right call. But the side effect is that the admin can no longer validate (and therefore can't Update) a URL-only change, even though the backend `preserve_existing_secrets` would happily keep the key on save. Model-name-only edits work fine because the URL still matches.

Consider allowing Save without re-validation when only the URL changed and the key is unchanged, or prompting the admin to re-enter the key explicitly when pointing an endpoint at a new host.

Anchored on the `canUseStoredSecret` path; Save is gated by `disabled={loading || validated !== true}` in the DialogFooter.

hedhoud:
Fixed the /config redaction regression in 1637c17. Config secrets now return full `<redacted>` values again, so we no longer expose the first 3 chars of DB/OIDC/API tokens. I kept the short prefix mask only for the model-endpoint API-key display path, where it helps identify the stored key without exposing the full value.\n\nI also updated the redaction test so this specific regression is covered. I left the internal-address validate guard as a separate hardening follow-up since it is a different path and the stored-key reuse leak is already closed by the saved-URL binding.

openrag/core/utils/redaction.py
hedhoud: Fixed in c6587e1. `redact_secret_mapping` now keeps the full non-secret `extra` shape and only masks fields that are actually secrets, recursively. So values like `temperature`, `enable_thinking`, and provider-specific kwargs stay in the read response and survive a normal Admin UI edit. I also updated the tests to cover this round trip.

ui/src/pages/admin/models.tsx
hedhoud: Fixed in c6587e1. The validate flow now uses the revealed stored key when it is available, so an admin can intentionally reveal the key and validate a changed endpoint URL without copy-pasting it. If the URL changed but the stored key is still hidden, the UI now asks the admin to reveal or enter the key instead of sending the stored-key reuse request that the backend must reject.

ui/src/lib/api/models.ts
```diff
@@ -55,8 +57,128 @@ export interface ValidateModelEndpointResponse {
...
+export function mergeModelEndpointApiKeyExtra(
+  extra: Record<string, unknown>,
+  apiKey: string,
+): Record<string, unknown> {
+  const prepared = prepareModelEndpointExtraForSubmit(extra);
+  const normalizedApiKey = prepareModelEndpointExtraForSubmit({ api_key: apiKey.trim() }).api_key;
+  if (typeof normalizedApiKey === "string" && normalizedApiKey) {
+    return { ...prepared, api_key: normalizedApiKey };
+  }
+  return prepared;
```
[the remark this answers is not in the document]
hedhoud: Fixed in f2ef2b7. The UI now sends an explicit empty `api_key` when an admin clears an existing stored key, instead of omitting it. I also updated the backend preservation helper so that explicit empty or null secret values really clear the stored secret, while omitted or redacted values still preserve it. Added focused tests for both sides.

ui/src/pages/admin/models.tsx
```diff
@@ -288,19 +419,47 @@ function EndpointDialog({
...
+        (!submittedApiKey || submittedApiKey === REDACTED_SECRET)
+      ) {
+        setValidated(false);
+        const msg = "Reveal or enter the API key before validating a changed endpoint URL.";
+        setValidationMsg(msg);
+        toast.error(msg);
+        return;
+      }
+      const canUseStoredSecret =
+        editing?.has_api_key === true && !apiKey && (!submittedApiKey || submittedApiKey === REDACTED_SECRET);
```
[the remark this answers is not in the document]
hedhoud: Fixed in 9b7bca0. Validation now treats an empty API key field as an explicit clear, so it probes the endpoint without reusing the old stored key. Stored-key reuse only happens when the field still contains the redacted placeholder.

openrag/core/utils/redaction.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def _preserve_existing_secret_lists(existing: list[Any], incoming: list[Any]) -> list[Any]:
+    merged = list(incoming)
+    for index, existing_item in enumerate(existing):
+        if index >= len(merged):
+            break
+        incoming_item = merged[index]
+        if isinstance(existing_item, Mapping) and isinstance(incoming_item, Mapping):
+            merged[index] = _preserve_existing_secrets(existing_item, incoming_item)
```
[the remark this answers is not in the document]
hedhoud: Fixed in 9b7bca0. List secret preservation no longer copies secrets by raw index. It now preserves a redacted list item only when it can match the existing item by non-secret identity, keeps the safe one-item case, and drops unbacked placeholders when the item cannot be matched. Added tests for the delete/reorder case.

openrag/core/utils/redaction.py
Ahmath-Gadji: Resolved in c6587e17 — `redact_secret_mapping` now masks secrets **in place** via `_redact_endpoint_extra` (recurses, keeps non-secret fields), so the read response preserves `implementation`, `temperature`, `enable_thinking`, and nested config (nested secrets masked, e.g. `auth.token` → `nes********`). The round-trip data loss is gone, and the nested `preserve_existing_secrets` machinery is now actually exercised. The updated `test_redact_secret_mapping_keeps_non_secret_endpoint_extra_shape` locks it in. 👍

ui/src/pages/admin/models.tsx
Ahmath-Gadji: Resolved in c6587e17 (+ f2ef2b72, 9b7bca0e). A changed endpoint URL with only the masked placeholder now short-circuits with a clear message — *"Reveal or enter the API key before validating a changed endpoint URL."* — instead of the cryptic backend 400, and a revealed key (`revealedApiKeyForEditing`) now feeds validation so the admin can proceed against the new host with the key sent explicitly. Explicit-clear (empty field → `api_key: ""`) is handled too. Good resolution. 👍

Ahmath-Gadji (approved):
Approving — all substantive findings are resolved and I validated the branch locally at `9b7bca0e`.

**Findings closed**
- ✅ **Endpoint `extra` data-loss** (`c6587e17`) — `redact_secret_mapping` now masks secrets in place instead of whitelisting, so non-secret config (`temperature`, `enable_thinking`, nested `auth`/`headers`) survives the read→edit→save round-trip. Locked in by `test_redact_secret_mapping_keeps_non_secret_endpoint_extra_shape`.
- ✅ **`/config` prefix leak** (`1637c174`) — `redact_secrets` returns full `<redacted>` for all secret fields; no more first-3-char disclosure of DB password / Fernet key / client secret.
- ✅ **URL-edit UX** (`c6587e17`, `f2ef2b72`) — changing an endpoint URL with a hidden key now gives a clear "reveal or enter the API key" prompt instead of a cryptic 400, a revealed key feeds validation, and explicit key-clearing is supported.

**Validation run locally**
- `pytest` redaction + model-endpoint-service + admin-routers → **59 passed**
- `vitest run src/lib/api/models.test.ts` → **20 passed**
- CI green (api-tests, tests, milvus-integration, lint, layer-import-guard)

**Non-blocking follow-up:** the draft-validate probe still has no private-range/`is_safe_url` guard, so an admin-supplied URL can reach internal addresses — admin-gated and mitigated by `follow_redirects=False`, so fine to land and harden later.

PR #619, opened 2026-07-02
fix(search): validate user-supplied filter expression

andyne13:
## What

Validates the user-supplied `filter` query parameter on the search routes before it is combined with the partition scope in the Milvus query.

## Why

The raw `filter` string is appended alongside the partition clause and wrapped as `(partition …) and (…)`. That wrapping only holds while the caller's expression has balanced parentheses — an unbalanced expression can rebalance the wrapper and widen the query beyond the caller's partitions. Tracked as `GHSA-3gvv-h672-gj2c`.

## How

New `validate_search_filter` (`core/utils/filter_validation.py`), called in `CommonSearchParams.__init__` so all three search routes are covered at one boundary. It is quote-aware (operators/parens inside string literals are ignored) and rejects:

- **unbalanced parentheses** — the core guarantee; keeps the caller's expression confined to the single `and`-joined operand it is wrapped in,
- references to the reserved **`partition`** field,
- bare **tautologies**,
- **over-long** input.

`or` / `not` stay allowed — once parentheses balance, the top-level `and` with the partition clause always confines results to the caller's partitions regardless of the operators inside.

Scope: the chat/retrieval path is unaffected (it builds filters from a structured, validated model, never a raw string); MCP builds its filter server-side. `_build_filter_expr` (the sink) is unchanged — validation is applied where untrusted input enters.

## Tests

- Unit tests for the validator: rejected cases (incl. the unbalanced-paren breakout, `partition` reference, tautologies, unterminated string) and allowed cases (incl. balanced `or`, quoted `)` / `partition`, `partition_group`, `IN`, `ISO`).
- End-to-end route test: a malformed filter returns **400** during dependency resolution, before retrieval runs.
- Full unit suite green (1569 passed); ruff + layer-import checks pass.

Refs `GHSA-3gvv-h672-gj2c`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Search requests now reject unsafe or malformed filter expressions with a clear 400-level error instead of passing them through.
  * Added protection against overly long filters, unbalanced parentheses, unterminated strings, tautologies, and reserved terms that could affect search scope.
  * Standardized API error responses during search requests.
* **Tests**
  * Added coverage for valid and invalid filter expressions, including cross-scope injection attempts and edge cases.

### DIFF
--- a/openrag/api/routers/user/search.py
+++ b/openrag/api/routers/user/search.py
@@ -17,6 +17,7 @@ from api.dependencies.auth import (
     require_partitions_viewer,
 )
 from api.dependencies.files import validate_file_id
+from core.utils.filter_validation import validate_search_filter
 from core.utils.logging import get_logger
 from di.providers import get_retrieval_service, get_workspace_service
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
@@ -58,6 +59,10 @@ class CommonSearchParams:
             description="""Milvus filter expression string.""",
         ),
     ):
+        # Reject filter expressions that could break out of the partition
+        # scope (unbalanced parens rebalancing the `(partition …) and (…)`
+        # wrapper) before the raw string reaches the store. Raises 400.
+        validate_search_filter(filter)
         self.text = text
         self.top_k = top_k
         self.similarity_threshold = similarity_threshold
--- a/openrag/api/routers/user/search.py
+++ b/openrag/api/routers/user/search.py
@@ -322,7 +322,10 @@ async def search_file(
         partition=partition, file_id=file_id, query_len=len(search_params.text), top_k=search_params.top_k
     )
 
-    filter = "file_id == {_file_id}" + (f" AND {search_params.filter}" if search_params.filter else "")
+    # Parenthesise the caller filter so it stays a single AND-operand and can't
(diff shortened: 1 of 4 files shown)


andyne13:
Thanks both — addressed in the latest commit:

- **Codex (file-route scope):** valid catch. The by-file route now parenthesises the caller filter — `file_id == {…} AND ({filter})` — so an `OR` predicate can't widen the `file_id` scope. Added a test capturing the filter handed to retrieval. (Note: the partition scope was already safe via the outer wrapper; this closes the intra-partition file-scope case.)

- **CodeRabbit (tautology via parens):** valid — `(1==1)` / `(true)` slipped past the literal check. Now strips wrapping parens before comparing, with tests for the wrapped forms. Kept it explicitly best-effort in the docstring, since tenant isolation is guaranteed by the balanced-paren + reserved-field rules rather than by the tautology reject.

- **CodeRabbit (assert the code, not just status):** done, with one tweak to match this repo's error contract — the body is `{"detail": "[CODE]: message"}` (no top-level `code` field), so the test now asserts `"INVALID_FILTER" in detail`.

hedhoud (approved):
Looks good to me. The filter is now validated before it reaches Milvus, and the file-search path also keeps the user filter inside the file scope. I checked the related tests and CI, and I don’t see anything blocking this PR.

PR #620, opened 2026-07-02
feat(api): pre-release security hardening bundle

andyne13:
## What

Three small, independent pre-release hardening changes on the API surface.

## Changes

**Security response headers** — new `SecurityHeadersMiddleware` sets `X-Content-Type-Options: nosniff`, `X-Frame-Options: SAMEORIGIN`, and `Referrer-Policy: strict-origin-when-cross-origin` on every response, plus `Strict-Transport-Security` **only when the request is HTTPS** (so local http development and http health probes are unaffected). `setdefault` preserves a stricter header a route sets for itself. A full CSP is intentionally omitted — the mounted admin UI / Chainlit rely on inline scripts/styles and would need per-UI tuning.

**Reject a wildcard CORS origin with credentials** — `sanitize_cors_origins` drops any `*` from the allowlist when `allow_credentials=True`, so a misconfigured `CORS_EXTRA_ORIGINS=*` can't cause Starlette to reflect the request `Origin` with credentials.

**Rate limiter parses only when enabled** — `RATE_LIMIT_*` values are now parsed only when the limiter is enabled, so a malformed value no longer crashes boot when rate limiting is turned off.

## Not included (by design)

- SSRF `is_safe_url` on the model-endpoint validate probe is handled in #615 — adding it to the config-endpoint `healthcheck.py` would break legitimate probing of internal inference endpoints.
- `PREFERRED_URL_SCHEME=https` is already the default in `infra/compose/.env.example`; adding it to the Helm values is a small follow-up (that chart is being reworked in #616).

## Tests

- `test_security_headers.py`, `test_cors_config.py`, and a rate-limit test asserting a malformed config with limiting disabled doesn't crash.
- Full unit suite green; ruff + layer-import checks pass.

_Note: the `main.py` CORS block is also touched by #616 (the `INDEXERUI` → `ADMIN_UI` rewire); the two compose — `sanitize_cors_origins` applies to whichever origin list that PR produces — so expect a trivial merge._

## Summary by CodeRabbit

* **New Features**
  * Added baseline security headers to API responses, including CORS preflight and unhandled error responses.
  * Improved CORS handling by sanitizing wildcard origins when credentialed requests are enabled.
  * Enabled the same security headers for the standalone Chainlit server.

* **Bug Fixes**
  * Rate limiting now ignores malformed limit configuration when rate limiting is disabled.
  * Default security headers no longer override stricter route-defined header values.

* **Tests**
  * Added/expanded unit tests covering CORS sanitization, security header behavior (including error paths), and disabled rate-limiting config handling.

### DIFF
--- a/openrag/api/cors_config.py
+++ b/openrag/api/cors_config.py
@@ -1,0 +1,33 @@
+"""CORS origin sanitisation.
+
+A wildcard origin combined with credentials is unsafe: Starlette's
+``CORSMiddleware`` reflects the request ``Origin`` *with* credentials in that
+case, which defeats the allowlist and lets any site make credentialed calls.
+This drops any ``"*"`` when credentials are enabled so a misconfigured
+``CORS_EXTRA_ORIGINS=*`` cannot silently open that hole.
+"""
+
+from __future__ import annotations
+
+from core.utils.logging import get_logger
+
+logger = get_logger()
+
+
+def sanitize_cors_origins(origins: list[str], *, allow_credentials: bool) -> list[str]:
+    """Return ``origins`` with any wildcard removed when credentials are on.
+
+    When ``allow_credentials`` is ``False`` the list is returned unchanged (a
+    wildcard is safe without credentials). When it is ``True`` and a ``"*"`` is
+    present, the wildcard is dropped and a warning is logged.
+    """
+    if not allow_credentials or "*" not in origins:
+        return origins
+    logger.warning(
+        "Dropping wildcard '*' from CORS origins because credentials are enabled; "
+        "list explicit origins in CORS_EXTRA_ORIGINS"
+    )
+    return [origin for origin in origins if origin != "*"]
+
+
+__all__ = ["sanitize_cors_origins"]
--- a/openrag/api/error_handlers.py
+++ b/openrag/api/error_handlers.py
@@ -25,6 +25,7 @@ from ``api/main.py``.
 
 from __future__ import annotations
 
+from api.middleware.security_headers import apply_security_headers
 from core.utils.exceptions import (
     AuthenticationError,
     AuthError,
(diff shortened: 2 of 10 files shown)


andyne13:
Thanks — all three verified as valid and addressed in the latest commit:

- **Preflights bypass the middleware:** `SecurityHeadersMiddleware` is now registered after CORS (outermost), so it wraps CORS preflight responses too.
- **Standalone Chainlit app:** added the same middleware to `chainlit_api.py`, so the Ray-Serve Chainlit origin gets the baseline headers (in embedded/compose mode Chainlit is mounted on the main app and was already covered).
- **Unhandled 500s:** Starlette generates those in the outer `ServerErrorMiddleware`, outside the user stack, so the 500 handler now sets the headers itself. Handled errors (400/403/OpenRAGError) already got them via `ExceptionMiddleware` inside the stack.

The header logic is extracted into a shared `apply_security_headers()` used by both the middleware and the 500 handler, and there's a new test asserting a 500 response carries the headers.

hedhoud (changes requested):
I found one small but real gap before I can approve.

Normal responses from the standalone Chainlit app get the new headers, but unhandled 500 responses can still bypass them. The main API has the extra 500-handler path that calls `apply_security_headers()`, but `chainlit_api.py` only adds the middleware.

Simple example: if `/chainlit/` raises unexpectedly in Ray Serve mode, that standalone origin can still return a 500 without `X-Content-Type-Options`, `X-Frame-Options`, or `Referrer-Policy`.

Could we add the same 500-header path for the standalone Chainlit app, or a small Chainlit-specific exception handler that calls `apply_security_headers()`?

andyne13:
Good catch — verified and fixed in 14d1b67f.

You're right: the standalone Chainlit app had the middleware but no 500-handler path, so its unhandled-500s (generated by Starlette's outer `ServerErrorMiddleware`, outside the user stack) came back without the headers — same class as the main-app 500 case, which I'd only covered on `api.main`.

Fix: `register_error_handlers(app)` on the Chainlit app, so its 500s go through the same handler that calls `apply_security_headers()`. As a bonus the Chainlit app (which had no error handling before) now gets consistent `OpenRAGError`/500 shaping too. ruff + layer-import + the chainlit/security-headers tests pass.

hedhoud (approved):
Rechecked the latest update. The previous gaps around CORS preflights, standalone Chainlit, and unhandled 500 responses are covered now, and CI plus the targeted local checks are green. Looks good to me.

PR #621, opened 2026-07-03
fix(indexing): keep topic tagging failures optional

hedhoud:
## Summary

This PR keeps topic tagging as an optional enrichment step during indexing.

We reproduced a case where a PDF was parsed, chunked, embedded, and ready to be stored, but the whole indexing task failed only because the topic-tagging LLM request was rejected by the model backend.

Topic tags are useful metadata, but they should not decide whether a document can be indexed. If the LLM used for topic extraction times out, trips the circuit breaker, or rejects a request, OpenRAG should keep indexing the document and simply store no topic tags for that file.

## Example

A deployment can use Mistral for topic tagging. If that backend rejects a request because of a model-specific chat-template option, the current behavior can fail the entire file upload.

The intended behavior is:

- PDF parsing still runs.
- Chunking still runs.
- Embedding still runs.
- Milvus insertion still runs.
- Topic tags are skipped for that document.
- The task can still complete.

This matches how optional enrichment should behave: missing tags are better than a failed document.

## Why

This avoids user-visible indexing failures caused by non-critical LLM enrichment errors. It is especially important for production-like setups where external model endpoints can reject individual requests or temporarily fail even though the core indexing pipeline is healthy.

The embedder remains strict. This PR does not make embedding optional, because embedding defines the vector space and a wrong or missing embedding would break retrieval quality.

## Validation

Ran the focused unit and pipeline checks:

- `uv run pytest tests/unit/core/indexing/test_topic_tags.py tests/unit/services/workers/test_pipeline_builder.py -q`
- `uv run ruff format --check openrag/core/indexing/topic_tags.py tests/unit/core/indexing/test_topic_tags.py`
- `uv run ruff check openrag/core/indexing/topic_tags.py tests/unit/core/indexing/test_topic_tags.py`

I also tested the behavior manually on the local scratch stack: after the change, a document that previously failed because topic tagging raised an inference error completed indexing, with topic tags skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved topic tagging resilience by handling additional inference and circuit breaker failures.
  * Topic tagging now returns an empty result on inference-related errors and circuit breaker open events instead of interrupting processing.
  * Adjusted inference request behavior so “thinking” and chat-template customization are only applied when supported by the selected model.
* **Tests**
  * Added async coverage to confirm topic tagging returns `[]` on inference errors.
  * Added a unit test ensuring chat-template kwargs are omitted for Mistral-specific model names.

### DIFF
--- a/openrag/core/indexing/topic_tags.py
+++ b/openrag/core/indexing/topic_tags.py
@@ -10,6 +10,7 @@ from collections.abc import Sequence
 
 from core.llm import LLM
 from core.models.chunk import Chunk
+from core.utils.exceptions import CircuitBreakerOpenError, InferenceError
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +61,15 @@ class TopicTagger:
                 else await operation
             )
             return _parse_topic_tags(_chat_response_text(response), max_tags=max_tags)
-        except (TimeoutError, OSError, RuntimeError, ValueError, TypeError) as exc:
+        except (
+            TimeoutError,
+            OSError,
+            RuntimeError,
+            ValueError,
+            TypeError,
+            InferenceError,
+            CircuitBreakerOpenError,
+        ) as exc:
             logger.warning("Error extracting topic tags for %s: %s", filename, exc)
             return []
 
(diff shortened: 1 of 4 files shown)


Ahmath-Gadji (approved):
LGTM

Ahmath-Gadji (approved):
LGTM

PR #624, opened 2026-07-06
docs(env): slim .env.example to essentials and fill env-var doc gaps

Ahmath-Gadji:
## Summary

Draft — more commits (API endpoint docs) to follow.

Simplifies `infra/compose/.env.example` down to only the variables needed to boot the default compose stack, and closes the gaps where variables shipped in `.env.example` but weren't documented anywhere.

## Changes

**`infra/compose/.env.example` (228 → ~65 lines)**
- Keeps only must-set-to-boot vars: LLM, VLM, embedder, required secrets (MinIO / Postgres), API + Chainlit auth, admin/indexer UI port, and the Ray operational block.
- Header points to the [env vars reference](https://linagora.github.io/openrag/documentation/env_vars/) for every other knob.
- The 4 `RAY_*` lines are kept deliberately: they're supplied *only* via `.env` (not baked into images/compose/entrypoint), and `RAY_ENABLE_UV_RUN_RUNTIME_ENV=0` is flagged critical for recent `uv`.

**`docs/assets/env_example.env` (quickstart-displayed snippet)**
- Synced to match — it previously omitted the required MinIO/Postgres secrets, so anyone copying from the quickstart display would have missed them.

**`docs/.../env_vars.md`**
- Documents vars that existed in `.env.example` but nowhere in the docs: MinIO credentials, rate limiting (`RATE_LIMIT_*`), the MCP server (`OPENRAG_MCP_*`), `MAX_PARTITIONS_PER_USER`, `MAX_UPLOAD_SIZE_MB`, `UVICORN_FORWARDED_ALLOW_IPS`, `APP_UID`.
- Adds a same-origin reverse-proxy mermaid diagram to the Admin UI section, explaining why `VITE_API_BASE_URL` is empty by default.
- All defaults verified against source (`mcp.py`, `rate_limit.py`, `partition_limits.py`, `files.py`, `main.py`, compose).

## Still TODO (later commits)
- New API endpoints missing from the API docs.

## Notes
- `GRAFANA_ADMIN_PASSWORD` is `${...:?}`-required by the monitoring profile but is neither in `.env.example` nor documented — flagged, out of scope for this pass.

## Summary by CodeRabbit

* **New Features**
  * Expanded API documentation and endpoint coverage, including extracts, pipeline presets, model endpoint management, and more detailed partition/file controls.
* **Documentation**
  * Updated deployment/quickstart to align with the current `infra/compose` workflow and streamlined compose commands.
  * Refreshed environment templates and config docs, including `PDFLOADER` (PyMuPDF default with scanned/OCR guidance via MarkerLoader), rate limiting, storage, access control, and MCP settings.
  * Removed outdated quick-start, MacOS, usage, and deprecated compose assets.
* **Bug Fixes**
  * Corrected shared environment file mounting behavior for compose-based setups.
* **Tests**
  * Updated unit tests to reflect the revised compose/admin UI configuration.


Ahmath-Gadji:
@codex review

docs/content/docs/documentation/API.mdx
```diff
@@ -266,26 +266,136 @@ Search within a particular file in a partition.
...
+{
+  "page_content": "The text content of the chunk…",
+  "metadata": {
+    "file_id": "doc-a-id",
+    "filename": "Document A.pdf",
+    "partition": "my_partition",
+    "page": 3,
+    "indexed_at": "2026-01-01T12:00:00Z"
+  }
+}
```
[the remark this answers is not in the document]
Ahmath-Gadji: Verified against the serializers: `GET /extract/{extract_id}` returns the chunk text as `page_content`, while `GET /partition/{partition}/chunks` returns it as `content` — they're genuinely different fields, so the docs were accurate. Rather than hide that, I documented the difference explicitly in 95476183. Harmonizing the two field names would be a breaking API change, out of scope for this docs-only PR.

docs/content/docs/documentation/API.mdx
```diff
@@ -266,26 +266,136 @@ Search within a particular file in a partition.
...
+
+Returns `{ "chunks": [ { "content", "metadata", "link", "embedding"? } ] }`.
+
+:::caution
+Without `file_id` or `limit` this can return a large amount of data for partitions with many documents.
+:::
+
+#### Partition Pipeline Config
+
+Each partition references an **indexation** and a **retrieval** [preset](#-pipeline-presets), plus an embedder and chat LLM.
```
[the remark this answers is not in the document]
Ahmath-Gadji: Verified false positive: `github-slugger` (what Astro/Starlight's slug plugin uses) slugifies `### 🧩 Pipeline Presets` → `-pipeline-presets`, so `#-pipeline-presets` resolves correctly. `npm run build` reports no broken-anchor warnings. No change needed.

infra/compose/.env.example
```diff
@@ -1,227 +1,105 @@
...
-# POSTGRES_USER=root
 
-# Prompts (templates ship inside the package at openrag/prompts/templates;
-# set PROMPTS_DIR only to override with a custom template directory)
-# PROMPTS_DIR=/path/to/custom/templates
 
-# Ray
-RAY_DEDUP_LOGS=0 # turns off ray log deduplication that appear across multiple processes
-RAY_ENABLE_RECORD_ACTOR_TASK_LOGGING=1 # # to enable logs at task level in ray dashboard
+# ── Ray (kept as-is by the compose stack; see the docs for what each does) ───
```
hedhoud, line 96: I think this minimal env still needs the shared env path. The compose file still mounts `/$SHARED_ENV:/ray_mount/.env`; with this template copied as `.env`, Compose warns that `SHARED_ENV` is empty and renders the mount as `/:/ray_mount/.env`. That is a risky default because the container gets the host root where it expects the env file. Can we add `SHARED_ENV=${PWD}/.env` back, or make the compose mount use a safer fallback?

docs/content/docs/getting_started/quickstart.mdx
```diff
@@ -24,124 +24,116 @@ cd openrag/
...
+</Tabs>
+
+:::note[Development builds]
+Add `--build` to rebuild the images from your working tree: `docker compose up --build -d`.
+:::
 
-## Ansible
+##### Inference services: external endpoints, or run them locally
+
+The stack does **not** start an embedder or reranker of its own — it expects them to be **external endpoints** you configure in `.env`, so you can reuse existing inference servers and keep the footprint small.
```
hedhoud: This section does not match what Compose does today. With the new env file, `docker compose config --services` still includes `vllm-gpu` and `reranker` by default, and the CPU profile includes the local inference services too. So users may think they are using only external endpoints while local inference containers still start. Can we align the docs with the real compose behavior, or change the compose profiles so this external-only path is true?

docs/content/docs/documentation/env_vars.md
```diff
@@ -22,10 +22,10 @@ Openrag loads all files into a pivot markdown file format before proceeding to c
 | `IMAGE_CAPTIONING_URL` | `bool` | `true` | If `true`, HTTP/HTTPS image URLs in markdown files are fetched and described by the VLM. |
 | `SAVE_MARKDOWN` | `bool` | `false` | If `true`, the pivot-format markdown produced during parsing is saved. Useful for debugging and verifying the correctness of the generated markdown. |
 |`SAVE_UPLOADED_FILES`|`bool`|`false`| When `true`, uploaded files are stored on disk. You must enable this option if you want Chainlit to show sources while chatting.|
-| `PDFLoader` | `str` | `MarkerLoader` | Specifies the PDF parsing engine to use. Available options: `PyMuPDFLoader`, `MarkerLoader` and `DotsOCRLoader`.|
+| `PDFLOADER` | `str` | `PyMuPDFLoader` | PDF parsing engine. `PyMuPDFLoader` (default) is a lightweight, fast, CPU-friendly backend for searchable PDFs. Switch to `MarkerLoader` for OCR / scanned documents, complex layouts and embedded images (heavier; GPU-friendly). Other options: `DoclingLoader`, `DotsOCRLoader`.|
```
hedhoud, line 25: This default looks right, but the Marker section just below still says `MarkerLoader` is the default. Small consistency fix: can we reword that sentence so it only describes Marker settings when `PDFLOADER=MarkerLoader` is selected?

infra/compose/.env.example
```diff
@@ -1,227 +1,105 @@
...
+# RERANKER_BASE_URL=http://reranker:7997
+# RERANKER_API_KEY=EMPTY
+
+
+# ── PDF parser (default: PyMuPDF — lightweight, CPU-friendly) ────────────────
+# Switch to MarkerLoader for OCR / scanned PDFs, complex layouts & embedded
+# images (heavier — more RAM/GPU). Other option: DoclingLoader.
+PDFLOADER=PyMuPDFLoader
+# Marker tuning (only when PDFLOADER=MarkerLoader):
+# MARKER_POOL_SIZE=1         # marker worker actors (≈ 1 per cluster node / Machine).
```
hedhoud: Small cleanup: this line has trailing whitespace, and `git diff --check` reports the same issue in the synced docs env snippet too.

hedhoud (changes requested):
Thanks for the cleanup here. I found two setup mismatches that I think should be fixed before approval. The other notes are small docs cleanup.

infra/compose/.env.example
Ahmath-Gadji: Fixed in f2d3488a — gave the mount a safe fallback `${SHARED_ENV:-./.env}:/ray_mount/.env` (matching the `env_file` directive right below it), so an unset SHARED_ENV no longer mounts host `/`. For the record: SHARED_ENV was never in `.env.example` — it only lived in the now-deleted `.env.ollama` — so the fallback is more robust than requiring every user to set the var.

docs/content/docs/getting_started/quickstart.mdx
Ahmath-Gadji: Fixed in 753e3e3a — the quickstart now states the bundled embedder (`vllm`) and reranker **do** start by default and documents how to switch to external endpoints, matching the real compose behavior.

docs/content/docs/documentation/env_vars.md
Ahmath-Gadji: Fixed in 753e3e3a — the Marker section no longer calls MarkerLoader the default; it now reads "these settings apply when `PDFLOADER=MarkerLoader` (the default is `PyMuPDFLoader`)".

infra/compose/.env.example
Ahmath-Gadji: Fixed in 753e3e3a — dropped the trailing whitespace in `.env.example` and its docs mirror; `git diff --check` is now clean.

hedhoud (approved):
Rechecked the latest update. The setup comments are fixed now: the env mount has a safe fallback, the quickstart matches the compose behavior, the PDF loader wording is consistent, and the whitespace cleanup is done. I also verified the compose config, docs build, and infra tests. LGTM.

PR #625, opened 2026-07-06
fix(inference): TEI reranker + streaming fixes, hardened for release

andyne13:
Consolidates the inference/streaming fixes for the OpenShift preprod release into `refactor/hexagonal`. Two production failures were seen on the deploy (external inference endpoints):

- Chat completions returned **`list index out of range`** during streaming.
- Reranking returned **HTTP 422 → 503** to the client.

## Root-cause fixes (authored by @ThibautChoppy, cherry-picked)
- **`fix: source filtering stream error`** — the OpenAI/litellm final *usage* chunk (sent with `stream_options.include_usage`) has `"choices": []`; `.get("choices", [{}])` only falls back on a *missing* key, not an empty list, so indexing raised `IndexError`. Fixed to fall back on the empty list too.
- **`chore: add TEI class for reranker`** — the external reranker is a Hugging Face **TEI** server, whose `/rerank` contract differs from Infinity/OpenAI (`texts` not `documents`, no `model`/`top_n`, bare-array response). Sending the old payload → 422. Adds a `TEIReranker` client + `TEIRerankerConfig` (discriminated union).
- **`fix: unit test for router`** — updates the reranker-providers assertion to include `tei`.

## Release-hardening + tests (this branch)
Bringing the above to the release bar:
- **Surface reranker error bodies** on all three clients (Infinity/OpenAI/TEI). A pydantic 422 body names the exact rejected field; the status code alone is undiagnosable — this is what made the original 422 take hours to pin down.
- **`TEIReranker`: catch `KeyError`** in the score sort, so a drifted wire format becomes a mapped `InferenceConnectionError` instead of an uncaught `KeyError`.
- **Add the missing `TEIReranker` unit tests** (bare-array parse + sort, the `texts` request contract, `top_k` truncation, missing-score error path, 422-body) — TEI shipped with zero coverage. Plus 422-body tests for Infinity/OpenAI and a `tei` registry assertion.
- **Chat streaming: log the exception traceback** on the generic handler, so a future unexpected streaming fault is recorded (the client only gets a generic message).

No behavioural change beyond error reporting.

## Not included
- Thibaut's 4th fork commit (`image repo… dockerhub push`) is intentionally **excluded**: it removes the linagora Docker Hub publish from the RC pipeline, which is a CI/release decision — not part of this app fix.

## Validation
- Full unit suite: **1626 passed**; ruff check + format clean; layer-import guard OK.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new TEI-based reranker provider and expanded the available reranker provider choices (including the admin preset default list).
* **Bug Fixes**
  * Improved robustness of streaming error handling while preserving the same client-facing error message.
  * Prevented failures when upstream streaming chunks contain an empty `choices` list.
  * Standardized reranker HTTP error handling to keep user-facing messages clean while capturing more detail in logs.
* **Documentation**
  * Updated reranker environment variable docs to include the TEI provider and its request/response and auth semantics.

### DIFF
--- a/docs/content/docs/documentation/env_vars.md
+++ b/docs/content/docs/documentation/env_vars.md
@@ -274,16 +274,16 @@ The retriever fetches relevant documents from the vector database based on query
 
 ### Reranker Configuration
 
-The reranker enhances search quality by re-scoring and reordering retrieved documents according to their relevance to the user's query. Two providers are supported: **Infinity** (default) and **OpenAI-compatible** endpoints.
+The reranker enhances search quality by re-scoring and reordering retrieved documents according to their relevance to the user's query. Three providers are supported: **Infinity** (default), **OpenAI-compatible** endpoints, and **Hugging Face Text Embeddings Inference (TEI)**.
 
 | Variable | Type | Default | Description |
 |----------|------|---------|-------------|
 | `RERANKER_ENABLED` | `bool` | true | Enable or disable the reranking mechanism |
-| `RERANKER_PROVIDER` | `str` | `infinity` | Reranker backend to use. Accepted values: `infinity`, `openai` |
-| `RERANKER_MODEL` | `str` | Alibaba-NLP/gte-multilingual-reranker-base | Model used for reranking documents.|
+| `RERANKER_PROVIDER` | `str` | `infinity` | Reranker backend to use. Accepted values: `infinity`, `openai`, `tei` |
+| `RERANKER_MODEL` | `str` | Alibaba-NLP/gte-multilingual-reranker-base | Model used for reranking documents. Ignored by the `tei` provider (a TEI instance serves a single fixed model) |
 | `RERANKER_TOP_K` | `int` | 10 | Number of top documents to return after reranking. Increase for better results if your LLM has a wider context window |
 | `RERANKER_BASE_URL` | `str` | `http://reranker:7997` | Base URL of the reranker service |
-| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service. Required when using the `openai` provider |
+| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service, sent as a `Bearer` token when set. Required when using the `openai` provider |
 | `RERANKER_SEMAPHORE` | `int` | 5 | Maximum number of concurrent reranking requests. Adjust based on your server capacity |
 
 #### Reranker Providers
@@ -291,7 +291,8 @@ The reranker enhances search quality by re-scoring and reordering retrieved docu
 | Provider | `RERANKER_PROVIDER` value | Description |
 |----------|--------------------------|-------------|
 | **Infinity** | `infinity` | Uses the [Infinity server](https://github.com/michaelfeil/infinity) via its native client. Default port: `7997` |
-| **OpenAI-compatible** | `openai` | Uses any OpenAI-compatible reranker endpoint (e.g. vLLM, LiteLLM, TEI). Default port: `8000` |
+| **OpenAI-compatible** | `openai` | Uses any reranker endpoint implementing the `{model, query, documents, top_n}` → `{results: [...]}` rerank contract (e.g. vLLM, LiteLLM). Default port: `8000` |
+| **TEI** | `tei` | Uses a [Hugging Face Text Embeddings Inference](https://github.com/huggingface/text-embeddings-inference) server via its native `/rerank` API (which is **not** OpenAI-compatible: `texts` instead of `documents`, no `model`/`top_n` fields, bare-array response). Default port: `8080`. Requests are batched to 32 texts to fit TEI's default `--max-client-batch-size` |
 
 ## Extra
 ### Prompts
--- a/docs/content/docs/documentation/env_vars.md
+++ b/docs/content/docs/documentation/env_vars.md
@@ -283,7 +283,8 @@ The reranker enhances search quality by re-scoring and reordering retrieved docu
 | `RERANKER_MODEL` | `str` | Alibaba-NLP/gte-multilingual-reranker-base | Model used for reranking documents. Ignored by the `tei` provider (a TEI instance serves a single fixed model) |
 | `RERANKER_TOP_K` | `int` | 10 | Number of top documents to return after reranking. Increase for better results if your LLM has a wider context window |
 | `RERANKER_BASE_URL` | `str` | `http://reranker:7997` | Base URL of the reranker service |
-| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service, sent as a `Bearer` token when set. Required when using the `openai` provider |
+| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service, sent as a `Bearer` token when set. Whether a key is required depends on your endpoint |
+| `RERANKER_TIMEOUT` | `float` | 60.0 | HTTP timeout in seconds for reranker requests |
 | `RERANKER_SEMAPHORE` | `int` | 5 | Maximum number of concurrent reranking requests. Adjust based on your server capacity |
 
 #### Reranker Providers
--- a/openrag/core/utils/source_filtering.py
+++ b/openrag/core/utils/source_filtering.py
@@ -127,7 +127,14 @@ async def stream_with_source_filtering(
         data = json.loads(line[len("data: ") :])
         data["model"] = model_name
 
-        choice = data.get("choices", [{}])[0]
+        # `choices` can be present but empty — e.g. the OpenAI/litellm final
+        # usage-report chunk sent when the caller requests
+        # `stream_options: {"include_usage": true}` has `"choices": []` and a
+        # top-level `"usage"` field. `.get("choices", [{}])` only falls back to
+        # the default when the key is *missing*, not when it's an empty list,
+        # so indexing straight into it raises IndexError on that chunk.
+        choices = data.get("choices") or [{}]
+        choice = choices[0]
         delta = choice.get("delta", {})
         content = delta.get("content", "") or ""
         finish_reason = choice.get("finish_reason")
--- a/openrag/services/inference/reranker_clients.py
+++ b/openrag/services/inference/reranker_clients.py
@@ -8,6 +8,9 @@ shapes (see each class' docstring).
 
 from __future__ import annotations
 
+import asyncio
+from typing import NoReturn
+
 import httpx
 from core.rerankers import Reranker, reranker_registry
 from core.utils.exceptions import InferenceConnectionError, InferenceTimeoutError
@@ -24,8 +27,8 @@ def _response_detail(response: httpx.Response, limit: int = 500) -> str:
 
     A non-2xx reranker reply (notably a 422 from a pydantic-validated server)
     carries the exact reason — which field/type it rejected — in its body. The
-    status code alone is undiagnosable, so surface a truncated body in the raised
-    error. Best-effort: never let logging/formatting mask the original failure."""
+    status code alone is undiagnosable, so the truncated body is logged.
+    Best-effort: never let logging/formatting mask the original failure."""
     try:
         body = response.text.strip().replace("\n", " ")
     except Exception:  # noqa: BLE001 — body may be unreadable; the status still helps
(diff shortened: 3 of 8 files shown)


openrag/services/inference/reranker_clients.py
```diff
@@ -74,6 +92,69 @@ async def aclose(self) -> None:
...
+            headers["Authorization"] = f"Bearer {api_key}"
+        self._client = httpx.AsyncClient(timeout=timeout, headers=headers)
+
+    @with_circuit_breaker("reranker")
+    @with_retry(max_attempts=2)
+    async def rerank(self, query: str, documents: list[str], top_k: int | None = None) -> list[tuple[int, float]]:
+        try:
+            resp = await self._client.post(
+                f"{self._endpoint}/rerank",
+                json={"query": query, "texts": documents, "raw_scores": False},
```
Ahmath-Gadji: **Blocker: a stock TEI rejects >32 texts per request.** `RetrieverPipeline._rerank_chunks` passes every retrieved chunk (`retriever.top_k` defaults to 50, `conf/config.yaml`), and this posts them all as a single `texts` array. TEI's default `--max-client-batch-size` is 32, so an out-of-the-box TEI 413s on every rerank of >32 chunks -> `InferenceConnectionError` on every chat/search query. The preprod instance presumably runs a raised limit, but anyone else flipping `RERANKER_PROVIDER=tei` hits this immediately. Could we split `documents` into <=32-text batches (offsetting the returned indices per batch), or at minimum document the required TEI flag?

Two smaller things on the same payload:
- No `"truncate": true` -- if any single text exceeds the model's max sequence length, TEI rejects the *whole* request (Infinity truncates silently). Doesn't bite with the default mGTE reranker (~8k tokens) + 512-char chunks, but does with e.g. `bge-reranker-base`. One-line hardening.
- `raw_scores: False` returns 0-1 sigmoid scores while `InfinityReranker` sends `raw_scores: True` (raw logits). Harmless today (scores are only used for ordering, and sigmoid is monotonic), but worth aligning or documenting on the `Reranker` ABC before anything starts consuming score magnitudes.

openrag/services/inference/reranker_clients.py
```diff
@@ -19,6 +19,22 @@
 logger = get_logger()
 
 
+def _response_detail(response: httpx.Response, limit: int = 500) -> str:
```
Ahmath-Gadji, line 25: This detail ends up client-visible: `InferenceConnectionError.message` is streamed verbatim to end users via `_make_sse_error(e.message, e.code)` in `chat.py`, and lands in the 503 `detail` for non-streaming requests. So up to 500 chars of the reranker backend's raw error body (stack traces, proxy error pages, internal hostnames) now ship to unprivileged API clients -- before this PR only the status code and endpoint leaked.

The diagnostic goal is right (the 422 body is what made the original bug hard to pin down), but I'd log the truncated body at the raise site and keep the exception message at `HTTP {status}`. That's exactly the pattern `ollama_client.py` already uses with its `_error_snippet()` helper (line 35) -- which this function also duplicates almost exactly (same whitespace-collapse + 500-char cut + truncation marker). Sharing that helper would solve both points.

openrag/core/utils/source_filtering.py
```diff
@@ -127,7 +127,14 @@ async def stream_with_source_filtering(
...
         data["model"] = model_name
 
-        choice = data.get("choices", [{}])[0]
+        # `choices` can be present but empty — e.g. the OpenAI/litellm final
+        # usage-report chunk sent when the caller requests
+        # `stream_options: {"include_usage": true}` has `"choices": []` and a
+        # top-level `"usage"` field. `.get("choices", [{}])` only falls back to
+        # the default when the key is *missing*, not when it's an empty list,
+        # so indexing straight into it raises IndexError on that chunk.
+        choices = data.get("choices") or [{}]
```
Ahmath-Gadji, line 136: This fixes the crash (strictly better than before), but the usage chunk now takes the final `else` branch and is emitted immediately, while the finish chunk is buffered as `chunk_template` and the tail content + finish chunk are only flushed at `[DONE]`. So with `stream_options: {"include_usage": true}` the emitted order becomes: content..., **usage chunk**, tail-content chunk, finish chunk, `[DONE]` -- OpenAI's contract puts the usage chunk last, and strict SDKs that treat `choices: []` as terminal will drop the buffered tail of the answer. The chunk also gets the nonstandard `"extra": "{}"` injected by the `else` branch, and `usage.completion_tokens` still counts the stripped `[Sources: ...]` tag.

Proper fix: detect the chunk (`choices == []` + top-level `usage`), hold it, and emit it between the finish chunk and `[DONE]`. Fine as a follow-up ticket rather than growing this PR, but worth recording.

openrag/services/inference/reranker_clients.py
```diff
@@ -114,8 +195,10 @@ async def rerank(self, query: str, documents: list[str], top_k: int | None = Non
         except httpx.TimeoutException as exc:
             raise InferenceTimeoutError(f"Reranker request timed out at {self._endpoint}") from exc
         except httpx.HTTPStatusError as exc:
+            detail = _response_detail(exc.response)
```
Ahmath-Gadji: This PR itself demonstrates the cost of the copy-paste in this module: adding `detail` meant patching the same 4-line `HTTPStatusError` handler in all three classes (plus a third copy of `__init__`/`aclose`, and a third copy-pasted `test_http_error_surfaces_response_body`). Not necessarily for this PR, but a shared `_post_rerank(client, url, payload, endpoint)` helper owning POST + raise_for_status + exception mapping would reduce each client to payload builder + response parser -- which is exactly the per-provider variation the module docstring identifies.

Ahmath-Gadji (commented):
Reviewed the full diff. The root-cause fixes are correct and the TEI test coverage is solid -- both production failures are genuinely addressed, and the streaming fix is strictly better than the pre-PR crash. Four inline comments, plus two things that live outside the diff:

- `openrag/api/routers/admin/presets.py:20` -- `_DEFAULT_RERANKER_PROVIDERS = ["infinity", "openai"]` wasn't updated. That fallback exists precisely for the "before DI imports providers" state its docstring names; in that state `GET /presets/options` omits `tei`, so the admin preset UI can't offer the provider the deployment actually runs. (The PR updates the registry-backed test assertion but misses this constant right next to it.)
- `docs/content/docs/documentation/env_vars.md` (line ~291 + the provider table below) -- `RERANKER_PROVIDER` accepted values still list only `infinity`, `openai`. `tei` should be added, ideally with its caveats (native `/rerank` contract, default port 8080, no model selector).

**Before merge I'd want:** the TEI batch limit and the error-body disclosure (both inline). The presets fallback and docs rows are one-liners. The usage-chunk ordering can be a follow-up ticket.

Verified non-issues while reviewing, for the record: `TEIReranker.__init__`'s `**_kwargs` correctly absorbs the factory's extra kwargs (`batch_size`), the `log.opt(exception=e)` change is correct Loguru usage, and the similar `choices[0]` patterns in `query_service.py` are non-streaming paths that can't receive the empty-choices usage chunk.

Ahmath-Gadji:
Pushed three commits addressing the review findings (adca5c94, 0e55f30a, 814e6643):

- **`adca5c94`** — `TEIReranker` now splits `/rerank` calls into <=32-text batches (TEI's default `--max-client-batch-size`) run concurrently, with each batch's local indices shifted back to full-list positions; adds `truncate: true` so one over-long text degrades to a truncated score instead of failing the whole request. Also moves the error-response body out of the client-visible exception message: it is now logged (`endpoint`/`status`/`body`) via a shared `_raise_reranker_http_error()` helper, and the raised `InferenceConnectionError` carries only the status code -- the message reaches API clients verbatim through the SSE error event and the 503 `detail`. The three `test_http_error_surfaces_response_body` tests now assert the body stays *out* of the message, and two new tests cover batching (50 docs -> 32+18 requests with offset-corrected indices) and cross-batch `top_k`.
- **`0e55f30a`** — adds `tei` to `_DEFAULT_RERANKER_PROVIDERS` so the registry-empty fallback of `GET /presets/options` matches the registry.
- **`814e6643`** — documents the `tei` provider in `env_vars.md`, and corrects the `openai` provider row, which listed TEI as an example of an OpenAI-compatible endpoint -- TEI's native `/rerank` contract is exactly what the `openai` client cannot talk to (the 422 this PR fixes).

Full unit suite passes (1628, was 1626 + 2 new tests); ruff check/format clean.

Not addressed here, flagged for follow-up:
1. **Usage-chunk ordering** (inline comment on `source_filtering.py`): with `stream_options.include_usage`, the `choices: []` chunk is forwarded before the synthesized tail-content/finish chunks instead of last-before-`[DONE]`. Needs a small buffering change in `stream_with_source_filtering`; worth its own ticket.
2. **`extern/reranker/tei.yaml`**: the compose include resolves `extern/reranker/${RERANKER_PROVIDER:-infinity}.yaml`, so a bundled-TEI deployment needs that file to exist (only `infinity.yaml`/`openai.yaml` today). Fine to defer since `tei` targets external endpoints.
3. **Score-scale divergence**: TEI requests `raw_scores: false` (0-1 sigmoid) while Infinity requests `raw_scores: true` (logits). Harmless today (scores are ordering-only and sigmoid is monotonic), but the `Reranker` ABC should eventually pin down a scale.
4. The remaining `__init__`/`aclose` duplication across the three clients and unifying `_response_detail` with `ollama_client._error_snippet` -- refactor material, not release-blocking.

docs/content/docs/documentation/env_vars.md
```diff
@@ -274,24 +274,25 @@ The retriever fetches relevant documents from the vector database based on query
...
 |----------|------|---------|-------------|
 | `RERANKER_ENABLED` | `bool` | true | Enable or disable the reranking mechanism |
-| `RERANKER_PROVIDER` | `str` | `infinity` | Reranker backend to use. Accepted values: `infinity`, `openai` |
-| `RERANKER_MODEL` | `str` | Alibaba-NLP/gte-multilingual-reranker-base | Model used for reranking documents.|
+| `RERANKER_PROVIDER` | `str` | `infinity` | Reranker backend to use. Accepted values: `infinity`, `openai`, `tei` |
+| `RERANKER_MODEL` | `str` | Alibaba-NLP/gte-multilingual-reranker-base | Model used for reranking documents. Ignored by the `tei` provider (a TEI instance serves a single fixed model) |
 | `RERANKER_TOP_K` | `int` | 10 | Number of top documents to return after reranking. Increase for better results if your LLM has a wider context window |
 | `RERANKER_BASE_URL` | `str` | `http://reranker:7997` | Base URL of the reranker service |
-| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service. Required when using the `openai` provider |
+| `RERANKER_API_KEY` | `str` | `EMPTY` | API key for the reranker service, sent as a `Bearer` token when set. Required when using the `openai` provider |
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 80c9de9d — reworded to “sent as a Bearer token when set; whether a key is required depends on your endpoint”, and added the missing `RERANKER_TIMEOUT` row (mapped in `loader.py`, default 60.0) flagged in the same review pass.

Ahmath-Gadji (approved):
Approving. The root-cause fixes are correct, and the review findings are addressed on the branch: TEI requests are now batched to fit a stock TEI's `--max-client-batch-size` (with `truncate: true` for parity with Infinity), reranker error bodies are logged instead of shipped to API clients, the presets fallback and env-var docs now know about `tei`, and CI is fully green (1628 unit tests, api-tests, milvus-integration, lint).

Disposition of the two remaining CodeRabbit nitpicks, for the record:

- **`semaphore` through TEI batching** — declined here: `reranker.semaphore` is loaded into config but consumed by no reranker client (Infinity and OpenAI included), unlike the LLM/VLM/audio semaphores wired in `runtime.py`. Enforcing it only in TEI would be an inconsistent special case; wiring (or removing) the setting across all three clients is follow-up material.
- **`asyncio.gather` → `TaskGroup`** — declined: `TaskGroup` wraps failures in an `ExceptionGroup`, which would break the `InferenceError` contract that `_retry._is_retryable` and the API error handlers rely on, unless unwrapped with `except*`. Marginal win at a typical fan-out of two batches.

Remaining follow-ups (non-blocking) are listed in my earlier comment: the usage-chunk ordering in `stream_with_source_filtering`, an `extern/reranker/tei.yaml` for the bundled-compose include, the Infinity/TEI `raw_scores` scale divergence, and the client `__init__`/`aclose` dedup.

PR #627, opened 2026-07-07
feat(loader)!: default PDF parser to PyMuPDFLoader and rename PDFLoader env to PDFLOADER

Ahmath-Gadji:
## What

- Switch the **default PDF backend** from `MarkerLoader` to **`PyMuPDFLoader`** — lightweight, fast, CPU-friendly, well suited to searchable PDFs and quick local testing.
- Rename the env override **`PDFLoader` → `PDFLOADER`** for casing consistency with the other loader vars (e.g. `AUDIOLOADER`).

Touches `conf/config.yaml`, the Pydantic defaults (`indexation.py`), the env-override map (`loader.py`), the Helm values, preset comments, and the API integration compose file.

## ⚠️ Breaking change

The `PDFLoader` environment variable is **no longer read**. Set **`PDFLOADER`** instead.

**Upgrade:**
- To keep the previous behaviour (OCR / scanned PDFs, complex layouts, embedded images), set `PDFLOADER=MarkerLoader`.
- `PyMuPDFLoader` (the new default) cannot process non-searchable / image-based PDFs and does not run OCR or extract embedded images.

## Notes

Docs for the rename + new default are handled separately in the docs PR (#624).

### DIFF
--- a/openrag/core/config/loader.py
+++ b/openrag/core/config/loader.py
@@ -96,7 +96,7 @@ _ENV_OVERRIDES: list[tuple[str, str, type]] = [
     ("IMAGE_CAPTIONING_URL", "loader.image_captioning_url", bool),
     ("SAVE_MARKDOWN", "loader.save_markdown", bool),
     ("SAVE_UPLOADED_FILES", "loader.save_uploaded_files", bool),
-    ("PDFLoader", "loader.file_loaders.pdf", str),
+    ("PDFLOADER", "loader.file_loaders.pdf", str),
     ("AUDIOLOADER", "loader.file_loaders.wav", str),
     ("MARKER_MAX_TASKS_PER_CHILD", "loader.marker_max_tasks_per_child", int),
     ("MARKER_POOL_SIZE", "loader.marker_pool_size", int),
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -193,10 +193,10 @@ loader:
     whisper_max_task_retry: 1
     whisper_retry_base_delay: 2.0
 
-  # Env: PDFLoader, AUDIOLOADER (per-extension overrides)
+  # Env: PDFLOADER, AUDIOLOADER (per-extension overrides)
   file_loaders:
     txt: TextLoader
-    pdf: MarkerLoader
+    pdf: PyMuPDFLoader
     eml: EmlLoader
     docx: DocxLoader
     pptx: PPTXLoader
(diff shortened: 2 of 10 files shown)


openrag/core/config/loader.py
```diff
@@ -96,7 +96,7 @@
     ("IMAGE_CAPTIONING_URL", "loader.image_captioning_url", bool),
     ("SAVE_MARKDOWN", "loader.save_markdown", bool),
     ("SAVE_UPLOADED_FILES", "loader.save_uploaded_files", bool),
-    ("PDFLoader", "loader.file_loaders.pdf", str),
+    ("PDFLOADER", "loader.file_loaders.pdf", str),
```
hedhoud, line 99: Since this is now a breaking env rename, I’d add a tiny config test around this mapping. One check for `PDFLOADER=MarkerLoader` overriding the default, and one check showing the old `PDFLoader` name is no longer honored, would make the contract clear and protect this from drifting later.

hedhoud (commented):
I found one small migration gap before approval: `infra/compose/.env.ollama` still has `PDFLoader=PyMuPDFLoader`. Since this PR intentionally stops reading `PDFLoader`, that example should move to `PDFLOADER=PyMuPDFLoader` too. Even if it currently falls back to the same default, leaving the old name there can confuse someone copying the env file and expecting the override to work.

openrag/core/config/loader.py
Ahmath-Gadji: Added in bb0f59cd — `tests/unit/core/config/test_pdf_loader_env.py` pins the contract: the `PyMuPDFLoader` default, the `PDFLOADER=MarkerLoader` override, and the old mixed-case `PDFLoader` name being ignored. The tests neutralize `PDFLOADER` with an empty string first so a local `.env` picked up by `load_dotenv()` can't leak in.

Ahmath-Gadji:
@hedhoud Good catch on `infra/compose/.env.ollama` — fixed in 6e85918f (`PDFLoader=` → `PDFLOADER=`). Also swept the repo for other lingering mixed-case `PDFLoader` references; that was the last one. The requested config tests landed in bb0f59cd.

hedhoud (approved):
Looks good to me now. The env rename is covered by tests, the compose example was updated, and CI is green. The remaining docs updates are already tracked in the separate docs PR, so I’m not blocking this one.

PR #629, opened 2026-07-07
perf(workers): lazily provision parser pools instead of pre-warming at startup

Ahmath-Gadji:
## What

Drop the eager `get_marker_pool()` and `init_audio_actor()` calls from `initialize_worker_bootstrap`. Parser pools are now created on first use instead of at startup.

## Why it's safe

Every parser pool consumer already self-provisions its Ray actor via `get_if_exists=True`:
- `MarkerLoader` → `MarkerPool`
- `DoclingLoader` → `DoclingPool`
- `WhisperActor` (OpenAI lang-detector)

The one gap was `LocalWhisperLoader`, which used a bare `ray.get_actor("WhisperPool")` that fails if nothing pre-created it. This PR makes it self-provision through `get_or_create_actor`, closing the gap — so both dropped bootstrap calls become redundant.

## Effect

Startup no longer spins up parser pools a deployment may never use — e.g. the Marker pool when the default PDF backend is PyMuPDF.

## Summary by CodeRabbit

* **New Features**
  * Parser workers now start lazily, provisioning each backend only when it’s first needed (including audio transcription).
  * Backend selection can happen per parsing preset at runtime without requiring pre-warmed pools.

* **Bug Fixes**
  * Improved worker startup reliability by ensuring parser pool restart/provisioning is correctly registered up front.
  * Updated whisper parsing to use the lazily created pool for transcription requests.

* **Tests**
  * Expanded unit coverage for lazy pool creation and factory registration behavior.

## Behavior note

For deployments whose configured backend is a pooled one (Marker / Docling / local Whisper), pool warm-up (heavy imports + Ray actor spin-up) moves from startup to the first parsed document. Boot gets faster for every deployment, but first-document latency after a cold start increases for those backends. The admin restart endpoint still knows every parser pool: bootstrap registers restart factories for all of them without creating actors, and restarting a never-created pool simply creates it.

### DIFF
--- a/openrag/services/workers/bootstrap.py
+++ b/openrag/services/workers/bootstrap.py
@@ -136,6 +136,4 @@ def initialize_worker_bootstrap(settings: "Settings") -> None:
     init_llm_semaphore()
     init_vlm_semaphore()
     init_audio_semaphore()
-    init_audio_actor()
-    get_marker_pool()
     get_task_state_manager()
--- a/openrag/services/workers/bootstrap.py
+++ b/openrag/services/workers/bootstrap.py
@@ -4,8 +4,13 @@ Imported at startup (explicitly by ``main.py``) to create the long-lived
 detached actors the request path looks up by name:
 
 * TaskStateManager  — shared task-state actor
-* MarkerPool / DoclingPool / WhisperPool / WhisperActor — GPU parsers
 * llmSemaphore / vlmSemaphore / audioSemaphore — cluster-wide rate limiters
+
+The GPU parser pools (MarkerPool / DoclingPool / WhisperPool / WhisperActor)
+are *not* created here: their loaders self-provision them on first use via
+``get_or_create_actor``. Bootstrap only registers their restart factories
+(see ``register_parser_pool_restart_factories``) so the admin restart
+endpoint can manage them from the API process.
 
 This is the **only** non-startup module outside ``services/workers/`` that
 imports Ray — the phase-9 plan permits Ray in ``services/workers/`` plus
@@ -64,32 +69,47 @@ def get_task_state_manager():
     return get_or_create_actor("TaskStateManager", TaskStateManager, lifetime="detached")
 
 
-def get_marker_pool():
-    from services.workers.parsers.docling_workers import DoclingPool
-    from services.workers.parsers.marker_workers import MarkerPool
+def register_parser_pool_restart_factories() -> None:
+    """Register restart factories for the lazily-created parser pools.
 
-    config = _require_settings()
-    pdf_loader = config.loader.file_loaders.pdf
-    match pdf_loader:
-        case "DoclingLoader":
-            return get_or_create_actor("DoclingPool", DoclingPool, lifetime="detached")
-        case "MarkerLoader":
-            return get_or_create_actor("MarkerPool", MarkerPool, lifetime="detached")
+    Parser pools are provisioned on first use, usually inside an indexer
+    worker process — but ``actor_creation_map`` is process-local, so the API
+    process would never learn how to (re)create them and
+    ``POST /cluster/{actor}/restart`` would 404 for every parser pool.
+    Register the factories for *all* pools here without creating any actor
+    (a per-preset ``parsing_strategy`` can bring up any of them at runtime,
+    not just the globally-configured backend). Restarting a pool that was
+    never created simply creates it.
 
+    Imports stay inside the factories so the API process doesn't pay for
+    the heavy parser dependencies (torch, marker, docling) at startup.
+    """
 
-def init_audio_actor():
-    from services.workers.parsers.whisper_workers import WhisperActor, WhisperPool, whisper_actor_options
+    def marker_pool():
+        from services.workers.parsers.marker_workers import MarkerPool
 
-    config = _require_settings()
-    use_whisper_lang_detector = config.loader.transcriber.use_whisper_lang_detector
-    file_loaders = config.loader.file_loaders
-    loader_values = set(file_loaders.values()) if file_loaders else set()
+        return get_or_create_actor("MarkerPool", MarkerPool, lifetime="detached")
 
-    if "LocalWhisperLoader" in loader_values:
+    def docling_pool():
+        from services.workers.parsers.docling_workers import DoclingPool
+
+        return get_or_create_actor("DoclingPool", DoclingPool, lifetime="detached")
+
+    def whisper_pool():
+        from services.workers.parsers.whisper_workers import WhisperPool
+
         return get_or_create_actor("WhisperPool", WhisperPool, lifetime="detached")
 
-    if "OpenAIAudioLoader" in loader_values and use_whisper_lang_detector:
+    def whisper_actor():
+        from services.workers.parsers.whisper_workers import WhisperActor, whisper_actor_options
+
+        config = _require_settings()
         return get_or_create_actor("WhisperActor", WhisperActor, lifetime="detached", **whisper_actor_options(config))
+
+    actor_creation_map["MarkerPool"] = marker_pool
+    actor_creation_map["DoclingPool"] = docling_pool
+    actor_creation_map["WhisperPool"] = whisper_pool
+    actor_creation_map["WhisperActor"] = whisper_actor
 
 
 def init_llm_semaphore():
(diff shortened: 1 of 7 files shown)


openrag/services/workers/bootstrap.py
```diff
@@ -136,6 +136,4 @@ def initialize_worker_bootstrap(settings: "Settings") -> None:
     init_llm_semaphore()
     init_vlm_semaphore()
     init_audio_semaphore()
-    init_audio_actor()
-    get_marker_pool()
     get_task_state_manager()
```
hedhoud, line 163: I like the lazy startup direction, but I think this drops one admin-side behavior by accident. `actor_creation_map` is process-local and the restart endpoint reads it from the API process. After removing the eager `init_audio_actor()` / `get_marker_pool()` calls, parser pools may be created later from an indexer/parser worker process, so the API process never registers restart factories for `MarkerPool`, `DoclingPool`, `WhisperPool`, or `WhisperActor`. That can make `/cluster/{actor}/restart` return `404 Unknown actor` for parser pools. Could we keep registering those restart factories during bootstrap without actually creating the actors?

openrag/services/workers/parsers/whisper_workers.py
```diff
@@ -200,7 +200,12 @@ class LocalWhisperLoader(BasePooledParser):
 
     def __init__(self):
         self.config = load_config()
-        self.whisper_actor: WhisperPool = ray.get_actor("WhisperPool", namespace="openrag")
+        # Lazily create the pool if bootstrap didn't — mirrors MarkerLoader so
+        # audio ingestion self-provisions WhisperPool instead of relying on the
+        # eager init_audio_actor() pre-warm at startup.
+        from services.workers.bootstrap import get_or_create_actor
+
+        self.whisper_actor: WhisperPool = get_or_create_actor("WhisperPool", WhisperPool, lifetime="detached")
```
hedhoud: This lazy creation matches Marker/Docling and looks like the right fix for local Whisper. Could we also add `LocalWhisperLoader` to the existing pool-loader test coverage? That would pin the contract that it uses `get_or_create_actor()` and does not rely on startup pre-warming anymore.

hedhoud (commented):
I left two notes. The lazy parser-pool direction makes sense, but there is one admin restart behavior that looks like it can regress and should be handled before approval.

openrag/services/workers/bootstrap.py
Ahmath-Gadji: Good catch — confirmed: `actor_creation_map` only gets populated in whichever process calls `get_or_create_actor`, so with lazy pools the API process never learned the parser-pool factories and `POST /cluster/{actor}/restart` 404'd for them.

Fixed in 6bbffa1e: bootstrap now calls `register_parser_pool_restart_factories()`, which registers factories for **all four** pools (MarkerPool, DoclingPool, WhisperPool, WhisperActor) into `actor_creation_map` without creating any actor. Registering all of them — rather than only the config-selected ones like the old eager code — also closes the pre-existing gap where a per-preset `parsing_strategy` could bring up a pool the restart endpoint didn't know about. Imports stay inside the factories so the API process still skips torch/marker/docling at startup, and `whisper_actor_options()` is recomputed at restart time so GPU detection stays fresh.

One behavior note: restarting a pool that was never created now just creates it, which seemed acceptable for an admin endpoint.

Contract pinned in `tests/unit/services/workers/test_bootstrap.py` (no actor creation at registration, all four factories registered, each factory creates its named detached pool).

openrag/services/workers/parsers/whisper_workers.py
Ahmath-Gadji: Done in 0abd80d4 — `LocalWhisperLoader` is now part of the parametrized contract test in `tests/unit/services/workers/parsers/test_pool_loaders.py`, which pins that it provisions `WhisperPool` via `get_or_create_actor(..., lifetime="detached")` and fails loudly if a get-only `ray.get_actor` sneaks back in.

To let the shared test assert uniformly, I renamed `LocalWhisperLoader.whisper_actor` → `.worker`, matching the MarkerLoader/DoclingLoader convention (the attribute was only used internally in `parse()`).

andyne13 (commented):
Independent verification against head `6bbffa1e` (base `refactor/hexagonal`). The lazy-provisioning direction is sound and the code checks out. Marker/Docling loaders already self-provisioned on the base branch, so removing the eager `get_marker_pool()`/`init_audio_actor()` warm-up is safe there.

@hedhoud's two notes both look fully resolved in the current head — confirming rather than re-opening them:
- **Restart 404 regression** (`bootstrap.py`): `register_parser_pool_restart_factories()` registers all four pools into the process-local `actor_creation_map` without creating actors, and `restart_ray_actor` (`di/workers.py:56`) reads that same map — so `/cluster/{actor}/restart` resolves for every parser pool. Registering all four (vs only the config-selected one) also closes the latent gap where a non-default/per-preset backend wasn't restartable. ✔
- **`LocalWhisperLoader` coverage + `.worker` rename**: now in the parametrized contract test. ✔

Three net-new points (none blocking):

1. **Also safe: the OpenAI lang-detector path.** Dropping `init_audio_actor()` also removed the eager `WhisperActor` create for the `OpenAIAudioLoader` + `use_whisper_lang_detector` case. That path self-provisions too — `detect_language_via_actor` uses `WhisperActor.options(name="WhisperActor", get_if_exists=True, ...)` (`whisper_workers.py:170`) — so no regression there either.

2. **Minor cleanup (outside this diff):** two test helpers still stub the now-removed function — `tests/unit/test_auth_router.py:47` and `tests/integration/api/test_oidc_lifecycle.py:252` both set `stub.get_marker_pool = lambda: None`. Harmless (they replace the whole module, so nothing calls it), but vestigial now that `get_marker_pool` is gone. Worth deleting those two lines so the stubs don't drift from the real module.

3. **Behavioral note for release notes:** for deployments whose configured backend is a pooled one (Marker / Docling / local Whisper), the pool warm-up (torch/marker/docling import + Ray actor spin-up) shifts from startup to the **first parse**. Intended trade-off — faster boot for the common PyMuPDF case — but first-document latency after a cold start increases; might be worth a line in the changelog.

LGTM aside from the tiny stub cleanup.

Ahmath-Gadji:
Thanks for the independent verification @andyne13 — all three points taken:

1. **Lang-detector path**: agreed, `detect_language_via_actor` already self-provisions via `get_if_exists=True`, so no gap there.
2. **Stub cleanup**: done in c7154409 — dropped the vestigial `stub.get_marker_pool` lines from `tests/unit/test_auth_router.py` and `tests/integration/api/test_oidc_lifecycle.py`.
3. **Changelog**: added a **Behavior note** section to the PR description covering the warm-up shift from startup to first parse for pooled backends (Marker / Docling / local Whisper), so it lands in the release notes.

andyne13:
@hedhoud — both of your notes are addressed now, so this should be ready for another look:

- **Restart-factory 404 regression** (the one you flagged as "before approval"): fixed via `register_parser_pool_restart_factories()`, which registers all four parser pools (MarkerPool / DoclingPool / WhisperPool / WhisperActor) into `actor_creation_map` without creating any actor — so `POST /cluster/{actor}/restart` resolves for each. It also closes the pre-existing gap where a non-default/per-preset backend wasn't restartable.
- **`LocalWhisperLoader` coverage + `.worker` rename**: added to the parametrized pool-loader contract test.
- Plus the two vestigial `get_marker_pool` stubs were dropped in c7154409.

CI is green (tests, api-tests, milvus-integration, lint, layer-import-guard) and it's mergeable/clean. Since your concern was the blocker, could you re-check and approve if it looks good to you? 🙏

openrag/services/workers/bootstrap.py
Ahmath-Gadji: Good catch — addressed in 6bbffa1e. Bootstrap now calls `register_parser_pool_restart_factories()`, which registers restart factories for **all** parser pools (`MarkerPool`/`DoclingPool`/`WhisperPool`/`WhisperActor`) in the API process **without creating** any actor. So `POST /cluster/{actor}/restart` still works even though the pools are now provisioned lazily inside worker processes.

openrag/services/workers/parsers/whisper_workers.py
Ahmath-Gadji: Added in 0abd80d4 — the pool-loader contract test now covers `LocalWhisperLoader`, pinning that it self-provisions `WhisperPool` via `get_or_create_actor()` and no longer relies on startup pre-warming (the field was also renamed `whisper_actor`→`worker` to match the other pooled loaders).

PR #631, opened 2026-07-07
fix(admin-ui): polish table controls

hedhoud:
## Summary

This PR makes the Admin UI table controls easier to scan and use without changing the overall layout or visual language.

It keeps table-related actions close to the tables they affect: refresh controls sit with the table toolbar, Jobs separates search from status filters, and bulk deletion now appears only when there is an active selection.

## Why

The previous bulk-selection banner took attention away from the table and made simple actions feel heavier than needed. The Jobs toolbar was also crowded because search and status filters were competing on the same row.

This keeps the interface closer to the existing OpenRAG design while making common table workflows clearer.

## Validation

- `npm test -- --run src/components/shared/data-table.test.tsx src/pages/admin/documents/list.test.tsx src/pages/admin/partitions/list.test.tsx src/lib/api/models.test.ts src/pages/admin/preset-config.test.ts src/pages/admin/overview.test.tsx`
- `npm run lint`
- `npm run build`
- Local Admin UI container rebuild
- Playwright smoke check on the Jobs page toolbar layout

## Summary by CodeRabbit

* **New Features**
  * Added controlled multi-row selection and bulk delete for partitions and documents, with page/visibility-aware “select all” checkboxes and confirm flow in the page header.
  * Added debounced client-side search and live filtered counts to the jobs admin list.
  * Added manual refresh controls with loading/spinner feedback on documents, partitions, and jobs.
* **Bug Fixes**
  * Selection is now synchronized per active partition/tab and restricted to deletable/configurable items.
  * Selection UI (counts/indeterminate “select all”) and bulk-selection clearing after deletion are now consistent.
* **Tests**
  * Expanded selection and bulk-action test coverage for tables and admin list pages.

### DIFF
--- a/ui/src/pages/admin/partitions/list.tsx
+++ b/ui/src/pages/admin/partitions/list.tsx
@@ -1,10 +1,11 @@
 import { useMemo, useState, useCallback, useEffect } from "react";
 import { Link, useSearchParams } from "react-router-dom";
 import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
-import { Plus, Pencil, Trash2, Search, ArrowUpDown, ArrowUp, ArrowDown, CheckCircle, XCircle, Loader2 } from "lucide-react";
+import { Plus, Pencil, Trash2, Search, ArrowUpDown, ArrowUp, ArrowDown, CheckCircle, XCircle, Loader2, RefreshCw } from "lucide-react";
 import { toast } from "sonner";
 import { Button } from "@/components/ui/button";
 import { Badge } from "@/components/ui/badge";
+import { Checkbox } from "@/components/ui/checkbox";
 import { Input } from "@/components/ui/input";
 import { Textarea } from "@/components/ui/textarea";
 import { Label } from "@/components/ui/label";
@@ -146,6 +147,7 @@ export default function PartitionListPage() {
   const [search, setSearch] = useState("");
   const [sortDir, setSortDir] = useState<SortDir>(null);
   const [sortColumn, setSortColumn] = useState<"name" | "created_at">("created_at");
+  const [selectedPartitionNames, setSelectedPartitionNames] = useState<string[]>([]);
 
   // `GET /partition/` is already membership-scoped server-side (admins with
   // SUPER_ADMIN_MODE see all; regular users see their memberships), so a single
@@ -153,6 +155,7 @@ export default function PartitionListPage() {
   const partitionsQuery = useQuery({
     queryKey: ["partitions"],
     queryFn: listPartitions,
+    refetchInterval: 2000,
   });
 
   const { data: presetsData } = useQuery({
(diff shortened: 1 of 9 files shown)


hedhoud:
@codex review

Ahmath-Gadji (commented):
Reviewed and verified against a live deployment — no blocking bugs found. One consistency point kept inline (the new 2s `refetchInterval` on the partitions list); measured the endpoint to confirm it's about consistency/log noise, not backend load.

ui/src/pages/admin/partitions/list.tsx
```diff
@@ -146,13 +148,16 @@ export default function PartitionListPage() {
...
+  const [selectedPartitionNames, setSelectedPartitionNames] = useState<string[]>([]);
+  const [pageIndex, setPageIndex] = useState(0);
 
   // `GET /partition/` is already membership-scoped server-side (admins with
   // SUPER_ADMIN_MODE see all; regular users see their memberships), so a single
   // query serves both — canManagePartitions only gates the admin-only columns/fields below.
   const partitionsQuery = useQuery({
     queryKey: ["partitions"],
     queryFn: listPartitions,
+    refetchInterval: 2000,
```
Ahmath-Gadji: This new 2s poll on `GET /partition/` isn't mentioned in the PR description and is 2.5× more aggressive than the 5s used by the Jobs and Documents pages — and this toolbar now has a manual refresh button anyway.

To be fair I checked this against a live deployment: the endpoint is deliberately lightweight (two Postgres queries, no Milvus/Ray) and answers in 5–9 ms, so this is not a backend-load problem. What remains is consistency and noise — the poll runs whenever the page is focused, and with debug logging every poll writes a log line (≈ 43k lines/day per open tab at 2s).

```suggestion
    refetchInterval: 5000,
```

(Or drop the interval entirely and rely on the refresh button + query invalidation.)

hedhoud: Fixed in 89e6e94f: partitions polling is back to 5s so it matches the rest of the list pages and avoids the extra request pressure.

hedhoud:
Follow-up on the review comments after 89e6e94f:

- Fixed the refresh flicker on Partitions and Documents by using local manual-refresh state instead of background polling state.
- Set Partitions polling back to 5s.
- Removed the zero-delay timeout around the non-writable document selection reset.
- Added accessible names to the new pagination icon buttons and updated the pagination test to use the button name.
- Kept the partitions checkbox column stable across pages by gating it on deletable rows in the filtered result set, not only the visible page.

Two points were intentional tradeoffs for this PR: partition selection is pruned by the search-filtered list so bulk delete only applies to what the user can see, and the Jobs table remount still resets sort along with pagination for a new tab/search context. I added a short code comment for the Jobs behavior.

ui/src/pages/admin/partitions/list.tsx
```diff
@@ -294,6 +349,23 @@ export default function PartitionListPage() {
     },
   });
 
+  const bulkDeleteMutation = useMutation({
+    mutationFn: async (names: string[]) => {
+      const results = await Promise.allSettled(names.map((partitionName) => deletePartition(partitionName)));
```
Ahmath-Gadji, line 354: **Nit (scalability): bulk delete fans out every request at once**

`Promise.allSettled(names.map(...))` fires all deletes concurrently. Select-all is page-scoped (~10 rows), but the selection accumulates across pages, so a large cleanup can launch dozens of simultaneous `DELETE /partition/{name}` calls — and partition deletion is heavy server-side (cascade delete + Milvus cleanup). Browser per-host connection limits soften this on HTTP/1.1, much less under HTTP/2. Same pattern on the documents page (`deleteFile` fan-out at `ui/src/pages/admin/documents/list.tsx:163`).

Not blocking — behavior is correct and the realistic worst case is an admin mass-cleanup. If it ever becomes a problem, a small concurrency cap (chunks of 3–5) or a sequential loop would bound the pressure without changing the UX.

Ahmath-Gadji (approved):
Re-reviewed at 89e6e94f. All previously raised issues are addressed: bulk-delete controls gated on write access with selection cleared on role loss, selection reset moved to the mutation settling, polling back to 5s across list pages, search debounced and cleared on tab change, and the DataTable selection props made all-or-nothing.

Validated locally on the head commit: 98/98 vitest, eslint and `tsc -b && vite build` clean. CI is fully green.

One non-blocking nit left inline (concurrent fan-out on bulk delete) — fine to address in a follow-up or ignore until it's a measured problem.

PR #633, opened 2026-07-07
chore: update template & general improvement

ThibautChoppy:
## Summary by CodeRabbit

* **New Features**
  * Added configurable secrets provisioning (values, external secrets, or Vault-backed) with safer wiring and provider-based rendering.
  * Added templated extra objects support.
  * Added improved Helm install notes and richer guidance on modelSpec updates.

* **Bug Fixes**
  * Standardized naming/labels across the stack and switched env/config/secret/PVC references to release-scoped names.
  * Updated health probes, security context merging, and conditional ingress exposure (app/admin UI/Ray).

* **Documentation**
  * Updated Kubernetes deployment guide for the new chart path and required chart version.

* **Tests**
  * Updated unit tests to match the new ingress rendering and secret reference behavior.


infra/charts/openrag-stack/templates/secrets-env.yaml
```diff
@@ -1,26 +1,78 @@
...
+{{- if eq $type "values" }}
+---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: rag-env-secrets
+  name: {{ $secretName }}
+  labels: {{- include "openrag-stack.labels" . | nindent 4 }}
 type: Opaque
 stringData:
```
Ahmath-Gadji, line 13: **Security regression: the fail-closed validation was deleted.** The previous version of this template required non-empty `AUTH_TOKEN` and `POSTGRES_PASSWORD` and rejected known placeholder values — including the literal string `CHANGE_ME_STRONG_PASSWORD`, which `values.yaml` now ships as the default Postgres password. With this PR, a default install renders:

```yaml
stringData:
  AUTH_TOKEN: ""
  POSTGRES_PASSWORD: "CHANGE_ME_STRONG_PASSWORD"
```

An empty `AUTH_TOKEN` means the app treats every request as the admin user, and `openrag.ingress.enabled` now defaults to `true` (it was `false`) — so out of the box this is a publicly routed, unauthenticated API with a published DB password. Please restore the required-secret + placeholder checks (they can live alongside the new `secretsProvider` switch: only the `values` provider needs them).

Also note this loop dropped the `tpl` rendering the old template applied to each secret value, so operators who templated values under `env.secrets` lose that ability.

infra/charts/openrag-stack/values.yaml
```diff
@@ -21,27 +41,63 @@ persistence:
...
+  # Kept in sync with the root fullnameOverride ("openrag") by hand — see the
+  # note above it. Without this, the postgresql sub-chart names its own
+  # Service from the real Release.Name instead, which won't match POSTGRES_HOST below.
+  fullnameOverride: "openrag-postgresql"
   auth:
     username: &pgUser root
-    # Must be supplied at install time; templates fail closed if it is empty.
-    password: &pgPass ""
+    # Override at install, e.g. --set postgresql.auth.password=$(openssl rand -hex 16)
+    password: &pgPass "CHANGE_ME_STRONG_PASSWORD"
```
Ahmath-Gadji: **The override suggested by this comment doesn't reach the app.** `env.secrets.POSTGRES_PASSWORD` (line 640) references this value via the YAML anchor `*pgPass`, but anchors are resolved when `values.yaml` is parsed — **before** Helm merges `--set`/`-f` overrides. Verified with:

```console
$ helm template . --set postgresql.auth.password=SuperSecret123 | grep POSTGRES_PASSWORD
  POSTGRES_PASSWORD: "CHANGE_ME_STRONG_PASSWORD"
```

The Postgres subchart gets `SuperSecret123` while the app keeps the placeholder → the app can't connect to the DB. The previous chart used `POSTGRES_PASSWORD: "{{ .Values.postgresql.auth.password }}"` + `tpl` in the secret template precisely so overrides propagate — please restore that indirection (the anchor only works for people editing values.yaml in place).

infra/charts/openrag-stack/values.yaml
```diff
@@ -1,18 +1,38 @@
...
-# outside the namespace. Disable only if you manage isolation elsewhere.
+# Default-deny ingress; allow only same-namespace traffic and the public HTTP
+# ports below. Isolates the Ray dashboard, Postgres, Milvus, etc.
 networkPolicy:
   enabled: true
   # Ports reachable from outside the namespace (e.g. via the Ingress controller).
   externalPorts:
-    - 8080  # openrag API/UI
+    - 8080  # openrag API/UI & admin-ui
+    - 8265  # ray dashboard
```
Ahmath-Gadji: **Security regression: this opens the unauthenticated Ray dashboard to the outside by default.** Combined with `ray.dashboardHost` flipping to `"0.0.0.0"`, port 8265 is now reachable from outside the namespace on a default install — while `openrag.rayDashboard.ingress.enabled` and `ray.ingress.enabled` both default to `false`, so nothing legitimate needs it. The Ray dashboard allows job submission (effectively RCE), which is exactly why the previous chart's NetworkPolicy comment called out 8265 as deliberately isolated — and why this PR itself moves the metrics port off 8080. Suggest dropping `8265` from the default list and documenting that it must be added only when a dashboard ingress is enabled.

infra/charts/openrag-stack/templates/admin-ui.yaml
```diff
@@ -2,65 +2,72 @@
...
-      automountServiceAccountToken: {{ .Values.security.automountServiceAccountToken }}
+      {{- with .Values.adminUi.podSecurityContext }}
       securityContext:
-        {{- toYaml .Values.security.podSecurityContext | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.adminUi.serviceAccountName }}
+      serviceAccountName: {{ tpl . $ }}
+      {{- end }}
+      {{- with .Values.adminUi.automountServiceAccountToken }}
```
Ahmath-Gadji: **Bug: `with` on a `false` boolean never renders.** Go templates treat `false` as empty, so `{{- with .Values.adminUi.automountServiceAccountToken }}` skips the block when the value is `false` — which is exactly the default. Verified: the rendered Deployment has no `automountServiceAccountToken` field at all, so every pod mounts a service-account token again (the old chart set `false` unconditionally). Same pattern in `infinity.yaml`, `openrag.yaml` and `raycluster.yaml` (head + workers). Render it directly like `postgres-migration-job.yaml` does, or guard with `kindIs "bool"` instead of `with`.

infra/charts/openrag-stack/templates/admin-ui.yaml
```diff
@@ -2,65 +2,72 @@
...
           securityContext:
-            {{- toYaml .Values.security.containerSecurityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - containerPort: {{ .Values.adminUi.service.targetPort }}
+              name: admin-ui
           # nginx serves the SPA at /app/; probe it so a crash-looping or stuck
           # pod is restarted and gated out of rollout traffic.
+          {{- if .Values.adminUi.probes.startup }}
```
Ahmath-Gadji, line 35: **Bug: key mismatch — the admin-ui probes are silently dropped.** This template reads `.Values.adminUi.probes.startup` / `.readiness` / `.liveness`, but `values.yaml` defines `probes.startupProbe` / `readinessProbe` / `livenessProbe` (values.yaml:385-398). Verified with `helm template`: the rendered admin-ui Deployment contains **zero probes**, losing the readiness/liveness probes the chart previously hardcoded. Renaming the values keys to match the `openrag` block (`startup:` / `readiness:` / `liveness:`) is probably the cleanest fix.

infra/charts/openrag-stack/values.yaml
```diff
@@ -21,27 +41,63 @@ persistence:
...
-    tag: "1.1.13"
+    registry: ghcr.io
+    repository: linagora/openrag-ray
+    tag: "v2.0.0-rc.1"
+  resources:
+    head:
+      requests:
+        cpu: "4"
+        memory: "8Gi"
+        nvidia.com/gpu: 1
```
Ahmath-Gadji, line 108: **Invalid PodSpec: extended resources can't be requests-only.** Kubernetes requires `nvidia.com/gpu` in `requests` to be matched by an equal `limits` entry; a pod with only a request is rejected at admission. So `ray.enabled=true` with default values produces head/worker pods that fail to create. It currently only works through `values-linagora.yaml`, whose `limits` merge on top of these `requests`. Please add matching `limits` here (same for the `worker` block below).

infra/charts/openrag-stack/values.yaml
```diff
@@ -211,25 +273,23 @@ vllm:
...
-                operator: In
-                values: ["vlm"]
+        nodeSelectorTerms: []
   routerSpec:
     enableRouter: false
 
 # === Infinity reranker ===
 reranker:
+  # Set to false to skip deploying the reranker and use an external one instead.
+  # Configure externalUrl to point to the external reranker endpoint.
```
Ahmath-Gadji: **`externalUrl` doesn't exist.** This comment documents a `reranker.externalUrl` value, but no such key is defined and no template consumes it. As a result, with `reranker.enabled=false` the ConfigMap still renders `RERANKER_ENABLED: "true"` and `RERANKER_BASE_URL: http://openrag-reranker:7997` pointing at a Service that was never created (verified with `helm template --set reranker.enabled=false`). Either implement `externalUrl` (use it for `RERANKER_BASE_URL` when set) or gate `RERANKER_ENABLED`/`RERANKER_BASE_URL` on `reranker.enabled` — and fix the comment either way.

infra/charts/openrag-stack/values-linagora.yaml
```diff
@@ -0,0 +1,148 @@
...
+    annotations:
+      nginx.ingress.kubernetes.io/rewrite-target: /$1
+      nginx.ingress.kubernetes.io/use-regex: "true"
+      nginx.ingress.kubernetes.io/proxy-body-size: "500m"
+
+ray:
+  # This overlay runs the separate RayCluster (GPU node-selectors/resources
+  # below) rather than the embedded ray.init() fallback — both this flag and
+  # env.config.ENABLE_RAY_SERVE must be enabled together (see values.yaml).
+  enabled: true
```
Ahmath-Gadji, line 36: **Wiring gap: `RAY_ADDRESS` is never set.** This overlay enables the RayCluster and `ENABLE_RAY_SERVE=true`, but `RAY_ADDRESS` is commented out in values.yaml's `env.config` — so the app pod calls `ray.init()` and starts an *embedded* Ray instead of connecting to the cluster deployed here, while the Ingress routes `/` to `openrag-raycluster-head-svc:ray-serve`, which nothing is serving. The old chart always set `RAY_ADDRESS: ray://raycluster-head-svc:10001`. This overlay (or the chart, whenever `ray.enabled=true`) should set `RAY_ADDRESS: 'ray://{{ include "openrag-stack.fullname" . }}-raycluster-head-svc:10001'`.

.github/workflows/build_rc.yml
```diff
@@ -230,4 +230,4 @@ jobs:
           docker buildx imagetools create \
             -t "${GHCR}-admin-ui:${VERSION}" \
             -t "linagoraai/openrag-admin-ui:${VERSION}" \
-            "${GHCR}-admin-ui:${STAGING_TAG}"
+            "${GHCR}-admin-ui:${STAGING_TAG}"
```
Ahmath-Gadji: This change only strips the file's trailing newline (`\ No newline at end of file`) — looks accidental; please revert so the workflow file keeps a POSIX-final newline.

Ahmath-Gadji (commented):
Nice modernization overall — the NOTES.txt list-merge warning, proxy fail-fast checks, `checksum/config` annotation and the pinned infinity tag are all welcome. I rendered the chart with `helm template` and left inline comments on a handful of issues that I think should be addressed before merge (a few are confirmed rendering bugs, a few are security regressions vs the previous chart).

Two concerns that don't fit a single line:

**Upgrade path / breaking renames.** Every resource is renamed (`rag-*` → `openrag-*`, `Release.Name-*` → fullname-based), and `postgresql`/`milvus` gain `fullnameOverride`s. For any existing release **not** named `openrag`, an upgrade renames the Postgres/Milvus StatefulSets and creates fresh empty PVCs; even for a release named `openrag`, the app PVCs change name (`rag-data` → `openrag-data`), orphaning existing data (the old PVCs only survive via `helm.sh/resource-policy: keep`). The selector-label change (`app.kubernetes.io/instance`: `.Release.Name` → fullname) is part of the same story. Could we get an upgrade/migration note in NOTES.txt or `docs/content/docs/documentation/kubernetes.md` spelling this out?

**venv coordination removed.** The `init-venv-sync` / `init-venv` init containers (`uv sync` + `.ready` flag on the shared RWX venv PVC) are gone from both the openrag Deployment and the RayCluster, but the venv PVC is still mounted at `/app/.venv` everywhere. What populates the venv now, and what prevents the head, workers and app pod from racing concurrent `uv` runs on the same PVC at first boot? If the `v2.0.0-rc.1` images bake the venv into the image, the PVC mount (which shadows it) should probably go too.

andyne13 (changes requested):
Thanks @ThibautChoppy. I resolved the `Chart.yaml` conflict with `develop` (kept your `version: 0.6.0` and the added metadata; set `appVersion: "2.0.0"` for current GA), which triggered CI. Everything is green **except `tests (3.12)`**, which fails 5 Helm **security-hardening** checks — this branch predates that hardening on `develop`, so the chart currently regresses it. Please reconcile with `develop`:

1. **Placeholder DB secret** — `postgresql.auth.password` must not default to `CHANGE_ME_STRONG_PASSWORD` (empty / secret ref).
2. **Ray dashboard** — `ray.dashboardHost` must default to `127.0.0.1`, not `0.0.0.0` (ShadowRay / CVE-2023-48022).
3. **securityContexts** — restore the restricted `securityContext` blocks on the workloads.
4. **Ingress** — `templates/ingress.yaml` was removed; keep the hardened template (not exposed by default + TLS).
5. **Migration Job** — reference the `rag-env` secret instead of inlining literal secret values.

The two test files are the exact spec: `tests/unit/infra/test_helm_security_hardening.py` and `tests/unit/infra/test_postgres_migration_job_template.py`. `uv run pytest tests/unit/infra/ -q` should be green before pushing (that re-runs CI). The rest of the PR looks good — and if it's easy, splitting the app-code commits from the Helm work would review faster, though not required.

infra/charts/openrag-stack/values-linagora.yaml
Ahmath-Gadji: Re-verified at HEAD (`a44256e8`) — still unaddressed. With this overlay (`ray.enabled: true` + `env.config.ENABLE_RAY_SERVE: "true"`) and no `RAY_ADDRESS`, `openrag/api/main.py` (L156-170) takes the *embedded*-Ray branch, so the app spins up its own in-pod Ray instead of joining the RayCluster deployed here. Meanwhile `templates/openrag.yaml` routes the `/` Ingress to `…-raycluster-head-svc:ray-serve`, but the head container only runs `ray start --head` (no Serve app is deployed there), so that backend serves nothing. Net effect: the whole `ray.enabled=true` serving path is dead.

Fix — set it in this overlay's `env.config` (these values are `tpl`-rendered, see `templates/configmap-env.yaml`):

```yaml
env:
  config:
    ENABLE_RAY_SERVE: "true"
    RAY_ADDRESS: 'ray://{{ include "openrag-stack.fullname" . }}-raycluster-head-svc:10001'
```

(`10001` is the `ray-client` port declared on the head in `templates/raycluster.yaml`; `fullname` resolves to `openrag` via `fullnameOverride`.)

infra/charts/openrag-stack/values-linagora.yaml
```diff
@@ -0,0 +1,148 @@
+persistence:
+  storageClass: longhorn
```
Ahmath-Gadji, line 2: **`storageClass: longhorn` here only reaches the chart's own 4 PVCs — not the postgresql/milvus sub-charts.**

`values.yaml` wires `postgresql.primary.persistence.storageClass` and milvus's `minio`/`etcd` storageClass through the `*storageClass` YAML alias to `persistence.storageClass`. YAML aliases are resolved when `values.yaml` is parsed — *before* Helm merges this overlay — so overriding the top-level `persistence.storageClass` here does not reflow into those sub-chart paths. (Same "anchor doesn't survive the values merge" class as the `POSTGRES_PASSWORD` anchor that was fixed; this half wasn't.)

Verified with `helm template . -f values-linagora.yaml`: only `openrag-{model-weights,data,logs,venv}` render `storageClassName: longhorn`; the postgresql / milvus / etcd / minio PVCs get no storageClass at all and fall back to the cluster default.

Fix — set them explicitly in this overlay:

```yaml
postgresql:
  primary:
    persistence:
      storageClass: longhorn
milvus:
  minio:
    persistence:
      storageClass: longhorn
  etcd:
    persistence:
      storageClass: longhorn
```

ThibautChoppy:
LGTM, waiting for the next release version for image tag update in valuefile before merge

andyne13 (commented):
Thanks @ThibautChoppy — I re-checked the five points from my earlier review and they're all addressed: the DB password placeholder is gone, `ray.dashboardHost` defaults to loopback, the shared `security` block is applied through the merge helpers, the ingress is off by default behind a `required` host guard, and the migration Job now goes through `envFrom` with no inlined literals. `tests/unit/infra/` is green at this head. The `secrets-env.yaml` work is really nice — reading `POSTGRES_PASSWORD` live from `postgresql.auth.password` instead of the YAML anchor is a subtle trap well avoided, and the presence-based merge helpers neatly dodge Sprig `merge`'s zero-value behaviour.

One thing I ran into while testing the built image, and I think it just needs a one-liner in `infra/docker/ui.Dockerfile`:

The new `chown -R 10001:10001` lines up with the chart's `podSecurityContext`, but the image still ends with `USER nginx`, which is UID 101 in `nginxinc/nginx-unprivileged`. Under Helm that's all fine since the pod forces `runAsUser: 10001` — it's only docker compose, where nothing overrides the user, that ends up running as 101 against dirs owned by 10001. Building from this branch:

```
$ docker exec <admin-ui> sh -c 'id; ls -ldn /var/cache/nginx; touch /var/cache/nginx/probe'
uid=101(nginx) gid=101(nginx)
drwxrwxr-x 1 10001 10001 /var/cache/nginx
touch: /var/cache/nginx/probe: Permission denied
```

Took me a while to spot, because the container starts and serves the UI perfectly happily — it only shows up once nginx spools a request body to `client_temp`, which is what `client_max_body_size 1024M` in the admin conf is there for. Either `USER 10001:10001` in the Dockerfile or `user: "10001:10001"` on the compose service should cover it, whichever you prefer.

Non-blocking, just worth a line in the docs while you're here: the PVCs/ConfigMap/Secret get renamed (`rag-data` → `{{ fullname }}-data`, etc.). `helm.sh/resource-policy: keep` means existing volumes survive, but they're no longer mounted, so an upgraded release would come up with empty volumes unless the operator sets `fullnameOverride` or migrates. An upgrade note would save someone a confusing afternoon.

Two small things for later, no need to touch them here: `mergeSecurityContext` is a shallow merge, so a component overriding `capabilities` would drop the shared `drop: [ALL]` (nothing does today); and the three `vllm` modelSpec entries still default to `tag: "latest"` — your own note says to pin, so the defaults could follow suit.

Happy to approve once the UID bit is sorted.

infra/charts/openrag-stack/values.yaml
```diff
@@ -250,103 +359,283 @@ reranker:
...
+  # the controller must then route this more specific "/app" prefix ahead of
+  # openrag.ingress's "/" catch-all (nginx-ingress does this automatically by
+  # merging same-host Ingress resources and sorting by path specificity —
+  # verify equivalent behavior before relying on it with another controller,
+  # e.g. OpenShift HAProxy Routes or a Gateway API implementation).
+  # There is no automatic "/" -> "/app/" redirect in this mode (unlike the
+  # compose front-door) — users must browse to /app/ directly.
+  # Off by default. This flag only adds the /app/ path to openrag.ingress's
+  # single Ingress resource (see templates/openrag.yaml) — it has no effect
+  # unless openrag.ingress.enabled=true too.
```
Ahmath-Gadji: **Stale/contradictory comment: `adminUi.ingress` docs describe an architecture that no longer exists.**

Lines 393-403 say admin-ui and openrag/ray-serve are exposed via *two separate* Ingress resources, and instruct setting the same `host` on both so nginx-ingress can merge them by path specificity. That's no longer accurate: `adminUi.ingress` has no `host`/`className`/`annotations`/`tls` fields at all, and there is only one Ingress resource now (`templates/openrag.yaml`) — `adminUi.ingress.enabled` just toggles whether the `/app/` path is added to it. The correct description is already present in lines 406-408, but the outdated paragraph above it (393-403) was left in place and directly contradicts it. An operator following the top of the comment would try to set `adminUi.ingress.host`, which the template silently ignores.

Suggest deleting lines 393-403 and keeping only the accurate lines 406-408 (perhaps noting that `openrag.ingress.host`/`className`/`tls` apply to both paths).

`infra/charts/openrag-stack/values.yaml:393-408`

Ahmath-Gadji:
@andyne13 — following up on your 2026-07-16 review. All 5 issues you flagged look resolved as of the current head (15db9a13):

1. **Placeholder DB secret** — `postgresql.auth.password` now defaults to `""`, and `secrets-env.yaml` fails the install closed if `env.secrets.POSTGRES_PASSWORD` is left unset or still a known placeholder (`CHANGE_ME_STRONG_PASSWORD`, etc.).
2. **Ray dashboard** — `ray.dashboardHost` defaults to `127.0.0.1` again, with a comment explicitly calling out CVE-2023-48022 / ShadowRay.
3. **securityContexts** — restored via a shared `security.podSecurityContext` / `containerSecurityContext` block (`runAsNonRoot`, `seccompProfile: RuntimeDefault`, `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`), merged per-workload through the new `openrag-stack.mergeSecurityContext` helper (presence-based, so an explicit `false`/`0` override isn't silently dropped).
4. **Ingress** — no longer a bare `templates/ingress.yaml`; `openrag.yaml` / `raycluster.yaml` each render their own optional Ingress, disabled by default and gated by a `required` host check so it can never render with a blank/wildcard host. TLS is supported.
5. **Migration Job** — `postgres-migration-job.yaml` now references the env Secret via `configMapRef` / `secretRef` (through the `openrag-stack.secretName` / `fullname` helpers) instead of inlining literal secret values.

`tests/unit/infra/test_helm_security_hardening.py` and `test_postgres_migration_job_template.py` assert all of the above and pass in CI; I also re-ran them locally against the current head (17/17 green), plus the full unit suite and a `helm template`/`helm lint` render.

The branch is also conflict-free against `develop` again (merged in the 2.0.1 release-version bump you were waiting on).

Could you take another look and update/clear your review when you get a chance? Thanks for catching these the first time around.

infra/charts/openrag-stack/values.yaml
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/cb932807.

Rewrote the comment to match the current single-Ingress implementation: no separate admin-ui Ingress, `adminUi.ingress.enabled` just toggles the `/app/` path onto `openrag.ingress`, and `openrag.ingress.host`/`className`/`tls` apply to both paths. Verified with `helm lint`/`helm template` and `tests/unit/infra/` (17/17 pass).

Ahmath-Gadji (dismissed):
LGTM

andyne13:
@Ahmath-Gadji thanks for the follow-up — agreed, and I confirmed the same five myself in [this comment](https://github.com/linagora/openrag/pull/633#pullrequestreview-4812144552) last night (I also re-ran `tests/unit/infra/` against the head merged with `develop`: 17 passed).

That comment is easy to miss though, because it also raised a **sixth** item that came out of building the image, and it's still open at `cb93280` — that's the one thing holding my approval, not the original five.

**`infra/docker/ui.Dockerfile`**: the `chown -R 10001:10001` matches the chart's `podSecurityContext`, but the image still ends with `USER nginx`, which is UID **101** in `nginxinc/nginx-unprivileged`. Under Helm that's fine — the pod forces `runAsUser: 10001`. Under docker compose nothing overrides the user, so it runs as 101 against directories owned by 10001. Built from this branch:

```
$ docker build -f infra/docker/ui.Dockerfile -t adminui:pr633 .
$ docker run -d --name t -p 18080:8080 adminui:pr633
$ docker exec t sh -c 'id; ls -ldn /var/cache/nginx; touch /var/cache/nginx/probe'
uid=101(nginx) gid=101(nginx) groups=101(nginx)
drwxrwxr-x 1 10001 10001 /var/cache/nginx
touch: /var/cache/nginx/probe: Permission denied
```

For reference, the base image ships that directory owned by 101, so the chown is what introduces it:

```
$ docker run --rm --entrypoint sh nginxinc/nginx-unprivileged:1.27-alpine -c 'ls -ldn /var/cache/nginx'
drwxrwxr-x 1 101 0 /var/cache/nginx
```

It's easy to miss because the container still starts and serves the UI normally (I got a 301 from `/` as expected) — it only bites when nginx spools a request body to `client_temp`, which is what `client_max_body_size 1024M` in the admin conf exists for. So uploads through the proxy in the compose deployment.

Either `USER 10001:10001` in the Dockerfile or `user: "10001:10001"` on the compose `admin-ui` service should cover it, whichever fits better.

Also non-blocking, from the same pass: the PVCs/ConfigMap/Secret are renamed (`rag-data` → `{{ fullname }}-data`, etc.). `helm.sh/resource-policy: keep` means existing volumes survive, but they're no longer mounted, so an upgraded release comes up with empty volumes unless the operator sets `fullnameOverride` or migrates — worth a line in the upgrade docs.

Happy to approve as soon as the UID bit is sorted.

andyne13:
@ThibautChoppy one last thing before I can approve — a one-liner in `infra/docker/ui.Dockerfile`, details and reproduction in [my earlier comment](https://github.com/linagora/openrag/pull/633#issuecomment-5128506462).

Short version: the new `chown -R 10001:10001` matches the chart's `podSecurityContext`, but the image still ends with `USER nginx`, which is UID **101** in `nginxinc/nginx-unprivileged`. Under Helm that's fine (the pod forces `runAsUser: 10001`); under docker compose nothing overrides the user, so it runs as 101 against directories owned by 10001:

```
$ docker exec <admin-ui> sh -c 'id; ls -ldn /var/cache/nginx; touch /var/cache/nginx/probe'
uid=101(nginx) gid=101(nginx)
drwxrwxr-x 1 10001 10001 /var/cache/nginx
touch: /var/cache/nginx/probe: Permission denied
```

Either `USER 10001:10001` in the Dockerfile or `user: "10001:10001"` on the compose `admin-ui` service works — whichever you prefer.

It's an easy one to miss because the container still starts and serves the UI normally; it only bites when nginx spools a request body to `client_temp`, which is what `client_max_body_size 1024M` in the admin conf is there for — so document uploads through the compose proxy.

Everything else from my 16 July review is confirmed resolved (thanks @Ahmath-Gadji for the follow-up), and `tests/unit/infra/` is green at this head — 17 passed against it merged with `develop`. Push that change and I'll approve straight away.

andyne13:
@ThibautChoppy thanks for the work here — the five points from my 16 July review are all resolved, and `tests/unit/infra/` is green at this head (17 passed against it merged with `develop`). One item left before I can approve, in `infra/docker/ui.Dockerfile`.

Correcting something from my earlier comment, where I framed this as compose-only: `admin-ui` is actually the one image that doesn't follow the arbitrary-UID pattern `api.Dockerfile` deliberately uses.

```
# api.Dockerfile — the OpenShift pattern
useradd --uid 10001 --gid 0 ... && chgrp -R 0 /app /opt/uv && chmod g=u ...
USER 10001
# values.yaml: openrag.podSecurityContext -> runAsUser: 10001, runAsGroup: 0

# ui.Dockerfile — group 10001, and the final user is still the base image's
chown -R 10001:10001 /var/cache/nginx /etc/nginx/conf.d /var/run && chmod -R g+w ...
USER nginx        # = UID 101 in nginxinc/nginx-unprivileged
# values.yaml: adminUi.podSecurityContext -> runAsUser: 10001, runAsGroup: 10001
```

**1. Under docker compose this is broken** — nothing overrides the user, so the container runs as 101 against directories owned by 10001. Built from this branch:

```
$ docker exec <admin-ui> sh -c 'id; ls -ldn /var/cache/nginx; touch /var/cache/nginx/probe'
uid=101(nginx) gid=101(nginx)
drwxrwxr-x 1 10001 10001 /var/cache/nginx
touch: /var/cache/nginx/probe: Permission denied
```

The base image ships that directory owned by 101, so the `chown` is what introduces it:

```
$ docker run --rm --entrypoint sh nginxinc/nginx-unprivileged:1.27-alpine -c 'ls -ldn /var/cache/nginx'
drwxrwxr-x 1 101 0 /var/cache/nginx
```

**2. On OpenShift it depends on your SCC** — and here I'd rather ask than assert, since you know the target better than I do. Under the default `restricted-v2` the UID is assigned arbitrarily from the namespace range with GID 0, which is exactly why `api.Dockerfile` does `--gid 0` + `chgrp -R 0` + `chmod g=u` and why `openrag.podSecurityContext` keeps `runAsGroup: 0`. `admin-ui` asks for `runAsUser: 10001` / `runAsGroup: 10001`, which only works where the SCC allows pinning that UID/GID. With a permissive SCC (`anyuid` or equivalent) it's fine; otherwise the pod is rejected, or runs as a UID that can't write those paths. I haven't verified this on a cluster — what does your deployment use?

**The fix that covers both**, aligning `admin-ui` with what the API image already does:

```diff
-RUN chown -R 10001:10001 /var/cache/nginx /etc/nginx/conf.d /var/run && \
+RUN chown -R 10001:0 /var/cache/nginx /etc/nginx/conf.d /var/run && \
     chmod -R g+w /var/cache/nginx /etc/nginx/conf.d /var/run

-USER nginx
+USER 10001:0
```

plus `adminUi.podSecurityContext.runAsGroup: 0` in the chart, matching `openrag`. Under compose the user then owns the paths; on OpenShift any arbitrary UID writes via group 0. (Changing only the `chown` isn't enough for compose — uid 101 isn't in group 0, so the `USER` line has to move too.)

Worth saying why this is easy to miss: the container starts and serves the UI normally, so neither CI nor a smoke test catches it. It only fails when nginx spools a request body to `client_temp` — i.e. document uploads through the proxy.

Non-blocking, for the docs whenever you touch them: the PVCs/ConfigMap/Secret are renamed (`rag-data` → `{{ fullname }}-data`, etc.). `helm.sh/resource-policy: keep` means existing volumes survive, but they're no longer mounted, so an upgraded release comes up with empty volumes unless the operator sets `fullnameOverride` or migrates. A line in the upgrade notes would save someone a confusing afternoon.

Let me know about the SCC, and I'll approve as soon as the UID part is sorted.

andyne13 (approved):
Approving.

**The admin-ui permission blocker is resolved** — re-verified empirically by replicating the final stage on the base image and probing `/var/cache/nginx`, `/etc/nginx/conf.d` and `/var/run`:

| variant | runs as | writable |
|---|---|---|
| before (`chown 10001:10001` + `USER nginx`) | uid=101 gid=101 | denied on all three |
| after (`chown 10001:0` + `USER 10001:0`) | uid=10001 gid=0 | OK on all three |
| after, arbitrary UID (restricted-v2 sim) | uid=1000670000 gid=0 | OK on all three |

Taking the portable variant over the minimal one-line fix was the right call: it repairs compose *and* makes the chart work under a restricted SCC, not only a `RunAsAny` one — which matters for a chart that ships publicly.

**`fix(helm): give the migration Job the same ServiceAccount as the app`** — correct. It renders the SA exactly as `templates/openrag.yaml` does, so a templated value resolves identically in both and an unset value behaves as before. The rationale holds: the Job pins `runAsUser: 10001`, which under the namespace's `default` SA lands on restricted-v2 and gets rejected — and as a pre-upgrade hook that aborts the whole release. Test covers it.

Six checks green on `70f3ab8e`, no unresolved threads.

Merge note: `Chart.yaml` moves to `version: 0.6.0` and keeps `appVersion: "2.0.1"` — appVersion and the image tags are set by the version bump, so nothing to change here.

Follow-up, not blocking: you flagged that `raycluster` and `infinity` weren't covered by the security-context changes. Worth a separate issue so it doesn't get lost — happy to file it.

andyne13:
Thanks Thibaut for the work here, and for the migration Job ServiceAccount fix. Everything is green on my side — approved, and we're merging this for 2.1.0.

PR #634, opened 2026-07-07
fix(chat): honor the partition chat_llm preset when chatting

Ahmath-Gadji:
## Summary

A partition's `chat_llm` model-endpoint preset was stored and exposed through the admin API (`PATCH /partition/{p}`) but **never used when chatting**: `QueryService` was constructed with a single LLM built from the global settings and answered every chat/completion with it. The preset's only runtime consumer was the multiQuery/hyde retriever fallback in `RetrievalService`.

### Fix

- `QueryService` now receives the shared named `llm_factory` and resolves the answer-generation LLM **per request** (`chat`, `chat_stream`, `complete`) via a new `_resolve_llm()`, mirroring the `_resolve_chat_history_depth` partition semantics:
  - no partition / `openrag-all` sentinel → default LLM
  - single partition → its `chat_llm`, falling back to the default when unset
  - multi-partition → the preset only when every partition that sets one names the same endpoint
- `chat_llm` is not validated on assignment (the endpoint can be renamed/deleted afterwards), so an unknown name **falls back to the default LLM with a warning** instead of failing the request.
- The factory caches clients per endpoint name and already gets evicted on endpoint update/rename, so there is no per-request construction cost.

### Observability

Debug-level breadcrumbs to trace preset resolution end-to-end (zero added volume at the default INFO level):

- one line per indexed file with the resolved captioning VLM, contextualization LLM, topic-tagging LLM and embedder names (`None` = stage disabled)
- one line per chat request naming the resolved `chat_llm` preset (or why the default applies)
- `VLLMClient` / `VLLMVision` / `OllamaClient` log a `<Class> ready [model=… | endpoint=…]` construction line, mirroring the existing `VLLMEmbedder` one — fires once per configured endpoint and maps a preset name to its base URL/model

Unresolvable preset names keep surfacing as warnings (`Skipping contextualization: cannot resolve LLM '…'`, `Partition chat_llm preset not found …`).

## Test plan

- [x] 5 new unit tests for `_resolve_llm` (preset used, default paths, multi-partition unanimity, deleted-endpoint fallback, end-to-end `chat()` answering with the preset LLM); pipeline test extended for the new resolution line — `tests/unit`: 1638 passed (the 2 `test_rate_limit.py` failures are pre-existing on the base branch)
- [x] Verified live on a compose stack: a partition wired to a fake `chat_llm` endpoint (`http://fake-chat-endpoint.invalid/v1`) fails with `Cannot reach LLM at http://fake-chat-endpoint.invalid/v1` on both streaming and non-streaming chat, while a preset-less partition answers from the default endpoint; an unknown preset name logs the fallback warning and still answers
- [x] Verified the indexing line live: `model endpoints resolved for indexing [… vlm=default | contextualization_llm=MistralOR | topic_tagging_llm=default | embedder=default]` with contextualization running through the named endpoint

## Summary by CodeRabbit

* **New Features**
  * Partitioned LLM selection for both query generation and final answering using per-partition `chat_llm`, with automatic fallback to the default model.
* **Bug Fixes**
  * Partition updates now correctly reset `chat_llm` to the default when set to `null`, and reject unknown LLM endpoint names.
  * Conflicting per-partition `chat_llm` selections now safely revert to defaults.
* **Logging Improvements**
  * Enhanced debug context during inference client initialization and indexing stage selection.
* **Tests**
  * Expanded unit coverage for partition-scoped model selection, schema validation, and logging.

### DIFF
--- a/openrag/services/orchestrators/query_service.py
+++ b/openrag/services/orchestrators/query_service.py
@@ -159,13 +159,17 @@ class QueryService:
         return self._default_chat_history_depth
 
     def _resolve_llm(self, partition: list[str] | None) -> LLM:
-        """Effective answer-generation LLM for this request.
+        """Effective LLM for this request — query generation and answering.
 
         Honors a partition's configured ``chat_llm`` model-endpoint preset
-        (set via the admin API) over the default LLM. Same partition
-        semantics as ``_resolve_chat_history_depth``: no partition and the
-        ``"all"`` sentinel use the default; a multi-partition request uses
-        the preset only when every partition that sets one names the same
+        (set via the admin API) over the default LLM. Resolved once per
+        request (``chat`` / ``chat_stream`` / ``complete``) and used for
+        both the query-contextualization call and the final answer;
+        map-reduce stays on the default LLM until its post-release
+        refactor. Same partition semantics as
+        ``_resolve_chat_history_depth``: no partition and the ``"all"``
+        sentinel use the default; a multi-partition request uses the
+        preset only when every partition that sets one names the same
         endpoint — with conflicting presets there is no single owning
         partition, so the default applies.
 
@@ -204,7 +208,8 @@ class QueryService:
     # Query generation (was RagPipeline.generate_query — no LangChain)
     # ------------------------------------------------------------------
 
-    async def generate_query(self, messages: list[dict]) -> SearchQueries:
+    async def generate_query(self, messages: list[dict], llm: LLM | None = None) -> SearchQueries:
+        llm = llm or self._llm
         last_user = messages[-1]["content"]
         if RAGMODE(self._rag_mode) is RAGMODE.SIMPLERAG:
             return SearchQueries(query_list=[Query(query=last_user)])
@@ -224,7 +229,7 @@ class QueryService:
         }
         for attempt in (1, 2):
             try:
-                resp = await self._llm.chat(llm_messages, **params)
+                resp = await llm.chat(llm_messages, **params)
                 content = resp["choices"][0]["message"]["content"]
                 return SearchQueries.model_validate_json(_json_slice(content))
             except Exception as exc:
@@ -242,6 +247,9 @@ class QueryService:
     # ------------------------------------------------------------------
 
     async def _infer_relevancy(self, query: str, doc) -> tuple[bool, str]:
+        # Deliberately pinned to the default LLM: map-reduce is slated for a
+        # full post-release refactor, and routing it through the partition
+        # chat_llm preset is part of that work.
         async with get_llm_semaphore():
             try:
                 resp = await self._llm.chat(
@@ -290,9 +298,9 @@ class QueryService:
     # Preparation (was RagPipeline._prepare_for_chat_completion)
     # ------------------------------------------------------------------
 
-    async def _prepare_chat(self, partition: list[str] | None, payload: dict):
+    async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: LLM | None = None):
         messages = payload["messages"][-self._resolve_chat_history_depth(partition) :]
-        queries = await self.generate_query(messages)
+        queries = await self.generate_query(messages, llm=llm)
 
         metadata = payload.get("metadata") or {}
         use_map_reduce = metadata.get("use_map_reduce", False)
@@ -378,9 +386,9 @@ class QueryService:
         doc_lists, web_lists = await asyncio.gather(rag, web)
         return doc_lists, web_lists
 
-    async def _prepare_completions(self, partition: list[str], payload: dict):
+    async def _prepare_completions(self, partition: list[str], payload: dict, llm: LLM | None = None):
         prompt = payload["prompt"]
-        queries = await self.generate_query([{"role": "user", "content": prompt}])
+        queries = await self.generate_query([{"role": "user", "content": prompt}], llm=llm)
         chunks = await self._retrieval.retrieve_multi(partitions=partition, search_queries=queries)
         docs = [c.to_langchain() for c in chunks]
         context, included = format_context(
@@ -460,14 +468,15 @@ class QueryService:
     ) -> dict:
         """Non-streaming chat completion → finalized OpenAI dict."""
         metadata = payload.get("metadata") or {}
+        llm = self._resolve_llm(partitions)
         if partitions is None and not metadata.get("websearch", False):
             docs, web_results = [], []
         else:
-            payload, docs, web_results = await self._prepare_chat(partitions, payload)
+            payload, docs, web_results = await self._prepare_chat(partitions, payload, llm)
         sources = prepare_sources(docs, web_results)
 
         payload["messages"] = self._sanitize_messages(payload["messages"])
-        chunk = await self._resolve_llm(partitions).chat(payload["messages"], **_sampling(payload))
+        chunk = await llm.chat(payload["messages"], **_sampling(payload))
         chunk["model"] = model_name
         content = chunk.get("choices", [{}])[0].get("message", {}).get("content", "") or ""
         clean, citations = extract_and_strip_sources_block(content)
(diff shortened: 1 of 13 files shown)


hedhoud:
@codex review

Ahmath-Gadji:
**Scope update** (66008bfc): the `chat_llm` preset now also covers **query generation** — `chat` / `chat_stream` / `complete` resolve the LLM once per request and hand it down to `generate_query`, so a partition-scoped request no longer sends its chat history to the default endpoint for query contextualization in ChatBotRag mode. Covered by a new unit test asserting both the query-gen call and the answer go through the preset LLM.

**Deliberately out of scope: map-reduce.** `_infer_relevancy` (the map-reduce relevancy/summarization calls) intentionally stays on the default LLM for this release — map-reduce is an entire feature due for a full refactor post-release, and preset routing will land as part of that work. There's a comment in the code marking this, so please don't flag it as an oversight in review.

Ahmath-Gadji:
**Assignment-time validation** (1529ee34): `chat_llm` is now validated when it is *assigned* — `POST /partition` and `PATCH /partition/{p}` return **422 `MODEL_ENDPOINT_NOT_FOUND`** when the name isn't in the LLM endpoint catalog, mirroring the existing `PRESET_NOT_FOUND` check for indexation/retrieval presets. Typos now fail at the API instead of silently answering with the default LLM.

Two deliberate boundaries:
- **Only the incoming value is checked.** A *stored* name that goes stale later (endpoint renamed/deleted after assignment) never blocks unrelated PATCHes — the fail-open runtime fallback from this PR still covers that case.
- **Explicit `null` now really clears the preset.** The service's None-filter used to swallow `chat_llm: null`, so the admin UI's "reset to default" was a silent no-op — fixed by treating `chat_llm` as a nullable column; blank strings normalize to null at the schema layer.

11 new unit tests (6 service-level: reject/accept/clear/stale-doesn't-block on update + reject/accept on create; 5 schema-level for the null/blank normalization) — `tests/unit` fully green (1652 passed, incl. the rate-limit pair).

hedhoud (commented):
I rechecked the latest commit. The chat_llm assignment validation helps, but I still left a few notes around runtime behavior and one small doc drift.

openrag/services/orchestrators/query_service.py
```diff
@@ -156,11 +158,58 @@ def _resolve_chat_history_depth(self, partition: list[str] | None) -> int:
...
+        both the query-contextualization call and the final answer;
+        map-reduce stays on the default LLM until its post-release
+        refactor. Same partition semantics as
+        ``_resolve_chat_history_depth``: no partition and the ``"all"``
+        sentinel use the default; a multi-partition request uses the
+        preset only when every partition that sets one names the same
+        endpoint — with conflicting presets there is no single owning
+        partition, so the default applies.
+
+        ``chat_llm`` is not validated against the endpoint catalog when it
```
hedhoud: Small doc drift after the latest commit: `chat_llm` is now validated when a partition is created or patched, so this paragraph is no longer fully accurate. The fallback is still useful for stale values after an endpoint is deleted or renamed, but I’d reword this so the next reader doesn’t think assignment accepts any unknown name.

openrag/services/orchestrators/query_service.py
```diff
@@ -416,14 +468,15 @@ async def chat(
     ) -> dict:
         """Non-streaming chat completion → finalized OpenAI dict."""
         metadata = payload.get("metadata") or {}
+        llm = self._resolve_llm(partitions)
```
hedhoud, line 473: One thing still worries me here: the HTTP layer runs the availability and token checks against the default LLM before we reach this selected `chat_llm`. So a partition with a valid custom chat model can still be blocked if the default LLM is down, and the request can be checked against the wrong context size. I think those preflight checks should use the same LLM that will actually answer the request.

openrag/services/workers/pipeline_builder.py
```diff
@@ -196,17 +205,17 @@ def _select_embedder(self, row: MutableMapping[str, Any]) -> Embedder:
...
-    def _select_vlm(self, config: IndexationPipelineConfig | None) -> VLM | None:
-        """Pick the captioning VLM instance (availability only — policy is in
-        ``_should_caption``)."""
+    def _select_vlm(self, config: IndexationPipelineConfig | None) -> tuple[VLM | None, str | None]:
+        """Pick the captioning VLM instance and its endpoint name for logging
+        (availability only — policy is in ``_should_caption``)."""
         if config is not None and self.vlm_factory is not None:
             if config.vlm:
-                return self.vlm_factory(config.vlm)
+                return self.vlm_factory(config.vlm), config.vlm
```
hedhoud: I’d make this path defensive like the contextualizer and topic tagger selectors. If this named VLM endpoint is deleted or renamed later, this raises and fails the whole indexing job. Since captioning is an enrichment step, falling back or skipping with a warning would be safer than breaking indexing for the partition.

openrag/services/workers/pipeline_builder.py
```diff
@@ -196,17 +205,17 @@ def _select_embedder(self, row: MutableMapping[str, Any]) -> Embedder:
...
             try:
-                return self.vlm_factory("default")
+                return self.vlm_factory("default"), "default"
             except KeyError:
-                return self.vlm
-        return self.vlm
+                return self.vlm, "default"
+        return self.vlm, "default"
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in abad29d8. `_select_vlm` now wraps the named-endpoint lookup in the same `try/except KeyError` + `logger.warning` the contextualizer/topic-tagger selectors use, falling back to the legacy default VLM. Updated the two tests that asserted the old fail-fast behavior (`test_pipeline_falls_back_when_explicit_named_vlm_is_missing` and the e2e `test_indexing_falls_back_to_default_vlm_when_named_vlm_is_missing`) and added a breadcrumb test covering resolved named endpoints.

openrag/services/workers/pipeline_builder.py
Ahmath-Gadji: Done in abad29d8 — made it defensive like the contextualizer/topic-tagger selectors: a named VLM whose endpoint was deleted/renamed after assignment now falls back to the default VLM with a warning instead of failing the indexing job.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Good catch — reworded in abad29d8. The docstring now says `chat_llm` is validated on assignment (create/PATCH) and that the runtime fallback exists for names that go stale afterwards (endpoint renamed/deleted), rather than implying assignment accepts any unknown name.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Confirmed and agreed this is a real gap: the `check_llm_model_availability` dependency validates only `config.llm`, and `check_tokens_limit` uses `get_max_model_tokens()`, a single global primed once at startup from the default model. So a partition with a healthy custom `chat_llm` can be rejected when the *default* endpoint is down, and its request is sized against the wrong context window.

I'm scoping this as a **follow-up rather than folding it into this PR**, for three reasons:
1. It's a pre-existing limitation of the HTTP preflight layer, independent of this PR's diff — this PR resolves the answer LLM in the *service*, which is what makes the gap reachable, but the fix lives entirely in `api/routers/user/chat.py` + `api/dependencies/llm.py`.
2. A correct fix has to resolve the partition→`chat_llm` mapping *before* the `Depends`, then run availability against that endpoint and cache/fetch `max_model_len` **per endpoint** (today it's one startup-primed global) — a non-trivial change to a dependency that runs on every chat request, deserving its own tests.
3. Keeping it separate avoids expanding this PR's blast radius right before merge.

Behaviorally, this PR is a strict improvement over `main` (the preset now actually answers); the preflight mismatch only degrades the default-down / non-default-context edge case, which existed implicitly before. I'll open a tracking issue for "run chat preflight checks against the resolved per-partition LLM" and link it here. Shout if you'd rather I pull it into this PR instead.

openrag/services/orchestrators/query_service.py
```diff
@@ -423,7 +467,7 @@ async def chat(
         sources = prepare_sources(docs, web_results)
 
         payload["messages"] = self._sanitize_messages(payload["messages"])
-        chunk = await self._llm.chat(payload["messages"], **_sampling(payload))
+        chunk = await self._resolve_llm(partitions).chat(payload["messages"], **_sampling(payload))
```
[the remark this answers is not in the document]
Ahmath-Gadji: Agreed — same finding as hedhoud's note on line 471. Tracking this as a follow-up: the fix belongs in the HTTP preflight layer (`check_llm_model_availability` + the startup-primed `get_max_model_tokens` global), which must resolve the partition's effective `chat_llm` before the `Depends` and size/validate against *that* endpoint per-partition. Out of scope for this PR (which only moves LLM resolution into the service); I'll open an issue and link it.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Tracking issue opened: #639.

openrag/services/orchestrators/query_service.py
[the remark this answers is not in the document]
Ahmath-Gadji: Tracking issue opened: #639.

hedhoud (approved):
Approved. The issues raised in review were handled well: the VLM fallback is now defensive, the chat_llm docs are aligned with the validation behavior, and the remaining preflight mismatch is tracked separately in #639. CI is green.

PR #641, opened 2026-07-08
fix(admin-ui): preserve public host port in nginx redirects

hedhoud:
## Summary

This fixes the Admin UI nginx front door so backend redirects keep the browser-facing host and mapped port.

In Docker port-mapped deployments, nginx was not preserving the public port when the backend generated redirects like the Chainlit trailing-slash redirect. That could send the browser back to the right host but the wrong port.

The header behavior is unchanged; the small test rename only makes the existing browser-direct behavior clearer.

## Validation

- uv run pytest tests/unit/infra/test_admin_ui_compose.py -q
- npm test -- --run src/components/layout/header.test.tsx --reporter=dot
- npm run lint
- npm run build

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed browser-direct chat links so they correctly use the configured API address, including when a port is present.
  * Improved request forwarding in the admin UI proxy so the original public host is preserved during redirects and API access.
  * Added coverage to help prevent regressions in host handling and chat-link behavior.

### DIFF
--- a/infra/compose/nginx/openrag-admin.conf
+++ b/infra/compose/nginx/openrag-admin.conf
@@ -56,7 +56,11 @@ server {
         resolver 127.0.0.11 valid=30s ipv6=off;
         set $openrag_upstream openrag;
         proxy_pass http://$openrag_upstream:8080;
-        proxy_set_header Host              $host;
+        # Preserve the browser-facing host exactly, including a mapped port
+        # such as :8068. The backend uses Host when it emits slash redirects
+        # (/chainlit -> /chainlit/), and $host would strip that port.
+        proxy_set_header Host              $http_host;
+        proxy_set_header X-Forwarded-Host  $http_host;
         proxy_set_header X-Real-IP         $remote_addr;
         proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $forwarded_scheme;
--- a/tests/unit/infra/test_admin_ui_compose.py
+++ b/tests/unit/infra/test_admin_ui_compose.py
@@ -1,3 +1,4 @@
+import re
 from pathlib import Path
 
 import yaml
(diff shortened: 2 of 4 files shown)


andyne13 (commented):
Reviewed the full branch state (not just the diff). The change is safe — one clarification and a couple of minor, non-blocking asks.

**What's actually changing.** The functional change is the nginx `Host $host → $http_host` (plus the new `X-Forwarded-Host $http_host`). `header.tsx` is only an import reorder and `header.test.tsx` only a test rename — the chat link is still `apiUrl("/chainlit/")`, so the header origin behavior is unchanged. Given that, the title/description ("keep Chat on the front-door origin / avoid the API base URL") reads as if the header changed; it'd be clearer to frame this as the nginx host/port-preservation fix so the intent matches the diff.

**The nginx fix is correct.** In a port-mapped deployment the browser hits `host:8068`, which Docker maps to nginx's internal port; when the backend emits a slash redirect (`/chainlit` → `/chainlit/`) it builds `Location` from the Host header. `$host` strips the port, so the redirect drops `:8068` and lands on the wrong port; `$http_host` carries the browser-facing `host:8068` verbatim and the redirect stays correct. `$server_port` wouldn't work here (it's nginx's internal listen port), so `$http_host` is the right choice. Also confirmed: a single `location /` proxies everything, so the header applies uniformly (no OIDC-vs-chat divergence), and OIDC `redirect_uri` comes from an explicit env var rather than the Host header, so this doesn't affect it.

**Minor / non-blocking:**
- The `header.tsx` import reorder is a no-op — consider dropping it to keep the diff to the actual fix. (The test rename is a nice clarity improvement, worth keeping.)
- `$http_host` forwards the raw client `Host` unvalidated (`server_name _`). Not a regression — the old `$host` also derived from the client Host — but for a hardened front door it's worth pinning `server_name` / sitting behind a trusted proxy so a spoofed Host can't be reflected in backend-built URLs.
- `test_admin_ui_compose.py` matches the exact multi-space `Host              $http_host` string, so it'll break on any reformat of the conf; a whitespace-tolerant check would be more robust.

Nice catch on the port-stripping — that redirect breakage is easy to miss.

andyne13 (approved):
Thanks for the quick turnaround — all the review points are addressed:
- Title now reflects the actual change (nginx host/port preservation).
- The no-op `header.tsx` import churn is dropped.
- The config test is whitespace-tolerant now; verified the regex matches the conf and doesn't catch the `X-Forwarded-Host` line.

The `$http_host` change correctly preserves the browser-facing port so the backend's slash-redirects (e.g. `/chainlit` → `/chainlit/`) keep the mapped port. CI is green. LGTM 👍

PR #644, opened 2026-07-08
fix(admin-ui): simplify account settings copy

hedhoud:
## Summary

This updates the Account Settings copy so it reads more like a product surface and less like generated help text.

The page keeps the same layout and behavior. Only the visible wording changes for the page description, sign-in section, and API token section.

## Validation

- `npm run lint`
- `npm run build`

Note: `npm ci` reported existing dependency audit findings while installing this worktree. This PR does not change dependencies.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Settings page text to better explain API access and authentication.
  * Clarified that sign-in is handled through the organization’s SSO and that passwords are not managed in the app.
  * Improved API token guidance for use with CLI tools and scripts, including a clearer note that regenerating a token immediately invalidates the previous one.

### DIFF
--- a/ui/src/pages/app/settings.tsx
+++ b/ui/src/pages/app/settings.tsx
@@ -33,7 +33,7 @@ export default function SettingsPage() {
 
   return (
     <div>
-      <PageHeader title="Account Settings" description="Manage your programmatic API access" />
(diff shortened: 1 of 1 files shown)


Ahmath-Gadji (approved):
LGTM

PR #645, opened 2026-07-08
feat(presets): show partition usage and guard deletion of in-use presets

andyne13:
## What

- Annotate every preset with `used_by_partitions` (single bulk aggregate query — no per-row N+1) and surface it on the presets list API and the admin UI.
- Refuse deleting a preset that any partition still references, returning **409 ConflictError** with a message telling the operator to reassign first.
- Admin UI: an amber "used by N partitions" / "unused" badge per preset card (reusing the Super-Admin banner's amber tokens), and a context-aware delete-confirmation dialog.

## Concurrency

The in-use check is enforced atomically at the repository layer, not just in the service:

- `PgPresetRepository.delete` locks `partitions` in **SHARE** mode, counts references, and only then deletes the preset.
- The assign path (`PgPartitionRepository.update_partition`) re-checks preset existence in the DB **under its own write transaction, in the same lock order** (`partitions` before `pipeline_presets`).

Together these serialize a concurrent delete/assign of the same preset without deadlocking, so a partition can never be left pointing at a deleted preset. Both interleavings are covered by unit tests.

## Tests

- **Backend** — repo guard (rollback on concurrent delete, commit when the preset exists, fast-path skip for non-preset updates), preset-service usage annotation + delete guard, 409 error-handler mapping; integration tests assert 409 on in-use delete and the `used_by_partitions` count.
- **UI** — vitest coverage for the badge states (singular/plural/unused) and the 409 toast.
- Full unit + UI suites green. No schema migration required — `used_by_partitions` is computed.

## Summary by CodeRabbit

* **New Features**
  * Preset list/details now include `used_by_partitions`, and the admin UI displays “unused” vs “used by N partitions” badges.
* **Bug Fixes**
  * Deleting an in-use preset is blocked with HTTP **409** and consistent conflict error details.
  * Partition updates that change preset references are now protected against concurrent preset deletion, preventing invalid preset pointers.
* **UI Improvements**
  * Delete confirmation messaging now varies based on whether a preset is currently in use.
* **Tests**
  * Expanded integration and unit coverage for usage counts, conflict behavior, and preset-reference safety.


openrag/services/persistence/preset_repo.py
```diff
@@ -117,11 +127,36 @@ async def rename(self, old_name: str, new_name: str, preset_type: str, config: d
...
+        write in the same transaction (touching ``partitions`` before
+        ``pipeline_presets``, the same order used here, so the two never
+        deadlock), which closes the window on both orderings — neither an
+        orphaned partition reference nor a lost 409 can result. Raises
+        :class:`ConflictError` (409) if the count is nonzero.
+        """
+        col = _preset_column(preset_type)
+        async with self.pool.acquire() as conn:
+            async with conn.transaction():
+                await conn.execute("LOCK TABLE partitions IN SHARE MODE")
```
hedhoud, line 156: I think there is still one lock-order gap here. Delete now locks `partitions` first, but rename still updates the preset row first and only touches `partitions` afterwards. If one request renames an unused preset while another request deletes the same preset, each transaction can end up waiting on the lock held by the other one.

Can we make rename follow the same order as delete, so it takes the `partitions` lock before touching the preset row? That would keep the preset mutations consistent and avoid this deadlock path.

andyne13: Good catch — you're right, that's a genuine ABBA deadlock: `delete()` locks `partitions` → `pipeline_presets`, while `rename()` did the reverse.

Fixed in f1258cf2 by making `rename()` repoint `partitions` **before** updating the `pipeline_presets` row, so both paths acquire the two locks in the same order.

One subtlety worth noting: taking an explicit `LOCK TABLE partitions IN SHARE MODE` at the top of `rename()` wouldn't fully close it — `rename()` still has to *upgrade* that SHARE to ROW EXCLUSIVE for its `UPDATE partitions`, and that upgrade conflicts with a concurrent `delete()` holding SHARE, so the two could still deadlock on the upgrade. Reordering the two writes gives the same consistent lock order without needing the explicit lock. Both statements stay in one transaction, so the transient state is never externally visible and a NotFound rolls the repoint back. Updated the rename unit test to assert the new order.

hedhoud (approved):
Looks good to me. I rechecked the preset usage counts, the in-use delete guard, the lock ordering around preset rename/delete and partition assignment, plus the Admin UI behavior. The targeted backend and UI tests pass locally, and CI is green.

PR #653, opened 2026-07-09
feat(indexation): disable topic tagging by default

Ahmath-Gadji:
## Why

Topic tagging is currently **on by default** for every partition, so every indexed document pays an extra LLM call to generate tags that nothing consumes: they are not surfaced in the API and not used in retrieval. The admin UI already acknowledges this — the toggle ships `disabled` with the hint *"Coming soon — generated tags aren't surfaced or used in retrieval yet."*

This makes the feature opt-in until it actually ships.

## What changed

| File | Change |
|---|---|
| `core/config/indexation_pipeline.py` | `enable_topic_tagging` default `True` → `False` |
| `services/orchestrators/preset_service.py` | seeded `default` indexation preset sets it `False` |
| `services/workers/indexer_pool.py` | `_required_llm_names` no longer resolves an LLM for an absent key |
| `services/workers/indexer_actor.py` | `_replace_topic_tags_if_needed` treats an absent key as disabled |
| `ui/src/pages/admin/presets.tsx` | toggle fallback `true` → `false`, matching the backend |

## Reviewer note — one behavior change beyond the flag

`_replace_topic_tags_if_needed` previously purged tags only when a preset set `enable_topic_tagging` **explicitly** to `False`. Now that an absent key *means* disabled, the purge must apply to absent-key presets too (`legal`, `finance`) — otherwise they would silently stop tagging but never clear tags written by an earlier run.

The cost is one indexed `DELETE FROM topic_tags WHERE document_id = ? AND partition = ?` per indexed document on the (now default) disabled path. It is a no-op for fresh files. Happy to scope it back to `is False` semantics if you'd rather not pay that on the hot path, at the price of leaving stale tags unreachable for those two presets.

## Upgrade impact

**Existing deployments keep their current behavior.** `seed_defaults()` only inserts when a preset type has zero rows, so a database that already seeded `default` with tagging on retains it. Only fresh deployments get the new default. Operators who want tagging back set `enable_topic_tagging: true` on the partition's indexation preset.

## Testing

- Added `test_indexation_pipeline_topic_tagging_defaults_off` pinning the model default.
- Added `test_seed_defaults_indexation_leaves_topic_tagging_off` covering the seeded presets.
- Updated `test_required_llm_names_mirrors_pipeline_selection` — `{}` now resolves no LLM, and an explicit `True` still resolves `default`.
- `uv run pytest tests/unit/` → 1663 passed. The 2 failures in `tests/unit/api/middleware/test_rate_limit.py` are timing-flaky and reproduce unchanged on `origin/main`.
- `uv run ruff check` / `format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Topic tagging is now off by default in pipeline configuration, seeded indexation presets, and the admin preset UI unless explicitly enabled.
  * Re-indexing now clears previously stored topic tags when topic tagging wasn’t enabled (including when the setting is omitted).
  * Indexer endpoint/model selection no longer assumes topic tagging is enabled when the flag is missing.

* **Tests**
  * Added/updated unit tests validating default-off behavior, preset seeding defaults, topic-tag purging when the key is absent, and required-model selection logic.

### DIFF
--- a/openrag/core/config/indexation_pipeline.py
+++ b/openrag/core/config/indexation_pipeline.py
@@ -52,8 +52,10 @@ class IndexationPipelineConfig(BaseModel):
         default_factory=lambda: ["person", "organization", "location", "event"],
     )
 
-    # Topic tagging
-    enable_topic_tagging: bool = True
+    # Topic tagging. Off by default: the generated tags are not yet surfaced in
+    # the API or used for retrieval, so enabling it only buys an extra LLM call
+    # per document. Partitions that want the tags opt in on their preset.
+    enable_topic_tagging: bool = False
     max_topic_tags: int = Field(default=7, ge=1, le=50)
     topic_tagging_llm: str | None = None
 
--- a/openrag/services/orchestrators/preset_service.py
+++ b/openrag/services/orchestrators/preset_service.py
@@ -52,7 +52,7 @@ _DEFAULT_SEEDS: dict[str, dict[str, dict[str, Any]]] = {
             # spin up that backend's Ray pool — e.g. forcing marker on a
             # pymupdf-configured deployment. Named presets below opt in explicitly.
             "enable_entity_extraction": True,
-            "enable_topic_tagging": True,
+            "enable_topic_tagging": False,
         },
         "legal": {
             "chunking": {"name": "recursive_splitter", "chunk_size": 1024, "chunk_overlap_rate": 0.25},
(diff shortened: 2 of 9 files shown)


Ahmath-Gadji:
Fixes #669

hedhoud (approved):
Looks good to me. The change is small and makes the behavior consistent: topic tagging is now opt-in instead of running an extra LLM call by default for a feature that is not exposed or used yet.

I checked the backend config, seeded presets, indexer LLM resolution, stale tag cleanup, and the Admin UI fallback. Explicit enable_topic_tagging: true still keeps the existing behavior, while missing/false now correctly disables it.

Validation I ran locally:
- focused backend tests: 94 passed
- ruff check / format check on touched backend files: passed
- Admin UI build: passed
- Admin UI lint: passed

Only note: the branch is behind develop, so it may need an update before merge depending on the repo policy.

PR #654, opened 2026-07-09
fix: stop chainlit's Socket.IO transport from tripping the rate limiter

Ahmath-Gadji:
## Problem

A chat session floods the API log with:

```
WARNING | api.middleware.rate_limit:dispatch:101 - Rate limit exceeded [path=/chainlit/ws/socket.io/ | tier=default | identity=ip:172.20.0.3]
```

Three defects compound here.

**1. Chainlit's Socket.IO traffic is rate limited, but can't be attributed.** `mount_chainlit()` mounts Chainlit's whole ASGI app — including its Socket.IO server at `/chainlit/ws/socket.io` — inside the API. That subtree is auth-bypassed (`is_bypass_path`), because Chainlit runs its own header-auth callback, so `AuthMiddleware` returns without setting `request.state.user`. `RateLimitMiddleware` then finds no user: `_is_admin()` is `False` even for an admin's browser, and `_identity()` falls back to `request.client.host`.

**2. That IP is the proxy, not the browser.** `entrypoint.sh` runs the bare `uvicorn` CLI, which reads `FORWARDED_ALLOW_IPS` — not the `UVICORN_FORWARDED_ALLOW_IPS` that `api/main.py` reads and that `env_vars.md` documents. The documented variable was a no-op on the default container path. So nginx's `X-Forwarded-For` is untrusted, `request.client.host` is the admin-ui container (`172.20.0.3`), and **every chat user in the deployment shares one 600/minute bucket.**

**3. The transport should never have been HTTP.** `openrag-admin.conf`'s `location /` proxied over nginx's default HTTP/1.0 with no `Upgrade`/`Connection` headers, so Socket.IO's WebSocket upgrade always failed and it stayed on long-polling. Under long-polling each emitted packet terminates the open `GET` and the client re-polls, so a single streamed answer costs hundreds of requests. The resulting 429s look like transport errors to engine.io, which reconnects — and issues more requests.

Note `SUPER_ADMIN_MODE` is unrelated: the limiter never reads it, and the `is_admin` bypass it resembles cannot fire on a path where auth never ran.

## Changes

One commit each:

- **`fix(rate-limit): exempt the auth-bypassed chainlit subtree`** — skip the limiter on `/chainlit` and `/assets/`, overridable via the new `RATE_LIMIT_EXEMPT_PATHS` (set-but-empty disables the exemption). A per-identity limit that cannot distinguish identities only throttles everyone at once. `/auth/*` stays limited — it's the brute-force surface and IP keying is the point there.
- **`fix(nginx): forward WebSocket upgrades to the API`** — `proxy_http_version 1.1` plus the standard `$connection_upgrade` map. With the upgrade succeeding, chainlit chat becomes one WebSocket, whose ASGI scope is `websocket`, which Starlette's `BaseHTTPMiddleware` skips — so it stops traversing the limiter entirely. This fixes the load; the exemption fixes the accounting.
- **`fix(entrypoint): honour UVICORN_FORWARDED_ALLOW_IPS on the uvicorn CLI path`** — forward the documented variable to `--forwarded-allow-ips` so per-user rate limits key on real clients. This gap also made OIDC session cookies ship `Secure=False` behind a TLS-terminating proxy, so it's worth fixing independently of the above.

## Verification

- `pytest tests/unit/api/` — 261 passed. The rate-limit suite grew from 10 to 15 tests: chainlit and `/assets/` exempt under a 2/minute budget, non-exempt routes still 429, env override replaces the defaults, set-but-empty disables them.
- `nginx -t` against `nginx:alpine` with the new conf mounted — syntax OK.
- New regression test asserts `entrypoint.sh` passes `--forwarded-allow-ips` from `UVICORN_FORWARDED_ALLOW_IPS`, mirroring the existing guard on `api/main.py`.
- Not yet exercised end-to-end against a live stack — the WebSocket-upgrade path in particular is verified by config review and `nginx -t`, not by watching a browser negotiate the upgrade.

## Operator notes

`UVICORN_FORWARDED_ALLOW_IPS` now takes effect where it previously didn't. It stays defaulted to `127.0.0.1` (unchanged behaviour, still ignores the proxy) — deployments wanting real client IPs must set it to their subnet or `*`. `.env.example` documents this, mirrored to `docs/assets/env_example.env`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable rate-limit path exemptions via `RATE_LIMIT_EXEMPT_PATHS`, with defaults for Chainlit and static assets.
  * Exemptions are applied early (before admin bypass) and won’t allow `/auth/` coverage.
  * Improved reverse-proxy correctness by explicitly trusting `UVICORN_FORWARDED_ALLOW_IPS`.

* **Documentation**
  * Expanded env var guidance for proxy trust and rate-limit exemptions, including security warnings against overly broad forwarded-header settings.

* **Tests**
  * Added unit tests for exemption boundaries/overrides, prevention of auth-route exemptions, and verification that the entrypoint passes `--forwarded-allow-ips` to Uvicorn.


Ahmath-Gadji:
Fixes #668

openrag/api/middleware/rate_limit.py
```diff
@@ -83,13 +113,20 @@ async def dispatch(self, request: Request, call_next):
         if not self.enabled:
             return await call_next(request)
 
+        path = request.url.path
+
+        # Exempt prefixes are checked before the admin bypass on purpose: these
+        # paths never run through AuthMiddleware, so request.state.user is unset
+        # and _is_admin() is False even for an admin's browser session.
+        if self._exempt and path.startswith(self._exempt):
```
hedhoud, line 140: I think we should make `/auth/` impossible to exempt here. Right now, if someone sets `RATE_LIMIT_EXEMPT_PATHS=/auth/` while tuning the Chainlit/assets exemptions, the auth limit is skipped completely. That would remove the brute-force protection on login routes by accident. Could we guard `/auth/` before applying exemptions, and add a small test for that case?

hedhoud (commented):
Small merge blocker: this PR now conflicts with develop in the nginx config because the WebSocket upgrade block was also added there. The config idea looks fine, but the branch needs to be updated and the duplicate/conflicting block cleaned up before it can merge.

openrag/api/middleware/rate_limit.py
Ahmath-Gadji: Good catch — fixed. Any `RATE_LIMIT_EXEMPT_PATHS` entry that would cover `/auth/` (an exact `/auth/` match, or something broader like a bare `/`) is now dropped at construction time, with a warning logged, so the auth tier can never be opted out via env var. Added `test_auth_cannot_be_exempted_via_env_override` and `test_auth_cannot_be_exempted_via_broad_prefix_override` in 91bca3ba.

Ahmath-Gadji:
Rebased onto develop — the WebSocket-upgrade commit here now overlaps with 0f8911fb (already on develop), so it's folded down to just the extra explanatory comment; no duplicate proxy_http_version/Upgrade/Connection directives remain. Re-verified with nginx -t against nginx:alpine. mergeable now reports MERGEABLE.

Also addressed the /auth/ exemption gap in the inline comment below.

openrag/api/middleware/rate_limit.py
```diff
@@ -34,13 +61,33 @@ def _env_flag(name: str, default: bool) -> bool:
...
     def __init__(self, app):
         super().__init__(app)
         self.enabled = _env_flag("RATE_LIMIT_ENABLED", True)
         self._limiter = MovingWindowRateLimiter(MemoryStorage())
+        exempt = _env_prefixes("RATE_LIMIT_EXEMPT_PATHS", DEFAULT_EXEMPT_PREFIXES)
+        # A prefix that is itself a prefix of "/auth/" (e.g. "/auth/", "/a", "/")
+        # would exempt the brute-force surface via path.startswith() below — drop
+        # those regardless of where they came from, so /auth/ can never be opted
+        # out of rate limiting by RATE_LIMIT_EXEMPT_PATHS.
+        unsafe = tuple(p for p in exempt if _PROTECTED_PREFIX.startswith(p))
```
hedhoud: Small gap here: this protects us from broad values like `/` or `/auth/`, but a more specific auth prefix still gets through. I checked locally with `RATE_LIMIT_AUTH=1/minute` and `RATE_LIMIT_EXEMPT_PATHS=/auth/login`; both `/auth/login` requests returned 200, so the login brute-force limit was skipped. Could we treat both directions as unsafe, for example drop a prefix when `/auth/`.startswith(prefix) or prefix.startswith(`/auth/`), and add a small regression test for `/auth/login`?

Ahmath-Gadji: Confirmed and fixed — good find, thanks for testing it locally. The guard only checked one direction; now it drops a prefix if it overlaps /auth/ either way (a prefix of /auth/, or a prefix starting with /auth/), so /auth/login specifically is caught too. Added test_auth_login_cannot_be_exempted_via_narrower_prefix_override in 2d351511.

hedhoud (approved):
LGTM

PR #662, opened 2026-07-15
fix(admin-ui): copy button silently fails to copy token in new-user dialog

### LINKED ISSUE #661: Copy-token button on new-user creation dialog reports success but doesn't copy to clipboard
## Bug

On the admin **Users** page, creating a new user (`New user` → pre-provision dialog) shows a one-time API token with a copy button. Clicking it shows the "Token copied to clipboard" success toast, but the token is **not** actually placed on the clipboard (pasting yields nothing / stale content) on both Firefox and Chrome.

## What works for comparison

The same `copyToClipboard()` helper (`ui/src/lib/utils.ts`) and the same success/error toast pattern is used in three places, but only one is broken:

| Flow | File | Copy button works? |
|---|---|---|
| Admin creates a new user | `ui/src/pages/admin/users/list.tsx` (`PreProvisionDialog`, ~L218) | ❌ No — reports success, clipboard not updated |
| Admin regenerates another user's token | `ui/src/pages/admin/users/detail.tsx` (`ApiTokenTab`, ~L520) | ✅ Yes |
| User regenerates their own token (Settings) | `ui/src/pages/app/settings.tsx` (~L25) | ✅ Yes |

## Likely lead

The broken flow is the only one of the three that renders the token/copy button **inside a Radix `Dialog`** (`ui/src/components/ui/dialog.tsx`, via `PreProvisionDialog`'s `<Dialog>/<DialogContent>`). Both working flows render the token in a plain `<Card>` on the page, with no dialog/portal/focus-trap involved.

`copyToClipboard()` (`ui/src/lib/utils.ts:44`) tries `navigator.clipboard.writeText()` first and falls back to a hidden-`<textarea>` + `document.execCommand("copy")` only when `navigator.clipboard` is unavailable (non-secure context). Since the bug reproduces on both Firefox and Chrome over what should be a secure context, the primary `navigator.clipboard.writeText()` path is presumably the one running and resolving without throwing — so `copyToClipboard()` returns `true` (hence the success toast) even though nothing ends up on the OS clipboard. That smells like a focus/timing issue caused by Radix `Dialog`'s focus-trap/portal machinery interfering with the clipboard write at the moment the button is clicked, rather than a bug in `copyToClipboard()` itself.

Worth checking during investigation:
- Does `document.hasFocus()` / `document.activeElement` look sane at the moment `copyToken()` in `PreProvisionDialog` (list.tsx:218) fires, versus in the two working flows?
- Does wrapping the copy button interaction differently (e.g. not inside `DialogPrimitive.Content`'s focus scope, or delaying focus trap re-assertion) resolve it?
- Whether `navigator.clipboard.writeText` throws (and we silently fall through to the `execCommand` fallback, which can return `true` without actually copying if the hidden textarea's `select()` loses out to Radix's focus trap stealing focus back).

## Repro steps

1. As admin, go to **Users** → **New user**.
2. Fill in the pre-provision form and submit.
3. In the "User created" dialog, click the copy icon next to the API token.
4. Toast shows "Token copied to clipboard".
5. Paste (Ctrl/Cmd+V) anywhere — the token is not there.

Tested on Firefox and Chrome.

## Expected

Clicking the copy button in the new-user dialog copies the token to the clipboard, consistent with the other two token-copy flows, and the toast accurately reflects whether the copy succeeded.

Ahmath-Gadji:
## Summary

- Fixes the "New user" pre-provision dialog's copy button — it toasted "Token copied to clipboard" without actually copying anything, on both Firefox and Chrome.
- Root cause: `copyToClipboard()`'s `execCommand("copy")` fallback appended its hidden `<textarea>` to `document.body`. When the button lives inside a focus-trapped container (the Radix `Dialog` used by `PreProvisionDialog`), the trap treats a `document.body` child as focus escaping the modal and yanks focus straight back — synchronously, before `ta.select()` + `execCommand("copy")` can read the selection. `execCommand` still returns `true` (nothing selected ≠ command failure), so the toast lies. The other two token-copy flows (admin regenerate, self-service regenerate in Settings) render the token in a plain `Card`, not inside a `Dialog`, so they never hit this.
- Fix: `copyToClipboard()` now takes an optional `anchor` element and inserts the temporary textarea next to it instead of at `document.body`, so it stays inside whatever focus-trapped subtree the button is already in. All three call sites now pass `e.currentTarget` (the clicked button) as the anchor.

## Test plan

- [x] `npx vitest run` — all 110 tests pass, including 3 new `copyToClipboard` tests covering: fallback textarea anchored next to the trigger (not `document.body`), fallback with no anchor still works, and the async `navigator.clipboard.writeText` path is preferred/unaffected.
- [x] `npx tsc -b` — clean.
- [x] `npx eslint` on changed files — clean.

Fixes #661

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved “Copy token” reliability across user administration and settings pages.
  - Clipboard fallback now keeps temporary copy elements within the relevant interface area, helping prevent focus-trap issues.
  - Copying works with both modern browser clipboard support and legacy fallback behavior.
  - Existing success and error notifications remain unchanged.

- **Tests**
  - Added coverage for modern clipboard copying, fallback behavior, anchored copying, cleanup, and missing-anchor scenarios.

### DIFF
--- a/ui/src/lib/utils.test.ts
+++ b/ui/src/lib/utils.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from "vitest";
-import { intOr, numOr } from "./utils";
+import { describe, it, expect, afterEach, vi } from "vitest";
+import { intOr, numOr, copyToClipboard } from "./utils";
 
 // These guard against `parseInt("")` / `parseFloat("")` → NaN, which
 // JSON.stringify serializes as `null` when sent to the API.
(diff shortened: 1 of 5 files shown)


hedhoud (dismissed):

hedhoud (commented):
The fix looks well scoped to the clipboard fallback, and the anchored textarea path is covered by tests. I don't see a blocker in this change, LGTM.

hedhoud (approved):

Issue #658, opened 2026-07-15
Deleted files/partitions can leave orphaned queryable chunks in Milvus; partition-name reuse can leak across tenants


andyne13:
## Summary

Deleting a file or a partition removes the Postgres rows **first**, then cleans up Milvus on a **best-effort basis with the exception swallowed**. There is no reconciliation job, so any failure in that window leaves orphaned chunks in Milvus that remain fully queryable. Because Milvus search is scoped only by the plain `partition` name string, recreating a partition with the same name can re-attach a previous tenant's surviving chunks — a cross-tenant exposure.

## Severity: high priority (correctness / tenant-isolation hardening) — not an every-request bug

The happy path is fine: when the Milvus delete succeeds (the normal case), the chunks are gone. This is a **failure-mode** gap — orphans only appear when the Milvus delete throws or the process dies between the Postgres commit and the Milvus delete. It is worth fixing promptly for the reasons below, but it is not an every-operation corruption like #657.

## Impact

- **Right-to-be-forgotten / correctness:** a "deleted" document whose Milvus chunks survived stays queryable and keeps being cited in RAG answers. The user believes the document is gone.
- **Cross-tenant exposure:** the Milvus `partition` field is the plain partition-name string (no epoch/UUID). After a partition delete leaves orphans, recreating a partition with the same name — a routine operation — makes the previous tenant's surviving chunks appear in the new partition's searches. In a single-tenant deployment this reduces to "a user's own deleted docs resurface"; in a multi-tenant one it is a cross-tenant leak.

## Relationship to #657

Shares the same root cause as #657: the Milvus lifecycle is not reliably coupled to Postgres. #657 is the every-re-index duplication case; this issue is the delete/cleanup case. A partition-epoch/UUID tag and idempotent delete-by-filter would address both.

## Root cause / code trace

Both delete paths commit the relational delete first, then attempt the Milvus delete and swallow any error:

- `openrag/services/workers/dispatcher.py:119-137` (`delete_file`):
  - `remove_file_from_partition(...)` (PG) → `query_ids_by_filter(...)` → `vector_store.delete(ids)` inside `try/except` that only logs `"... reconciliation task required"`.
- `openrag/services/orchestrators/partition_service.py:245-273` (`delete_partition`):
  - `_partition_repo.delete_partition(...)` (PG) → `query_ids_by_filter({"partition": ...})` → `vector_store.delete(ids)` inside `try/except` that only logs `"... reconciliation task required"`.

[...]

PR #670, opened 2026-07-15
fix(indexer): re-index no longer duplicates a file's chunks in Milvus

### LINKED ISSUE #657: Re-indexing a file duplicates its chunks in Milvus (replace path never deletes old chunks)
## Summary

Re-indexing an existing file (`PUT /indexer/partition/{partition}/file/{file_id}`, i.e. `replace=True`) **never deletes the file's previous chunks from Milvus**. The old chunk set is left in place and a full new chunk set is inserted alongside it. Duplication compounds on every re-index; search then returns duplicated and stale content, and vector storage grows without bound.

This is a regression introduced by the hexagonal / Ray-removal refactor: the delete step existed in the legacy `Indexer.add_file(replace=True)` flow and was dropped when the worker pipeline replaced it. The stale route comment still describes the old behaviour (see below).

## Impact

- **Severity: critical** — corrupts a first-class, documented operation (re-index / replace).
- Retrieval returns the same content multiple times and can surface stale content from a superseded version of the file.
- Milvus storage for a repeatedly re-indexed file grows linearly with the number of re-indexes.
- First-time indexing is **not** affected — only the `replace=True` path.
- Not a data-loss bug: the new content is indexed correctly; the problem is duplication of the old chunks.

## Root cause / code trace

The replace path updates the Postgres catalog row and topic tags, runs the pipeline, and stores chunks — but no stage deletes the prior Milvus chunks.

- Route: `openrag/api/routers/admin/indexing.py:283` calls `service.add_file(replace=True)`. The comment at `indexing.py:276-278` claims an "insert-before-delete" flow that no longer exists.
- `IndexingService.add_file` (`openrag/services/orchestrators/indexing_service.py:174-210`) → `dispatcher.dispatch_indexing(replace=True)`.
- `WorkerDispatcher.dispatch_indexing` (`openrag/services/workers/dispatcher.py:62-118`) only sets task state and submits to the pool — no delete.
- `IndexerWorkerActor.process_file` (`openrag/services/workers/indexer_pool.py:200-244`) — no delete.
- `IndexerWorker.process_file` (`openrag/services/workers/indexer_actor.py:43-105`) — on `replace=True` it updates only the PG catalog row (`_write_catalog_record`) and PG topic tags (`_replace_topic_tags_if_needed`).
- Pipeline stages (`openrag/services/workers/pipeline_builder.py`): parse → caption → chunk → contextualize → topic_tag → embed → store. **There is no delete stage.**
- `store_stage` (`openrag/services/workers/stages/store.py:43`) → `vector_store.upsert(chunks)`.
- `MilvusStore.upsert` (`openrag/services/storage/milvus_store.py:711-765`) is a plain `insert` into an `auto_id=True` collection (schema at `milvus_store.py:276`). Every insert mints new Milvus-assigned INT64 primary keys, so a second index of the same file cannot overwrite the first — the rows co-exist. (The method is named `upsert` but performs a pure `insert` with no business-key dedup.)

The only Milvus delete in the worker layer is `WorkerDispatcher.delete_file` (`dispatcher.py:119-137`), which filters by `{"partition", "file_id"}` and is invoked **only** by the `DELETE` endpoint — never by the replace path.

## Steps to reproduce

1. Index a file into a partition (`POST /indexer/partition/{p}/file/{id}`).
2. Note the chunk count for that `file_id` (e.g. search or query by filter `{"partition": p, "file_id": id}`).
3. Re-index the same file (`PUT /indexer/partition/{p}/file/{id}`).
4. Query chunks for that `file_id` again → the count has roughly doubled; the original chunks are still present.
5. Repeat step 3 → count keeps growing linearly.

## Suggested fix

Delete the file's existing Milvus chunks at the front of the replace path before inserting the new ones, mirroring `WorkerDispatcher.delete_file`:

- Before the store stage (or in the dispatcher when `replace=True`), run `query_ids_by_filter({"partition": partition, "file_id": file_id})` + `vector_store.delete(ids)`.
- Prefer a delete-by-filter over the previous "snapshot-then-delete" approach so it is idempotent and safe to re-run.
- Note the ordering trade-off: deleting old chunks before inserting new ones leaves a brief window with no chunks if the re-index fails; the legacy flow inserted first, then deleted the old IDs. Whichever ordering is chosen, the delete must actually run.
- Add a regression test that asserts the chunk count for a `file_id` is stable (not cumulative) across repeated re-indexes.

## Also worth fixing alongside

The stale route comment at `openrag/api/routers/admin/indexing.py:276-278` should be removed/corrected, since it currently asserts a delete behaviour that the code does not perform.

andyne13:
## What

Re-indexing an existing file (`PUT /indexer/partition/{partition}/file/{file_id}`, `replace=True`) no longer duplicates the file's chunks in Milvus. Previously the replace path only rewrote the Postgres catalog row and topic tags, then inserted a fresh full set of chunks — but the Milvus collection is `auto_id`, so a plain insert can never overwrite the old chunks. Every re-index therefore left the previous chunks in place and appended a new copy, compounding on each run: search returned duplicated + stale content and vector storage grew without bound.

This was a regression introduced by the hexagonal / Ray-removal refactor (v2.0.0): the legacy `Indexer.add_file(replace=True)` flow did insert-before-delete; the new worker pipeline dropped the delete step.

## How

`IndexingPipeline.run` (`services/workers/pipeline_builder.py`) is now **insert-before-delete** on `replace=True`:

1. **Snapshot** the file's existing chunk ids *before* the store stage runs — `query_ids_by_filter({"partition": partition, "file_id": file_id})`, scoped strictly to this file in this partition.
2. **Store** the new chunks (unchanged store stage).
3. **Delete** exactly the snapshotted old ids *after* the new chunks are safely stored — `delete(old_ids)`.

This reuses the same `query_ids_by_filter` + `delete` primitives that `WorkerDispatcher.delete_file` already uses in production, so the Milvus mechanics are proven.

### Safety properties

- **No empty window.** Snapshot-before / delete-after means a re-index that fails at or before the store stage leaves the old chunks intact — the file is never left with zero chunks.
- **Snapshot failure is safe.** If the pre-insert lookup fails (e.g. Milvus hiccup), cleanup is skipped and indexing still completes. Leftover duplicates are recoverable; losing the freshly-indexed content is not.
- **Delete failure is best-effort.** The new chunks are already stored, so a cleanup delete failure only leaves recoverable duplicates behind (logged for reconciliation) — mirrors `delete_file`.
- **Scoped.** The snapshot/delete is filtered by `{partition, file_id}`, so a re-index never touches other files or other tenants.

First-time indexing (`replace=False`) is completely untouched — no extra query, no delete.

## Known limitations (flagged, not fixed here)

Milvus has no transactions, so this fix is intentionally scoped to the common case. Both seams below are **strictly better than the pre-fix behaviour** (which duplicated on *every* re-index) and are called out in a code comment on the replace block:

- **Not atomic under concurrency.** Two overlapping re-indexes of the *same* file both snapshot the same old ids and both keep their new set, leaving duplicates. Serializing replace per `(partition, file_id)` (e.g. a Postgres advisory lock) belongs with the durable job/lifecycle work — see #658/#660.
- **No reconciler for the crash window.** A process crash between store and delete orphans the old chunks; the delete is best-effort and logged (same contract as `WorkerDispatcher.delete_file`), but the reconciliation job doesn't exist yet — tracked in #658/#660.

These are deliberately out of scope for a v2.0.1 patch: they need the partition-epoch / reconciliation machinery, not a bigger change to the hot re-index path.

## Tests

- **Unit** (`tests/unit/services/workers/test_pipeline_builder.py`, 5 new): insert-before-delete ordering (`query → upsert → delete`), `{partition, file_id}` scoping, no-delete-when-file-had-no-prior-chunks, first-time index issues no query/delete, snapshot-failure still indexes + skips cleanup, delete-failure is best-effort.
- **Integration vs. real Milvus 2.6** (`tests/integration/test_reindex_no_duplicate_integration.py`, 2 new): drives the actual pipeline — re-indexing a 3-chunk file twice keeps the count at 3 (not 6, then 9) and the survivors are the freshly-stored set; re-indexing one file leaves another file's chunks untouched. Auto-skips when Milvus isn't reachable.

Full unit suite green (1668 passed); ruff + layer-import guard clean.

Fixes #657

## Summary by CodeRabbit

* **Bug Fixes**
  * Re-indexing with `replace=True` no longer creates duplicate chunks.
  * Replacement is correctly scoped to the selected file and partition.
  * Cleanup is safer: prior chunks are removed only after new chunks are successfully stored, avoiding the “empty window” case and using best-effort cleanup if snapshot/query/delete encounters errors.

* **Tests**
  * Added integration tests for repeated re-indexing, partition scoping, and re-indexing a single file.
  * Added unit tests validating insert-before-delete semantics and best-effort behavior when snapshot/query/delete fails.

### DIFF
--- a/openrag/api/routers/admin/indexing.py
+++ b/openrag/api/routers/admin/indexing.py
@@ -273,9 +273,10 @@ async def put_file(
             detail=f"'{file_id}' not found in partition '{partition}'",
         )
 
-    # No Milvus deletion here. The Indexer's add_file(replace=True) flow uses
-    # insert-before-delete: it snapshots old chunk IDs, inserts new chunks,
-    # then deletes old ones — so the file is never left in a half-replaced state.
+    # No Milvus deletion here. The indexing pipeline is insert-before-delete on
+    # replace: it snapshots the file's existing chunk ids, stores the new chunks,
+    # then deletes the old set (see ``IndexingPipeline.run``, #657) — so the file
+    # is never left in a half-replaced state.
     original_filename = file.filename
     file.filename = sanitize_filename(file.filename)
     file_path = await save_file_to_disk(file, Path(config.paths.data_dir), with_random_prefix=True)
--- a/openrag/api/routers/admin/indexing.py
+++ b/openrag/api/routers/admin/indexing.py
@@ -275,8 +275,10 @@ async def put_file(
 
     # No Milvus deletion here. The indexing pipeline is insert-before-delete on
     # replace: it snapshots the file's existing chunk ids, stores the new chunks,
-    # then deletes the old set (see ``IndexingPipeline.run``, #657) — so the file
-    # is never left in a half-replaced state.
+    # then deletes the old set (see ``IndexingPipeline.run``, #657). The guarantee
+    # is "no empty window" — a failed/crashed re-index keeps the old chunks rather
+    # than zero — not atomicity: a crash between store and delete can leave stale
+    # duplicates behind (recoverable; #658/#660 reconciliation covers that).
     original_filename = file.filename
     file.filename = sanitize_filename(file.filename)
     file_path = await save_file_to_disk(file, Path(config.paths.data_dir), with_random_prefix=True)
--- a/openrag/services/workers/pipeline_builder.py
+++ b/openrag/services/workers/pipeline_builder.py
@@ -187,7 +187,18 @@ class IndexingPipeline:
                     per_chunk_timeout=self.timeouts.store_per_chunk,
                 ),
             )
-            if replace and old_chunk_ids:
+            # BUG (#657 follow-up): ``store_stage`` completes successfully even
+            # when it stores zero chunks — an empty/whitespace-only file, a
+            # parser that extracts no text, etc. all legitimately chunk down to
+            # ``[]`` without raising (see chunk_stage / BaseChunker.chunk). If the
+            # delete below fired on ``old_chunk_ids`` alone, a re-index that
+            # produces no new chunks would delete the *entire* previous chunk set
+            # and leave the file with zero chunks in Milvus — worse than the
+            # pre-fix duplication bug, and a violation of the "no empty window"
+            # guarantee this whole insert-before-delete design is built on.
+            # Gating on ``stored_count`` ensures cleanup only runs once we know
+            # the new set actually replaced the old one.
+            if replace and old_chunk_ids and row.get("stored_count"):
                 await self._delete_replaced_chunks(row, old_chunk_ids)
             return row
         finally:
--- a/openrag/services/workers/pipeline_builder.py
+++ b/openrag/services/workers/pipeline_builder.py
@@ -15,6 +15,7 @@ from core.models.document import Document, DocumentType
 from core.utils.logging import get_logger
 from core.vector_stores.vector_store import VectorStore
 from core.vlm.vlm import VLM
+from services.workers.stages._common import run_with_optional_timeout
 from services.workers.stages.caption import caption_stage
 from services.workers.stages.chunk import chunk_stage
 from services.workers.stages.contextualize import contextualize_stage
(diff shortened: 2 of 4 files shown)


hedhoud (dismissed):
Looks good from my side. The replace flow is scoped to the file and partition, and the tests cover the main safety cases around duplicate chunks.

Small note before merge: the branch is still behind develop, so it should be updated once before landing.

Ahmath-Gadji (approved):
Reviewed the re-index insert-before-delete fix for #657 — solid design, correctly scoped by `{partition, file_id}`, and consistent with `WorkerDispatcher.delete_file`'s existing best-effort contract. The concurrency/crash-window seams are honestly flagged as known limitations tracked in #658/#660, which is the right call for a v2.0.1 patch.

Found and fixed one gap during review: the cleanup delete didn't check that `store_stage` actually stored new chunks, so a re-index producing zero chunks (blank file, parser extracting no text, etc.) would wipe the file's existing content instead of leaving it in place. Pushed 850a1cc0 to gate the delete on `stored_count`, plus a regression test (`test_reindex_with_zero_new_chunks_keeps_old_chunks`).

All CI checks are green (api-tests, tests, milvus-integration, lint, layer-import-guard). LGTM.

PR #673, opened 2026-07-15
fix(indexing): stop IMAGE_CAPTIONING env var from silently overriding partition presets

### LINKED ISSUE #672: IMAGE_CAPTIONING env flag silently overrides partition presets with no visibility
## Bug

`IndexingPipeline._should_caption` (`openrag/services/workers/pipeline_builder.py`) gates embedded-image captioning with:

```python
per_partition = config.enable_image_captioning if config is not None else True
return self.image_captioning and per_partition
```

`self.image_captioning` comes from the live `IMAGE_CAPTIONING` env var (`config.loader.image_captioning`), re-checked on every indexing run. If an operator sets `IMAGE_CAPTIONING=false` at the deployment level, it silently disables captioning for **every** partition, even ones whose preset explicitly has `enable_image_captioning: true` — with no error, no warning, and nothing visible via the preset API to explain why captioning isn't running.

This is inconsistent with how the two sibling toggles behave:

- `enable_contextualization` (`CONTEXTUAL_RETRIEVAL`) and `enable_reranker` (`reranker.enabled`) are folded into the `default` preset **once, at first-boot seed time** (`PresetService._finalize_seed`). After that, the preset alone is authoritative — flipping the env var later has no effect, and admins have full, visible control per partition.
- `enable_image_captioning` was hardcoded to `true` in every seed and instead re-gated against the live env var on every single indexing run, which meant an admin could never turn captioning on for a partition without also flipping the global env var (and vice versa, could never rely on the partition preset alone).

### Repro

1. Deploy with `IMAGE_CAPTIONING=false`.
2. Create/keep a partition preset with `enable_image_captioning: true`.
3. Upload a PDF with embedded images to that partition.
4. No captioning runs, no logs mention it (`vlm: null` in the "model endpoints resolved for indexing" debug log, no `ms_caption` entry in the stage-timings log) — nothing signals that the global env var, not the preset, is the reason.

### Fix

Fold `IMAGE_CAPTIONING` into the `default` indexation preset's `enable_image_captioning` at seed time (mirroring `enable_contextualization`/`enable_reranker`), and make `_should_caption` trust the preset alone at runtime. See PR.

Ahmath-Gadji:
## Summary

- `IndexingPipeline._should_caption` re-ANDed the live `IMAGE_CAPTIONING` env flag against each partition's preset on every indexing run, so a deployment-wide `IMAGE_CAPTIONING=false` silently disabled captioning for every partition — even ones whose preset explicitly set `enable_image_captioning: true` — with no visibility into why.
- This was inconsistent with `enable_contextualization`/`enable_reranker`, which fold their env toggle into the `default` preset once at first-boot seed time (`PresetService._finalize_seed`), after which the preset alone is authoritative.
- Fixes it by folding `IMAGE_CAPTIONING` into the `default` indexation preset's `enable_image_captioning` at seed time the same way, and making `_should_caption` trust the preset alone at runtime (standalone image files still always caption, unchanged).
  - **Breaking change / upgrade note:** `IMAGE_CAPTIONING` is now only consulted once, at first-boot seed time, to initialize the `default` preset's `enable_image_captioning`. After that, the preset alone controls captioning — the env var is no longer re-checked at runtime. If you have an existing deployment that currently relies on `IMAGE_CAPTIONING=false` to keep embedded-image captioning off, set `enable_image_captioning` to `false` on the affected preset(s) via the admin UI after upgrading; otherwise captioning will resume on the next indexing run. No automatic migration is provided, since all known major deployments run with image captioning enabled.

Fixes #672

## Test plan

- [x] `uv run pytest tests/unit/` — 1665 passed
- [x] `uv run ruff check` / `ruff format` — clean
- [x] Added `test_seed_defaults_default_indexation_inherits_image_captioning_{disabled,enabled}` mirroring the existing contextualization seed tests
- [x] Updated the two pipeline_builder tests that exercised the old pipeline-level global flag to instead exercise `IndexationPipelineConfig(enable_image_captioning=False)`

## Summary by CodeRabbit

* **New Features / Behavior Changes**
  * Image captioning is now governed per indexing preset (per partition) rather than a single global pipeline flag.
  * The `default` preset now inherits the global loader image-captioning setting at seeding time; the `legal` preset keeps its explicit value.
  * Standalone images captioning follows the selected preset setting; embedded-image captioning is gated by the partition’s configuration.
* **Tests**
  * Added and updated unit tests to cover default vs. legal presets and partition-specific captioning behavior.

### DIFF
--- a/openrag/services/workers/pipeline_builder.py
+++ b/openrag/services/workers/pipeline_builder.py
@@ -53,10 +53,6 @@ class IndexingPipeline:
     embedder: Embedder
     vector_store: VectorStore
     vlm: VLM | None = None
-    # Global gate for captioning images *embedded* in other documents
-    # (mirrors ``config.loader.image_captioning``). Standalone image files are
-    # always captioned when a VLM is available, regardless of this flag.
-    image_captioning: bool = True
     contextualizer: ChunkContextualizer | None = None
     topic_tagger: TopicTagger | None = None
     timeouts: PipelineTimeouts = PipelineTimeouts()
@@ -312,14 +308,17 @@ class IndexingPipeline:
 
         A standalone image file's caption is its only text content, so it is
         always captioned when a VLM is available (legacy ``ImageLoader``
-        parity). Images embedded in other documents are gated by the global
-        ``image_captioning`` flag and the per-partition setting.
+        parity). Images embedded in other documents are gated solely by the
+        per-partition ``enable_image_captioning`` setting — the deployment's
+        ``IMAGE_CAPTIONING`` env flag only seeds that setting's default on the
+        ``default`` preset at first boot (see ``PresetService._finalize_seed``);
+        it is not re-checked here, so a preset can enable/disable captioning
+        independent of the current env value.
         """
         document = row.get("document")
         if isinstance(document, Document) and document.content_type is DocumentType.IMAGE:
             return True
-        per_partition = config.enable_image_captioning if config is not None else True
-        return self.image_captioning and per_partition
+        return config.enable_image_captioning if config is not None else True
 
     def _select_contextualizer(
         self, config: IndexationPipelineConfig | None
@@ -367,7 +366,6 @@ def build_indexing_pipeline(
     embedder: Embedder,
     vector_store: VectorStore,
     vlm: VLM | None = None,
-    image_captioning: bool = True,
     contextualizer: ChunkContextualizer | None = None,
     topic_tagger: TopicTagger | None = None,
     timeouts: PipelineTimeouts | None = None,
@@ -387,7 +385,6 @@ def build_indexing_pipeline(
         embedder=embedder,
         vector_store=vector_store,
         vlm=vlm,
-        image_captioning=image_captioning,
         contextualizer=contextualizer,
         topic_tagger=topic_tagger,
         timeouts=timeouts or PipelineTimeouts(),
--- a/openrag/services/orchestrators/preset_service.py
+++ b/openrag/services/orchestrators/preset_service.py
@@ -34,12 +34,14 @@ _VALID_PRESET_TYPES = frozenset({"indexation", "retrieval"})
 _RESERVED_PRESET_NAME = "default"
 
 _DEFAULT_SEEDS: dict[str, dict[str, dict[str, Any]]] = {
-    # ``enable_contextualization`` and ``parsing_strategy`` are intentionally
-    # omitted from the ``default`` indexation preset below. The former is injected
-    # at seed time from the global ``chunker.contextual_retrieval``
-    # (``CONTEXTUAL_RETRIEVAL``) toggle in _finalize_seed; the latter stays unset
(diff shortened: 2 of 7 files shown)


openrag/services/workers/pipeline_builder.py
```diff
@@ -230,14 +226,17 @@ def _should_caption(self, row: MutableMapping[str, Any], config: IndexationPipel
...
+        ``default`` preset at first boot (see ``PresetService._finalize_seed``);
+        it is not re-checked here, so a preset can enable/disable captioning
+        independent of the current env value.
         """
         document = row.get("document")
         if isinstance(document, Document) and document.content_type is DocumentType.IMAGE:
             return True
-        per_partition = config.enable_image_captioning if config is not None else True
-        return self.image_captioning and per_partition
+        return config.enable_image_captioning if config is not None else True
```
hedhoud, line 321: One upgrade case worries me here. On a fresh install the default preset now gets the env value, but existing deployments already have a default preset row that was seeded with image captioning enabled. If they currently rely on `IMAGE_CAPTIONING=false` to keep embedded image captioning off, this line will start captioning after upgrade because the preset still says enabled. That can add unexpected VLM load/cost on the next indexing run. Could we add a one-time migration/backfill for the old untouched default preset, or at least make the upgrade path explicit so operators know they must flip the default preset before deploying this?

hedhoud (commented):
I found one upgrade edge case worth checking before merge. The fresh-install path looks covered and the focused tests are green.

openrag/services/workers/pipeline_builder.py
Ahmath-Gadji: Deployment shouldn'''t need to depend on `IMAGE_CAPTIONING` for this anymore. The env var is only a one-time seed for the `default` preset'''s `enable_image_captioning` at first boot — after that, the preset alone is authoritative, same contract as `enable_contextualization`/`enable_reranker`. If an operator wants captioning off, they toggle `enable_image_captioning` on the preset via the admin UI.

All of our major deployments currently run with image captioning enabled, so I don'''t think a backfill/migration script is warranted here for what is, as far as we know, a non-existent case (a deployment that intentionally set `IMAGE_CAPTIONING=false` and never touched the preset).

To make the upgrade path explicit for anyone who is in that situation, I'''ve labeled the PR `breaking-change` and added an upgrade note under the fix bullet in the description: if you rely on `IMAGE_CAPTIONING=false`, disable `enable_image_captioning` on the affected preset(s) via the admin UI after upgrading. That'''ll surface in the release notes.

hedhoud:
@codex review

hedhoud (approved):

hedhoud (approved):
Looks good to me. I checked the captioning flow from preset seeding through indexing, and the behavior now matches the preset model: IMAGE_CAPTIONING only seeds the default preset, then the preset controls runtime captioning. The upgrade note covers the one behavior change for existing deployments, and CI is green.

Issue #676, opened 2026-07-16
Reconcile orphaned in-flight jobs on startup + live-Ray e2e coverage for the durable job/quota paths


Ahmath-Gadji:
## Summary

The durable-job work (#660) and the atomic quota reserve (#664) both land with the same two known gaps, deliberately scoped out of each. Both are recorded here so they are not lost.

## 1. Startup reconciliation of orphaned in-flight jobs

#660 makes Postgres the source of truth for job state, but nothing reconciles rows left behind by a crash. A job row only leaves the active states (`QUEUED`/`SERIALIZING`/`CHUNKING`/`INSERTING`) when a worker writes a terminal transition, so a process death mid-flight leaves the row active forever — retention sweeps terminal rows only.

This has a concrete consequence for #664. The quota reserve charges `users.file_count` at admission; the request owns the slot until dispatch, the worker owns it after. Hard process death between those points (worker OOM, node crash, API crash between reserve and dispatch) releases neither, and the slot leaks. A leak is silent and permanently narrows the quota of the affected user — the failure mode the #664 design notes call out as the main risk. There is no durable record to reconcile against today, which is exactly what the `jobs` table introduced by #660 now provides.

**Suggested direction**

- On startup, sweep jobs still in an active state with no live Ray task and drive them to a terminal state (re-queue, fail, or clean up).
- Release the quota slot for any job reconciled to a non-`COMPLETED` terminal state, so `file_count` self-heals.
- Reconcile orphaned uploads in `paths.data_dir` for the same jobs (cleanup is currently tied to task completion).

Once this lands, the reserve/release lifecycle is closed on every path, including hard crashes.

## 2. Live-Ray end-to-end verification

The dispatch → worker → Postgres path in #660 is covered at the unit level and by repo-level integration tests against a real Postgres, but was never exercised through a live Ray cluster. The same applies to the worker-side quota release from #664, which runs inside `IndexerWorker.process_file` under Ray.

**Suggested direction**

- An end-to-end test (or a documented manual runbook) that dispatches a real file through a live Ray cluster and asserts:
  - the durable `jobs` row transitions `QUEUED → SERIALIZING → COMPLETED`;
  - a cancelled task lands `CANCELLED` and not `FAILED` (the `set_failed_if_not_cancelled` verdict is now honoured by the worker, and this is the path that proves it);
  - `users.file_count` is correct after success, failure, and cancellation.

## Related

- #660 — durable job state (the `jobs` table this reconciliation would sweep)
[...]

PR #677, opened 2026-07-16
fix: atomic file-quota reserve + durable indexation job state

### LINKED ISSUE #660: Persist indexing job state to Postgres; retire the in-memory-only TaskStateManager (unbounded, wiped on restart)
## Summary (expanded scope — P0 job durability)

Indexing job/task state lives **only** in the in-memory `TaskStateManager` Ray actor. It is unbounded (grows forever, no TTL/cap), volatile (wiped on any restart), and unpersisted (the Postgres-backed job repository is a stub with zero writers and no backing table). This makes in-flight batches unobservable and un-cancellable across restarts, orphans uploads on disk, and can leave Milvus/Postgres inconsistent with no reconciliation. This issue is the concrete realization of the already-agreed "persist indexation job history" P0.

> Scope note: this supersedes the original narrower framing of this issue (unbounded memory growth only). The eviction/TTL fix and the persistence fix are the same workstream — a TTL-only patch would be thrown away once persistence lands — so they are tracked together here.

### 1. Unbounded in-memory growth (the original report)
- `TaskStateManager.tasks` / `user_index` (`openrag/services/workers/task_state.py:36-37`) are insert-only — no `pop`/`del`/TTL/cap anywhere in the class.
- Each dispatched file leaves a permanent `TaskInfo`; on failure it also stores a full traceback string (`set_failed_if_not_cancelled`, `task_state.py:64-65`).
- The actor is `lifetime="detached"` (`openrag/services/workers/bootstrap.py:69`), so it persists across API restarts and grows until OOM.

### 2. No persistence — `job_repo` is a stub with zero writers
- `PgJobRepository` (`openrag/services/persistence/job_repo.py`) is an explicit **stub**: every method (`create_job`, `update_job`, `get_job`, `list_jobs`) raises `stub_not_implemented`. There is **no `jobs` table** in `schema.py`.
- The ingest path never references it — job state is written exclusively to the in-memory actor (no `job_repo` reference anywhere under `services/workers/` or `indexing_service.py`).

### 3. Restart consequences
On any restart mid-batch:
- The entire in-flight batch becomes **unobservable** (status polling returns nothing) and **un-cancellable** (object refs are gone).
- Uploads **orphan** on disk (`data_dir`), since cleanup is tied to task completion.
- Partially-indexed files can leave Milvus and Postgres **inconsistent**, with no reconciliation job (see the Milvus↔Postgres lifecycle epic).

## Suggested direction

- Implement the real `PgJobRepository` against a new `jobs` table (idempotent migration per repo conventions); write job lifecycle transitions (`QUEUED → … → COMPLETED/FAILED/CANCELLED`) durably.
- Keep the in-memory actor as a hot cache if useful, but make Postgres the source of truth so state survives restarts and is operator-visible.
- Bound retention (evict/archive terminal jobs) so the durable store doesn't grow unbounded either; bound stored traceback length.
- On startup, reconcile in-flight/orphaned jobs (re-queue, fail, or clean up) so a restart mid-batch is recoverable.

## Related

- Enables correct cross-restart cancellation and observability that admission control and quota integrity also rely on (job-durability epic).
- The Milvus/Postgres inconsistency facet is shared with the lifecycle-consistency epic.

### LINKED ISSUE #664: File-quota check/admit is a TOCTOU race (quota bypass under concurrent uploads)
## Summary

The per-user file-quota enforcement is a check/admit TOCTOU race. The upload dependency reads the user's pending-task count from the in-memory `TaskStateManager` and the durable `users.file_count` from the request's user snapshot, decides admission, and only increments the durable `file_count` at the **end** of indexing. Concurrent uploads all observe the same pre-increment state and all pass → quota bypass under burst. A restart additionally zeroes the in-memory pending count, reopening the gate.

## Severity: medium — quota bypass under concurrency

The SQL increment itself is atomic; the defect is at **check/admit**, not the arithmetic.

## Root cause / code trace

- `check_user_file_quota` (`openrag/api/dependencies/auth.py:257-300`) computes `pending_count = await job_service.get_user_pending_task_count(user_id)` — served from the **in-memory** `TaskStateManager.get_user_pending_task_count` (`openrag/services/workers/task_state.py:158-162`, reads `self.user_index` / `self.tasks`).
- It then calls `auth_service.validate_file_quota(user, pending_task_count=pending_count, ...)`, where the user's durable `file_count` is the snapshot carried on `request.state.user` from auth middleware.
- The durable `users.file_count` is only incremented at the **end** of indexing, in `add_file_to_partition` (`openrag/services/persistence/document_repo.py:349`, `file_count = file_count + 1`).

So between the check and the increment there is a wide window. N concurrent `POST /indexer/partition/{p}/file/{id}` requests each read the same `file_count` + pending and all admit.

## Failure scenarios

1. **Burst bypass:** a user at quota `N` fires `K` uploads in parallel; all `K` read the pre-increment count and pass, overshooting the quota by roughly the concurrency width.
2. **Restart reopens the gate:** the in-memory `TaskStateManager` is wiped on restart, so `pending_count` returns `0` for everyone until durable counts catch up — and tasks lost mid-flight never increment `file_count`, so the gate stays wider than reality.

## Suggested fix

- Make admission an **atomic DB-side reserve** rather than a read-then-check: e.g. `UPDATE users SET file_count = file_count + 1 WHERE id = :id AND (file_quota < 0 OR file_count < effective_quota) RETURNING file_count`, and reject with `429`/`403` when no row is returned. Release (decrement) on indexing failure/cancellation.
- This removes the dependency on the volatile in-memory pending count as a correctness input.

## Dependency note

This can and should be fixed **independently** of the job-durability epic — the atomic-reserve approach does not require job persistence. It is grouped under the same epic only because it shares the volatile-in-memory-state root cause; do not defer it behind the persistence work.

Ahmath-Gadji:
Two related fixes that share a root cause — indexing state living only in the in-memory `TaskStateManager` Ray actor — plus the follow-ups from review. The first two commits are the substance, quota first, then durability; the rest are fixes found reviewing them.

Closes #664
Closes #660

> **Rebased onto `develop` @ d450f149**, which includes #685 (bulk cancellation) and #671 (milvus cleanup before catalog delete). Both rewrote paths this PR also rewrites, so each merge needed its own review: the #685 interaction is described in §13–15, and the #671 interaction — 14 overlapping files, including one collision that would have broken every `replace` re-index — in §16. A finding from the #685 review is split out into #700, stacked on this branch.

---

## 1. `fix(quota)`: reserve file slots atomically at admission (#664)

The per-user file quota was a check/admit TOCTOU race: `check_user_file_quota` compared a stale `users.file_count` snapshot plus an in-memory pending count against the quota, and the durable count was only incremented at the *end* of indexing. Concurrent uploads all observed the same pre-increment state and all passed; a restart zeroed the pending count and reopened the gate.

Admission is now a single conditional UPDATE (`try_reserve_file_slot`), so the read and write happen under one row lock:

```sql
UPDATE users SET file_count = file_count + 1
 WHERE id = $1
   AND (is_admin
        OR COALESCE(file_quota, $2::int) < 0
        OR file_count < COALESCE(file_quota, $2::int))
RETURNING file_count
```

The predicate reproduces the documented resolved-quota semantics exactly. Note the `COALESCE` resolves the per-user quota against the global default **before** the `< 0` test — the naive `file_quota < 0` from the issue body would let a negative global default silently override an explicit per-user limit. No row returned means no slot.

`file_count` therefore becomes a **reserved + completed** counter, so the completion-time increment in `add_file_to_partition` is removed (it would double-count), and every reservation now needs an owner:

- **Before dispatch the request owns it.** `check_user_file_quota` is a yield dependency whose teardown releases anything uncommitted — covering the 409 duplicate, upload rejection, workspace validation, dispatch errors, and client disconnects.
- **After dispatch the worker owns it.** `IndexerWorker.process_file` releases in a `finally` — deliberately not an `except`, which cancellation would skip, since `ray.cancel` raises `CancelledError` (a `BaseException`). This covers indexing failure, cancellation, and the duplicate-at-catalog race.
- The MCP tools (`index_url`, `copy_file`) have no dependency chain, so they reserve inline. `put_file` (replace re-index) never reserves; it reuses an existing row.

The in-memory pending count is no longer a correctness input anywhere, and `get_current_user_info` stops adding it to `file_count` — reserved uploads are already counted, so that double-counted in-flight files.

**Status code is 403**, not the 429 the issue floated: `validate_file_quota` already raised `FILE_QUOTA_EXCEEDED` as 403 and the API tests assert it. Easy to flip if reviewers prefer 429.

**Admins and unlimited users are incremented too.** Removing the completion-time increment while reserving only for capped users would have frozen every admin's `file_count` at zero, and `/users/info` displays it. This keeps release symmetric for all callers.

## 2. `feat(jobs)`: persist indexation job state in Postgres (#660)

Job state lived only in the in-memory actor: unbounded (insert-only, no TTL/cap), volatile (wiped on restart), and unpersisted (`PgJobRepository` was a stub whose every method raised, with no `jobs` table and zero writers).

- **Migration `d4e5f6a7b8c9`** adds a `jobs` table (one row per dispatched task, `id` = `task_id`, status CHECK over the 7 states, three indexes). Guarded with `table_exists`/`index_exists` in both directions, because `metadata.create_all()` runs at startup and an unguarded CREATE would raise `DuplicateTable`. Constraints stay unnamed so the migration and `create_all` converge on the same Postgres-default names. No FK to `partitions` — a job row is a historical record and must outlive the partition it targeted.
- **Real `PgJobRepository`** replaces the stub. `IndexationJob` becomes a per-task record reusing `DocumentStatus` (matches the actor's taxonomy verbatim; `JobStatus` lacked SERIALIZING/CHUNKING/INSERTING/CANCELLED).
- **Lifecycle writes**: `QUEUED` from the dispatcher *before* submit (so a crash between submit and the worker's first write still leaves the job visible rather than silently in-flight), `SERIALIZING`/`COMPLETED`/`FAILED` from the worker, `CANCELLED` from `cancel_task`. All durable writes are best-effort by design — this is bookkeeping about the work, not the work; a Postgres blip must not fail a file that indexed correctly, nor mask the real exception on the failure path.
- **Reads** hit Postgres and fall back to the actor — on a failure *or* on an empty result (see §6). This is required, not optional: evicting terminal entries from the actor without it would make completed work vanish from the queue views.
- **Bounding**, both sides: terminal jobs swept by age (7d) + row cap (10k), throttled to once/5min and run inline from the dispatch path (no new background task to own across replicas); the actor evicts terminal tasks by cap (2000) + TTL (1h), never in-flight ones — their `object_ref` is not serializable. Stored tracebacks are truncated (4000 chars, tail kept).

**Incidental fix:** the worker was discarding `set_failed_if_not_cancelled`'s return value. It now honours the verdict, so a durable `FAILED` cannot overwrite a cancellation the user already asked for and already saw.

### On the in-memory pending count

`get_user_pending_task_count` was intentionally **not** routed through Postgres. A job row only leaves the active states when a worker writes a terminal transition, so a job orphaned by a crash would occupy the quota of that user *permanently* (retention sweeps terminal rows only) — the in-memory count is wrong in the opposite, self-healing direction. §1 makes this moot by removing the pending count from the quota decision entirely, so the combined design has no reader of it as a correctness input.

## 3–12. Review follow-ups

Found reviewing the two commits above; each is standalone and independently reviewable.

- **`fix(quota)`: release the reserved slot when the SERIALIZING write fails.** The pre-flight `set_state` ran *outside* the try whose `finally` releases, so an unreachable state actor leaked the slot permanently — the request had already handed ownership over at dispatch.
- **`fix(quota)`: reserve for the MCP `index_url` / `copy_file` tools.** They create `files` rows without passing through the dependency, so removing the completion-time increment made MCP files invisible on the way in but still counted on the way out — driving `file_count` *below* reality and handing back free slots on a repeatable loop. Measured at quota 5: upload 5 via HTTP, index 3 via MCP, delete those 3 → the user holds 5 files with 3 free slots.
- **`fix(jobs)`: fail closed when listing tasks for an anonymous non-admin.** `list_jobs(user_id=None)` means "every job". Not reachable today, but the MCP `_USER_ID` ContextVar defaults to `None`.
- **`fix(jobs)`: treat an empty durable read as a miss, not an empty queue.** The fallback only triggered when the repo *raised* — `[]`/`{}` are not `None`, so zero rows read as authoritative. The actor is detached and survives the API restart that first deploys this, so every pre-cutover task is live in the cache and absent from `jobs`: `/queue/tasks` answered `[]` and `/queue/info` reported 0 active while workers were indexing. Both aggregate readers now fall through on empty; `get_task_details` keeps `is not None`, since for a single row `None` already is the miss.
- **`fix(workers)`: drop a late write instead of resurrecting an evicted task.** `set_error`/`set_details`/`set_object_ref` went through `_ensure_task`, which recreates on miss; the recreated entry has `state=None`, never re-enters `terminal_at`, and is never evictable again — the unbounded growth #660 exists to fix. Not reachable today (the dispatcher writes `set_state` first), but silent and permanent if it ever is.
- **`fix(jobs)`: upper-case a job status before it reaches the CHECK.** `_expand_status` normalized case, `_status_value` did not; a lower-case status would violate `ck_jobs_status`, and best-effort writes would swallow it, freezing the job at its previous status.
- **`refactor(workers)`: use the project logger in `indexer_actor`.**
- **`test(workers)`: pin the retention bounds and prove the throttle recovers.** The shipped 7d/10k were asserted nowhere, and the throttle test also passed for an implementation that purges once per process and never again.
- **`test(repos)`: revive the delete-cascade `file_count` assertion.** Dead on develop (`create_legacy_user` signature drift), and it guards the very counter this branch redefines. Fixing the signature alone would have made it pass *vacuously* (0 → 0) now that the insert no longer increments, so it reserves the slots at admission the way a real upload does and asserts 2 → 0.
- **`test(quota)`: pin that dispatch tells the worker the slot is reserved.** The release paths were covered, but nothing asserted the signal that arms them: `commit_quota_reservation` only stops the request's teardown from releasing, while `quota_reserved=True` is what makes the worker take ownership. Dropping that keyword from the router leaves the rest of the unit suite green — teardown declines to release (committed) and the worker declines too (never told it owns one), so every failed upload would leak a slot in silence. It now fails exactly one test.
- **`docs`: stop two comments overstating their guarantees** — the setup-failure release in `indexer_pool` releases through the store whose init just failed; `truncate_error_text` bounds the retained text, not the returned string.
- **`fix(jobs)`: stop `list_tasks` truncating the queue silently.** `_LIST_LIMIT` caps the page at 1000 against a retention cap of 10k, so the cap is reachable and the route returned a short list with nothing to say it was short. It now reads one row past the limit to tell an exactly-full page from a truncated one, drops the probe, and logs a warning. Real pagination is the proper fix and is left as follow-up.

## 13–15. Found reviewing the rebase onto #685

#685 landed a bulk-cancel UI and rewrote `cancel_task` to claim the cancellation *before* signalling `ray.cancel`. That is the same code path this branch makes durable, so the merge needed its own review.

- **`fix(workers)`: mark a claimed cancellation terminal so it can be evicted.** `set_cancelled_if_active` (from #685) is the only state write that does not call `_mark_terminal` (this branch's eviction hook). Eviction is driven entirely off `terminal_at`, and nothing writes that task's state again — `ray.cancel` raises `CancelledError`, a `BaseException` that `process_file`'s `except Exception` sails past. So every user-initiated cancel retained a `TaskInfo`, its `user_index` entry and its pinned `object_ref` for the lifetime of the detached actor: verbatim the unbounded growth #660 exists to fix, and reachable in bulk from #685's cancel-all button. Neither change is wrong alone; the gap is the merge. `test_cancelled_tasks_are_evictable` missed it because it drives `set_state(..., "CANCELLED")`, a path no cancellation actually takes.
- **`fix(jobs)`: keep a cancellation sticky in the durable row too.** `mark_failed_if_not_cancelled` arbitrates FAILED-vs-CANCELLED in SQL, but the worker's SERIALIZING/COMPLETED writes were blind UPDATEs racing the cancel's blind UPDATE — so only the direction where the cancel *loses* was covered. When the cancel wins: COMPLETED-over-CANCELLED leaves the actor and the table disagreeing permanently (and the two read paths answer differently for the same task); SERIALIZING-over-CANCELLED is worse, stranding the row non-terminal after `ray.cancel` has killed the only writer that could finish it — and retention sweeps terminal rows only, so it never ages out. `update_job` now carries the same guard, mirroring the stickiness `TaskStateManager.set_state` already enforces in memory.
- **`docs(quota)`: name the residual ambiguity at the submit boundary.** A `submit` timeout leaves ownership of the reserved slot genuinely undecidable — the worker may or may not have started. Teardown releases, which is right for the common case and off by one the other way; both directions undercount rather than leak.

A fourth finding from the same review — the **cancel-before-start slot leak**, which #685's cancel-all makes reachable at scale — is fixed in **#700**, stacked on this branch. It is a change to the ownership model rather than a bug fix, so it is kept separate and reviewable on its own.

## Verification

Re-run on the rebased branch:

- `ruff check` / `ruff format --check` clean; `scripts/check_layer_imports.py` → `layer import guard: OK`.
- **Unit: 1914 passed, 0 failed.**
- **Integration against real Postgres: 120 passed, 1 failed.** The 1 (`test_create_is_idempotent_per_name` — `create_partition` raises `PARTITION_EXISTS` where the test expects the conflict swallowed) is pre-existing, confirmed identical on clean `origin/develop` with the same DSN, and fixed separately in #691.
- **On the "6 failed" reported in the previous revision:** those were an artefact of the *checkout*, not the code. `load_dotenv()` (`core/config/loader.py:271`) is called with no argument, so it walks up from the CWD and takes the closest `.env`; a checkout without one of its own inherits a parent directory's, and an empty `RERANKER_PROVIDER=` there breaks a pydantic discriminator. A checkout with its own `.env` shadows it and is fully green. Unrelated to this PR either way, but it means test runs are not hermetic depending on checkout location — worth a separate look.
- The quota concurrency claim is measured, not asserted: against a real Postgres, 20 parallel admits at quota 1 and quota 5 grant exactly 1 and exactly 5, with contiguous returned counts. The regression test was checked against a read-then-check reimplementation of the reserve — it overshoots the quota by a nondeterministic margin, observed between 14 and all 20 of the 20 concurrent admits across runs and machines — i.e. the test reliably fails on the bug it guards, though not at a fixed count.
- Migration verified end-to-end against real Postgres: `create_all` → `upgrade head` does not raise, re-upgrade is a no-op, downgrade drops and re-upgrade re-creates, and the two schemas are byte-identical across columns, PK, FK, indexes and check constraints. Names converge on `jobs_pkey` / `jobs_user_id_fkey`. Single head (`d4e5f6a7b8c9`).
- Two guards were mutation-checked rather than assumed: dropping `quota_reserved=True` from the router, and dropping the status upper-casing in `_status_value`, each fail exactly one test.

## Known gaps — tracked in #676

- **No startup reconciliation.** A hard process death between reserve and dispatch leaks a quota slot, and an orphaned job row stays active forever. There is now a durable record to reconcile against; the sweep itself is follow-up work.
  - Retention only sweeps *terminal* rows, so an orphaned row is never purged; and `cancel_task` returns early when the actor no longer holds an `object_ref`, so an operator cannot force one terminal by hand either. The `active` roll-up in `/queue/info` therefore drifts upward across such events until reconciliation lands.
- **A cancel of a not-yet-started task leaks its reserved quota slot** — fixed in **#700**, stacked on this branch, and left out of this one so the ownership-model change stays separately reviewable. After dispatch the worker owns the slot and only releases it in `process_file`'s `finally`, so if `ray.cancel` retires the task before that body runs the release never happens — and `cancel_task` cannot release itself without double-releasing a task cancelled mid-flight (whose `finally` does run). Because the cancel writes a *terminal* `CANCELLED` row, indistinguishable from a clean cancel, recovering the slot on this branch alone needs the #676 reconciliation to **recount** `file_count` (completed files + active job rows), not merely sweep orphaned *active* rows.
  - Worth weighing when reviewing: #685's cancel-all fires one cancel per selected task, so on this branch the leak scales with a single click.
- **A dispatch that partially succeeds** (pool accepts the job, `set_object_ref` then raises) double-releases the slot: the worker consumes it *and* teardown releases it. Chosen deliberately here — under-counting is recoverable, whereas a leak permanently locks the user out. Also closed by #700.
- **No live-Ray e2e.** The dispatch → worker → Postgres path is proven at unit and repo-integration level only.

## Review notes

- `JobService.list_tasks` caps at 1000 rows and now says so when it truncates (§12). Filtering happens in SQL before the cap, so the cap cannot leak another user's jobs.
- `JobStatus` is now unused; left in place rather than change a public model export. `IndexationJob.total_documents` *was* dropped — it had no readers and the model is internal (it appears in no API response).
- The actor's TTL is swept lazily, only when some task settles — an idle deployment keeps its last terminal entries cached past the hour. Harmless (a terminal state is immutable, so a stale read is not a wrong read) and the cap is the real memory bound; noted because the comment used to claim otherwise.

---

## 16. Found reviewing the rebase onto #671

Rebased onto `develop` @ `d450f149`, which includes **#671** (`fix/658-milvus-cleanup-before-catalog-delete`). #671 rewrote the indexing dispatch, cancellation and delete paths that this branch also rewrites — 14 files overlap — so, as with #685 in §13–15, the merge needed its own review rather than a textual resolution. Seven of the 33 commits conflicted; the decisions that were not mechanical are below.

The single most important one is §16.1: a naive merge is not merely untidy, it breaks every `replace` re-index.

### 16.1 `_write_catalog_record` — two incompatible meanings for one bool

The two branches gave the same return value different meanings:

- **This branch**: *"was a **new** row created?"* — deliberately `False` on a `replace`, because that is the verdict that decides whether the reserved quota slot was consumed (#664).
- **#671**: *"did the catalog write land?"* — `True` on either branch, because it added a fail-closed `if not wrote_catalog: raise RuntimeError("Catalog row was not written after vector indexing")` on that value, so a `require_existing_partition` rejection surfaces instead of silently no-opping.

Merged textually, `replace` keeps returning `False` and #671's new guard turns **every re-index into a hard failure**. Nothing in either test suite catches it: this branch's tests assert the return value directly, and #671's exercise the insert path.

The two questions are now separate. `_write_catalog_record` reports whether the write landed (#671's meaning, so the fail-closed guard is correct), and the quota verdict is derived at the call site from `replace`:

```python
if not replace:
    release_slot = False
```

The duplicate-at-catalog race — the other case that used to return `False` — now reaches the same outcome by a different route: `add_file_to_partition` returns `False` for the loser, the fail-closed guard raises, and `process_file`'s `finally` releases the slot on the way out. `test_duplicate_at_catalog_releases_the_slot` was updated to assert both halves (it raises **and** the slot comes back) so neither can regress silently, and `test_a_replace_reindex_does_not_consume_the_reservation` was rewritten against the new contract as `test_a_replace_reindex_updates_the_row_without_creating_one`.

Verified live, not just in tests: a `PUT` re-index over an existing file returns `202` and settles `COMPLETED`, with `users.file_count` unchanged.

### 16.2 `4ec0a634` is obsoleted by #671, and its test contradicted the new behaviour

`fix(quota): don't report a dispatch failure once the worker has started` argued that once `submit` returns, the worker owns the reserved slot and will run to completion — so a failing `set_object_ref` must not fail the dispatch, or the router skips `commit_quota_reservation` and teardown releases a slot the worker also owns.

**#671 invalidated the premise.** `dispatch_indexing` now rolls the dispatch back: `_cancel_submitted_task` cancels the worker and `_cleanup_submitted_vectors` sweeps what it wrote. The worker no longer runs to completion, so the error is truthful and must propagate — and #671's `test_dispatch_indexing_cancels_worker_when_ref_registration_fails` asserts exactly that.

Kept #671's behaviour and removed this branch's `test_dispatch_does_not_fail_once_the_worker_has_started`, which asserted the opposite of a test already on `develop`. The residual cost is a **double release** (the cancelled worker's `finally` and the request teardown both give the slot back), which under-counts rather than leaks — the direction this branch already chose deliberately (see *Known gaps*), and which #700 closes.

Worth noting the two signals from that one call are genuinely different and only one may fail the dispatch: a raised exception means the state actor is unreachable, while `False` is #671's deliberate delete-fence refusal. They happen to want the same handling here only because #671 rolls back in both cases.

### 16.3 The durable QUEUED write moves behind #671's admission gate

`set_state(QUEUED)` + `set_details(...)` were collapsed into a single atomic `_set_queued_details(...) -> bool`, which **can now refuse**: it returns `False` when a delete fence covers the file, and forces the in-memory state to `CANCELLED`.

The durable `create_job(QUEUED)` therefore moved *after* that gate. Writing it before would leave a row for a job that was never admitted — and since no worker ever runs, nothing settles it: non-terminal forever (retention sweeps terminal rows only) and counted active in `/queue/info` for good. Exactly the orphan class §16.5 exists to close, introduced by the rebase rather than fixed by it.

### 16.4 `quota_reserved` threads through #671's `submit_kwargs`, unconditionally

#671 hoisted the nine inline `submit.remote(...)` keywords into a `submit_kwargs` dict, and adds `require_existing_partition` to it **only when true** so a rolling-deploy actor without the kwarg still accepts the common case.

`quota_reserved` is passed unconditionally, like the nine base kwargs. The conditional pattern exists for rolling-deploy tolerance, and `_is_legacy_require_existing_partition_rejection` is hard-coded to that one kwarg name; generalising it for a flag that ships in the same release would be speculative. Flagging the asymmetry since it is deliberate: an actor old enough to reject `quota_reserved` is old enough to predate the whole quota model.

### 16.5 The dispatch-orphan fix lands in #671's failure path, not around `submit`

The fix from §16's sibling commit (`settle a task that never reaches a worker`) originally wrapped the `submit` call. On `develop` that call lives in `_submit_indexing_task_once`, which `_submit_indexing_task` **retries** for legacy actors — settling there would write a spurious `FAILED` for an attempt that is about to be retried successfully.

It moved into `_mark_submit_failed`, which #671 already calls on the failure path and which is gated on `_cancel_submitted_task` confirming the task really was cancelled. So the durable `FAILED` inherits that gate and a task that actually completed is never recorded failed in either store.

One improvement kept from this branch: #671's `_mark_submit_failed` drove its actor write through an unguarded `self._call`, so an unreachable state actor there replaced the exception the caller needs (*"the pool is down"*) with one about the bookkeeping (*"the actor is gone"*). It is now guarded, matching how every other durable write on this branch behaves.

### 16.6 `_live_task` composes with #671's `_record_details`

`b7e0b966` stops `set_error` / `set_details` / `set_object_ref` recreating an evicted task, while #671 refactored `set_details` onto a shared `_record_details` helper. Merged so `set_details` drops the write on a miss *and* still uses the shared helper. `_ensure_task` now has exactly two callers — `set_state` and #671's `set_queued_details` — and both are legitimate creation points for a task admission.

### 16.7 `copy_file` keeps both changes

#671 added `strip_internal_metadata` so `_openrag*` keys cannot leak into copied chunks; this branch made the return value the quota-commit verdict. Both kept. The empty-source path is an explicit `return False` rather than the bare `return` #671 left — under a `-> bool` signature that returned `None`, which is accidentally falsy but not something to rely on.

### Not addressed — a gap this rebase opens

**Cancellations from the delete path write no durable row.** #671 added `task_cancellation.cancel_active_indexing_tasks`, which `delete_file` and `delete_partition` call to cancel every active task for a file or partition; it writes `set_state(task_id, "CANCELLED")` directly to the actor. That path did not exist at the merge base, so this branch does not mirror it to Postgres — a delete now silently cancels tasks with no durable record, and the two read paths disagree for those task ids until the actor evicts them.

Left out deliberately: it is a new writer to cover rather than a merge conflict to resolve, and `_cancel_refs` writes its state *after* `ray.cancel` and a settle-wait, which is the opposite ordering from `cancel_task` (§14). Making that path durable should be a deliberate decision about which ordering is right, not a rebase artefact. **Tracked in #676** alongside the other reconciliation work.

### Verification after the rebase

- `ruff check` / `ruff format --check` clean; `layer import guard: OK`.
- **Unit: 1914 passed, 0 failed.**
- **Integration against real Postgres: 120 passed, 1 failed** — `test_create_is_idempotent_per_name`, pre-existing and unrelated, fixed separately in #691.
- Re-ran the live stack on the rebased code: clean boot, durable job history survived an API restart, a new index settles `COMPLETED`, a `replace` re-index settles `COMPLETED` without consuming a slot (§16.1), and a cancel writes a terminal `CANCELLED` and returns the slot.

## Summary by CodeRabbit

* **New Features**
  * Added durable indexation job history with lifecycle statuses, error details, filtering, queue counts, cancellation tracking, and retention.
  * Job status and details remain available after in-memory task data is evicted or services restart.
  * Added atomic file-quota reservations for uploads, URL indexing, and file copies.

* **Bug Fixes**
  * Prevented concurrent uploads from exceeding quotas or double-counting in-progress files.
  * Ensured reservations are released after failed, cancelled, duplicate, or incomplete operations.
  * Limited stored error details to prevent excessive diagnostic data growth.

* **Documentation**
  * Clarified quota usage, reservation ownership, and release behavior.


EnjoyBacon7 (commented):
## Review: two bugs fixed on the branch, two design gaps left for you

Reviewed both commits, ran the suites, and verified the claims in the description rather than taking them on trust. The core of this is solid — see the end for what I checked and confirmed. Four commits pushed to the branch (`704d00f6..dcffc1fd`); two findings below are **not** fixed because they're your design calls, not mine.

---

### Fixed: MCP paths bypass the quota entirely (`00da466e`)

This is the one I'd have blocked a merge on.

Removing the completion-time increment assumes every new `files` row is preceded by a reservation. `MCPService.index_url` (`mcp_service.py:623`) and `MCPService.copy_file` (`mcp_service.py:555`) create rows without one — they aren't HTTP routes, so they never pass through the `check_user_file_quota` dependency.

It's worse than an unenforced quota, because the delete path still decrements unconditionally on `created_by` (`document_repo.py:201`, `:225`, `:374`). MCP files are invisible going *in* but counted going *out*, driving the counter below reality. Measured against a real Postgres, at `file_quota = 5`:

```
upload 5 via HTTP   -> file_count=5   (correctly at quota, 6th rejected)
index 3 via MCP     -> file_count=5   (3 real files invisible to the quota)
delete those 3      -> file_count=2   (count driven below reality)
=> user holds 5 files, has 3 free slots again. Repeatable => quota unbounded.
```

`GREATEST(..., 0)` doesn't save this — the count is driven *down*, not negative.

Both tools now reserve at admission and release anything unconsumed. `index_url` reserves before the download so an over-quota call costs no bandwidth; `copy_file` is synchronous, so its slot settles on the `created` verdict. I made `auth_service`/`default_file_quota` **required** constructor args rather than optional — a missed wiring would silently reopen exactly this hole.

Note this is the failure mode your own `CLAUDE.md` warns about ("any new code path that *creates a file row without reserving* … leaks the counter"), so I extended that section to name the MCP tools.

### Fixed: quota slot leaks when the SERIALIZING write fails (`7d0dd791`)

`process_file` takes ownership of the slot at dispatch, but `set_state.remote(task_id, "SERIALIZING")` ran *before* the `try` whose `finally` releases it. `TaskStateManager` is a detached actor that can be unreachable (restart, node loss), so that await can raise — and nothing else covers the window:

- the request already handed ownership over (`commit_quota_reservation` runs once dispatch returns a task id), so its teardown won't release;
- `IndexerWorkerActor.process_file`'s `except BaseException` only wraps the catalog/registry setup, not the worker call.

Result: the slot leaks permanently and narrows that user's quota until an admin fixes `file_count` by hand. Moved the pre-flight writes inside the `try`; regression test added. Your pool comment already names this exact window for the file-purge concern — the release just hadn't been extended to it.

---

### Not fixed — your call

**1. The durable `FAILED` write is gated on the volatile actor's verdict** (`indexer_actor.py`, the `if failed:` gate)

`set_failed_if_not_cancelled` returns `False` for **two** different reasons (`task_state.py:126-133`): the task was cancelled (correct to skip), **or `info is None`** — the actor has no record of it. So if the actor is restarted or has evicted the entry between `SERIALIZING` and the exception, no durable `FAILED` is written and the row is stranded in `SERIALIZING` forever.

This inverts the stated design: the source of truth defers to the cache for a decision the cache is allowed to forget.

The clean fix is to move the arbitration into SQL rather than distinguish the two in the actor's return value — make the durable transition conditional:

```sql
UPDATE jobs SET status='FAILED', error=$2, completed_at=$3, updated_at=now()
 WHERE id=$1 AND status <> 'CANCELLED'
```

Then Postgres arbitrates the durable row and the actor arbitrates only its own cache, which is what "Postgres is the source of truth" should mean. I left this alone because `test_failing_after_eviction_does_not_resurrect_the_task` shows the no-resurrect behaviour is deliberate, and reworking your failure arbitration felt like your decision rather than a review fix.

**2. Jobs stuck in active states are never purged** — partly #676, but with a consequence worth calling out

`purge_terminal_jobs` only matches terminal states, and every durable write is best-effort (`_record_job`, `_update_job` both swallow). So a worker OOM, a lost node, or a Postgres blip at the terminal write leaves a row no retention pass can ever reclaim — contradicting the module docstring's "bounded rows … otherwise the durable store would just reproduce the in-memory leak on disk".

The extra consequence beyond #676: it permanently inflates `get_queue_info`'s `active` count, which reads `count_by_status` from Postgres. This is precisely the hazard you documented as the reason to keep `get_user_pending_task_count` on the actor (`job_service.py:171-179`) — the same reasoning applies to `get_queue_info` but wasn't applied there. Worth either an age-based sweep for active rows or a note in #676.

**Minor:** `_LIST_LIMIT = 1000` silently truncates an unpaginated route whose retention cap is 10k — no indicator to the caller that rows were dropped.

---

### Also pushed

- `dcffc1fd` — `list_tasks` fails closed for an anonymous non-admin. `list_jobs(user_id=None)` means *every* job, and there's no guard for `is_admin=False, user_id=None`. Unreachable today (the HTTP path always resolves an id, MCP 403s first), but the MCP `_USER_ID` ContextVar's unset default is exactly `None` — the escalating value. One middleware-ordering change from a data leak.
- `704d00f6` — logging/naming consistency in `indexer_actor`: the two new helpers each invented an idiom (function-local `get_logger` import; stdlib `logging.getLogger(__name__)` with %s-formatting) where every other module under `services/workers/` binds `logger = get_logger()` at import. Also dropped the underscore from `release_quota_slot`, since `indexer_pool` imports it — it's module API, and its neighbour `delete_uploaded_file` is already public.

### Claims I verified

- **The 6 unit / 2 integration failures are genuinely pre-existing and environmental.** Confirmed both directions: the same 6 fail on a clean `origin/develop` checked out *inside* a directory whose parent holds a `.env`, and all pass on the same commit checked out *outside* it. Your `load_dotenv()` diagnosis is exactly right.
- **The atomicity claim holds under harsher conditions than the test asserts.** Re-derived independently with 50 concurrent racers on separate pool connections: quota 7 → exactly 7 granted, quota 1 → 1, quota 0 → 0, admin → all 50, unlimited → all 50, returned counts contiguous every time. The `COALESCE`-before-`< 0` ordering is right, and the point about a negative global default not overriding an explicit per-user limit is a good catch.
- Migration chain: `b7c1d2e3f4a5` really is the current head, single head after this revision, up/down idempotent, `create_all` → `upgrade head` converges.
- `update_job`'s `enumerate(..., start=2)` parameter numbering is correct — dict ordering keeps assignments and `updates.values()` aligned.
- `TaskStateManager` eviction is sound under stress (50k tasks/100 users → caps hold at 2000, no dangling `user_index`, in-flight never evicted).

Final state: **1735 unit passed** (+10 from the new tests), integration **111 passed**, ruff/format/layer-guard clean, merges cleanly into current `develop`.

The acknowledged double-release on a partially-succeeded dispatch I left alone — agreed that under-counting beats a permanent lockout.

hedhoud:
@codex review

Ahmath-Gadji:
Closing this. It won't merge as it stands: it's months behind `develop`, CONFLICTING, 42 files, and it bundles two separate concerns. Its job-state half has landed, or is landing, in smaller PRs.

**Follow-up: #1023** makes Postgres the authoritative store for job state, which was the remaining part of #660. Continue review there.

**Already on `develop`**
- **#903**: bounds the in-memory `TaskStateManager` at 2,000 terminal tasks / 1 h, the numbers used here.
- **#904**: persists job state to Postgres. `39964a8f` ports this PR's `jobs` schema: `user_id` FK with `ON DELETE SET NULL`, the seven-state `ck_jobs_status` CHECK, `(status, created_at)`, the `COALESCE(completed_at, created_at)` retention index, and `started_at` / `completed_at`, so queue wait can be derived.

**Still open**
- **Read precedence (#660)**: on `develop` the actor still wins when both the actor and Postgres hold a task (`job_service.list_tasks` merges `{**durable, **actor}`). This PR had it the other way round: Postgres authoritative, actor as fallback. **#1023** (@hedhoud) makes that change on top of #904.
- **Quota TOCTOU (#664)**: not fixed. `check_user_file_quota` still reads `file_count` plus the in-memory pending count, so concurrent uploads can still get past the quota. The atomic `try_reserve_file_slot` reserve, and the rule for who owns the slot once it's reserved, from §1 and §3–4 should go in their own PR against current `develop`. The job-state work doesn't block it. #664 stays open for that.
- **#700** is stacked on this branch, so it has to be rebased onto that new quota PR (or folded into it). I'm keeping `fix/660-durable-job-state` so #700 keeps its base.
- **#676** (startup reconciliation of orphaned jobs, live-Ray e2e) still applies. The gap from §16 also still applies there: cancellations from the delete path write no durable row.

Thanks to everyone who reviewed this. The findings in §13–16 on the #685 and #671 merges still hold for whoever picks up the quota split.

PR #678, opened 2026-07-16
Fix Settings auth copy for token sessions

### LINKED ISSUE #538: Admin UI settings copy should match the active auth mode
Context
The Settings page is where admins check how access is configured for the current deployment. It should reflect the authentication mode that is actually running.

Problem
On the tested Admin UI deployment, login is token-based, but the Settings page says authentication is handled by SSO. That can confuse operators when they are validating token auth, debugging API access, or explaining how users should sign in.

Suggested fix
Make the Settings copy depend on the active auth mode. For token auth, explain bearer-token usage and token regeneration. For SSO/OIDC, keep the current SSO-oriented wording.

Acceptance
- Token-auth deployments no longer mention SSO as the active login mechanism.
- SSO/OIDC deployments still show SSO-specific guidance.
- The API token section remains clear in both modes.

hedhoud:
## Context

The Settings page always described sign-in as SSO, even when the current session was authenticated with a bearer token. That made token-mode deployments look misconfigured.

## Change

Use the existing protected auth endpoint to detect whether the current session is token-based or OIDC-based, then keep the existing Settings layout while swapping only the sign-in explanation.

Closes #538.

## Validation

- npm test -- settings.test.tsx
- npm run lint
- npm run build

## Summary by CodeRabbit

* **New Features**
  * Settings now displays authentication-specific sign-in guidance for SSO, bearer-token, and deployment-managed authentication.
  * Added support for retrieving the current authentication method and optional session expiration.

* **Tests**
  * Added coverage to verify the correct sign-in guidance appears for SSO and bearer-token authentication modes.

### DIFF
--- a/ui/src/lib/api/account.ts
+++ b/ui/src/lib/api/account.ts
@@ -17,9 +17,21 @@ export interface MyInfo {
   external_user_id?: string | null;
 }
 
+export interface CurrentAuthInfo {
+  user_id: number | null;
+  email?: string | null;
+  auth_method: "token" | "oidc" | string;
+  session_expires_at?: string | null;
+}
+
 /** Current authenticated user (from the bearer token / session). */
 export function getMyInfo() {
   return request<MyInfo>("/users/info");
+}
+
+/** Current authentication method resolved by the backend middleware. */
+export function getCurrentAuthInfo() {
+  return request<CurrentAuthInfo>("/auth/me");
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Ahmath-Gadji (approved):
Tested manually and confirms the fix works (token-mode Settings now shows bearer-token copy instead of the SSO text).

Reviewed the diff: `/auth/me` is registered unconditionally and runs through the standard AuthMiddleware, so `auth_method` resolves correctly in both token and OIDC deployments. Test coverage added for both branches, CI is green.

Two purely cosmetic nitpicks, not blocking:
- Brief flash of the fallback copy while the auth-info query is loading (CodeRabbit already flagged this).
- The file header comment in account.ts is slightly stale now that token-mode auth is also handled there.

LGTM, approving.

PR #679, opened 2026-07-16
Fix accessible names for icon-only Admin UI actions

### LINKED ISSUE #539: Admin UI icon-only controls should have accessible names
Context
The Admin UI uses icon-only controls for repeated actions such as pagination, account menu access, and row-level actions.

Problem
Some of these controls do not expose an accessible name. This makes the UI harder to use with screen readers and also makes automated tests more fragile because the action is not discoverable by role/name.

Suggested fix
Give every icon-only control a meaningful accessible label and, where useful, a tooltip with the same action name. The label should describe the action, not the icon.

Acceptance
- Icon-only buttons can be selected by an accessible role/name.
- Pagination controls announce previous/next clearly.
- Account/menu controls expose a clear action name.
- Tooltips and accessible labels stay consistent.

hedhoud:
## Context

Some icon-only actions in the Admin UI did not expose a clear accessible name, so screen-reader users and role/name tests could not reliably find them.

## Change

Add focused accessible labels and matching titles to the affected pagination, table action, copy-token, and restart buttons without changing their layout or behavior.

Closes #539.

## Validation

- npm test -- data-table.test.tsx list.test.tsx
- npm run lint
- npm run build

## Summary by CodeRabbit

- **Accessibility**
  - Added descriptive labels and tooltips to pagination, document, partition, system, and API token action buttons.
  - Improved screen-reader identification for view, edit, delete, restart, copy, and navigation controls.

- **Tests**
  - Added coverage confirming accessible labels for table pagination and administrative row actions.

### DIFF
--- a/ui/src/components/shared/data-table.test.tsx
+++ b/ui/src/components/shared/data-table.test.tsx
@@ -13,14 +13,18 @@ const columns: ColumnDef<Row, unknown>[] = [
 
 const makeRows = (n: number): Row[] => Array.from({ length: n }, (_, i) => ({ id: i + 1 }));
 
-// The pager renders only a prev + next icon button (no text), so "next" is the
-// last button on screen.
 const clickNextPage = () => {
-  const buttons = screen.getAllByRole("button");
-  fireEvent.click(buttons[buttons.length - 1]);
+  fireEvent.click(screen.getByRole("button", { name: /next page/i }));
 };
 
 describe("DataTable pagination", () => {
+  it("labels pagination icon buttons", () => {
+    render(<DataTable columns={columns} data={makeRows(25)} />);
+
+    expect(screen.getByRole("button", { name: /previous page/i })).toBeTruthy();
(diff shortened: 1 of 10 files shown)


Ahmath-Gadji (commented):
Clean, well-scoped a11y fix — nice job closing #539.

- Every previously-unlabeled icon-only control (pagination prev/next, document/partition row actions, actor restart, copy-token) now has a meaningful `aria-label`, and tests were added/hardened for each (including replacing the brittle "click the last button on screen" pagination test with a role/name query).
- Checked for overlap with `fix/538-admin-ui-settings-auth-copy` (also touches `settings.tsx`) — no shared lines, should merge cleanly.

One minor nit: in `system.tsx`, the restart button's `aria-label` is per-actor (`Restart ${a.name}`) but the `title` stays the generic "Restart actor". Not a bug, just a small inconsistency — might as well make the title match the label while you're in there.

Ahmath-Gadji (approved):
Nit addressed, CI green. LGTM.

PR #681, opened 2026-07-16
Fix Create Partition dialog height on small screens

### LINKED ISSUE #541: Admin UI Create Partition dialog should behave better on smaller viewports
Context
Create Partition is one of the first workflows an admin uses when setting up or testing OpenRAG.

Problem
On smaller viewports, the dialog can become taller than the visible area and the action buttons are awkward to reach. This makes the form feel fragile even though the workflow itself works.

Suggested fix
Make the dialog responsive: keep the title and action buttons usable, and let the form body scroll when there is not enough vertical space.

Acceptance
- The dialog is usable on laptop-sized and smaller viewports.
- Cancel/Create remain reachable without page-level layout issues.
- Keyboard users can tab through the form and actions predictably.

hedhoud:
## Context

The Create Partition dialog could become awkward on shorter screens because the actions moved with the full dialog content.

## Change

Keep the existing form and layout, but constrain this dialog to the viewport and make only the form body scroll so the title and Cancel/Create actions remain reachable.

Closes #541.

## Validation

- npm test -- partitions/list.test.tsx
- npm run lint
- npm run build

### DIFF
--- a/ui/src/pages/admin/partitions/list.test.tsx
+++ b/ui/src/pages/admin/partitions/list.test.tsx
@@ -87,6 +87,22 @@ describe("PartitionListPage bulk actions", () => {
     });
   });
 
+  it("keeps the create dialog actions reachable with a scrollable form body", async () => {
+    renderPartitions();
+
+    await userEvent.click(screen.getByRole("button", { name: /create personal partition/i }));
+
+    const dialog = screen.getByRole("dialog");
+    expect(dialog.className).toContain("max-h-[calc(100vh-2rem)]");
+    expect(dialog.className).toContain("overflow-hidden");
+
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji (approved):
Reviewed the diff (pure Tailwind layout fix, no logic changes) plus issue #541's acceptance criteria — dialog stays within viewport, Cancel/Create remain reachable, tab order unaffected since only CSS changed.

Verified locally (not covered by CI, which only runs Python jobs):
- `npx vitest run` — 19 files / 123 tests pass, including the new test
- `npx eslint` on changed files — clean
- `npx tsc -b` — clean
- `npx vite build` — succeeds

No issues found. Approving.

PR #688, opened 2026-07-17
fix(chat): resolve LLM token preflight against the partition's chat_llm preset

### LINKED ISSUE #639: chat preflight checks run against the default LLM, not the resolved per-partition chat_llm
### Context

Follow-up from #634, which made partition chat completions honor the per-partition `chat_llm` model-endpoint preset (query generation + answer generation resolve the effective LLM in `QueryService._resolve_llm`).

The **HTTP preflight layer still runs against the default LLM only**, so it is now out of step with the LLM that actually answers:

1. **Availability check** — `check_llm_model_availability` (`openrag/api/dependencies/llm.py`) validates only `config.llm` (the default endpoint/model). A partition with a healthy custom `chat_llm` is rejected (503/404) when the *default* endpoint is down, even though the request would never touch it.
2. **Token / context sizing** — `check_tokens_limit` uses `get_max_model_tokens()` (`openrag/api/routers/user/chat.py`), a single global primed once at startup from `config.llm.model`'s `max_model_len`. A partition request is admitted/rejected against the *default* model's context window, not the `chat_llm` model's — over-strict if the custom model has a larger window, over-permissive (fails downstream) if smaller.

### Proposed direction

- Resolve the request's effective LLM (partition → `chat_llm`, mirroring `QueryService._resolve_llm`) **before** the availability/token preflight, and validate against that endpoint.
- Move availability off a body-blind `Depends` (or make it body-aware), since the partition→endpoint mapping isn't known until the request body is parsed.
- Cache `max_model_len` **per endpoint name** instead of one startup-primed global; fall back to `config.llm_context.max_llm_context_size` when unknown.
- Cover with tests: default-down + healthy custom `chat_llm` (should succeed), and per-endpoint context sizing.

### Notes

- Pre-existing limitation of the preflight layer; #634 only makes it reachable by routing answers through non-default endpoints. #634 is a strict improvement over `main` — this only affects the default-down / non-default-context edge cases.
- Raised in review by @hedhoud and Codex on #634.

Ahmath-Gadji:
## Summary

Closes #639.

Follow-up to #634, which made partition chat completions honor the per-partition `chat_llm` model-endpoint preset (query generation + answer generation resolve the effective LLM in `QueryService._resolve_llm`). The HTTP preflight layer still ran against the default LLM only, out of step with the LLM that actually answers.

- **Per-endpoint token budgets** — `max_llm_context_size` / `max_output_tokens` are now admin-configurable per LLM endpoint (stored in the endpoint's `extra`), editable from the admin UI (with inline tooltips explaining each field), falling back to the global `LLMContextConfig` when unset.
- **Resolved-LLM preflight** — `check_tokens_limit` / `get_max_model_tokens` now resolve the partition's `chat_llm` preset endpoint (when the request is scoped to partitions that agree on one) and check against *that* endpoint's budget, instead of always the default endpoint's.
- **Dropped the default-only availability preflight** — `check_llm_model_availability` (a `Depends()` on both `/v1/chat/completions` and `/v1/completions`) probed only `config.llm`'s `/v1/models` before every request, so it validated the *wrong* endpoint once a partition's `chat_llm` preset — or a non-`.env` catalog default (#755) — answered: a partition on a healthy custom LLM was falsely rejected (503/404) whenever the *default* endpoint happened to be down, and its "Validating model" log misattributed the answering model. The liveness it emulated already exists against the actually-resolved endpoint — the generation call maps upstream failures to `InferenceConnectionError` / `InferenceTimeoutError` / `InferenceError` (503/504/502), surfaced by the global exception handler (non-streaming) and the streaming SSE error path — so removing it loses no error contract (no test pinned the 404/503 behaviour) and drops a per-request `/v1/models` round-trip on the healthy path. `get_openai_models` stays; the token preflight still uses it.
  - ⚠️ **Breaking (error contract, hence the `breaking-change` label)** — for an unreachable/misconfigured LLM, chat/completions no longer fail *before* generation. **Streaming** requests now return `200` with an SSE `error` event (was a pre-stream `5xx`, since headers are already sent once streaming starts). **Non-streaming** requests still return `503`/`504` for a genuinely-down *resolved* endpoint, but no longer the preflight's `404` ("model not available at endpoint") / `500` ("LLM configuration incomplete"), and no longer reject a request just because the *default* endpoint is down. All arguably fixes (the old guard was default-only), but they change status codes on the OpenAI-compatible surface, so integrators keying on them should be aware.
- Extracted `resolve_partition_chat_llm()` as a shared consensus rule used by both `QueryService._resolve_llm` (which LLM answers) and the chat-completions preflight (what budget it's checked against), so the two can't fall out of sync.
- `ModelsConfig` accessors generalized (`default_llm_context_size()` → `llm_context_size(name)`, etc.) to read any catalogued endpoint's budget, not just `"default"`.
- **Per-endpoint auto-probing** — `prime_max_model_tokens` now probes every registered LLM endpoint's `/v1/models` for `max_model_len` (deduplicating endpoints that alias the same config, e.g. the `"default"` name), instead of only the default endpoint. Cached per endpoint name in `_max_model_tokens_by_name`.
- **Cache kept fresh on writes** — `config.models.llm` already refreshes synchronously on every model-endpoint CRUD write (`ModelEndpointService.load_all()`), but the auto-probe cache didn't. `api/routers/admin/model_endpoints.py` now re-primes it after create/update/delete/set-default for LLM endpoints (best-effort — a probe failure never fails the admin's request), so a newly added or edited endpoint doesn't wait for a process restart to get sized correctly.
- Admin UI copy updated: the token budget fields apply whenever the endpoint answers a request (default LLM or a partition's chat LLM preset).

**Note:** this PR also carries the groundwork commit (per-endpoint token budgets) from the original #642 scaffold, whose branch was deleted before merging — it was never landed on `develop`.

## Related

**#755** (`fix/754-chat-llm-default-endpoint`, **merged**) fixed a separate bug on the *default* path — `_resolve_llm` returned the static `.env` `self._llm` instead of the live catalog default (`is_default=True`), so promoting a new default endpoint had no effect on the default chat path — and refactored `_resolve_llm` into helpers (`_agreed_partition_chat_llm` / `_default_llm` / `_default_llm_name`). This PR is now rebased on top of it: `_agreed_partition_chat_llm` **delegates the partition-consensus decision to this PR's shared `resolve_partition_chat_llm()`** instead of re-deriving it inline, so the answering-LLM resolution and the token preflight share one rule.

## Future work (separate PR)

- Remove the global `MAX_LLM_CONTEXT_SIZE` / `MAX_OUTPUT_TOKENS` env vars / `LLMContextConfig` fallback once every deployment is expected to carry per-endpoint budgets (admin-configured or auto-probed).
- Reorganize endpoint resolution into a dedicated service: `_resolve_llm_endpoint_name` (preflight) and `QueryService._resolve_llm` (answering) now agree on the endpoint *name* via the shared `resolve_partition_chat_llm`, but each still rebuilds the resolved client / budget separately. Centralize llm/vlm/embedder/reranker resolution in one place and propagate the resolved endpoint down through the call chain, instead of every consumer re-resolving it. At that point, reconsider whether a pre-generation liveness/status guard should return — this time against the *resolved* endpoint — to restore a non-`200` status for the unreachable-LLM streaming case flagged above.

## Test plan
- [x] `uv run pytest tests/unit/` — 2113 passed
- [x] `uv run ruff check openrag/ tests/` — clean
- [x] `npx vitest run` (ui) — passed
- [x] `npx tsc -b` (ui) — clean
- [ ] Manual verification in the admin UI (not yet done)

## Summary by CodeRabbit

* **New Features**
  * Added vendor selection with branded icons (and neutral fallbacks) in the Admin model endpoint UI.
  * Introduced per-endpoint LLM token budgets (max context size and max output tokens) with partition-aware chat token limiting and endpoint-scoped defaults.
* **Bug Fixes**
  * Improved auto-probing and caching of LLM token limits across distinct endpoints, including alias reuse and safer failure isolation.
  * Token-limit validation now uses the resolved partition preset before enforcing limits for both chat and completion endpoints.
  * Admin model-endpoint updates now refresh probed token limits in the background (best-effort).
  * Strengthened validation for LLM token-budget fields to reject invalid values.
* **Tests**
  * Added coverage for background token-cache refresh and LLM token-budget validation.


openrag/api/routers/admin/model_endpoints.py
```diff
@@ -31,6 +32,27 @@ def _same_endpoint_url(left: str, right: str) -> bool:
...
+    ``config.models.llm`` itself is already refreshed synchronously inside
+    the service call (``ModelEndpointService.load_all()``), but the
+    ``/v1/models`` auto-probe cache (``chat._max_model_tokens_by_name``) is
+    a separate cache that ``prime_max_model_tokens`` otherwise only
+    populates once at process startup — without this, a newly added or
+    edited LLM endpoint keeps falling back to the global context-size
+    default until the next restart. No-op for non-LLM endpoint types.
+    Best-effort: a probe failure must not fail the admin's CRUD request.
+    """
+    if model_type != "llm":
```
aditykris, line 56: Just a suggetion , will rewriting the code like this make it more elegant ??

```
if model_type == "llm":
    try:
        await prime_max_model_tokens()
    except Exception:
        logger.exception("Failed to refresh auto-probed LLM token cache...")

return 
```

openrag/api/routers/user/chat.py
```diff
@@ -149,43 +178,85 @@ def is_direct_llm_model(
...
+    """
+    return resolve_partition_chat_llm(partitions, config.partitions) or "default"
+
+
+def _effective_max_output_tokens(config: "Settings", partitions: list[str] | None = None) -> int:
+    """Default output-token budget: the resolved LLM endpoint's admin-configured
+    value when set, else the global ``llm_context`` fallback."""
+    models = getattr(config, "models", None)
+    configured = (
+        models.llm_output_tokens(_resolve_llm_endpoint_name(config, partitions)) if models is not None else None
```
aditykris: `is not None`  feels redundant .

Can we rewrite it either this :

`models.llm_output_tokens(_resolve_llm_endpoint_name(config, partitions)) if models else None `

or 

`models and models.llm_output_tokens(_resolve_llm_endpoint_name(config, partitions))`


### Another point to note:

**`getattr(config, "models", None)` + `is not None` guards against something that can't happen**

```python
def _effective_max_output_tokens(config: "Settings", partitions: list[str] | None = None) -> int:
    models = getattr(config, "models", None)
    configured = (
        models.llm_output_tokens(_resolve_llm_endpoint_name(config, partitions)) if models is not None else None
    )
    return configured or int(config.llm_context.max_output_tokens)
```

Settings.models (core/config/root.py:69) is models: ModelsConfig = Field(default_factory=ModelsConfig) — not Optional, always populated with a default on every Settings instance. There's no code path where config.models is missing, so getattr(..., None) + if models is not None can never actually take the else None branch in practice — it's dead defensive code standing in for a plain attribute access.

It's also inconsistent with its neighbor a few lines down, added in this same PR:

```
def get_max_model_tokens(partitions: list[str] | None = None, settings: "Settings | None" = None) -> int:
    ...
    configured = config.models.llm_context_size(name)
```

get_max_model_tokens accesses config.models directly, no guard. Two functions in the same file, added in the same PR, treating the same attribute with different trust levels — that inconsistency is a stronger signal something's off than the redundancy itself.

Suggest simplifying to:

```
def _effective_max_output_tokens(config: "Settings", partitions: list[str] | None = None) -> int:
    configured = config.models.llm_output_tokens(_resolve_llm_endpoint_name(config, partitions))
    return configured or int(config.llm_context.max_output_tokens)
```

openrag/api/routers/user/chat.py
```diff
@@ -149,43 +178,85 @@ def is_direct_llm_model(
...
+    ``prime_max_model_tokens`` — cached per endpoint name, not just the
+    default) > the global ``llm_context`` fallback.
+    """
+    config = _runtime_config(settings)
+    name = _resolve_llm_endpoint_name(config, partitions)
+    configured = config.models.llm_context_size(name)
+    if configured is not None:
+        return configured
+    probed = _max_model_tokens_by_name.get(name)
+    if probed is not None:
```
aditykris: I think we can rewrite here as well. 

_max_model_tokens_by_name only ever holds positive ints (max_model_len from /v1/models, never legitimately 0), so or and is not None are equivalent here.

```
    probed = _max_model_tokens_by_name.get(name)
    return probed or int(config.llm_context.max_llm_context_size)
```

Full function reads more consistently too:

```
def get_max_model_tokens(partitions: list[str] | None = None, settings: "Settings | None" = None) -> int:
    ...
    config = _runtime_config(settings)
    name = _resolve_llm_endpoint_name(config, partitions)
    configured = config.models.llm_context_size(name)
    if configured is not None:
        return configured
    probed = _max_model_tokens_by_name.get(name)
    return probed or int(config.llm_context.max_llm_context_size)
```
configured (from llm_context_size) is admin-entered and validated >0 at the schema layer (_validate_llm_token_extra in model_endpoint_schemas.py), so its is not None check is a real distinction worth keeping explicit only the probed/fallback pair collapses cleanly.

aditykris (dismissed):

openrag/api/routers/user/chat.py
Ahmath-Gadji: Confirmed — `Settings.models` is `Field(default_factory=ModelsConfig)`, never `Optional`, so the `is not None` branch was dead. Simplified as suggested in edd51a1e.

openrag/api/routers/user/chat.py
Ahmath-Gadji: Good catch — `_max_model_tokens_by_name` only ever holds positive ints from `_fetch_max_model_tokens` (never a legitimate 0), so `or` is safe here and matches the fallback pattern used elsewhere in this module. Simplified in edd51a1e.

openrag/api/routers/admin/model_endpoints.py
Ahmath-Gadji: I'd rather keep the guard clause here: `if model_type != "llm": return` exits early with no nesting, whereas wrapping the body in `if model_type == "llm":` adds an indent level and needs a trailing bare `return` (a no-op) to read cleanly as "nothing else happens after." Left as-is.

openrag/api/routers/admin/model_endpoints.py
```diff
@@ -31,6 +32,27 @@ def _same_endpoint_url(left: str, right: str) -> bool:
...
+        return
+    try:
+        await prime_max_model_tokens()
+    except Exception:
+        logger.exception("Failed to refresh auto-probed LLM token cache after endpoint write")
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 89b3c138 — `_reprime_llm_token_cache` is now scheduled via FastAPI `BackgroundTasks.add_task(...)` in all four CRUD handlers, so the response returns before the serial `/v1/models` probes run. The helper keeps its non-LLM no-op and best-effort exception logging, and the existing reprime tests still pass (the ASGI test client runs the background task within the request's ASGI call).

openrag/api/routers/user/chat.py
```diff
@@ -44,25 +44,53 @@
...
+        if identity not in probed_by_identity:
+            probed_by_identity[identity] = await _fetch_max_model_tokens(
+                base_url=endpoint.endpoint,
+                model_id=endpoint.model_name,
+                api_key=endpoint.extra.get("api_key", ""),
+            )
+        value = probed_by_identity[identity]
+        if value is not None:
+            results[name] = value
+    _max_model_tokens_by_name = results
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 894993a8 — `prime_max_model_tokens` now iterates `list(config.models.llm.items())`, a snapshot taken before the first `await`, so a concurrent `load_all()` (`dict.clear()` + `dict.update()`) can no longer raise `RuntimeError: dictionary changed size during iteration`. Dedup/probe/cache behaviour is unchanged.

ui/src/pages/admin/models.tsx
```diff
@@ -294,13 +340,18 @@ function EndpointDialog({
...
         setName(editing.name);
         setEndpoint(editing.endpoint);
         setModelName(editing.model_name || "");
         setBatchSize(String(editing.batch_size));
         setTimeout(String(editing.timeout));
         setApiKey(displayApiKey);
+        setMaxContextSize(llmContext.maxContextSize);
+        setMaxOutputTokens(llmContext.maxOutputTokens);
+        setVendor(implementation || DEFAULT_VENDOR_BY_TYPE[editing.model_type]);
         setExtraJson(JSON.stringify(displayExtra, null, 2));
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in a6cc399f — the dialog now carves the LLM budget keys out of `extra` (both the display path and the validation comparator) only when the edited endpoint's `model_type === "llm"`. Non-LLM endpoints keep their `extra` untouched, so editing and saving an embedder/reranker/VLM no longer drops same-named keys.

ui/src/pages/admin/models.tsx
```diff
@@ -62,6 +72,34 @@ type RevealedApiKey = {
 
 const normalizeEndpointUrl = (value: string) => value.trim().replace(/\/+$/, "");
 
+// System-wide fallback values (MAX_LLM_CONTEXT_SIZE / MAX_OUTPUT_TOKENS env
+// vars — see core/config/endpoints.py:LLMContextConfig) shown as placeholders
+// so admins can see what an unset override falls back to.
+const DEFAULT_MAX_CONTEXT_SIZE = "8192";
+const DEFAULT_MAX_OUTPUT_TOKENS = "1024";
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 5812ac5a — the Max context size / Max output tokens inputs now use a generic `System default` placeholder instead of the hard-coded `8192` / `1024`, which are only the built-in fallbacks and are overridable per deployment via `MAX_LLM_CONTEXT_SIZE` / `MAX_OUTPUT_TOKENS`.

ui/src/pages/admin/models.tsx
```diff
@@ -572,6 +634,42 @@ function EndpointDialog({
               <Input type="number" step="0.1" value={timeout} onChange={(e) => setTimeout(e.target.value)} />
             </div>
           </div>
+          {isLlm && (
+            <div className="grid grid-cols-2 gap-4">
+              <div className="space-y-2">
+                <LabelWithInfo
+                  label="Max context size"
+                  tooltip="Fallback maximum token limit for chat/completion requests answered by this endpoint. At startup, the /v1/models endpoint is queried for the model's max_model_len; if that query fails this value is used instead. Requests whose total token count (prompt + max_tokens) exceeds the limit are rejected with a 413 error."
+                />
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in d2118b79 — the tooltip now states that a configured value takes precedence over the auto-probed `max_model_len` (leave blank to use the probed value, or the global default if the probe fails), matching `get_max_model_tokens`'s actual precedence.

openrag/api/routers/admin/model_endpoints.py
```diff
@@ -31,19 +32,49 @@ def _same_endpoint_url(left: str, right: str) -> bool:
...
     body: CreateModelEndpointRequest,
+    background_tasks: BackgroundTasks,
     service=Depends(get_model_endpoint_service),
 ):
     """Register a named inference endpoint."""
     now = datetime.now(UTC)
     row = ModelEndpointRow(**body.model_dump(), created_at=now, updated_at=now)
-    return await service.create_model_endpoint(row)
+    result = await service.create_model_endpoint(row)
+    background_tasks.add_task(_reprime_llm_token_cache, body.model_type)
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in c0e70485 — `prime_max_model_tokens` now runs under a module-level `asyncio.Lock`, taking the registry snapshot and probing while holding it. Overlapping refreshes no longer interleave, so the last refresh to run always observes the latest config and publishes the newest result; an older, slower probe can no longer overwrite a newer cache. Added a concurrent-refresh regression test asserting the probes are serialized (max one in flight — it reaches 2 without the lock).

Note: this last-writer-wins window predates the background-task change (two concurrent inline `await prime_max_model_tokens()` calls could already interleave); the lock closes it either way.

andyne13 (commented):
Reviewed end to end (preflight resolution, the dropped availability guard, token-budget validation, auto-probe cache, and the UI), traced the answering vs. preflight paths against each other, and verified the claims against the code. This is solid, well-tested work — the shared-consensus design is the right call and the core invariant holds. A few findings, one of which I'd like to see addressed before merge.

## What I verified as correct
- **Preflight ↔ answering stay in sync in the normal cases.** Both `QueryService._resolve_llm` (`query_service.py`) and the preflight `_resolve_llm_endpoint_name` (`chat.py`) delegate to the same `resolve_partition_chat_llm(partitions, config.partitions)` (`core/models/preset.py`), on the same `partitions` object — so single-preset, multi-partition-agreement, disagreement→default, `all`, no-partition, and no-preset all resolve the budget against the endpoint that actually answers. Nicely covered (`test_partition_preset_endpoint_budget_wins_over_default`, `test_conflicting_partition_presets_fall_back_to_default`, `test_preset_models`).
- **Dropping `check_llm_model_availability` is safe.** No test pins the old 404/500 behaviour (grep across the tree is empty; the one 404 test is `get_partition_name`'s "model not found", untouched). Liveness is re-emitted against the *resolved* endpoint: non-streaming maps `Inference{Connection,Timeout,}Error → 503/504/502` via the global handler, streaming returns 200 + an SSE `error` event (headers already flushed). And the old guard was itself the #639 bug (default-only probe). Good.
- **Token-budget validation is sound.** `_validate_llm_token_extra` checks `isinstance(bool)` *before* the int check, so `True`/`False` are rejected (the classic bool-is-int trap is handled), along with 0/negative/float/`"4096"`; the read side coerces garbage to `None` → global fallback. Parametrized tests cover it.
- **Auto-probe dedup + failure handling** — dedup by `id(endpoint)` (so the `"default"` alias is probed once), and a failed probe omits the key rather than caching/poisoning it → clean fallback. Good.

## Findings

**1. (correctness, please address) Preflight budget and the answering endpoint diverge for a *stale* `chat_llm` preset.** `resolve_partition_chat_llm` returns the stored name without checking catalog membership. If a partition's `chat_llm` names an endpoint that was later deleted/renamed:
- Answering: `_resolve_llm` catches the factory `KeyError` and falls through to the catalog **default** endpoint (as its docstring intends).
- Preflight: `get_max_model_tokens` resolves that same dangling name, finds no configured/probed value, and falls back to the **global** `llm_context.max_llm_context_size`.

So the 413 guard is checked against the global budget while the request is answered by the default endpoint — whose configured/probed window may be smaller. A request can clear preflight and then exceed the real window downstream, which is exactly the preflight↔answer drift this PR sets out to prevent (just in the stale-preset corner). It's reachable — nothing cascades `partitions.chat_llm` when an endpoint is deleted — and currently untested. Suggested fix: have the preflight fall to `"default"` when the resolved name isn't in `config.models.llm` (mirroring `_resolve_llm`'s KeyError fallback), or drop non-catalogued names inside `resolve_partition_chat_llm` so both sides converge.

**2. (concurrency — matches CodeRabbit's open comment on `model_endpoints.py:76`) The reprime race is still open.** `prime_max_model_tokens` rebinds `_max_model_tokens_by_name` atomically (no half-built cache — good), but there's no serialization across the four CRUD handlers now scheduling it via `BackgroundTasks`. Two overlapping writes can race so the slower/older reprime rebinds last and republishes a stale snapshot (e.g. a just-deleted endpoint's probed value). Low severity — it self-heals on the next reprime and an admin-configured budget overrides the probed tier — but the serialization/generation-guard + regression test CodeRabbit asked for aren't in yet. A generation counter or an `asyncio.Lock` around the rebind would close it.

**3. (test coverage) The breaking-change status codes aren't pinned.** The unknown-model case now shifts 404→503 (documented), and the streaming path returns 200 + SSE `error`. Nothing in the chat-route tests asserts these — they lean on client-level `test_vllm_client`. Given this is a change on the OpenAI-compatible surface, a test locking the non-streaming 503/504 and the streaming SSE-`error` envelope for a down *resolved* endpoint would keep the new contract from silently regressing.

## Nits
- `chat.py` still has the `model_extra` fallback (`… or model_data.get("model_extra", {}).get("max_model_len")`) CodeRabbit flagged as unreachable — trivial/harmless, just noting it's still there.
- Minor asymmetry: an explicit JSON `null` in a budget key is rejected 422 on write, while the read path treats `None` as a graceful fallback; and `False` is handled but not explicitly tested.

Overall: 👍 — the invariant holds and the removal is well-reasoned. My one real ask is #1 (align the preflight's stale-name fallback with the answer's default), with #2 (the still-open reprime race) and a status-code test as the follow-ups.

Ahmath-Gadji:
@coderabbitai Declining the outside-diff suggestion to relocate the token-validation tests into mirrored `tests/unit/api/routers/user/` modules.

`tests/unit/test_token_validation.py` is deliberately placed flat — its module docstring documents the reason: a test module under `.../api/routers/user/` importing from `api.routers.user.chat` shadows the third-party `openai` package during import resolution and causes a circular import. This PR only *appends* a concurrent-refresh test to that existing file, following its established convention; it does not introduce the placement. The flagged ranges span nearly the whole file, so relocating them is a pre-existing, out-of-scope refactor that would risk reintroducing that import cycle. Happy to revisit the file placement separately if the shadowing can be resolved.

Ahmath-Gadji:
@andyne13 thanks for the thorough end-to-end review — I appreciate the trace against the answering vs. preflight paths. All three findings are addressed:

**#1 (correctness — stale `chat_llm` preflight divergence) — fixed in fba2ba41.** You nailed it: a partition preset naming a deleted/renamed endpoint answered via the catalog **default** (`_resolve_llm`'s `KeyError` fallback) but preflighted against the **global** budget. `_resolve_llm_endpoint_name` now falls back to `"default"` when the resolved name isn't in `config.models.llm`, mirroring `_resolve_llm` so both sides converge. Regression test (`test_stale_partition_preset_falls_back_to_default_endpoint_budget`) pins it: default endpoint configured to 4096, global at 8192 — the stale preset now resolves 4096 (what actually answers), and returns 8192 without the fix.

**#2 (reprime race) — already closed in c0e70485** (it landed just before your review, so you likely saw the prior state). `prime_max_model_tokens` now runs under a module-level `asyncio.Lock`, taking the snapshot and probing while holding it, so the last refresh to run observes the latest registry and wins — no older probe can republish a stale snapshot. `test_concurrent_refreshes_are_serialized` asserts max one probe in flight (reaches 2 without the lock). CodeRabbit re-verified this one.

**#3 (breaking-change status contract) — pinned in 8c6afd23**, at the mechanisms the chat route relies on:
- Non-streaming: `InferenceConnectionError` / `InferenceTimeoutError` / `InferenceError` → 503 / 504 and upstream-status passthrough (502), end-to-end through the global handler wiring in `test_error_handlers`.
- Streaming: `_make_sse_error` emits the OpenAI-shaped `error` + `[DONE]` envelope (the response stays 200 once headers are flushed).

**Nits:** the null-vs-`None` asymmetry is intentional — an explicit JSON `null` on *write* is almost certainly a mistake, so it's rejected 422 to fail loud, whereas a missing/`None` on *read* is the normal "unset → global fallback" path. `False` already takes the same `isinstance(bool)` reject path as `True` (which is tested). The dead `model_extra` fallback and an explicit `False` test case I'll fold into a small follow-up to keep this PR's diff focused — both are harmless as-is.

CI is green on 8c6afd23. Thanks again 🙏

andyne13:
Thanks for the fast turnaround — re-verified against the head code:

- **#1 (stale-preset divergence)** — 👍 `_resolve_llm_endpoint_name` now falls back to `"default"` when the resolved name isn't in `config.models.llm`, so preflight and `_resolve_llm` converge on the same endpoint. The `test_stale_partition_preset_falls_back_to_default_endpoint_budget` case revert-proves it (4096 vs the global 8192). Solid.
- **#2 (reprime race)** — 👍 the `asyncio.Lock` around snapshot→probe→publish closes it, and `test_concurrent_refreshes_are_serialized` (max one probe in flight) revert-proves it. Confirmed I'd seen the pre-lock state.

One small note on **#3**: the non-streaming 503/504/502 mapping is now pinned in `test_error_handlers`, but I couldn't find a test covering the **streaming** side — nothing references `_make_sse_error` or asserts the 200 + `error` + `[DONE]` envelope for a down resolved endpoint. Not a blocker, and fine to fold into the same follow-up as the `model_extra`/explicit-`False` cleanup — just flagging so the streaming contract doesn't silently go unpinned.

Nice work on the rest.

Ahmath-Gadji:
@andyne13 good catch — and you were right, my "#3 streaming pinned" claim was wrong. The test existed in my working tree but never made it into 8c6afd23 (that commit touched only `test_error_handlers.py`), so the streaming contract genuinely was unpinned. Apologies for the inaccurate status.

Now actually landed in 4f7e5f70 — and I verified the **committed blob** this time rather than just the working tree:

- `test_make_sse_error_envelope` asserts `_make_sse_error` emits the OpenAI-shaped `error` payload (`message` / `type` / `param` / `code`) followed by the `[DONE]` terminator — i.e. once headers are flushed the response stays 200 and the failure has to surface *inside* the stream.

So both halves of the availability-preflight removal's error contract are pinned now: 503/504/502 on the non-streaming path (`test_error_handlers`), and the SSE `error` + `[DONE]` envelope on the streaming path.

Follow-up list is unchanged and still small: the dead `model_extra` fallback and an explicit `False` validation case.

openrag/core/config/model_endpoints.py
```diff
@@ -28,6 +28,28 @@ class ModelEndpointConfig(BaseModel):
...
+        return None
+    return value if value > 0 else None
+
+
+# Well-known ``extra`` keys carrying the LLM's admin-configurable token budgets
+# (surfaced as first-class fields in the admin UI). Kept in ``extra`` rather than
+# as dedicated columns so they stay LLM-only and need no migration, mirroring the
+# existing ``max_model_len`` / ``embed_concurrency`` convention.
+LLM_CONTEXT_SIZE_KEY = "max_llm_context_size"
+LLM_OUTPUT_TOKENS_KEY = "max_output_tokens"
```
hedhoud, line 50: These budget values stay in `extra`, and the model factory forwards `extra` into the LLM client. The client then sends them in every request, even though `max_llm_context_size` and `max_output_tokens` are not OpenAI request fields. A strict provider can start returning 400 errors as soon as an admin saves a budget. Could we treat these as OpenRAG-only settings and remove them before building the client? A test checking the actual request body would help prevent this from coming back.

openrag/api/schemas/user/chat.py
```diff
@@ -6,7 +6,13 @@
...
     from core.config import load_config
 
-    return load_config().llm_context.max_output_tokens
+    config = load_config()
+    # Prefer the default LLM endpoint's admin-configured budget over the global
+    # llm_context fallback (both editable, but the per-endpoint value wins).
+    # This factory runs at request-body parse time, before the partition is
+    # resolved, so it can only ever see the default endpoint — the
+    # partition-aware check happens later in check_tokens_limit.
+    return config.models.llm_output_tokens() or config.llm_context.max_output_tokens
```
hedhoud: This chooses the output default before we know which partition LLM will answer. For example, if the default endpoint uses 1024 but the partition endpoint uses 4096, a request without `max_tokens` is already filled with 1024 here, so the partition-aware fallback never runs and 1024 is sent downstream. Could we leave it unset, or use `model_fields_set`, and apply the default after the partition endpoint is resolved?

hedhoud (commented):
I found two cases that should be fixed before approval.

openrag/core/config/model_endpoints.py
Ahmath-Gadji: Confirmed and fixed in 8dc5e942 — you were exactly right, and it was reaching the wire. I captured the actual outbound HTTP body before the fix:

```
POST /chat/completions  body keys = [max_llm_context_size, max_output_tokens, messages, model, stream]
POST /completions       body keys = [max_llm_context_size, max_output_tokens, model, prompt]
```

Both paths, because `_chat_payload_kwargs` starts from `{**self._defaults, **kwargs}` and `generate()` splats `self._defaults` directly — and the factory was putting the budgets there by splatting all of `extra` into the constructor.

Fix: strip `max_llm_context_size` / `max_output_tokens` alongside `implementation` before building the client (`_NON_CONSTRUCTOR_EXTRA_KEYS` in `di/factories.py`). They're OpenRAG-side sizing settings read by the token preflight via `ModelsConfig.llm_context_size` / `llm_output_tokens`, never by the client. Verified end-to-end through the real factory + registry + `VLLMClient` afterwards:

```
chat         body keys = [messages, model, stream]   LEAKED = NONE
completions  body keys = [model, prompt]             LEAKED = NONE
```

`api_key` still reaches the client (it becomes the `Authorization` header). No client constructor consumes either budget key, so stripping them is safe for every component kind.

And yes — this is the same leak class as `batch_size` in #712, which the surrounding code already documents. Added the request-body regression test you asked for (`test_llm_token_budgets_are_not_forwarded_as_kwargs`), sitting right next to the existing #712 one so the pattern is hard to reintroduce.

openrag/api/schemas/user/chat.py
Ahmath-Gadji: Confirmed and fixed in 3f6dd8f4 — good catch, this defeated the partition-aware path exactly as you described.

The `default_factory` ran at body-parse time, so in `validate_tokens_limit`:

```python
requested_tokens = request.max_tokens or default_output_tokens
```

the left branch was *always* taken for a request that omits `max_tokens`, leaving the partition-aware `_effective_max_output_tokens` dead code. And because the payload forwards `max_tokens` verbatim, the default endpoint's cap was also what got sent downstream — so a partition endpoint allowing 4096 was silently capped at 1024.

Fix, along the lines you suggested: `max_tokens` is now left unset in the schema (`Field(None)`), and the router fills it via `_apply_default_max_tokens(request, config, partitions)` immediately after the partition is resolved and before the preflight — so the preflight *and* the downstream payload both use the answering endpoint's budget. An explicit client-supplied value is never overwritten.

`default_max_tokens` is gone (it could only ever see the default endpoint). `TestDefaultMaxTokensFactory` is replaced by `TestApplyDefaultMaxTokens`, covering: schema leaves it unset, partition preset wins (asserts 4096 — the old parse-time path yields 1024, so it revert-proves the fix), no-partition uses the default endpoint, global fallback, and explicit value honoured.

Full unit suite green (2128 passed).

andyne13:
Re-verified the latest head against the code — all five earlier findings plus hedhoud's two are genuinely closed, each with a revert-proving test:

- #1 stale-preset preflight → `_resolve_llm_endpoint_name` falls back to `"default"` when the name isn't in `config.models.llm`.
- #2 reprime race → `asyncio.Lock` around snapshot+probe+publish (`test_concurrent_refreshes_are_serialized` asserts one probe in flight).
- #3 streaming envelope → `test_make_sse_error_envelope`.
- hedhoud A (budget keys on the wire) → `_NON_CONSTRUCTOR_EXTRA_KEYS` strips them before the constructor; `test_llm_token_budgets_are_not_forwarded_as_kwargs` asserts they're absent from the kwargs.
- hedhoud B (parse-time default) → `_apply_default_max_tokens` runs after partition resolution, and it's applied on both `/v1/chat/completions` and `/v1/completions`.

Also confirmed the two earlier nits are already handled — the `model_extra` fallback is gone (only the explanatory comment remains in `_fetch_max_model_tokens`), and `False` is in the token-budget parametrize now (`[0, -1, 2.5, "4096", True, False]`).

One suggestion worth considering (not a blocker), on the leak-class root cause:

`_NON_CONSTRUCTOR_EXTRA_KEYS` is a **denylist**. Tracing the path it guards: `make_component_factory` splats `extra` (minus the denylist) into `registry.create(impl, **kwargs)` (`di/factories.py:105`); `VLLMClient.__init__` absorbs any kwarg it doesn't explicitly name into `self._defaults` (`vllm_client.py:85`), which is then splatted into every request payload (`{**self._defaults, **kwargs, ...}`, `vllm_client.py:124` / `:137`). So the denylist has to enumerate every non-constructor key, and any *new* key persisted into `extra` later would leak into the request body by default — the same gap that let `batch_size` through (#712) and these budgets just now. An **allowlist** at the factory (forward only recognized constructor args), or having the client reject unknown kwargs instead of defaulting them, would close the class once rather than one key at a time.

Minor / optional: the removed-preflight error contract is pinned at the mechanism level — `_make_sse_error` is asserted directly, and the exception→status map (503/504/502) via synthetic routes in `test_error_handlers` — but nothing drives the actual chat route with a down *resolved* endpoint. A single route-level test (non-streaming 503/504, streaming 200 + SSE `error`) would lock the new contract end-to-end.

Nice work — the substance is all there.

Ahmath-Gadji:
@andyne13 thanks for re-verifying the whole set — and for pushing past the individual leak to the class. Neither of these is in yet; both are going on the follow-up list rather than into this PR. Detail on each, because #1 turns out not to be quite an allowlist.

## 1. The denylist → the root cause is real, but a constructor-arg allowlist would break a deliberate passthrough

Your trace is exactly right, and the conclusion — that the denylist has to enumerate every non-constructor key, so a *new* `extra` key leaks by default — is the actual defect. `batch_size` (#712) and these budgets are the same bug twice.

Where it gets awkward is the proposed shape. `self._defaults` isn't an accident of absorbing stray kwargs; it's the mechanism by which per-endpoint sampling params reach the wire, and the legacy path depends on it. `_build_query_service` splats `settings.llm.model_dump()` minus `base_url`/`model`/`api_key` into `create_llm` (`di/container.py:511-517`, mirrored at `:557-563`). I resolved what that actually forwards:

```
passed to create_llm:                            [enable_thinking, logprobs, max_retries, temperature, timeout]
-> absorbed as named __init__ params:            [enable_thinking, max_retries, timeout]
-> falls into **kwargs => _defaults => the wire: [logprobs, temperature]
```

`temperature` and `logprobs` land in `_defaults` and get splatted into every payload *on purpose* — both are legitimate OpenAI request fields, and that's how `LLMParamsConfig` is meant to configure an endpoint. So:

- **"client rejects unknown kwargs"** would reject `temperature`/`logprobs` unless they're enumerated in `__init__`.
- **"allowlist recognized constructor args at the factory"** would strip them, silently dropping configured sampling defaults — a quieter regression than the one it fixes.

The split that matters isn't *recognized constructor arg vs. not*, it's **OpenRAG-side setting vs. wire-bound sampling param** — and the second set is open-ended and vendor-extensible (`top_p`, `seed`, `stop`, `frequency_penalty`, vLLM's own extensions), so enumerating *it* is just the denylist inverted, with a worse failure mode.

What closes the class without enumerating either side is making the OpenRAG-only keys **structurally** identifiable instead of listed — a reserved namespace in `extra` (a `openrag_*` prefix, or a nested `openrag: {...}` sub-dict) that the factory strips by rule. Then a new OpenRAG-side key cannot leak by construction, while an unprefixed key still passes through as a sampling param, which is the intended escape hatch.

The reason that's a follow-up and not a commit here: `max_llm_context_size` / `max_output_tokens` are already persisted in `extra` in the DB and written by the admin UI, so moving them under a namespace needs a read-side back-compat shim (or a migration) plus a UI change — well outside this PR's diff. The denylist is correct for the keys that exist today; it's the *default* that's wrong, and that's the part worth fixing properly.

## 2. Route-level error contract — agreed, and it should follow the flat test placement

Agreed, and your characterisation is accurate: `test_error_handlers` raises from synthetic routes and `test_make_sse_error_envelope` asserts the emitter directly, so the mapping is pinned at the mechanism, not through the chat route. Nothing currently drives `/v1/chat/completions` with a down *resolved* endpoint.

The mechanism for it already exists — `async_client_factory` (`tests/unit/api/conftest.py`) mounts a purpose-built app with just the router under test and dependency overrides, so the test would mount the chat router, override the LLM to raise `InferenceConnectionError` / `InferenceTimeoutError`, and assert 503/504 non-streaming plus 200 + SSE `error` + `[DONE]` streaming.

One placement constraint to note up front, so it doesn't get discovered halfway in: it'd need the same flat placement as `tests/unit/test_token_validation.py`, whose docstring documents why — a test module under `tests/unit/api/routers/user/` shadows the third-party `openai` package during import resolution and causes a circular import. Not a blocker, just the reason it won't sit in the mirrored path.

Follow-up list is now: the `extra` namespace/allowlist rework (1) and the route-level contract test (2). Both non-blocking for this PR, as you called them.

openrag/api/routers/admin/model_endpoints.py
hedhoud: Moving the refresh to a background task keeps admin writes fast, but it also creates a correctness window: the registry/client cache changes synchronously, while the probed token-limit cache can still hold the previous snapshot until the background task completes. An immediate chat request after a default/update/delete can be answered by the new endpoint but validated against the old context limit when no explicit `max_llm_context_size` is set. Could we synchronously invalidate the affected probed entry, version the cache, or otherwise force a safe fallback until the refresh finishes?

openrag/api/schemas/admin/model_endpoint_schemas.py
```diff
@@ -53,6 +75,12 @@ def validate_endpoint(cls, value: str) -> str:
         """Normalize the endpoint URL."""
         return _normalize_endpoint(value)
 
+    @field_validator("extra")
```
hedhoud, line 88: This validator runs for every endpoint type, not only LLMs. The UI now preserves same-named non-LLM `extra` keys, but the API will still reject an embedder/reranker/VLM payload containing `max_output_tokens` or `max_llm_context_size` unless the value is a positive int. That makes these names globally reserved, even for provider-specific metadata. Can we make this validation conditional on `model_type == "llm"`, or move these budgets into an LLM-specific namespace?

hedhoud (commented):
Re-reviewed the current head. Leaving two remaining behavioral concerns that still look valid to me.

openrag/api/routers/admin/model_endpoints.py
Ahmath-Gadji: Confirmed and fixed in 9757d0b8 — you were right, and moving the reprime off the request path is what opened it.

`ModelEndpointService.load_all()` runs synchronously inside the service call (and `set_default` then evicts the client cache), so by the time the CRUD response returns the registry already describes the new endpoint while `_max_model_tokens_by_name` still holds the previous one's `max_model_len`. In `get_max_model_tokens` the `configured` tier reads the live registry but `probed` reads that stale dict, so with no admin-set `max_llm_context_size` a chat request in the window is answered by the new endpoint and preflighted against the old one's limit — exactly the divergence this PR exists to close.

Fix follows your first suggestion (invalidate synchronously) plus the second (version the cache), because they cover different halves:

- **Inline invalidation.** `invalidate_max_model_tokens()` now runs in the request handler, before the background task is scheduled — both halves behind one `_refresh_llm_token_cache(background_tasks, model_type)` helper so the four call sites can't drift apart. The window now falls through to the endpoint's admin-configured size (read live, so unaffected) or the global `llm_context` fallback — the same conservative "not probed yet" state the cache is in before startup priming — rather than a confidently wrong value belonging to a different endpoint. The slow part still runs after the response, so a dead endpoint can't stall an admin write. Not locked, deliberately: rebinding the dict is atomic, and taking `_max_model_tokens_lock` here would reintroduce the stall.
- **Generation counter.** Invalidation alone isn't enough — a probe that started *before* the write finishes afterwards and republishes pre-write values straight over the invalidation. `prime_max_model_tokens` now captures the generation before probing and discards its results if it moved, so an in-flight refresh can't silently re-open the window the write just closed.

One thing that fell out of this: the probe was calling `get_openai_models` without a timeout, so it used the helper's generic 30s default and ignored `ModelEndpointConfig.timeout` entirely (the availability check you removed *did* pass it). Since probes run serially under the lock, an unreachable endpoint was setting how long the cache stayed empty — it was effectively controlling the width of your window. Each probe now gets its own endpoint's configured timeout.

Revert-proved rather than just asserted — dropping the inline invalidation fails 4 tests: the ordering test (`["invalidated", "primed"]`, which is the whole point — invalidating *inside* the background task would land no earlier than the re-probe it precedes) and all three mutation verbs (update / delete / set-default). There's also a test that a non-LLM write doesn't clear the cache as a side effect, one that a mid-probe invalidation discards its results, and one that the guard doesn't block the normal path.

Full unit suite green (2187 passed).

openrag/api/schemas/admin/model_endpoint_schemas.py
Ahmath-Gadji: Confirmed and fixed in 570db233 — and the collision you point at with the UI change is the part that makes it bite.

The validator ran on every payload regardless of type, so both names were reserved across the whole registry: an embedder / reranker / VLM carrying same-named provider metadata of another shape got a 422. Combined with the UI fix earlier in this PR — which now deliberately preserves those keys verbatim for non-LLM endpoints — the endpoint resubmits them on its next save, so a pre-existing endpoint holding e.g. `max_output_tokens: "auto"` would have become uneditable through the admin UI. The two changes were pulling against each other.

Scoped it to LLMs, which lands differently on the two schemas:

- **`CreateModelEndpointRequest`** declares `model_type` before `extra`, so the field validator reads it from `ValidationInfo.data` and scopes itself. (Absent only if `model_type` failed its own validation, in which case the request is rejected anyway.)
- **`UpdateModelEndpointRequest`** has no `model_type` at all — it's a path parameter — so it *cannot* scope the check, and validating there unconditionally is precisely what made the names global. The check moves to the update route, which reads the type from the path and raises the domain `ValidationError` (already mapped to 422 by the global handler, so the status is unchanged from the pydantic path). The schema keeps a comment explaining why it deliberately has no validator, so it doesn't get "helpfully" restored later.

The rejection contract for LLM endpoints is unchanged — same rule, same 422, still ahead of the service call.

Revert-proved in both directions, since scoping a validation can easily turn into deleting it:

| reverted | fails |
|---|---|
| create-side scoping | 36 — non-LLM types rejected again |
| route-side check dropped | 2 — LLM budgets stop being validated |
| route-side check made unconditional | 6 — non-LLM names reserved again |

The schema-level update test moved to the route rather than being deleted, and the create side gained an `embedder`/`reranker`/`vlm` × both keys × the full bad-value matrix asserting the names stay unreserved.

Full unit suite green (2187 passed).

hedhoud (approved):
Re-checked the two previously raised issues on the current head. Both are addressed, and the focused backend test set passed. Approving.

PR #694, opened 2026-07-17
fix(indexing): use the configured image-captioning prompt instead of the bare VLM default

### LINKED ISSUE #692: Image captioning ignores the configured prompt template (image_describer); always uses the bare "Describe this image in detail." fallback
The VLM image-captioning prompt template is never wired into the pipeline, so captioning silently falls back to the hardcoded default instead of the shipped `image_captioning_tmpl.txt`.

## Evidence
- `caption_stage` reads the prompt from the row: `prompt = row.get("caption_prompt")` (`openrag/services/workers/stages/caption.py`), but **`row["caption_prompt"]` is never written anywhere** in the codebase.
- With `prompt=None`, `VllmClient.caption_image` uses its fallback: `prompt or "Describe this image in detail."` (`openrag/services/inference/vllm_client.py`).
- The config fields meant to select the template have **no consumers**:
  - `PromptsConfig.image_describer = "image_captioning_tmpl.txt"` (`core/config/infrastructure.py`) — referenced nowhere.
  - `IndexationPipelineConfig.vlm_caption_prompt_name` (`core/config/indexation_pipeline.py`) — referenced nowhere.
- The extract path (`ParserFileSerializer.serialize`) also calls `_caption_document(processed, self._vlm, None)` — same bare fallback.

This looks like wiring that was dropped during the hexagonal refactor: the template file and config keys survived, but the code that loaded the template and set `caption_prompt` did not.

## Impact
- The rich shipped prompt (transcribe text, format tables as markdown, emit `[Image Placeholder]` for non-informative images) is **not used** — captions are generic free-form descriptions.
- The chunker depends on it: `BaseChunker._prepare_md_elements` **skips image blocks containing `[image placeholder]`**, which the bare prompt never produces — so non-informative images now pollute chunks instead of being dropped, and tables inside images are no longer transcribed into markdown.

## Proposed fix
Mirror how the contextualizer prompt is wired (`indexer_pool.py` uses `load_template_by_key(cfg.paths.prompts_dir, cfg.prompts, "chunk_contextualizer")`):
1. Load the caption template once at indexer setup via `load_template_by_key(..., "image_describer")`.
2. Inject it into the caption stage (through the existing `row["caption_prompt"]` seam, honoring a per-row override).
3. Do the same on the extract path (`ParserFileSerializer`).

Follow-up (separate): honor the per-preset `vlm_caption_prompt_name` override (needs prompt-store lookup, like named model endpoints).

andyne13:
Closes #692.

## Problem
The VLM image-captioning prompt template was never wired into the pipeline, so captioning silently used the client fallback `"Describe this image in detail."` instead of the shipped `image_captioning_tmpl.txt`. `caption_stage` reads `row["caption_prompt"]`, but nothing ever set it; `PromptsConfig.image_describer` and `IndexationPipelineConfig.vlm_caption_prompt_name` had no consumers (wiring dropped in the hexagonal refactor).

Impact: the rich prompt (transcribe text, format tables as markdown, emit `[Image Placeholder]` for non-informative images) was not used — and the chunker depends on it (`BaseChunker._prepare_md_elements` skips blocks containing `[image placeholder]`, which the bare prompt never produces).

## Fix
Mirror how the contextualizer prompt is loaded (`load_template_by_key(cfg.paths.prompts_dir, cfg.prompts, "chunk_contextualizer")`):
- New `load_caption_prompt(config)` loads `prompts.image_describer`, degrading to `None` (VLM built-in default) on a missing template / partial config so indexing never fails.
- Injected on both caption paths: the indexing pipeline (new `IndexingPipeline.caption_prompt`, set on `row["caption_prompt"]` and honoring a per-row override) and the extract path (`ParserFileSerializer`).

The `row["caption_prompt"]` seam is intentionally the hook for the planned DB-backed prompt management (a future `PromptService.resolve(partition, "image_captioning")` populates it per-partition; this disk load is that design's tier-3 fallback) — noted in code comments.

## Tests
- New: template reaches the VLM; an explicit per-row `caption_prompt` overrides the configured default.
- All `tests/unit/services/workers/` pass (176); `ruff check` / `ruff format --check` / `check_layer_imports.py` green.

## Verification caveat
Verified at the wiring level (the VLM now receives the template) via unit tests. True end-to-end captioning (a real caption containing `[Image Placeholder]`, chunker skipping it) needs a VLM+GPU and should be smoke-tested on a deployment before merge.

Follow-up (separate): honor the per-preset `vlm_caption_prompt_name` override — deferred, and likely superseded by the `partition_prompts` mechanism in the prompt-management plan.

## Summary by CodeRabbit

* **New Features**
  * Image captioning now supports a configurable prompt template, which is loaded during worker/pipeline setup and forwarded to the captioning stage.
  * A pipeline default caption prompt can be applied automatically, while per-row/per-image caption prompt values override it when provided.
* **Bug Fixes**
  * Caption prompt loading no longer relies on a global captioning VLM being configured; if the template can’t be loaded, captioning safely falls back to the VLM’s default.
* **Tests**
  * Added unit tests to verify prompt forwarding, per-row override precedence, and prompt loading without a global caption VLM default.

### DIFF
--- a/openrag/services/workers/indexer_pool.py
+++ b/openrag/services/workers/indexer_pool.py
@@ -35,7 +35,11 @@ class IndexerWorkerActor:
         from core.utils.logging import get_logger
         from services.storage.milvus_store import MilvusVectorStore
         from services.storage.postgres_store import PostgresStore
-        from services.workers.parsers.parser_dispatcher import build_caption_vlm, build_parser_dispatcher
+        from services.workers.parsers.parser_dispatcher import (
+            build_caption_vlm,
+            build_parser_dispatcher,
+            load_caption_prompt,
+        )
         from services.workers.pipeline_builder import build_indexing_pipeline
 
         cfg = load_config()
@@ -43,6 +47,7 @@ class IndexerWorkerActor:
         parser = build_parser_dispatcher(cfg)
         parser_factory = _build_parser_factory(parser)
         vlm = build_caption_vlm(cfg)
+        caption_prompt = load_caption_prompt(cfg) if vlm is not None else None
         chunker = _build_chunker(cfg)
         embedder_factory = _build_embedder_factory(cfg)
         vlm_factory = _build_vlm_factory(cfg)
@@ -68,6 +73,7 @@ class IndexerWorkerActor:
             embedder=embedder,
             vector_store=self._vector_store,
             vlm=vlm,
+            caption_prompt=caption_prompt,
             timeouts=_build_pipeline_timeouts(cfg),
             chunker_factory=_build_chunker_from_config,
             parser_factory=parser_factory,
--- a/openrag/services/workers/indexer_pool.py
+++ b/openrag/services/workers/indexer_pool.py
@@ -47,7 +47,14 @@ class IndexerWorkerActor:
         parser = build_parser_dispatcher(cfg)
         parser_factory = _build_parser_factory(parser)
         vlm = build_caption_vlm(cfg)
-        caption_prompt = load_caption_prompt(cfg) if vlm is not None else None
+        # Loaded unconditionally: a preset can caption through a *named* VLM
+        # endpoint (resolved per-row via vlm_factory, see
+        # IndexingPipeline._select_vlm) even when no global default is
+        # configured, so gating this on `vlm is not None` would silently skip
+        # the prompt for that deployment shape. load_caption_prompt already
+        # degrades to None on any load failure, so there's no safety
+        # trade-off in always calling it.
+        caption_prompt = load_caption_prompt(cfg)
         chunker = _build_chunker(cfg)
         embedder_factory = _build_embedder_factory(cfg)
         vlm_factory = _build_vlm_factory(cfg)
--- a/openrag/services/workers/parsers/file_serializer.py
+++ b/openrag/services/workers/parsers/file_serializer.py
@@ -25,11 +25,16 @@ class ParserFileSerializer(FileSerializer):
 
     def __init__(self) -> None:
         from core.config import load_config
-        from services.workers.parsers.parser_dispatcher import build_caption_vlm, build_parser_dispatcher
+        from services.workers.parsers.parser_dispatcher import (
+            build_caption_vlm,
+            build_parser_dispatcher,
+            load_caption_prompt,
+        )
 
         config = load_config()
         self._dispatcher = build_parser_dispatcher(config)
         self._vlm = build_caption_vlm(config)
+        self._caption_prompt = load_caption_prompt(config) if self._vlm is not None else None
         # Global gate for captioning images embedded in other documents.
         # Standalone image files are always captioned (see ``serialize``).
         self._image_captioning = bool(config.loader.image_captioning)
@@ -57,7 +62,7 @@ class ParserFileSerializer(FileSerializer):
             and (document.content_type is DocumentType.IMAGE or self._image_captioning)
         )
         if should_caption:
-            processed = await _caption_document(processed, self._vlm, None)
+            processed = await _caption_document(processed, self._vlm, self._caption_prompt)
 
         return "\n\n".join(block.text for block in processed.text_blocks if block.text)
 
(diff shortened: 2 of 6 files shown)


hedhoud:
One small gap I think we should close before merging: the caption prompt is currently loaded only when the legacy/env VLM is configured. If a deployment uses only a VLM registered from the Admin UI, the VLM can still be resolved dynamically, but the caption prompt stays empty and we fall back to the basic default prompt again.

So I’d load the caption prompt independently from the legacy VLM check, so both env-based and Admin UI / DB-backed VLMs use the same configured image-captioning prompt.

openrag/services/workers/indexer_pool.py
```diff
@@ -35,14 +35,19 @@ def __init__(self) -> None:
...
+            load_caption_prompt,
+        )
         from services.workers.pipeline_builder import build_indexing_pipeline
 
         cfg = load_config()
 
         parser = build_parser_dispatcher(cfg)
         parser_factory = _build_parser_factory(parser)
         vlm = build_caption_vlm(cfg)
+        caption_prompt = load_caption_prompt(cfg) if vlm is not None else None
```
Ahmath-Gadji: **Bug: caption prompt gate misses named/per-preset VLM endpoints (no global default configured)**

`caption_prompt` is only loaded when `build_caption_vlm(cfg)` (the *global* `cfg.vlm.base_url`) resolves to a client:

```python
vlm = build_caption_vlm(cfg)
caption_prompt = load_caption_prompt(cfg) if vlm is not None else None
```

But `vlm_factory` (`_build_vlm_factory(cfg)`, built unconditionally a few lines below) lets an indexation preset caption through a **named** VLM endpoint registered independently of the global default — `IndexationPipelineConfig.vlm: str | None` (`endpoint name; None = use global default`, `openrag/core/config/indexation_pipeline.py:33`), resolved per-row in `IndexingPipeline._select_vlm` via `self.vlm_factory(name)`.

So for a deployment that only registers a named VLM endpoint (via the model-endpoint admin API, `openrag/api/routers/admin/model_endpoints.py`) and points a preset at it — with `VLM_BASE_URL`/global default left unset — `build_caption_vlm(cfg)` returns `None`, `caption_prompt` is never loaded, and `IndexingPipeline.run()`'s `if self.caption_prompt is not None: row.setdefault(...)` never fires even though a real VLM (resolved via `vlm_factory`) does caption the row. That preset silently reproduces the exact bug this PR fixes (#692) — bare client default instead of the configured `image_describer` template.

`load_caption_prompt` already degrades gracefully to `None` on any error, so the `vlm is not None` gate isn't buying safety — it's just an incomplete predictor of "will anything ever caption", since named VLMs can be added later via the admin API/DB (hydrated lazily, per the comment at `indexer_pool.py:89-94`) independent of the global default.

`openrag/services/workers/indexer_pool.py:50`

(Note: `file_serializer.py`'s equivalent gate is fine — that path has no `vlm_factory`/named-VLM override, `self._vlm` is the only VLM it ever captions with.)

Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/c871c062924b17394b6b110f23517f0117060582.

Removed the `vlm is not None` gate — `load_caption_prompt(cfg)` is now called unconditionally in `IndexerWorkerActor.__init__`. It already degrades to `None` on any load failure, so there's no downside to always loading it, and now a preset that only captions through a named/per-row VLM endpoint (no global default configured) still gets the configured `image_describer` template. Added `test_indexer_pool_loads_caption_prompt_without_global_vlm_default` to cover the regression.

Ahmath-Gadji:
@hedhoud Good catch — flagged and fixed the same gap independently in commit https://github.com/linagora/openrag/commit/c871c062924b17394b6b110f23517f0117060582 (see the inline thread on `indexer_pool.py:50` for details): `load_caption_prompt(cfg)` is now called unconditionally instead of being gated on the legacy/env VLM, so Admin-UI/DB-registered named VLM endpoints get the configured prompt too. Pushed and CI is re-running.

Note on scope: this only makes the *same* disk-loaded `image_describer` template reach captioning regardless of which VLM (legacy/env or named) is used. Configuring a **different** prompt per named/Admin-UI VLM endpoint isn't supported yet — that's a new feature and will be handled in an upcoming release.

Ahmath-Gadji (approved):
LGTM

PR #700, opened 2026-07-20
fix(quota): arbitrate the reserved file-slot release instead of leaking it

Ahmath-Gadji:
Closes the cancel-before-start quota-slot leak that #677 documents as a known gap.

> **Stacked on #677** (base = `fix/660-durable-job-state`). It builds directly on the reserve/release machinery introduced there, so it cannot land first. Review #677 before this. Once #677 merges, this retargets to `develop` cleanly.

## The leak

After dispatch the worker owns the uploader's reserved file slot and hands it back in `IndexerWorker.process_file`'s `finally`. When `ray.cancel` retires a task *before* that body runs, the `finally` never executes and nothing else gives the slot back. The `CANCELLED` row written by `cancel_task` is terminal, so the loss is indistinguishable from a clean cancel — the uploader's quota is permanently one narrower, silently.

`cancel_task` could not simply release. A task cancelled *mid-flight* does run its `finally`, so releasing in both places would drive `file_count` below reality and hand the user free quota. The two cases were distinguishable only by how far the worker happened to get — which is exactly the kind of thing the caller cannot observe.

## The fix

`TaskStateManager` is a single-threaded Ray actor that already arbitrates precisely this shape of race (`set_cancelled_if_active`, `set_failed_if_not_cancelled`). A one-shot `claim_quota_release` compare-and-set is the idiomatic primitive here:

- `process_file`'s `finally` asks before releasing.
- `cancel_task` asks after claiming the cancellation.
- The `indexer_pool` setup-failure release asks too — a cancellation *during* setup reaches both it and `cancel_task`.

Exactly one caller is told to release; the outcome no longer depends on worker progress. The dispatcher records `quota_reserved` on the task at admission so the actor knows whether a slot is outstanding at all (`put_file` reuses an existing row and reserves nothing).

**This also closes the dispatch-partial-success double-release** that #677 accepts as a deliberate trade-off: the worker and the request teardown can no longer both consume the same reservation.

## Failure behaviour

Biased the same way the rest of the reserve/release design is — **an unreachable arbiter releases anyway.** An undercount is recoverable and self-heals under the #676 recount; a leak is permanent and silent. `_release_cancelled_slot` is best-effort throughout, because a cancellation the user already asked for and already saw must not fail on quota bookkeeping.

The one case that stays open is genuinely un-arbitrable and belongs to #676: outright process death between reserve and dispatch, and an unreachable Postgres inside the release itself.

## Why now rather than in #676

#685 shipped a cancel-all UI that fires one `cancelTask` per selected task (`Promise.allSettled(activeTaskIds.map(...))`). The leak was previously one slot per manual cancel; it is now N slots from a single click, which moves it out of "acceptable known gap" territory.

## Verification

- `ruff check` / `ruff format --check` clean; `scripts/check_layer_imports.py` → `layer import guard: OK`.
- **Unit: 1839 passed, 0 failed** (1831 on the #677 base, +8 here).
- **Integration against real Postgres: 119 passed, 1 failed** — unchanged from the base; the failure is the pre-existing `test_create_is_idempotent_per_name`, fixed separately in #691.
- New tests cover both directions of the race and the degraded paths: the claim is grantable exactly once; a task that reserved nothing has nothing to claim; an unknown task claims rather than leaks; the cancel releases when the worker never started; the cancel declines when the worker already owns it; a failing release does not fail the cancellation; and an unreachable state actor still releases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reserved file-slot handling for indexing tasks canceled before execution by using single-ownership arbitration for quota release.
  * Prevented duplicate quota releases while ensuring the correct path returns reserved slots when a worker loses or fails during release.
  * Added best-effort safeguards so release failures during cancellation don’t block task completion.
* **Tests**
  * Expanded unit coverage for quota-release claiming, cancellation scenarios, and quota state arbitration edge cases.
* **Documentation**
  * Clarified quota reservation/release ownership behavior after dispatch and handling of unreachable arbitration.

### DIFF
--- a/openrag/services/workers/dispatcher.py
+++ b/openrag/services/workers/dispatcher.py
@@ -63,6 +63,7 @@ class WorkerDispatcher(IndexingDispatcher):
         collection: str,
         timeout: float = DEFAULT_TIMEOUT,
         job_repo: Any = None,
+        user_repo: Any = None,
     ) -> None:
         self._pool = pool
         self._tsm = task_state_manager
@@ -76,6 +77,10 @@ class WorkerDispatcher(IndexingDispatcher):
         # is absent, job state degrades to the in-memory actor — the pre-#660
         # behaviour.
         self._job_repo = job_repo
+        # Optional for the same reason as ``job_repo``. Without it a cancel that
+        # beats the worker to the task cannot return the reserved slot, which is
+        # the pre-#664 behaviour rather than a new failure.
+        self._user_repo = user_repo
         self._last_job_purge_at: float | None = None
 
     async def _call(self, future: Any, task_description: str) -> Any:
@@ -95,6 +100,7 @@ class WorkerDispatcher(IndexingDispatcher):
         partition: str,
         metadata: dict[str, Any],
         user_id: int | None,
+        quota_reserved: bool = False,
     ) -> bool:
         remote = _remote_actor_method(self._tsm, "set_queued_details")
         if remote is not None:
@@ -105,6 +111,7 @@ class WorkerDispatcher(IndexingDispatcher):
                     partition=partition,
                     metadata=metadata,
                     user_id=user_id,
+                    quota_reserved=quota_reserved,
                 ),
                 task_description=f"set_queued_details({task_id})",
             )
@@ -121,6 +128,7 @@ class WorkerDispatcher(IndexingDispatcher):
                 partition=partition,
                 metadata=metadata,
                 user_id=user_id,
+                quota_reserved=quota_reserved,
             ),
             task_description=f"set_details({task_id})",
         )
@@ -168,6 +176,9 @@ class WorkerDispatcher(IndexingDispatcher):
             partition=partition,
             metadata=user_metadata,
             user_id=user.get("id") if user else None,
+            # Tells the actor a slot is outstanding for this task, so
+            # ``claim_quota_release`` can arbitrate who gives it back.
+            quota_reserved=quota_reserved,
         )
         if not accepted:
             raise RuntimeError(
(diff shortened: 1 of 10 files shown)


openrag/services/workers/dispatcher.py
```diff
@@ -392,17 +400,49 @@ async def cancel_task(self, task_id: str) -> bool:
...
+            return
+        try:
+            details = await self._call(
+                self._tsm.get_details.remote(task_id),
+                task_description=f"get_details({task_id})",
+            )
+            user_id = (details or {}).get("user_id")
+            if user_id is None:
+                return
+            claimed = await self._call(
```
hedhoud, line 729: This claim does not distinguish a task that never started from one that has already created its catalog row. If cancellation lands after add_file_to_partition succeeds but before COMPLETED is reported, the dispatcher can win this claim and decrement file_count even though the file remains stored. Could the worker take ownership when processing starts, or otherwise mark the reservation consumed atomically, so only a true cancel-before-start releases here? This also needs a regression test for that timing window.

openrag/services/workers/dispatcher.py
```diff
@@ -107,6 +112,9 @@ async def dispatch_indexing(
                 partition=partition,
                 metadata=user_metadata,
                 user_id=user.get("id") if user else None,
+                # Tells the actor a slot is outstanding for this task, so
+                # ``claim_quota_release`` can arbitrate who gives it back.
+                quota_reserved=quota_reserved,
```
hedhoud, line 131: TaskStateManager is detached and reused under the same actor name. During a rolling upgrade, the existing actor will not accept the new quota_reserved argument and also will not expose claim_quota_release, so new indexing requests can fail until Ray is manually restarted. Could we version the actor name or add an explicit compatibility/recreation guard before changing this contract?

openrag/services/workers/dispatcher.py
```diff
@@ -392,17 +400,49 @@ async def cancel_task(self, task_id: str) -> bool:
...
+            if user_id is None:
+                return
+            claimed = await self._call(
+                self._tsm.claim_quota_release.remote(task_id),
+                task_description=f"claim_quota_release({task_id})",
+            )
+            if not claimed:
+                return
+            await self._user_repo.release_file_slot(user_id)
+        except Exception as exc:  # noqa: BLE001 - the cancellation itself already succeeded
```
hedhoud, line 736: If get_details succeeds but this arbitration RPC raises or times out, the outer handler only logs and leaves the slot reserved. That reintroduces the permanent leak this PR is meant to close and differs from the worker-side fallback. Could we release with the already-resolved user_id when the claim cannot be made, and cover that path with a regression test?

hedhoud (changes requested):
The arbitration direction makes sense, but two correctness blockers and one degraded-path leak remain. I’ve left the details inline.

PR #701, opened 2026-07-20
feat(admin): show job creation time and duration

hedhoud:
## Context

The Jobs view shows task status but gives operators no indication of when a task was created or how long it has been processing. This makes it difficult to distinguish recent work from slow or stalled indexing.

## Change

The Jobs table now shows creation time and total duration. Duration advances for active tasks using the existing polling cycle and stops when a task completes, fails, or is cancelled. Existing filtering, permissions, and cancellation behavior remain unchanged.

## Validation

- Backend timing and job-service tests pass.
- All 136 UI tests pass.
- UI lint and production build pass.
- Python lint and formatting checks pass.

The broader backend run completed 511 tests successfully; nine unrelated Ray pipeline tests remain blocked by local permissions on `db/.gitignore`.

## Summary by CodeRabbit

* **New Features**
  * Updated the admin task list to show **Created** (localized timestamp) and **Duration** (human-readable) columns.
  * Task listing API responses now include **created time** and **duration** fields when available.
* **Bug Fixes**
  * Improved resilience when timing metadata is missing or malformed (legacy cases).
  * Ensures task timing is tracked accurately even after restarts, while keeping timing metadata hidden from task details.

### DIFF
--- a/openrag/services/workers/dispatcher.py
+++ b/openrag/services/workers/dispatcher.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+import asyncio
 import traceback
 import uuid
+from datetime import UTC, datetime
 from typing import Any
 
 from core.indexing.dispatcher import IndexingDispatcher
+from core.models.catalog import TASK_CREATED_AT_METADATA_KEY, TASK_FINISHED_AT_METADATA_KEY
 from core.utils.conts import is_internal_metadata_key, strip_internal_metadata
 from core.utils.logging import get_logger
 from ray.exceptions import TaskCancelledError
@@ -56,6 +59,7 @@ class WorkerDispatcher(IndexingDispatcher):
         self._workspace_repo = workspace_repo
         self._collection = collection
         self._timeout = timeout
+        self._completion_tasks: set[asyncio.Task[None]] = set()
 
     async def _call(self, future: Any, task_description: str) -> Any:
         from services.workers.ray_utils import call_ray_actor_with_timeout
@@ -140,12 +144,16 @@ class WorkerDispatcher(IndexingDispatcher):
         task_id = uuid.uuid4().hex
 
         user_metadata = {key: value for key, value in metadata.items() if key not in {"file_id", "source"}}
+        user_metadata[TASK_CREATED_AT_METADATA_KEY] = _utc_now_iso()
+        task_details = {
+            "file_id": metadata.get("file_id"),
+            "partition": partition,
+            "metadata": user_metadata,
+            "user_id": user.get("id") if user else None,
+        }
         accepted = await self._set_queued_details(
             task_id,
-            file_id=metadata.get("file_id"),
-            partition=partition,
-            metadata=user_metadata,
-            user_id=user.get("id") if user else None,
+            **task_details,
         )
         if not accepted:
             raise RuntimeError(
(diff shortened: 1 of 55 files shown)


ui/src/pages/admin/jobs/list.tsx
```diff
@@ -109,6 +125,20 @@ const columns: ColumnDef<TaskListItem, unknown>[] = [
       <TruncatedValue value={str(row.original.details?.partition)} className="max-w-[180px] sm:max-w-[240px]" />
     ),
   },
+  {
+    accessorKey: "created_at",
+    header: "Created",
```
Ahmath-Gadji: **Suggestion: make the Created column sortable**

Same idea as above — `created_at` is already an ISO string (sorts lexically = chronologically, same as `documents/list.tsx`'s "Indexed" column), so swapping `header: "Created"` for `header: ({ column }) => <SortableHeader column={column} title="Created" />` would enable sorting jobs by creation time with minimal change.

Ahmath-Gadji:
**Suggestion: make the File column sortable too**

`ui/src/pages/admin/jobs/list.tsx:112-120` — the File column (`id: "file"`) uses a plain string header. Same idea as the inline suggestion on the Created column: the shared `DataTable` already supports column sorting via `SortableHeader`, and `documents/list.tsx`'s "Filename" column already does this (`accessorFn: (f) => fileLabel(f).toLowerCase()` + `header: ({ column }) => <SortableHeader column={column} title="Filename" />`). Wiring the same pattern here would let operators sort jobs by filename, alongside sorting by Created.

ui/src/pages/admin/jobs/list.tsx
hedhoud: Done. The Created column now uses the shared sortable header and sorts by its ISO timestamp, so the order is chronological.

hedhoud:
Done. The File column now uses the shared sortable header and sorts by the normalized filename. I also added a test covering both File and Created sorting.

Ahmath-Gadji:
**Scope concern: the merge commit bundles unrelated, unreviewed actor-versioning code**

Commit `ea3afcaf` ("Merge develop and refresh job timing actors") isn't just a merge — I diffed it against the actual tip of `develop` (`d450f149`, confirmed current via the GitHub API) and `TaskStateManagerV2`, `TASK_STATE_ACTOR_GENERATION`, the file-delete fencing (`begin_file_delete`/`end_file_delete`), and `set_queued_details` do not exist anywhere on `develop`. This is new code, invented and bundled directly into this commit, spanning `task_state.py`, `indexer_pool.py`, `dispatcher.py`, and `cluster.py` — well outside this PR's stated scope (job creation time/duration + sortable columns).

This also isn't just a scope issue: @chatgpt-codex-connector already flagged two **P1** findings in that new code (stale detached actor after deploy leaving the Jobs table blank until a state-dropping manual restart, and un-drained legacy V1 workers racing V2 — including a scenario where a deleted file's chunks get written back by a stale V1 worker). Those read as real correctness/data-safety concerns, not nitpicks.

Given the size and risk of this addition relative to the rest of the PR, could this actor-versioning work be split into its own PR for dedicated review, or otherwise have the P1s addressed before this merges?

openrag/services/workers/dispatcher.py
```diff
@@ -180,17 +188,43 @@ async def dispatch_indexing(
...
+            return
+        tracker = asyncio.create_task(self._record_completion(task_id, task, task_details))
+        self._completion_tasks.add(tracker)
+        tracker.add_done_callback(self._completion_tasks.discard)
+
+    async def _record_completion(self, task_id: str, task: Any, task_details: dict[str, Any]) -> None:
+        try:
+            await asyncio.gather(task, return_exceptions=True)
+        finally:
+            await self._record_finished_at(task_id, task_details)
```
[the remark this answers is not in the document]
hedhoud: Fixed. Completion tracking now survives API restarts and recovers jobs missed during restart, so completed jobs keep their duration.

Ahmath-Gadji (dismissed):
Reviewed the full history end-to-end, including the follow-up commits addressing sorting and the actor-versioning concerns.

- **Feature**: `created_at`/`duration_ms` now flow through `details.metadata` (`_openrag_job_created_at`/`_openrag_job_finished_at`), read via the existing `set_details`/`get_details` contract — no dependency on redeploying the detached `TaskStateManager` actor. Displayed in the Jobs table with graceful degradation for tasks lacking timing data.
- **Sorting**: File and Created columns are sortable via the shared `SortableHeader`, matching the pattern used on `documents/list.tsx`. Tested.
- **P1s resolved**: the `TaskStateManagerV2`/actor-generation scheme is fully removed — confirmed `task_state.py`, `indexer_pool.py`, and `cluster.py` are now byte-identical to `develop`'s current tip. No more stale-actor or dual-generation-worker race.
- **P2 resolved**: completion tracking now runs in a dedicated, durable `TaskCompletionTracker` Ray actor (detached, `max_restarts=-1`) with a `recover()` path that reattaches to in-flight tasks and backfills terminal ones missed during API downtime, without overwriting an already-recorded completion time (tested).
- **Test coverage**: metadata extraction, malformed-metadata tolerance, live duration for active tasks, internal-key stripping from public API responses, and the tracker's completion/recovery/backfill/non-overwrite paths are all covered.
- CI green across tests, lint, api-tests, milvus-integration, layer-import-guard.

Nice work tightening the scope back down after the initial merge pulled in unrelated changes. Approving.

Ahmath-Gadji (approved):
LGTM

PR #705, opened 2026-07-20
fix(retrieval): extract chat content so multiQuery and hyde stop crashing

### LINKED ISSUE #703: multiQuery and hyde retrieval strategies crash on every query — LLM.chat() returns a dict, both treat it as str
## Problem

The `multiQuery` and `hyde` retrieval strategies raise `AttributeError` on **every** query. Two of the three selectable retrieval strategies cannot work.

`LLM.chat()` is declared to return a `dict` and does so, but both expansion retrievers treat the return value as a `str`.

**Contract** — `openrag/core/llm/llm.py:18`:
```python
async def chat(self, messages: list[dict[str, str]], **kwargs) -> dict:
```
`VLLMClient.chat` returns `_parse_response(resp)` i.e. `resp.json()` (`openrag/services/inference/vllm_client.py:159`).

**Misuse 1** — `openrag/core/retrieval/retriever.py:135-138`:
```python
response = await self.llm.chat([{"role": "user", "content": prompt}])
queries = split_multi_query_response(response)[: self.k_queries]
```
`split_multi_query_response` does `response.split(separator)` (`openrag/core/prompts/query_rewriter.py:35`) → `AttributeError: 'dict' object has no attribute 'split'`.

**Misuse 2** — `openrag/core/retrieval/retriever.py:193`:
```python
hyde = (await self.get_hyde(query)).strip()
```
`get_hyde` returns `await self.llm.chat(...)` unchanged (`:186`) → `AttributeError: 'dict' object has no attribute 'strip'`.

The correct extraction already exists elsewhere in the same client — `vllm_client.py:413` does `_parse_response(resp)["choices"][0]["message"]["content"]`. This is an omission, not a contract ambiguity.

## Reachability

Not a dead code path. `type` is a per-partition preset field:

- `openrag/core/config/retrieval_pipeline.py:20` — `type: Literal["single", "multiQuery", "hyde"] = "single"`
- Built at `openrag/services/orchestrators/retrieval_service.py:103,110,136,147`
- Exposed in the admin UI as a live Strategy dropdown (`ui/src/pages/admin/presets.tsx:627`); selecting a non-`single` value even reveals a dedicated "Expansion LLM" selector

Any partition whose preset selects `multiQuery` or `hyde` fails every chat and search request.

## Why CI is green

The test double violates the port contract — `tests/unit/core/retrieval/test_retriever.py:58`:
```python
async def chat(self, messages: list[dict], **kwargs) -> str:
```
The fake returns `str` where the real port returns `dict`, so the crash is unreachable in tests. This is the same class of gap as #692: the fake is more permissive than reality.

## Suggested fix

1. Add a small helper (e.g. `_content(resp)`) that extracts `resp["choices"][0]["message"]["content"]`, and use it at both call sites.
2. Change the test double's `chat` to return the OpenAI dict shape so the contract is enforced by tests.
3. Add a regression test asserting each strategy returns queries for a dict-shaped response.

## Verified on

`develop` @ `e61479b4`.

andyne13:
Closes #703.

## Problem

`LLM.chat` returns the raw provider payload (a `dict`, `openrag/core/llm/llm.py:18`), but both expansion retrievers treated it as a `str`:

- `retriever.py:137` — `split_multi_query_response(response)` → `.split()` on a dict
- `retriever.py:193` — `get_hyde()` returned the dict straight into `.strip()`

Every `multiQuery` / `hyde` query raised `AttributeError`. `type` is a per-partition preset field (`retrieval_pipeline.py:20`) exposed as a live Strategy dropdown in the admin UI, so two of the three selectable strategies could not work — any partition configured that way failed every chat and search request.

## Fix

Add `chat_content()` alongside the `LLM` port and use it at both call sites. The same extraction already existed inline elsewhere (`vllm_client.py:413`, `query_service.py:235`); this gives it one name.

It returns `""` on a malformed/empty payload rather than raising, so a non-compliant provider degrades to "no expansion" (callers already fall back to the seed query) instead of failing the whole search.

## Why this shipped green

The test double was **more permissive than the port it stood in for** — `test_retriever.py:58` declared `async def chat(...) -> str` while the real port returns `dict`, making the crash unreachable in tests. Same class of gap as #692.

`FakeLLM` now returns the OpenAI-shaped dict the real port declares.

## Verification

Reverting **only** the production call sites while keeping the corrected fake fails 10 tests — 7 of which passed before this PR:

```
FAILED test_hyde_retriever_uses_hyde_only_by_default
FAILED test_hyde_retriever_combine_appends_original_query
FAILED test_multi_query_retriever_splits_llm_response
FAILED test_multi_query_retriever_caps_response_to_k_queries
... 10 failed, 18 passed
```

With the fix in place: **1810 passed**, `ruff check` / `ruff format --check` / `check_layer_imports.py` all green.

New coverage: `chat_content` extraction, 7 malformed-payload shapes, and end-to-end dict handling for both strategies.

## Note

Base is `develop` and the repo default branch is `main`, so `Closes #703` will not auto-close on merge — needs closing manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of chat responses during retrieval.
  * Multi-query and HyDE retrieval now correctly process structured LLM responses.
  * Added safe fallbacks when responses are malformed or missing content, including retaining the original query when needed.

* **Tests**
  * Added coverage for valid, malformed, and unexpected chat response formats.

### DIFF
--- a/openrag/core/llm/llm.py
+++ b/openrag/core/llm/llm.py
@@ -4,6 +4,27 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator
+from typing import Any
+
+
+def chat_content(response: Any) -> str:
+    """Extract the assistant message text from an OpenAI-shaped chat response.
+
+    ``LLM.chat`` returns the raw provider payload (a ``dict``), not a string —
+    callers that want the text must reach into ``choices[0].message.content``.
+    Doing that inline is easy to forget: ``multiQuery``/``hyde`` both treated the
+    dict as a ``str`` and raised ``AttributeError`` on every query (#703).
+
+    Returns ``""`` for a malformed or empty payload rather than raising, so a
+    non-compliant provider degrades to "no expansion" instead of failing the
+    whole search — callers already handle an empty result by falling back to the
+    original query.
+    """
+    try:
+        content = response["choices"][0]["message"]["content"]
+    except (KeyError, IndexError, TypeError):
+        return ""
+    return content if isinstance(content, str) else ""
 
 
 class LLM(ABC):
--- a/openrag/core/retrieval/retriever.py
+++ b/openrag/core/retrieval/retriever.py
@@ -22,7 +22,7 @@ from abc import ABC, abstractmethod
 from itertools import chain as ichain
 from typing import Any
 
-from core.llm.llm import LLM
+from core.llm.llm import LLM, chat_content
 from core.models.chunk import Chunk
 from core.prompts.query_rewriter import (
     build_hyde_prompt,
@@ -135,7 +135,7 @@ class MultiQueryRetriever(BaseRetriever):
         response = await self.llm.chat([{"role": "user", "content": prompt}])
         # Cap to k_queries — a non-compliant LLM response can otherwise fan
         # out far more searches than configured.
-        queries = split_multi_query_response(response)[: self.k_queries]
+        queries = split_multi_query_response(chat_content(response))[: self.k_queries]
         return queries or [query]
 
     async def retrieve(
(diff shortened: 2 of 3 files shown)


hedhoud:
@codex review

hedhoud (approved):
Reviewed the retrieval flow and provider response handling. MultiQuery and HyDE now handle the LLM response correctly and fall back safely when content is missing. Tests pass against the current develop branch. Approved.

PR #716, opened 2026-07-20
fix(security): block caller-set 'source' on REST metadata writes

### LINKED ISSUE #713: PATCH file metadata allows overwriting 'source' — cross-tenant document read via /static (upload and MCP paths both block it)
## Problem

The REST `PATCH` metadata path lets a caller overwrite `source`, while the two sibling write paths both block it.

`openrag/services/orchestrators/indexing_service.py:236-245`:
```python
async def update_metadata(self, file_id, metadata, partition, user) -> None:
    metadata = dict(metadata or {})
    metadata["file_id"] = file_id
    await self._dispatcher.update_file_metadata(file_id, metadata, partition, user)
```
`dispatcher.update_file_metadata` (`:364-370`) applies only `strip_internal_metadata`, which drops keys prefixed `_openrag` (`core/utils/conts.py:13-20`). `source` is **not** covered — verified:
```
input : {'source': '/app/data/other_tenant_file.pdf', 'note': 'x', '_openrag_internal': 'y'}
output: {'source': '/app/data/other_tenant_file.pdf', 'note': 'x'}
>>> source SURVIVES: True
```

Both sibling paths do protect it:
- upload — `dispatcher.dispatch_indexing:142`: `if key not in {"file_id", "source"}`
- MCP — `mcp_service.py:64-74`: `_PROTECTED_METADATA_KEYS` includes `source`

## Impact

`source` is the filesystem path read by `GET /static/{extract_id}` (`api/routers/user/download.py:56-63`). Uploads are stored flat in one shared `data_dir` across tenants, and `/static` authorizes on the **chunk's** partition — still the attacker's own after a PATCH.

So a user with `editor` on any partition they legitimately hold can repoint `source` at another tenant's document and read it. Confinement to `DATA_DIR` holds (`is_relative_to` check), so this is cross-tenant document read, not arbitrary host-file read.

`created_by` is equally writable, which corrupts quota accounting.

## Fix

Apply the same protected-key filter inside `IndexingService.update_metadata` (and `copy_file`) so all three transports share one guard, rather than each call site re-implementing it.

andyne13:
Closes #713.

## Problem

`PATCH /partition/{partition}/file/{file_id}` let a partition **editor** overwrite `source` — the filesystem path served by `GET /static/{extract_id}`. That route authorizes on the **chunk's** partition, which this write leaves untouched, so an editor could point `source` at another tenant's upload in the shared flat data dir and read it back. `created_by` was equally writable, which corrupts quota accounting.

The other two write paths already blocked these keys:

| Path | Guard | Blocked `source`? |
|---|---|---|
| Upload — `dispatcher.dispatch_indexing:142` | inline `{"file_id", "source"}` | yes |
| MCP tools — `mcp_service._strip_protected_metadata` | `_PROTECTED_METADATA_KEYS` | yes |
| **REST PATCH — `IndexingService.update_metadata`** | none | **no** |

`strip_internal_metadata` doesn't help — it only drops keys prefixed `_openrag` (`core/utils/conts.py:13-20`), so `source` passes straight through to `entity.update(public_metadata)`.

## Fix

Rather than add a third copy of the filter, the key set and helper move into `core/utils/conts.py` as `PROTECTED_METADATA_KEYS` / `strip_protected_metadata`, and the transports share it. `copy_file` gets the same guard — it carries caller-supplied metadata onto the new rows.

`partition` is deliberately **not** in the set: the MCP update tool uses it as an authorized move control and re-checks editor access on the destination.

## Tests

Two regression tests assert the *effect* — that the keys are absent from what reaches the dispatcher — rather than that a filter was called:

- `test_update_metadata_drops_protected_keys` — all 7 protected keys stripped, user keys preserved
- `test_copy_file_drops_protected_keys`

Full suite green (1802), `ruff check` / `ruff format --check` / `check_layer_imports.py` all pass.

## Known follow-up

`dispatch_indexing` still hardcodes its own `{"file_id", "source"}` filter instead of using the shared set — a third copy of the same idea. Folding it in touches the upload path, so it's left out of a security patch deliberately. Worth a follow-up issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented caller-supplied metadata from overwriting server-managed fields during REST update and file copy operations.
  * Server-managed keys such as `source` are now automatically stripped from incoming metadata to avoid unintended persistence.
* **Tests**
  * Added unit tests covering metadata hardening for both update and copy paths, verifying protected keys are removed while required fields (like target identifiers) are still set.

### DIFF
--- a/openrag/core/utils/conts.py
+++ b/openrag/core/utils/conts.py
@@ -18,3 +18,32 @@ def is_internal_metadata_key(key: object) -> bool:
 
 def strip_internal_metadata(row: dict) -> dict:
     return {key: value for key, value in row.items() if not is_internal_metadata_key(key)}
+
+
+# Catalog/store-managed fields a caller must never be able to spoof through the
+# free-form metadata dict on a write path. ``source`` is the load-bearing one:
+# it is the filesystem path served by ``GET /static/{extract_id}``, so letting a
+# caller set it turns a metadata write into a cross-tenant file read (#713).
+#
+# ``partition`` is intentionally NOT here — the MCP update tool uses it as an
+# authorized move control and re-checks editor access on the destination.
+#
+# Lives in core/ (not in one orchestrator) so every transport shares one guard:
+# the upload path, the MCP tools, and the REST PATCH path previously each
+# re-implemented this and the REST one simply omitted it.
+PROTECTED_METADATA_KEYS: frozenset[str] = frozenset(
+    {"file_id", "source", "created_by", "file_size", "file_count", "_id", "vector", "text"}
+)
+
+
+def strip_protected_metadata(metadata: dict | None) -> tuple[dict, list[str]]:
+    """Drop server-managed keys from caller-supplied metadata.
+
+    Returns the cleaned dict and the sorted list of dropped keys so the caller
+    can log the rejection with its own context. Never mutates the input.
+    """
+    md = dict(metadata or {})
+    removed = sorted(k for k in md if k in PROTECTED_METADATA_KEYS)
+    for key in removed:
+        del md[key]
+    return md, removed
--- a/openrag/core/utils/conts.py
+++ b/openrag/core/utils/conts.py
@@ -25,6 +25,10 @@ def strip_internal_metadata(row: dict) -> dict:
 # it is the filesystem path served by ``GET /static/{extract_id}``, so letting a
 # caller set it turns a metadata write into a cross-tenant file read (#713).
 #
+# ``content_sha256`` is the server-computed dedup hash: a caller-set value would
+# corrupt dedup (spoofed ``DOCUMENT_CONTENT_EXISTS``), so it is protected here and
+# re-set from the server side on the copy path after this strip runs.
+#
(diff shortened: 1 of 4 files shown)


hedhoud:
Reviewed the metadata update and copy paths end to end. Protected fields are removed before reaching storage, valid custom metadata is preserved, and the tests pass against the current develop branch. No blocking issues found.

andyne13:
Good catch — fixed in 12faaba5. Switched to `logger.bind(keys=removed).warning(...)` so the dropped keys actually reach the structured log (previously `keys=removed` was treated as a str.format arg with no `{keys}` placeholder, so it was dropped). Matches the project's `logger.bind(...)` convention. This one matters for #713 specifically since the log is the audit trail for spoof attempts.

hedhoud (approved):
Approved. Protected metadata is removed before REST, copy, and MCP writes reach storage, including the server-owned content hash, while authorized partition moves and normal custom metadata remain intact. The latest structured-logging suggestion is useful but non-blocking.

PR #717, opened 2026-07-20
fix(security): ship web-search TLS verification enabled

### LINKED ISSUE #382: Web search fetcher disables TLS verification by default
`openrag/config/models.py:466` declares `_BaseWebSearchConfig.fetch_verify_ssl: bool = False`, and `openrag/components/websearch/content_fetcher.py:126` passes that value straight into `httpx.AsyncClient(verify=...)`. With the shipped default, every server-side fetch of a web search result skips TLS verification, leaving citation contents open to silent man-in-the-middle tampering even when `WEBSEARCH_API_TOKEN` is configured against an HTTPS provider.

Change the default to `True` and document a single opt-out env var for testing only. If a deployment legitimately needs to allow self-signed hosts, accept a list of hostnames rather than a global on/off, and refuse to start when both `fetch_verify_ssl=False` and the global default are active in the same process.

### LINKED ISSUE #714: Web search TLS verification is off at runtime: conf/config.yaml overrides the #404 fix, and its regression test bypasses the loader
## Problem

TLS verification for web-search page fetches is **off at runtime**, despite #382 having been fixed by merged PR #404.

Proven through the real loader:
```
load_config().websearch.fetch_verify_ssl = False
```

## Why — the fix was incomplete on the day it merged

Timeline:

| When | What |
|---|---|
| **2026-03-22** | `9779fea1` "YAML-first config" moves defaults into `conf/config.yaml`, writing `fetch_verify_ssl: false` (matching the Python default of the time, which was `False`) |
| **2026-05-21** | PR #404 fixes #382 — changes the **Python** default `False` → `True` and adds a regression test |

PR #404 never touched `conf/config.yaml`. Current state:

```python
openrag/core/config/retrieval.py:130   fetch_verify_ssl: bool = True     # fixed by #404
```
```yaml
conf/config.yaml:346                     fetch_verify_ssl: false           # still the old value
```

The loader reads YAML and then validates with Pydantic, so **the YAML value wins** and the Pydantic default is never reached. This is not a later regression — the fix simply never took effect.

## The regression test asserts the wrong thing

`tests/unit/services/websearch/test_content_fetcher.py:298-307`:
```python
def test_fetch_verify_ssl_default_is_true():
    from core.config.retrieval import StaanWebSearchConfig
    cfg = StaanWebSearchConfig()          # constructed directly — bypasses the YAML loader
    assert cfg.fetch_verify_ssl is True
```
It instantiates the model directly, so it asserts the **declared default**, never the **effective loaded value**. It has been green since May while TLS verification is disabled in production — the same "assert the declaration, not the effect" pattern as #692.

## Impact

Out of the box, OpenRag fetches URLs returned by the web-search provider with TLS verification disabled, and injects that text into the LLM context as a cited source. Consumed at `services/websearch/content_fetcher.py:102,232` (`verify=self.verify_ssl`).

The docs also state the default is `false`, so documentation currently matches the bug rather than the intent.

## Fix

1. Set `fetch_verify_ssl: true` in `conf/config.yaml` and correct the env-var docs.
2. Change the regression test to assert the value **through `load_config()`**, not on a directly-constructed model — otherwise the same class of bug recurs for any YAML-backed setting.

/cc @EnjoyBacon7 (filed #382), since this concerns the fix in #404.

andyne13:
Closes #714.

## Problem

TLS verification was **off at runtime** for web-search page fetches, whose text is injected into the LLM context as a cited source.

```
load_config().websearch.fetch_verify_ssl = False    # before this PR
```

`conf/config.yaml:346` shipped `fetch_verify_ssl: false`, and the loader merges YAML over the Pydantic default — so `core/config/retrieval.py:130` declaring `True` never took effect.

## Why the existing fix didn't hold

Not a regression. The timeline runs the other way:

| When | What |
|---|---|
| 2026-03-22 | `9779fea1` "YAML-first config" moves defaults into `conf/config.yaml`, writing `false` — correct at the time, since the Python default was then `False` |
| 2026-05-21 | PR #404 fixes #382 by flipping the **Python** default to `True`, and adds a regression test |

PR #404 never touched the YAML. The fix was incomplete the day it merged, two months after the refactor it might otherwise be blamed on.

## Why the regression test didn't catch it

```python
def test_fetch_verify_ssl_default_is_true():
    cfg = StaanWebSearchConfig()      # constructed directly — bypasses the YAML loader
    assert cfg.fetch_verify_ssl is True
```

It asserts the **declared default**, never the **effective loaded value**, so it has been green since May while verification was disabled in production.

## Fix

- `conf/config.yaml` → `fetch_verify_ssl: true`
- `docs/.../env_vars.md` corrected (it documented the default as `false`, matching the bug rather than the intent)
- New test `test_fetch_verify_ssl_is_true_in_the_shipped_yaml` reads the shipped YAML directly

**Verification of the test itself:** reverting `config.yaml` to `false` fails the new test **while the original test still passes** — which is exactly the blindness being fixed.

## Reviewer notes

Two judgement calls worth a second opinion:

1. The new test asserts the **shipped YAML** rather than going through `load_config()`. `load_config()` would be the truer end-to-end check, but it's env-sensitive — a stray `WEBSEARCH_FETCH_VERIFY_SSL` in CI could flip it either way. The YAML assertion is deterministic but narrower: it catches this class of bug and would *not* catch a future loader-precedence bug.
2. `Path(__file__).parents[4]` is path-fragile if the test file moves. A repo-root fixture would be sturdier; kept direct to hold the change small.

43 websearch tests pass; full suite green.

## Summary by CodeRabbit

* **Security**
  * Web Search now verifies TLS certificates when fetching web content by default, improving connection security.

* **Documentation**
  * Updated the Web Search configuration documentation to reflect the new certificate verification default.

* **Tests**
  * Added coverage to ensure the shipped configuration enables TLS certificate verification.

### DIFF
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -343,7 +343,7 @@ websearch:
   fetch_max_results: 3
   fetch_timeout: 1.0
   fetch_max_tokens: 500
-  fetch_verify_ssl: false
+  fetch_verify_ssl: true
 
 # --- MCP server ---
 # Env: OPENRAG_MCP_SERVER_NAME, OPENRAG_MCP_HOST, OPENRAG_MCP_PORT,
--- a/docs/content/docs/documentation/env_vars.md
+++ b/docs/content/docs/documentation/env_vars.md
@@ -481,7 +481,7 @@ Web search allows the LLM to augment RAG document context with live web results.
 | `WEBSEARCH_FETCH_MAX_RESULTS` | `int` | `3` | Number of top URLs to fetch content from (the remaining results use their search snippet). |
 | `WEBSEARCH_FETCH_TIMEOUT` | `float` | `1.0` | Per-URL timeout in seconds for content fetching. URLs that don't respond within this time fall back to their snippet. |
 | `WEBSEARCH_FETCH_MAX_TOKENS` | `int` | `500` | Maximum approximate tokens of content to extract per page. Content is truncated at word boundaries. |
-| `WEBSEARCH_FETCH_VERIFY_SSL` | `bool` | `false` | Whether to verify SSL certificates when fetching page content. |
+| `WEBSEARCH_FETCH_VERIFY_SSL` | `bool` | `true` | Whether to verify SSL certificates when fetching page content. Set to `false` only for internal CAs — fetched pages are injected into the LLM context as cited sources. |
 
 :::tip[How to Enable Web Search?]
 When chatting, you can enable web search through the OpenAI-compatible API by setting `"websearch": true` in the `metadata` field of the request body. See the [API documentation](/openrag/documentation/api/#extra-arguments) for examples.
--- a/tests/unit/services/websearch/test_content_fetcher.py
+++ b/tests/unit/services/websearch/test_content_fetcher.py
@@ -1,5 +1,6 @@
 import asyncio
 import socket
+from pathlib import Path
 
 import httpx
 import pytest
@@ -305,6 +306,24 @@ def test_fetch_verify_ssl_default_is_true():
 
     cfg = StaanWebSearchConfig()
     assert cfg.fetch_verify_ssl is True
+
+
+def test_fetch_verify_ssl_is_true_in_the_shipped_yaml():
+    """The *effective* value must be True, not just the Pydantic default.
+
+    Regression for #714. The test above constructs the model directly, so it
+    passes on the declared default and never sees ``conf/config.yaml``. Since
+    the loader merges YAML over that default, `conf/config.yaml` shipping
+    ``fetch_verify_ssl: false`` silently disabled TLS verification at runtime
+    while this file stayed green — the #404 fix never took effect.
+
+    Assert the shipped YAML directly so a config-level override cannot
+    re-disable it without turning a test red.
+    """
+    import yaml
+
+    conf = yaml.safe_load((Path(__file__).parents[4] / "conf" / "config.yaml").read_text())
+    assert conf["websearch"]["fetch_verify_ssl"] is True
 
 
 # ---------------------------------------------------------------------------


Ahmath-Gadji:
Reviewed. Fix is correct and minimal — the YAML-over-Pydantic-default merge order fully explains why #404's fix never took effect, and the new test genuinely catches the regression (verified: reverting `config.yaml` to `false` fails the new test while `test_fetch_verify_ssl_default_is_true` stays green).

On the two judgement calls raised in the PR description:

1. **Shipped-YAML assertion vs. `load_config()`.** Agree with the narrower, deterministic test. An env-var-sensitive assertion through `load_config()` would be a better end-to-end check in principle, but flaky-by-environment in CI is worse than narrow-by-design — and this test's whole purpose is to catch *this* class of bug (declared default silently overridden by shipped config), which it does.
2. **`Path(__file__).parents[4]`.** Confirmed the arithmetic is currently correct (`tests/unit/services/websearch/` → 4 parents → repo root). It's fragile to a future file move, but a moved test would just make the path wrong and the test error loudly (`FileNotFoundError`), not silently pass — low blast radius. Fine to keep as-is rather than introducing a `repo_root` fixture for one call site.

Also checked repo-wide for other places that might ship the stale `fetch_verify_ssl: false` default (`.env.example`, docs mirror, integration tests) — none found. CI is green across the board (api-tests, lint, milvus-integration, tests, CodeRabbit). No blocking issues from my pass.

Ahmath-Gadji:
Labeling this **breaking-change** for release notes — worth an explicit upgrade note.

**Upgrade impact:** deployments with web search enabled (`WEBSEARCH_API_TOKEN` set) that fetch pages behind an internal/self-signed CA were silently relying on the buggy `fetch_verify_ssl: false` default. After this upgrade those fetches will fail TLS verification and silently fall back to snippet-only content for those results — no crash, no error surfaced (confirmed: `_fetch_single` catches the failure and returns `None`; `format_web_context` falls back to `result.snippet`), but a quiet quality regression with no warning to the operator.

Suggested upgrade note: operators whose web-search fetch targets sit behind an internal/private CA should set `WEBSEARCH_FETCH_VERIFY_SSL=false` explicitly before/after upgrading, per the updated `env_vars.md` guidance.

Ahmath-Gadji (approved):
Approving. Fix is correct, minimal, and fully tested (43/43 websearch unit tests pass locally, CI green across api-tests/lint/milvus-integration/tests/CodeRabbit). Verified the bug narrative against the config loader's merge order and the test's revert-check claim. Only follow-up was a release-notes gap (upgrade impact for internal-CA deployments), now covered by the `breaking-change` label and comment above.

PR #718, opened 2026-07-20
fix(inference): make embedder transport failures retryable

### LINKED ISSUE #704: Embedder retry never fires — EmbeddingAPIError hardcodes status_code=500, which is not in the retryable set
## Problem

`@with_retry(max_attempts=3)` on the embedder is dead code — it can never fire. A single transient embedder failure (429 / 503 / connection reset / timeout) aborts `embed()`, which cancels sibling batches and fails the whole file.

`EmbeddingAPIError` hardcodes `status_code=500` — `openrag/core/utils/exceptions.py:300-301`:
```python
def __init__(self, message: str, **kwargs):
    super().__init__(message, code="EMBEDDING_API_ERROR", status_code=500, **kwargs)
```
(the `EmbeddingError` base also defaults to `500`, `:293`)

The retry predicate only accepts `{429, 502, 503, 504}` — `openrag/services/inference/_retry.py:14-24`:
```python
_RETRYABLE_STATUS_CODES = {429, 502, 503, 504}

def _is_retryable(exc: BaseException) -> bool:
    if isinstance(exc, (httpx.TimeoutException, httpx.ConnectError)):
        return True
    ...
    if isinstance(exc, OpenRAGError):
        return exc.status_code in _RETRYABLE_STATUS_CODES
    return False
```
`500` is not in the set, so `EmbeddingAPIError` is never retried.

The `httpx` branches cannot rescue it either: `_embed_batch` is the decorated function (`vllm_client.py:282-283`), and it translates every transport failure to `EmbeddingAPIError` **inside its own body** (`:296-312`). By the time the exception reaches the retry decorator it is already an `OpenRAGError` with `status_code=500`.

## This is specific to the embedder

Every other inference client maps to a retryable code, which is what makes this an oversight rather than a policy:

| Exception | status_code | Retried |
|---|---|---|
| `InferenceError` (`exceptions.py:227`) | 503 | yes |
| `InferenceTimeoutError` (`:249`) | 504 | yes |
| `InferenceConnectionError` (`:256`) | 503 | yes |
| **`EmbeddingAPIError` (`:301`)** | **500** | **no** |

The LLM path also preserves the upstream status (`vllm_client.py:134-138`, `status_code=exc.response.status_code`); the embedder discards it. So LLM, VLM and reranker retries all work — only the embedder's is inert.

## Impact

Scales with batch size. During bulk ingestion, routine embedder backpressure (a 429 or a brief 503) fails the entire file instead of retrying, and `embed()` cancels its sibling batches on the first failure. At production batch sizes this turns transient load into mass ingestion failure, while `@with_retry(max_attempts=3)` in the source reads as though it is handled.

## Suggested fix

Mirror the `Inference*` pattern in `VLLMEmbedder._embed_batch` (`vllm_client.py:296-312`):

- connection error → `503`
- timeout → `504`
- HTTP status error → preserve `exc.response.status_code`

Either by passing `status_code=` through `EmbeddingAPIError`, or by adding `EmbeddingConnectionError` / `EmbeddingTimeoutError` subclasses alongside the existing `Inference*` ones.

Add a regression test asserting `_embed_batch` is actually re-invoked on a 429/503 (assert the call count, not merely that the exception type propagates).

## Verified on

`develop` @ `e61479b4`.

andyne13:
Closes #704.

## Problem

`@with_retry(max_attempts=3)` on the embedder was **dead code** — it could never fire.

`_is_retryable` (`services/inference/_retry.py:14-24`) retries an `OpenRAGError` only when its status is in `{429, 502, 503, 504}`. `EmbeddingAPIError` hardcoded `status_code=500` (`core/utils/exceptions.py:301`), which isn't in that set.

The `httpx` branches of the predicate can't rescue it either: `_embed_batch` is the decorated function and it translates every transport failure into `EmbeddingAPIError` **inside its own body**, so the decorator only ever sees the already-translated 500.

**Impact:** a single transient 429/503 aborted `embed()`, which cancels its sibling batches and fails the whole file. During bulk ingestion that turns routine embedder backpressure into mass failure — while `@with_retry(max_attempts=3)` in the source reads as though it's handled.

## Why this is an oversight, not a policy

Every other inference client maps to a retryable code:

| Exception | status_code | Retried |
|---|---|---|
| `InferenceError` | 503 | yes |
| `InferenceTimeoutError` | 504 | yes |
| `InferenceConnectionError` | 503 | yes |
| **`EmbeddingAPIError`** (before) | **500** | **no** |

The LLM path also preserves the upstream status (`vllm_client.py:139`); the embedder discarded it.

## Fix

Mirror the `Inference*` family:

- `EmbeddingConnectionError` → 503
- `EmbeddingTimeoutError` → 504
- `EmbeddingAPIError` gains a `status_code` override; the HTTP branch now preserves the upstream status

## Scope note — ollama is included

`ollama_client.embed` had the **identical** bug: same `@with_retry(max_attempts=3)`, all three raisers on a hardcoded 500. It's reachable via `{"implementation": "ollama"}` (`core/config/model_endpoints.py:19`), so it's fixed in the same change rather than left as a known-broken sibling. Happy to split it out if reviewers prefer.

## Behavioural change beyond "retries now work"

Preserving the upstream status means a **4xx now fails fast** instead of burning three attempts. That's correct, but it is a change in behaviour worth noting.

## Tests

They assert the retry **re-invokes** (call counts), not merely that an exception type changed:

- `test_retries_then_succeeds_on_transient_503` — 2 retries then success, `calls == 3`
- `test_retries_on_429_rate_limit` — `calls == 2`
- `test_does_not_retry_client_error` — `calls == 1` on a 400

**Verification:** reverting only the production fix while keeping these tests fails the 503 and 429 cases, while the 4xx case still passes — so they detect this specific bug rather than agreeing with the new code.

An autouse fixture swaps the tenacity wait for `wait_none()` (the decorator is applied at import time, so patching the module function has no effect) — 6.5s → 0.23s.

Full suite green (1802); ruff check / format / layer-import-guard all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding error handling by distinguishing timeout vs connection failures with dedicated retryable exception types.
  * Embedding failures now surface the upstream HTTP status code so retry logic can correctly treat transient/service/rate-limit errors as retryable and client errors as non-retryable.

* **Tests**
  * Added unit tests to verify vLLM and Ollama embedder retries on transient timeouts and connection/reset scenarios, and confirm no retries occur for non-retryable client errors.

### DIFF
--- a/openrag/core/utils/exceptions.py
+++ b/openrag/core/utils/exceptions.py
@@ -295,10 +295,38 @@ class EmbeddingError(OpenRAGError):
 
 
 class EmbeddingAPIError(EmbeddingError):
-    """API error with the embedding provider."""
+    """API error with the embedding provider.
+
+    ``status_code`` is overridable so transport failures can carry a *retryable*
+    code. ``_retry._is_retryable`` only retries ``OpenRAGError`` when the status
+    is in {429, 502, 503, 504}; hardcoding 500 here made ``@with_retry`` on the
+    embedder dead code (#704). Callers should prefer the two subclasses below.
+    """
+
+    def __init__(self, message: str, *, status_code: int = 500, **kwargs):
+        super().__init__(message, code="EMBEDDING_API_ERROR", status_code=status_code, **kwargs)
+
+
+class EmbeddingTimeoutError(EmbeddingAPIError):
+    """Embedding request timed out. Maps to HTTP 504 — retryable.
+
+    Mirrors :class:`InferenceTimeoutError`, which the LLM/VLM/reranker clients
+    already use; the embedder was the only client whose transport failures
+    landed on a non-retryable 500.
+    """
 
     def __init__(self, message: str, **kwargs):
-        super().__init__(message, code="EMBEDDING_API_ERROR", status_code=500, **kwargs)
+        super().__init__(message, status_code=504, **kwargs)
+
+
+class EmbeddingConnectionError(EmbeddingAPIError):
+    """Cannot reach the embedding provider. Maps to HTTP 503 — retryable.
+
+    Mirrors :class:`InferenceConnectionError`.
+    """
+
+    def __init__(self, message: str, **kwargs):
+        super().__init__(message, status_code=503, **kwargs)
 
 
 class EmbeddingResponseError(EmbeddingError):
--- a/openrag/services/inference/ollama_client.py
+++ b/openrag/services/inference/ollama_client.py
@@ -166,16 +166,18 @@ class OllamaEmbedder(Embedder):
             resp = await self._client.post(f"{self._endpoint}/embeddings", json=body)
             resp.raise_for_status()
         # Retryable statuses so @with_retry above actually fires (#704).
-        except httpx.ConnectError as exc:
-            raise EmbeddingConnectionError(
-                f"Cannot reach Ollama embedder at {self._endpoint}",
+        # TimeoutException before TransportError (it is a subclass); the broader
+        # net catches a mid-request connection reset (httpx ReadError) too.
+        except httpx.TimeoutException as exc:
+            raise EmbeddingTimeoutError(
+                f"Ollama embedder request timed out at {self._endpoint}",
                 model_name=self._model,
                 base_url=self._endpoint,
                 error=str(exc),
             ) from exc
-        except httpx.TimeoutException as exc:
-            raise EmbeddingTimeoutError(
-                f"Ollama embedder request timed out at {self._endpoint}",
+        except httpx.TransportError as exc:
+            raise EmbeddingConnectionError(
+                f"Cannot reach Ollama embedder at {self._endpoint}",
                 model_name=self._model,
                 base_url=self._endpoint,
                 error=str(exc),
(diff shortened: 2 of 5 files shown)


hedhoud (approved):
The retry behavior is consistent across both embedders, and client errors still fail fast. I also verified it against the latest develop branch and the full unit suite passes. Approved.

hedhoud (changes requested):
There is one transport failure path still missing here. A connection reset after the request starts is raised by httpx as ReadError, not ConnectError. I reproduced it with both the VLLM and Ollama embedders: each made one call and let the raw ReadError escape, so no retry happened. Since this fix is meant to recover from transient connection failures, could we handle the relevant network errors as retryable and add a regression test for a connection reset?

andyne13:
Fixed in 7a254b27. Broadened the catch from `httpx.ConnectError` to `httpx.TransportError` in both VLLMEmbedder and OllamaEmbedder, so a mid-request connection reset (`ReadError`) is now retried. `TimeoutException` is caught first (it's a TransportError subclass, keeps its 504) and `HTTPStatusError` is unaffected (4xx still fail-fast). Added regression tests for both a ReadError and a ConnectError that assert the retry re-invokes — reverting the broadening fails them. Thanks for the catch.

hedhoud (approved):
The connection-reset path is fixed for both embedders, and the regression tests verify that the request is retried. I also checked the change against the current develop branch and found no remaining issue. LGTM.

PR #726, opened 2026-07-20
test(auth): assert the role literal each authorization wrapper enforces

### LINKED ISSUE #725: Authorization wrappers' role literals are untested — 5 of 6 can be made allow-all with zero test failures
## Problem

The partition-authorization *helper* is well tested, but the **role literal inside each wrapper is not** — five of six wrappers can be made allow-all without turning a single test red.

Established by mutation testing against the full suite (1731 tests at the time):

| Mutation | Tests failed |
|---|---|
| Neuter `ensure_partition_role` (the shared helper) | several — `test_ensure_partition_role_*` |
| **Neuter only the six thin wrappers** | **2** — both `require_partitions_viewer` (the plural, all-scope variant) |

So `require_partition_owner`, `require_partition_editor`, `require_partition_viewer`, `require_admin`, and `require_admin_or_self` can each be made to allow everything with **zero** failures.

The reason: every test reference to them is a `dependency_overrides` entry that *replaces* the dependency (e.g. `tests/unit/api/routers/admin/test_phase14_partition_routes.py:59-60`), so the wrapper body — including the role string it passes — is never executed. `require_admin_or_self`, which guards token regeneration, appears **zero** times in `tests/`.

Two further mutations that changed nothing:

- dropping the `SUPER_ADMIN_MODE` conjunct at `api/dependencies/auth.py:68` (making super-admin permanently on) → all tests passed; the only super-admin test is allow-only, with no deny case for `super_admin_mode=False, is_admin=True`
- forcing `pending_task_count=0` in the quota check → all tests passed; both enforcing cases already use `pending_count=0`, so the in-flight half of the quota — the very thing that closes the #664 race — is unverified

## Impact

The classic one-word copy-paste slip (`"owner"` → `"viewer"`) across three byte-identical wrappers ships green. Nothing in CI distinguishes a correctly-scoped guard from an allow-all one.

## Fix

Call each wrapper **directly** (not via `dependency_overrides`) with a spy on the helper, asserting the role literal it passes:

```python
async def test_require_partition_editor_passes_editor_role(spy):
    await require_partition_editor(...)
    assert spy.called_with_role == "editor"
```

Plus a deny case for `SUPER_ADMIN_MODE=False, is_admin=True`, and a quota case where `pending_task_count` alone crosses the limit.

andyne13:
Closes #725.

**Tests only — no production change.**

## Problem

The shared `ensure_partition_role` helper is well covered (5 tests), but the **role literal inside each wrapper is not**. Every existing test of the wrappers replaces them via `dependency_overrides` (e.g. `test_phase14_partition_routes.py:59-60`), so the wrapper body never executes.

Mutation testing against the full suite:

| Mutation | Failed before |
|---|---|
| Neuter the six thin wrappers | **2** — both the plural `require_partitions_viewer` |
| Drop the `SUPER_ADMIN_MODE` conjunct (super-admin permanently on) | **0** |
| Force `pending_task_count = 0` in the quota check | **0** |

So `require_partition_owner`, `_editor`, `_viewer`, `require_admin` and `require_admin_or_self` could each be made allow-all with zero failures. `require_admin_or_self` — which guards token regeneration — had **zero** references in `tests/`.

## Change

- Call each wrapper **directly** with a spy on `ensure_partition_role`, asserting the role it forwards (`viewer` / `editor` / `owner`, and `viewer` per-partition for the plural variant)
- Direct coverage for `require_admin` and a 4-case matrix for `require_admin_or_self`
- The missing `SUPER_ADMIN_MODE=False, is_admin=True` **deny** case (only the allow side was tested)
- A quota case where **pending tasks alone** exhaust the limit — the in-flight half that #664 / #700 is about

The quota tests use the **real `AuthService`**, not a fake, so they assert production semantics rather than a stand-in's.

## Verification

Each mutation from the issue, re-run against these tests:

```
require_partition_owner  "owner"  -> "viewer"    1 failed  (was 0)
require_partition_editor "editor" -> "viewer"    1 failed  (was 0)
SUPER_ADMIN_MODE conjunct dropped                1 failed  (was 0)
pending_task_count forced to 0                   1 failed  (was 0)
require_admin_or_self allow-all                  3 failed  (was 0)
baseline (unmutated)                            28 passed
```

Full suite: **1815 passed**. `ruff check` / `ruff format --check` / `check_layer_imports.py` green.

## Why this one matters

It's the only issue in the v2.0.1 batch that *prevents* bugs rather than fixing one. A one-word slip (`"owner"` → `"viewer"`) across three byte-identical wrappers currently ships green — this makes that impossible.

## Summary by CodeRabbit

* **Tests**
  * Expanded authentication and authorization test coverage for administrator access, self-service permissions, role requirements, and missing-user scenarios.
  * Added coverage for super-admin mode behavior, including partition access checks.
  * Added quota enforcement tests covering pending tasks and combined file and task counts.

### DIFF
--- a/tests/unit/api/dependencies/test_auth.py
+++ b/tests/unit/api/dependencies/test_auth.py
@@ -285,3 +285,211 @@ async def test_check_user_file_quota_default_zero_enforces_zero():
     assert exc.value.status_code == 403
     assert exc.value.detail == "File quota exceeded"
     assert denied_job_service.pending_checks == [8]
+
+
+# ---------------------------------------------------------------------------
+# #725 — the role literal each wrapper passes must itself be asserted.
+#
+# `ensure_partition_role` is well covered, but every existing test of the thin
+# wrappers replaces them via `dependency_overrides`, so the wrapper body — and
+# the role string it hardcodes — never executes. Mutation testing showed five of
+# six wrappers could be made allow-all with zero failures, and a one-word slip
+# ("owner" -> "viewer") across three byte-identical wrappers would ship green.
+#
+# These call each wrapper DIRECTLY and spy on the role it forwards.
+# ---------------------------------------------------------------------------
+
+
+class RoleSpy:
+    """Records the required_role each wrapper forwards to ensure_partition_role."""
+
+    def __init__(self) -> None:
+        self.roles: list[str] = []
+        self.partitions: list[str] = []
+
+    async def __call__(self, partition, user, user_partitions, required_role, **kwargs):
+        self.partitions.append(partition)
+        self.roles.append(required_role)
+        return True
+
+
+@pytest.fixture
+def role_spy(monkeypatch):
+    spy = RoleSpy()
+    monkeypatch.setattr("api.dependencies.auth.ensure_partition_role", spy)
+    return spy
+
+
+@pytest.mark.parametrize(
+    ("wrapper_name", "expected_role"),
+    [
+        ("require_partition_viewer", "viewer"),
+        ("require_partition_editor", "editor"),
+        ("require_partition_owner", "owner"),
+    ],
+)
+@pytest.mark.asyncio
+async def test_partition_wrappers_forward_their_own_role(role_spy, wrapper_name, expected_role):
+    import api.dependencies.auth as auth_deps
+
+    wrapper = getattr(auth_deps, wrapper_name)
+    user = {"id": 1, "is_admin": False}
+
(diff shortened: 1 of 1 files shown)


hedhoud (approved):
The new tests cover the authorization decisions that were previously unprotected, including role forwarding, admin boundaries, and pending-task quota enforcement. I verified them against the latest develop branch and the full unit suite passes. Approved.

PR #729, opened 2026-07-20
fix(docling): make the docling pool and workers restartable (max_restarts)

### LINKED ISSUE #722: Docling pools lack max_restarts — a single OOM/CUDA crash permanently disables docling parsing
## Problem

`DoclingPool` and `DoclingWorker` are declared with a bare `@ray.remote` and no `max_restarts`, so a single crash permanently disables Docling parsing.

`services/workers/parsers/docling_workers.py:58` and `:83`:
```python
@ray.remote            # DoclingWorker
@ray.remote            # DoclingPool
```
Compare the sibling pools, which both set it:
```python
services/workers/parsers/marker_workers.py:259:  @ray.remote(max_restarts=5)
services/workers/parsers/whisper_workers.py:45:   @ray.remote(max_restarts=...)
```

Worker handles are created once in `__init__` and re-queued unconditionally after use (`:117`) — nothing checks liveness or evicts a dead handle.

## Impact

Ray's default is `max_restarts=0`, so an OOM or CUDA fault kills the actor permanently. With the default `docling_pool_size=1` there is exactly one worker, so every subsequent Docling parse draws the same dead handle and fails. Recovery requires restarting the container.

Any partition whose preset sets `parsing_strategy: docling` is then broken until an operator notices — and because parse failures surface per-file, it reads as "some PDFs fail" rather than "the pool is dead".

## Fix

Add `max_restarts=5` to both decorators to match marker/whisper, and name the workers (`get_if_exists=True`) so a pool restart reattaches instead of orphaning the previous set — the pattern `IndexerPool` already uses (`indexer_pool.py:262-270`).

Distinct from #581 (tuning slow defaults / adding concurrency), which does not address restart resilience.

andyne13:
Closes #722.

## Problem

`DoclingPool` and `DoclingWorker` used a bare `@ray.remote`, so Ray's default `max_restarts=0` applied — a single OOM or CUDA fault killed the actor permanently.

`DoclingPool` is detached, and its worker handles are queued once in `__init__` and re-queued unconditionally after each use (`docling_workers.py:106-114`) with no liveness check. With the default `docling_pool_size=1` there is exactly one worker, so every subsequent docling parse drew the same dead handle and failed until the container was restarted. Because failures surface per-file, it reads as "some PDFs fail" rather than "the pool is dead".

## Sibling parity

Both other GPU parsers already survive crashes — docling was the sole gap:

| | restart policy |
|---|---|
| `MarkerPool` | `@ray.remote(max_restarts=5)` (`marker_workers.py:259`) |
| `MarkerWorker` | `.options(..., max_restarts=5)` (`marker_workers.py:270`) |
| `WhisperActor` | `max_restarts=5` via `whisper_actor_options` (`whisper_workers.py:45`) |
| **`DoclingPool` / `DoclingWorker`** | **none** |

## Fix

Mirror marker exactly: `max_restarts=5` on the `DoclingPool` decorator and on the `DoclingWorker.options(...)` call. A restarted Ray actor keeps its actor id, so the handle already sitting in the queue self-heals instead of staying dead — which is precisely the "every retry draws the same dead handle" failure.

## Tests

- `test_docling_pool_sets_max_restarts` — asserts the pool metadata; reverting the decorator fails it
- `test_docling_pool_restart_policy_matches_marker` — pins docling to parity with `MarkerPool` so the two can't silently drift apart again

The worker-side `.options(max_restarts=5)` is applied at actor-creation time, so it is not statically inspectable without a Ray runtime — it mirrors `MarkerWorker:270` and is verified by reading.

Full suite: **1812 passed**; ruff check / format / layer-import-guard green.

## Out of scope (follow-up)

Marker also has an active `ensure_worker_pool_healthy` / `is_pool_broken` recreate loop that docling lacks entirely. That's a larger enhancement; `max_restarts` is the minimal fix for the permanent-death bug this issue is about.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF processing resilience by automatically restarting failed processing workers up to five times.
  * Reduced the risk of interrupted document processing after worker or supervisor crashes.

* **Tests**
  * Added coverage to verify worker restart behavior and configuration consistency.

### DIFF
--- a/openrag/services/workers/parsers/docling_workers.py
+++ b/openrag/services/workers/parsers/docling_workers.py
@@ -80,7 +80,9 @@ class DoclingWorker:
             return await asyncio.to_thread(self.converter.convert, str(file_path))
 
 
-@ray.remote
+# max_restarts mirrors MarkerPool: the detached pool supervisor must survive its
+# own crashes rather than staying permanently dead (#722). Ray's default is 0.
+@ray.remote(max_restarts=5)
 class DoclingPool:
     def __init__(self):
         self.logger = get_logger()
@@ -88,8 +90,12 @@ class DoclingPool:
         self.pool_size = self.config.loader.docling_pool_size
         self.max_tasks_per_worker = self.config.loader.docling_max_tasks_per_worker
 
+        # max_restarts mirrors MarkerWorker: a worker that OOMs/CUDA-faults is
+        # restarted in place (same actor id), so the handle already queued keeps
+        # working instead of every retry drawing the same dead handle (#722).
         self.actors = [
-            DoclingWorker.options(num_gpus=_docling_num_gpus(self.config)).remote() for _ in range(self.pool_size)
+            DoclingWorker.options(num_gpus=_docling_num_gpus(self.config), max_restarts=5).remote()
+            for _ in range(self.pool_size)
         ]
         self._queue: asyncio.Queue[ray.actor.ActorHandle] = asyncio.Queue()
 
--- a/tests/unit/services/workers/parsers/test_docling_workers.py
+++ b/tests/unit/services/workers/parsers/test_docling_workers.py
@@ -30,3 +30,21 @@ def test_docling_num_gpus_zero_when_not_requested(monkeypatch):
     monkeypatch.setattr(docling_workers.ray, "cluster_resources", lambda: {"GPU": 1.0})
 
     assert docling_workers._docling_num_gpus(_config(docling_num_gpus=0)) == 0
+
+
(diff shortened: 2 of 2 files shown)


hedhoud (approved):
I reviewed the restart behavior and tested the affected parser pool paths locally. The change is focused, matches the existing Marker policy, and the current implementation recovers without leaving the previous worker behind. All focused tests and CI checks are green. The worker-side option could use a direct regression assertion later, but that is not blocking this fix. This is ready to merge from my side.

PR #731, opened 2026-07-21
fix: require chat_history_depth >= 1 and default it to 4

### LINKED ISSUE #730: Chat History Depth: misleading default of 0 and no minimum guard
## Problem

The Create/Edit Partition forms in the admin UI default **Chat History Depth** to `0` and allow any non-negative value, with no explanation of what `0` actually does.

Investigation of the backend (`QueryService._resolve_chat_history_depth` in `openrag/services/orchestrators/query_service.py`) showed:

- A stored value of `0` is a legacy sentinel meaning "inherit the global default" (`conf/config.yaml: rag.chat_history_depth`, currently `4`) — it does **not** mean "no history", but the UI gives no indication of this, so an admin reading "0" reasonably assumes history is disabled.
- A stored value of `1` is **not** special-cased: `messages[-1:]` slices down to just the current user turn, i.e. zero prior context — a real "no history" footgun that isn't caught anywhere today.
- Neither the Pydantic schemas (`CreatePartitionRequest`/`UpdatePartitionRequest` in `openrag/api/schemas/admin/partition_schemas.py`) nor the DB column enforce any lower bound beyond `>= 0`.

## Fix

- UI: default Chat History Depth to `4` (matching the global default) in both the Create Partition dialog (`ui/src/pages/admin/partitions/list.tsx`) and the Edit Partition form (`ui/src/pages/admin/partitions/detail.tsx`), with a required, `min=1` input and a submit-time guard.
- Backend: require `chat_history_depth >= 1` on `CreatePartitionRequest`/`UpdatePartitionRequest`, drop the `0` sentinel from new writes, and normalize any pre-existing `0`-valued rows to `4` at read time (`PartitionService.resolve_partition_row` / `list_partition_summaries` / `_partition_detail`) so legacy partitions display and behave consistently without a data migration.

## Test plan

- [x] Backend unit tests updated/added for the new `ge=1` constraint and the legacy-zero normalization (`tests/unit/api/schemas/admin/test_phase14_schemas.py`, `tests/unit/core/models/test_preset_models.py`, `tests/unit/services/orchestrators/test_partition_preset_resolution.py`, `tests/unit/services/orchestrators/test_partition_service.py`)
- [x] Frontend typecheck + existing partition tests pass

Ahmath-Gadji:
## Summary
- Chat History Depth defaulted to `0` in the Create/Edit Partition UI with no lower bound and no explanation. `0` was actually a backend sentinel meaning "inherit the global default" (4), not "no history" — but `1` wasn't special-cased and would genuinely drop all prior conversation context.
- UI: default is now `4`, input is `required`/`min=1` with a submit-time guard, and an info tooltip explains what the field does.
- Backend: `CreatePartitionRequest`/`UpdatePartitionRequest` now enforce `chat_history_depth >= 1`; `PartitionRow`/`PartitionConfig` match. Any pre-existing partition still holding the legacy `0` is normalized to `4` at read time (no DB migration needed).

Closes #730

## Test plan
- [x] `uv run pytest tests/unit/` — 1828 passed
- [x] `uv run ruff check openrag/ tests/` — clean
- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run` on the touched partition test files — passing
- [x] New/updated unit tests cover: rejecting 0/negative at the schema and model level, and the legacy-zero → 4 normalization in `resolve_partition_row`

## Summary by CodeRabbit

* **Improvements**
  * Chat History Depth now defaults to **4** and enforces a **minimum of 1** for partition creation, updates, and resolved partition responses.
  * Values below **1** are rejected both in the API schema and during partition/query handling.
  * Legacy partitions storing a depth of **0** are automatically normalized to the current default (**4**), including when global settings are invalid.
* **Admin UI**
  * Create/edit dialogs now require Chat History Depth **≥ 1**, disable saving for invalid values, and show an error toast if attempted.
  * Added an info tooltip to explain the setting.

### DIFF
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -270,7 +270,11 @@ class PartitionService:
                 "indexation_preset": r.get("indexation_preset") or "default",
                 "retrieval_preset": r.get("retrieval_preset") or "default",
                 "dimension": r.get("dimension"),
-                "chat_history_depth": r.get("chat_history_depth") or 0,
+                # `or 4`, not `or 0`: normalizes pre-existing rows stored under the old
+                # "0 = inherit global default" scheme to the concrete value they already
+                # resolve to at chat time (see QueryService._resolve_chat_history_depth).
+                # New writes can no longer produce 0 (schema now requires >=1).
+                "chat_history_depth": r.get("chat_history_depth") or 4,
                 "chat_llm": r.get("chat_llm"),
                 "created_at": created.isoformat() if hasattr(created, "isoformat") else created,
                 "document_count": counts.get(name, 0),
@@ -287,7 +291,7 @@ class PartitionService:
         embedder: str = "default",
         indexation_preset: str = "default",
         retrieval_preset: str = "default",
-        chat_history_depth: int = 0,
+        chat_history_depth: int = 4,
         chat_llm: str | None = None,
     ) -> None:
         """Create a partition owned by ``user_id`` with preset references.
@@ -501,7 +505,8 @@ class PartitionService:
             "retrieval_pipeline": cfg.retrieval.model_dump(mode="json"),
             "dimension": row.get("dimension"),
             "created_at": row.get("created_at"),
-            "chat_history_depth": row.get("chat_history_depth") or 0,
+            # See list_partition_summaries() above for why this normalizes to 4, not 0.
+            "chat_history_depth": row.get("chat_history_depth") or 4,
             "chat_llm": row.get("chat_llm"),
         }
 
@@ -536,7 +541,10 @@ class PartitionService:
             indexation=IndexationPipelineConfig(**idx_preset),
             retrieval=RetrievalPipelineConfig(**ret_preset),
             collection_name=row.get("collection_name"),
-            chat_history_depth=row.get("chat_history_depth") or 0,
+            # See list_partition_summaries() above for why this normalizes to 4, not 0.
+            # This value feeds Settings.partitions, so it's what
+            # QueryService._resolve_chat_history_depth actually reads at chat time.
+            chat_history_depth=row.get("chat_history_depth") or 4,
             chat_llm=row.get("chat_llm"),
         )
 
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -270,11 +270,7 @@ class PartitionService:
                 "indexation_preset": r.get("indexation_preset") or "default",
                 "retrieval_preset": r.get("retrieval_preset") or "default",
                 "dimension": r.get("dimension"),
-                # `or 4`, not `or 0`: normalizes pre-existing rows stored under the old
-                # "0 = inherit global default" scheme to the concrete value they already
-                # resolve to at chat time (see QueryService._resolve_chat_history_depth).
-                # New writes can no longer produce 0 (schema now requires >=1).
-                "chat_history_depth": r.get("chat_history_depth") or 4,
+                "chat_history_depth": r.get("chat_history_depth") or self._legacy_chat_history_depth_fallback(),
                 "chat_llm": r.get("chat_llm"),
                 "created_at": created.isoformat() if hasattr(created, "isoformat") else created,
                 "document_count": counts.get(name, 0),
@@ -505,8 +501,7 @@ class PartitionService:
             "retrieval_pipeline": cfg.retrieval.model_dump(mode="json"),
             "dimension": row.get("dimension"),
             "created_at": row.get("created_at"),
-            # See list_partition_summaries() above for why this normalizes to 4, not 0.
-            "chat_history_depth": row.get("chat_history_depth") or 4,
+            "chat_history_depth": row.get("chat_history_depth") or self._legacy_chat_history_depth_fallback(),
             "chat_llm": row.get("chat_llm"),
         }
 
@@ -541,10 +536,9 @@ class PartitionService:
             indexation=IndexationPipelineConfig(**idx_preset),
             retrieval=RetrievalPipelineConfig(**ret_preset),
             collection_name=row.get("collection_name"),
-            # See list_partition_summaries() above for why this normalizes to 4, not 0.
             # This value feeds Settings.partitions, so it's what
             # QueryService._resolve_chat_history_depth actually reads at chat time.
-            chat_history_depth=row.get("chat_history_depth") or 4,
+            chat_history_depth=row.get("chat_history_depth") or self._legacy_chat_history_depth_fallback(),
             chat_llm=row.get("chat_llm"),
         )
 
@@ -574,6 +568,19 @@ class PartitionService:
         if self._config is None:
             raise ConfigError("PartitionService was constructed without a config; preset resolution unavailable.")
         return self._config
+
+    def _legacy_chat_history_depth_fallback(self) -> int:
+        """Value substituted for a partition row still holding the pre-guard ``0``.
+
+        ``0`` used to mean "inherit the global default" (``config.rag.chat_history_depth``,
+        see ``QueryService._resolve_chat_history_depth``). New writes can no longer
+        produce ``0`` (``ge=1`` on ``CreatePartitionRequest``/``UpdatePartitionRequest``),
+        but a row written before that guard existed may still have it stored — read it
+        from the live config, not a hardcoded constant, so a legacy row keeps resolving
+        to the *current* global default exactly as it did before, even if an operator
+        changes ``rag.chat_history_depth`` later.
+        """
+        return self._config.rag.chat_history_depth if self._config is not None else 4
 
     # ------------------------------------------------------------------
     # File / chunk reads
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -569,6 +569,9 @@ class PartitionService:
             raise ConfigError("PartitionService was constructed without a config; preset resolution unavailable.")
         return self._config
 
+    #: Depth used when neither a partition row nor the global config yields a usable value.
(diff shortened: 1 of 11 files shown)


hedhoud:
@codex review

Ahmath-Gadji:
Self-review follow-up (42f32ba8): the first commit's "normalize a legacy `chat_history_depth=0` row to the global default" logic hardcoded the literal `4` in three read paths, even though the comments claimed it mirrored `QueryService._resolve_chat_history_depth`'s global-default fallback. That was only true by coincidence (`conf/config.yaml`'s `rag.chat_history_depth` also happens to be `4`) — if an operator changed that config value, legacy partitions would have silently frozen at the old default instead of tracking it.

Fixed by reading `config.rag.chat_history_depth` live via a small `_legacy_chat_history_depth_fallback()` helper, with a `4` fallback only when the service has no config at all. Added a regression test (`test_resolve_partition_row_legacy_zero_tracks_current_global_default`) that sets the global default to `9` and asserts a legacy-zero row resolves to `9`, not a hardcoded `4`. Full unit suite (1829 tests) + ruff + tsc still green.

openrag/services/orchestrators/partition_service.py
```diff
@@ -536,7 +541,10 @@ def resolve_partition_row(self, row: dict) -> PartitionConfig:
             indexation=IndexationPipelineConfig(**idx_preset),
             retrieval=RetrievalPipelineConfig(**ret_preset),
             collection_name=row.get("collection_name"),
-            chat_history_depth=row.get("chat_history_depth") or 0,
+            # See list_partition_summaries() above for why this normalizes to 4, not 0.
+            # This value feeds Settings.partitions, so it's what
+            # QueryService._resolve_chat_history_depth actually reads at chat time.
+            chat_history_depth=row.get("chat_history_depth") or 4,
```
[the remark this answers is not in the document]
Ahmath-Gadji: Handled — this was the same bug I caught in my own self-review (commit 42f32ba8, pushed before this Codex review ran, though it looks like the review picked up the pre-fix commit 575b05e9). `chat_history_depth` for legacy-zero rows now resolves via `_legacy_chat_history_depth_fallback()`, which reads `self._config.rag.chat_history_depth` live instead of hardcoding 4 — see `partition_service.py:572-583` on the current head.

On the seeded `default` partition specifically: `seed_default_partition()` does leave `0` in the DB (it calls the repo layer directly, bypassing the orchestrator's explicit-write path), but `DIContainer.initialize()` always calls `load_partitions()` immediately after seeding (`container.py:204-205`), which runs every row — including that one — through `resolve_partition_row()` and the same live-config fallback. So it resolves correctly and dynamically, not frozen at 4.

Regression test added: `test_resolve_partition_row_legacy_zero_tracks_current_global_default` (sets the global default to 9, asserts a legacy-zero row resolves to 9).

openrag/services/orchestrators/partition_service.py
```diff
@@ -567,6 +569,19 @@ def _require_config(self) -> Settings:
...
+
+        ``0`` used to mean "inherit the global default" (``config.rag.chat_history_depth``,
+        see ``QueryService._resolve_chat_history_depth``). New writes can no longer
+        produce ``0`` (``ge=1`` on ``CreatePartitionRequest``/``UpdatePartitionRequest``),
+        but a row written before that guard existed may still have it stored — read it
+        from the live config, not a hardcoded constant, so a legacy row keeps resolving
+        to the *current* global default exactly as it did before, even if an operator
+        changes ``rag.chat_history_depth`` later.
+        """
+        return self._config.rag.chat_history_depth if self._config is not None else 4
```
hedhoud: `RAGConfig.chat_history_depth` still accepts `0`. When a default or legacy partition also stores `0`, this fallback returns `0` and the new `PartitionConfig(ge=1)` validation makes `load_partitions()` fail during startup. Please either enforce the same minimum in the global config with a clear configuration error, or normalize the fallback before building `PartitionConfig`.

ui/src/pages/admin/partitions/detail.tsx
```diff
@@ -266,17 +271,37 @@ function GeneralTab({ partition }: { partition: PartitionConfig }) {
             </div>
           </div>
           <div className="space-y-2">
-            <Label>Chat History Depth</Label>
+            <Label className="flex items-center gap-1.5">
+              Chat History Depth *
+              <TooltipProvider>
+                <Tooltip>
+                  <TooltipTrigger asChild>
```
hedhoud: This help text is not reachable from the keyboard because `asChild` makes the non-focusable SVG the trigger. Please use a button trigger here and in the create dialog so the explanation is available without a pointer.

hedhoud (changes requested):
Two issues should be addressed before merge: legacy global configuration can now break partition loading, and the new help text is not keyboard-accessible.

openrag/services/orchestrators/partition_service.py
Ahmath-Gadji: Fixed in 6930f88c. `_legacy_chat_history_depth_fallback()` now clamps the live global value: if `rag.chat_history_depth` is `< 1` (the config carries no lower bound, so `0`/negative is possible) it falls back to the default `4` instead of propagating the value into `PartitionConfig(ge=1)`. That closes the startup crash — a legacy-zero row (including the seeded `default`) can no longer make `load_partitions()` raise `ValidationError`, regardless of the configured global depth. Valid values (>=1) still track the live config as before. Added a parametrized regression test (`test_resolve_partition_row_legacy_zero_clamps_invalid_global_default`) covering global `0` and `-1`.

ui/src/pages/admin/partitions/detail.tsx
Ahmath-Gadji: Fixed in 6930f88c. Replaced `<TooltipTrigger asChild>` with `<TooltipTrigger type="button" className="cursor-help">` in both the edit form (`detail.tsx`) and the create dialog (`list.tsx`), so Radix renders a focusable native `<button>` instead of putting the non-focusable `<Info>` SVG as the trigger. The tooltip is now reachable and toggleable from the keyboard.

openrag/services/orchestrators/query_service.py
```diff
@@ -138,9 +138,16 @@ def _resolve_chat_history_depth(self, partition: list[str] | None) -> int:
...
-        admin API) over the global default. A partition value of ``0`` means
-        "inherit the global default" — it never reaches the ``messages[-depth:]``
-        slice, where ``0`` would otherwise select the *entire* history.
+        admin API) over the global default. The admin API now requires an
+        explicit ``chat_history_depth >= 1`` on create/update (``ge=1`` in
+        ``CreatePartitionRequest``/``UpdatePartitionRequest``), and
+        ``PartitionService.resolve_partition_row`` normalizes any pre-existing
+        row still holding the legacy ``0`` sentinel to the global default
+        value before it reaches ``Settings.partitions``. So in practice
+        ``cfg.chat_history_depth`` is always >= 1 here; the ``> 0`` filter
```
hedhoud, line 158: One small edge case is still possible here. The partition value is now protected, but the global `rag.chat_history_depth` config can still be `0` or negative. For chats without a specific partition, or with `all`, we return that raw global value; if it is `0`, `messages[-0:]` keeps the full history, which is the behavior this PR is trying to avoid. Could we either validate the global config with the same minimum, or clamp the default to `4` when the configured value is less than `1`? A small test with global depth `0` would cover it.

hedhoud (commented):
I left one small follow-up comment on the remaining global-default edge case.

openrag/services/orchestrators/partition_service.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/6930f88cbcec574303e54711aaf75aa1863c39f3.

`_legacy_chat_history_depth_fallback()` now clamps to the hardcoded default of 4 when `self._config.rag.chat_history_depth` itself is < 1, instead of propagating an invalid global config value straight into `PartitionConfig(ge=1)`. So a misconfigured `rag.chat_history_depth: 0` (or negative) deployment can no longer crash `load_partitions()` at startup — legacy-zero rows fall back to 4 in that case. Regression test added: `test_resolve_partition_row_legacy_zero_clamps_invalid_global_default` (parametrized over global depth 0 and -1).

ui/src/pages/admin/partitions/detail.tsx
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/6930f88cbcec574303e54711aaf75aa1863c39f3.

Dropped `asChild` on both `TooltipTrigger`s (this one in `detail.tsx` and the matching one in `list.tsx`'s create dialog) so Radix renders its default focusable `<button type="button">` around the Info icon instead of making the raw SVG the trigger. Help text is now reachable via keyboard in both the edit form and the create dialog.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/083cea2c3cbead2a754189ee83cbe97c93672044.

Went with the clamp option, mirroring `PartitionService`'s pattern: `QueryService.__init__` now clamps `self._default_chat_history_depth` to a hardcoded `4` when `config.rag.chat_history_depth < 1`, so a misconfigured global default can no longer reach `messages[-depth:]` as `0`/negative for no-partition or `all`-sentinel chats. Regression test added: `test_default_chat_history_depth_clamps_invalid_global_config` (parametrized over global depth 0 and -1), asserting both `_default_chat_history_depth` and `_resolve_chat_history_depth(None)` land on 4.

hedhoud (approved):
LGTM. I rechecked the latest commit: the global chat history fallback is now clamped safely, the legacy partition fallback still behaves correctly, and the tooltip accessibility fix is in place.

PR #733, opened 2026-07-21
fix: make file count deletion atomic

### LINKED ISSUE #724: file_count decrement is not guarded by the DELETE rowcount — concurrent file/partition delete double-decrements quota
## Problem

`file_count` is decremented without checking that the `DELETE` actually removed a row, so a file deleted concurrently with its partition is counted twice.

`services/persistence/document_repo.py:370-382` (same shape at `:190-204`):
```python
row = await conn.fetchrow(
    "SELECT id, created_by FROM files WHERE file_id = $1 AND partition_name = $2", file_id, partition)
if row is None:
    return False
await conn.execute("DELETE FROM files WHERE id = $1", row["id"])      # rowcount discarded
if row["created_by"] is not None:
    await conn.execute(
        "UPDATE users SET file_count = GREATEST(file_count - 1, 0) WHERE id = $1", row["created_by"])
```

The `SELECT` takes no row lock and the `DELETE`'s affected-row count is never inspected.

## Failure scenario

Under READ COMMITTED, with `PgPartitionRepository.delete_partition` (`partition_repo.py:146-168`) already mid-transaction having snapshotted `COUNT(*) GROUP BY created_by`:

1. `delete_file` runs its `SELECT` and sees the row.
2. Its `DELETE` blocks on the partition-delete's lock, then commits affecting **0 rows**.
3. The decrement fires regardless.

Both transactions decrement for the same file. A user who owned 10 files loses 11 — and `GREATEST(..., 0)` clamps it to 0 rather than going negative, so the drift is **silent and unrecoverable** without manual SQL. The user's effective quota is permanently wrong in whichever direction the race landed.

## Fix

Make the decrement conditional on the delete actually removing the row:
```python
deleted = await conn.fetchrow("DELETE FROM files WHERE id = $1 RETURNING created_by", row["id"])
if deleted is not None and deleted["created_by"] is not None:
    ...decrement...
```
or take `SELECT ... FOR UPDATE` in the initial fetch.

Distinct from #664, which is the check/admit TOCTOU on the *upload* side; this is a lost-update on the delete side. #379 (closed) covered Milvus/PG divergence, not the counter.

hedhoud:
## Context

Concurrent file, bulk, and partition deletion could decrement a user's file count more than once for the same file. The zero clamp hid the drift, leaving quota decisions based on incorrect data.

## Fix

Deletion accounting now follows only rows actually removed by the transaction. Partition deletion is serialized, uploader counters are updated consistently, and retries remain no-ops.

Closes #724

## Validation

- Full unit suite: 1854 passed
- PostgreSQL concurrency and rollback coverage: 8 passed
- Ruff lint and formatting checks passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed per-user `file_count` tracking during document, file, and partition deletions.
  * Prevented duplicate counter decrements when deletions overlap or run concurrently.
  * Ensured deletions roll back safely if updating usage counts fails.
  * Clamped counter decrements to avoid negative `file_count`.
  * Improved bulk and unscoped deletion handling, including entries without uploader information.
* **Reliability**
  * Strengthened atomic deletion behavior under concurrent activity.
* **Tests**
  * Added new integration and unit coverage for counter correctness and rollback behavior.

### DIFF
--- a/openrag/services/persistence/document_repo.py
+++ b/openrag/services/persistence/document_repo.py
@@ -27,6 +27,8 @@ from typing import TYPE_CHECKING, Any
 
 from core.models.catalog import DocumentRecord, DocumentStatus
 from core.ports.document_repo import DocumentRepository
+from core.utils.logging import get_logger
+from services.persistence.file_count import decrement_file_counts
 
 if TYPE_CHECKING:
     from datetime import datetime
@@ -37,6 +39,8 @@ if TYPE_CHECKING:
 # on every connection, so reading a JSON column yields a Python dict and
 # binding a dict to a JSON parameter is encoded transparently. The repo
 # therefore never calls ``json.dumps`` itself.
+
+logger = get_logger()
 
 
 class PgDocumentRepository(DocumentRepository):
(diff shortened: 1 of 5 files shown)


Ahmath-Gadji (approved):
Approved. Correct fix for the double-decrement race in per-user `file_count` accounting.

- `DELETE … RETURNING created_by` makes "deleted" and "counted" the same event, eliminating the read-then-write race across all deletion paths.
- Deadlock avoidance is deliberate and correct: `decrement_file_counts` applies `UPDATE users` in sorted user-id order, and `delete_partition` serializes via `SELECT … FOR UPDATE`.
- Strong test coverage — the integration tests poll `pg_locks` to deterministically force the race the old code failed, plus an injected-trigger rollback test.

Optional (non-blocking) follow-ups: the per-uploader `UPDATE` fan-out in `decrement_file_counts` could be collapsed into a single `UPDATE … FROM unnest(...)`, and a one-line comment noting why `delete_documents_by_partition` intentionally omits the partition lock would help future readers.

Issue #707, opened 2026-07-20
rrf_k is configured and UI-editable but never passed to rrf_reranking (fusion constant hardcoded to 60)


andyne13:
## Problem

`rrf_k` is configured, range-validated, and editable per-preset in the admin UI — and never reaches the fusion function. The RRF constant is permanently hardcoded.

Whole-repo grep returns exactly four hits:
```
openrag/core/config/retrieval_pipeline.py:30:  rrf_k: int = Field(default=60, gt=0, le=1000)
ui/src/pages/admin/presets.tsx:772-773:        value/onChange for "rrf_k"
```
Both fusion call sites omit `k`:
```python
openrag/services/orchestrators/retrieval_service.py:338:  fused = rrf_reranking(doc_lists, key_fn=_chunk_key)
openrag/core/retrieval/pipeline.py:149:            fused = rrf_reranking(ranked_lists, key_fn=_chunk_key)
```
`rrf_reranking(..., k: int = 60)` — so `k` is always 60.

Worse, the two fusion layers disagree with no config path to reconcile them: `openrag/services/storage/milvus_store.py:104` uses `RRF_K = 100` for Milvus-native hybrid fusion.

## Impact

An operator tunes RRF dampening per partition, the value persists to the `pipeline_presets` JSONB, the UI reads it back correctly — and ranking never changes.

## Fix

Thread `pipeline_cfg.rrf_k` into `RetrieverPipeline` and pass `k=` at both call sites; reconcile the Milvus-native `RRF_K` with the same setting, or drop the field and the UI control.

Verified on `develop` @ `e61479b4`.

PR #734, opened 2026-07-21
fix(retrieval): expand "all" to per-partition pipelines (right embedder + top_n), bounded

### LINKED ISSUE #708: Multi-partition (all) search uses the default embedder and drops reranker top_n
## Problem

`_pipeline_groups_for_partitions` short-circuits the `all` / no-preset path to the deployment-default pipeline, losing **both** the per-partition embedder and the reranker `top_n`.

`openrag/services/orchestrators/retrieval_service.py:213-217`:
```python
def _pipeline_groups_for_partitions(self, partitions):
    if not partitions or "all" in partitions or not self._partition_configs():
        return [(["all"] if "all" in partitions else partitions, self._pipeline, None)]
    return [([partition], pipeline, default_top_k) for partition in partitions ...]
```

### 1. Wrong embedder
The named branch resolves the partition's embedder (`:167-168`, `self._searcher_factory(partition_cfg.embedder)`), but `self._pipeline` is built once in `di/container.py:475-491` from `settings.embedder`. Index time *does* honour `partition_cfg.embedder` (`indexing_service.py:200-209` → `pipeline_builder._select_embedder`).

So a partition indexed with a named embedder is queried with a **different model's** query vector. The collection is shared and `COSINE` (`milvus_store.py:343`), so results come back — just near-random. Recall collapses with nothing logged. Single-partition search of the same partition works fine, which makes it read as "all-search is just noisier".

### 2. Reranker top_n dropped
`default_top_k` is `None` on this path, and `core/retrieval/pipeline.py:129` only truncates `if top_k is not None` (the reranker itself is called with `top_k=None` at `:119-129`). Named partitions apply `top_n` correctly; `all` — and any deployment with no partition presets — does not. The full candidate set flows into `format_context`, where the token budget silently decides how many sources survive, with no log line (`core/prompts/chat_prompt_builder.py:75`).

## Impact

Affects `GET /search?partitions=all` and `model: openrag-all`, plus every deployment that hasn't created partition presets.

## Fix

Expand `all` to the caller's accessible partition list and build one pipeline per partition (as the non-`all` branch already does), rather than collapsing to `self._pipeline`. Fall back to `self._config.reranker.top_k` instead of `None`.

andyne13:
Closes #708.

Two commits: the correctness fix, then a production-safety cap found by an adversarial pre-PR audit.

## 1. Wrong embedder + dropped top_n on the `all` path

`_pipeline_groups_for_partitions` collapsed the `"all"` case to the default-embedder legacy pipeline (`self._pipeline`, `default_top_k=None`). So a partition indexed with a named embedder was queried with a **different model's vector** (near-random recall — the collection is COSINE, so results still come back), and the reranker `top_n` was silently dropped (a token heuristic decided the count instead).

Fix: expand `"all"` to the hydrated partition set and reuse the **existing** per-partition fan-out + `fuse()` — the same path regular multi-partition users already run — so each partition is retrieved with its own embedder and `top_n`. The single-legacy-pipeline fallback is kept only when zero partitions exist.

### Scope / security (verified, not assumed)
`"all"` only reaches this layer **post-authorization**: `get_partition_name` (`api/dependencies/llm.py:103-110`) expands `openrag-all` to a regular user's concrete membership list upstream, so they already take the per-partition branch. The literal `["all"]` is a **SUPER_ADMIN_MODE** admin, for whom every partition is legitimately in scope — so expanding to all hydrated partitions is correct and leaks nothing. The `/search` route uses a separate raw-search path and is unaffected. `config.partitions` is hydrated from DB presets by `PartitionService.load_partitions()`, so per-partition config exists for every partition; the legacy collapse only fires on a fresh system with zero partitions.

## 2. Bounded fan-out (production safety)

The expansion is correct but, without a cap, a super-admin `openrag-all` launched **one embed+Milvus search per partition** in a single unbounded `asyncio.gather` — proportional to the partition count, and ×sub-queries for multiQuery/hyde. On a large multi-tenant deployment (the SaaS-orgs direction), one admin request could exhaust the Milvus/embedder pools and starve the event loop (~4,500 concurrent ops at 500 partitions × 3 sub-queries × 3).

Fix: a per-call semaphore over the per-partition leaf coroutines — `retriever.max_partition_concurrency` (default 16, env `RETRIEVER_MAX_PARTITION_CONCURRENCY`). A fast path keeps fan-outs **within** the cap fully parallel and byte-identical to before, so regular multi-partition users are untouched; only oversized fan-outs are throttled.

Design notes:
- The cap is **per-call, not a shared instance semaphore** — it bounds a single `retrieve()` call's partition fan-out without coupling concurrent requests. Under `retrieve_per_query` (multiQuery/hyde) the N sub-queries each get their own cap, so worst-case concurrency is `N × max_partition_concurrency` — deliberately not a flat per-request cap, but still bounded and no longer partition-count-proportional (the blowup being fixed); N is small (`k_queries`, default 3).
- It wraps only the **leaf** coroutines, so it composes across the `retrieve_per_query → retrieve` nesting: the awaited coroutines hold no permit while waiting for one → no cross-level deadlock.

## Verified

- Security: no cross-tenant leak (traced every caller of the `["all"]` path).
- Concurrency race: snapshot-and-build is fully synchronous (no `await` between `list(configs.keys())` and pipeline construction), so a concurrent `load_partitions`/`pop` cannot interleave — no `PartitionNotFoundError`, no "dict changed size".
- Tests assert the *effects*, and each fails against its pre-fix code:
  - each partition searched with its **own** embedder, scoped to itself (never the default searcher, never `partition=["all"]`)
  - `top_n` truncates on the `all` path
  - the cap engages (max concurrent == limit over a larger fan-out) and within-cap fan-outs stay fully parallel
  - zero-partition fallback unchanged
- Full suite: **1846 passed**; ruff check / format / layer-import-guard green.

## Deliberately out of scope (tracked follow-ups)

- **Group by embedder+config** (#735) — in a homogeneous deployment the N per-partition searches could collapse to one `partition=[list]` query. Bigger change that also touches the regular-user path.
- **`return_exceptions` on the fan-out** (#736) — one failing partition currently aborts the whole `all`-search (pre-existing for regular multi-partition users too).

## Summary by CodeRabbit

* **New Features**
  * “All partitions” retrieval now expands into per-partition searches using each partition’s configured settings.
  * Added `RETRIEVER_MAX_PARTITION_CONCURRENCY` (default: 16) to cap concurrent per-partition retrievals within a request.
  * The concurrency cap is enforced for both single-query and multi-query retrieval.
* **Bug Fixes**
  * Partition-specific `top_n` truncation is applied correctly when searching across all partitions.
  * Legacy fallback behavior remains unchanged when no partition configurations are available.

### DIFF
--- a/docs/content/docs/documentation/env_vars.md
+++ b/docs/content/docs/documentation/env_vars.md
@@ -305,6 +305,7 @@ The retriever fetches relevant documents from the vector database based on query
 | `RELATED_LIMIT` | `int` | 10 | Maximum number of related/ancestor chunks fetched per matched result when expansion is enabled. |
 | `MAX_DEPTH` | `int` | 10 | Maximum ancestor depth traversed when `INCLUDE_ANCESTORS` is enabled. |
 | `RETRIEVER_ALLOW_FILTERLESS_FALLBACK` | `bool` | true | When a temporally-filtered retrieval returns no documents, re-run the query without the filter. Set to `false` for strict temporal retrieval. |
+| `RETRIEVER_MAX_PARTITION_CONCURRENCY` | `int` | 16 | Upper bound on how many per-partition retrievals run concurrently within a single request. Bounds fan-out for multi-partition / `openrag-all` searches; small fan-outs stay fully parallel. |
 
 #### Retrieval Strategies
 
--- a/docs/content/docs/documentation/env_vars.md
+++ b/docs/content/docs/documentation/env_vars.md
@@ -306,7 +306,7 @@ The retriever fetches relevant documents from the vector database based on query
 | `RELATED_LIMIT` | `int` | 10 | Maximum number of related/ancestor chunks fetched per matched result when expansion is enabled. |
 | `MAX_DEPTH` | `int` | 10 | Maximum ancestor depth traversed when `INCLUDE_ANCESTORS` is enabled. |
 | `RETRIEVER_ALLOW_FILTERLESS_FALLBACK` | `bool` | true | When a temporally-filtered retrieval returns no documents, re-run the query without the filter. Set to `false` for strict temporal retrieval. |
-| `RETRIEVER_MAX_PARTITION_CONCURRENCY` | `int` | 16 | Upper bound on how many per-partition retrievals run concurrently within a single request. Bounds fan-out for multi-partition / `openrag-all` searches; small fan-outs stay fully parallel. |
+| `RETRIEVER_MAX_PARTITION_CONCURRENCY` | `int` | 16 | Upper bound on how many per-partition retrievals run concurrently **per retrieval call**. Bounds fan-out for multi-partition / `openrag-all` searches; small fan-outs stay fully parallel. Note: a multi-query request (multiQuery/hyde) issues one such call per sub-query, so peak concurrency can reach `N × this value`, not a flat per-request cap. |
 
 #### Retrieval Strategies
 
--- a/openrag/core/config/loader.py
+++ b/openrag/core/config/loader.py
@@ -154,6 +154,7 @@ _ENV_OVERRIDES: list[tuple[str, str, type]] = [
     ("RELATED_LIMIT", "retriever.related_limit", int),
     ("MAX_DEPTH", "retriever.max_ancestor_depth", int),
     ("RETRIEVER_ALLOW_FILTERLESS_FALLBACK", "retriever.allow_filterless_fallback", bool),
+    ("RETRIEVER_MAX_PARTITION_CONCURRENCY", "retriever.max_partition_concurrency", int),
     # RAG
     ("RAG_MODE", "rag.mode", str),
     # WebSearch
--- a/openrag/core/config/retrieval.py
+++ b/openrag/core/config/retrieval.py
@@ -67,6 +67,13 @@ class _BaseRetrieverConfig(ConfigMixin):
     # document hierarchies.
     max_ancestor_depth_cap: int = 1000
     allow_filterless_fallback: bool = True
+    # Upper bound on how many per-partition retrievals run concurrently within a
+    # single request. A multi-partition search (notably a SUPER_ADMIN_MODE
+    # ``openrag-all`` that expands to every partition) fans out one embed+search
+    # per partition; without a cap that is proportional to the partition count
+    # and can exhaust the Milvus/embedder pools from one request. Small fan-outs
+    # (the common case) stay fully parallel — the cap only engages past it.
+    max_partition_concurrency: int = Field(default=16, gt=0)
 
 
 class SingleRetrieverConfig(_BaseRetrieverConfig):
--- a/openrag/services/orchestrators/retrieval_service.py
+++ b/openrag/services/orchestrators/retrieval_service.py
@@ -213,7 +213,21 @@ class RetrievalService:
     def _pipeline_groups_for_partitions(
         self, partitions: list[str]
     ) -> list[tuple[list[str], RetrieverPipeline, int | None]]:
-        if not partitions or "all" in partitions or not self._partition_configs():
+        configs = self._partition_configs()
+        # Expand the "all" sentinel to concrete partitions so each is retrieved
+        # with its own embedder and top_n, via the same per-partition fan-out the
+        # named-partition path already uses (#708). Collapsing to self._pipeline
+        # embedded the query with the deployment-default embedder — near-random
+        # recall for any partition indexed with a different one — and dropped the
+        # reranker top_n (its default_top_k was None). "all" only reaches this
+        # layer post-authorization (a SUPER_ADMIN_MODE admin; regular users are
+        # already expanded to their memberships upstream), so every hydrated
+        # partition is in scope.
+        if "all" in partitions and configs:
+            partitions = list(configs.keys())
+        elif not partitions or not configs:
+            # Nothing to expand (no partitions exist yet) — keep the single
+            # legacy pipeline; there is no per-partition config to honour.
             return [(["all"] if "all" in partitions else partitions, self._pipeline, None)]
         return [
             ([partition], pipeline, default_top_k)
--- a/openrag/services/orchestrators/retrieval_service.py
+++ b/openrag/services/orchestrators/retrieval_service.py
@@ -165,7 +165,10 @@ class RetrievalService:
         return getattr(self._config.retriever, name, default)
 
     def _pipeline_for_partition(self, partition: str) -> tuple[RetrieverPipeline, int | None]:
-        if partition == "all" or not self._partition_configs():
+        # Callers only ever pass a concrete partition name — the "all" sentinel is
+        # expanded to concrete keys by _pipeline_groups_for_partitions before this
+        # runs. With no per-partition configs at all, fall back to the legacy pipeline.
+        if not self._partition_configs():
             return self._pipeline, None
 
         partition_cfg = self._require_partition_config(partition)
@@ -300,7 +303,7 @@ class RetrievalService:
         its own leaves; the coroutines being awaited hold no permit while
         waiting for one, so there is no cross-level deadlock).
         """
-        limit = getattr(self._config.retriever, "max_partition_concurrency", 16) or 16
+        limit = self._config.retriever.max_partition_concurrency
         if len(coros) <= limit:
             return await asyncio.gather(*coros)
 
--- a/openrag/services/orchestrators/retrieval_service.py
+++ b/openrag/services/orchestrators/retrieval_service.py
@@ -284,6 +284,34 @@ class RetrievalService:
     # Pipeline retrieval (powers QueryService — 8C.2)
     # ------------------------------------------------------------------
 
+    async def _gather_partition_groups(self, coros: list) -> list:
+        """Await one coroutine per partition group, bounding concurrency.
+
+        Small fan-outs (the common case: a handful of partitions) run fully
+        parallel via a plain gather — no added overhead, byte-identical to the
+        prior behaviour. Only a fan-out larger than ``max_partition_concurrency``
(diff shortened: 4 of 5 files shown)


Ahmath-Gadji:
## Review

Verified locally: checked out the PR head in a worktree, ran the new/modified unit tests (`tests/unit/services/orchestrators/test_retrieval_service.py`, 16 passed) and `ruff check` on the changed files (clean).

### Overview
Fixes #708: `_pipeline_groups_for_partitions` previously collapsed the `"all"` partition sentinel to the single legacy pipeline (deployment-default embedder, `default_top_k=None`), so a `SUPER_ADMIN_MODE` `openrag-all` search on a partition indexed with a non-default embedder silently searched with the wrong vector space and dropped the reranker's `top_n`. The fix expands `"all"` to the hydrated partition list and reuses the existing per-partition fan-out (`_pipeline_for_partition` + `fuse()`), so each partition is searched with its own embedder/top_n — same path regular multi-partition users already take. A second commit adds `retriever.max_partition_concurrency` (default 16, env `RETRIEVER_MAX_PARTITION_CONCURRENCY`) as a per-call semaphore around the leaf coroutines, since without it the expansion turns one admin request into one embed+Milvus call per partition.

### Correctness
- Traced the `"all"` sentinel's only caller path (`get_partition_name` in `api/dependencies/llm.py`): regular users are already expanded to concrete memberships upstream, so the literal `["all"]` only reaches this layer for a `SUPER_ADMIN_MODE` admin — expanding to every hydrated partition is the correct scope, not a leak.
- The zero-partition fallback (fresh system, no configs yet) correctly still routes to the single legacy pipeline searching `partition=["all"]` — covered by `test_retrieve_all_falls_back_to_legacy_when_no_partitions_exist`.
- `_pipeline_groups_for_partitions` has no `await` between snapshotting `configs` and building each partition's pipeline, so a concurrent `load_partitions`/`partitions.pop` (both `async def` on the same event loop) cannot interleave mid-build — no `PartitionNotFoundError`/`dict changed size` race. Confirmed by reading `partition_service.py`.
- Reused `_pipeline_for_partition` already existed and is exercised by the current single-named-partition path, so this isn't new untested machinery — it's the existing per-partition builder now also driving the `"all"` case.

### Minor / non-blocking observations
- `_pipeline_for_partition`'s `partition == "all"` branch (`retrieval_service.py:168`) is now unreachable dead code — after this change, the loop in `_pipeline_groups_for_partitions` only ever calls it with concrete partition names (the `"all"` case is handled and returned before the list comprehension runs, and there's no other caller). Not a bug, just worth a follow-up cleanup.
- The concurrency cap is explicitly per-call (a fresh `asyncio.Semaphore(limit)` inside `_gather_partition_groups` on each invocation), not shared instance-wide. That's fine and avoids cross-request coupling, but note `retrieve_per_query` fans out over sub-queries via a plain unbounded `asyncio.gather` at the top level, and each of those concurrently calls `self.retrieve(...)`, which creates its *own* `limit`-sized semaphore. So for a flow that hits `retrieve_per_query` with N queries, worst-case concurrent partition-level operations is `N × max_partition_concurrency`, not a flat `max_partition_concurrency` for the whole request. Given typical N is small (a handful of sub-queries), this still prevents the original partition-count-proportional blowup, but it's worth confirming that's the intended guarantee rather than a strict per-request cap — the PR description's framing ("bounds one request's fan-out") could be read either way.
- New env var `RETRIEVER_MAX_PARTITION_CONCURRENCY` isn't added to `infra/compose/.env.example`, though that file's header states it intentionally documents only a curated subset of knobs (other retriever settings like `RETRIEVER_ALLOW_FILTERLESS_FALLBACK` aren't listed either), so this is consistent with existing convention, not a gap.

### Tests
Good coverage of the actual regression: per-partition embedder isolation, `top_n` truncation on the `all` path, the zero-partition fallback, and both the fast-path (fully parallel, within cap) and bounded (semaphore-throttled) fan-out cases, with each test asserting on the fix's actual runtime effect (live-concurrency tracking) rather than just call counts. No bugs found — approving.

Ahmath-Gadji:
One more nit: the new `RETRIEVER_MAX_PARTITION_CONCURRENCY` env var isn't documented in `docs/content/docs/documentation/env_vars.md`. Unlike `.env.example` (which intentionally lists only a curated subset), this file's "Retriever Configuration" table (around line 294) documents every other retriever knob, including `RETRIEVER_ALLOW_FILTERLESS_FALLBACK` right above where this one would go. Worth adding a row there, e.g.:

`| \`RETRIEVER_MAX_PARTITION_CONCURRENCY\` | \`int\` | 16 | Upper bound on how many per-partition retrievals run concurrently within a single request. Bounds fan-out for multi-partition/\`openrag-all\` searches; small fan-outs stay fully parallel. |`

Ahmath-Gadji:
Nit on `retrieval_service.py` (the `_gather_partition_groups` limit line, not a bug): `getattr(self._config.retriever, "max_partition_concurrency", 16) or 16` can just be `self._config.retriever.max_partition_concurrency`. Every real retriever config variant (`SingleRetrieverConfig`/`MultiQueryRetrieverConfig`/`HydeRetrieverConfig`) inherits the field from `_BaseRetrieverConfig` with `Field(default=16, gt=0)`, so it's always present and pydantic already guarantees it's a positive int at config-load time — the `getattr` default and the `or 16` are both guarding against a state that can't occur, which reads as if the attribute might be missing/falsy when it can't be.

The only reason to keep it as-is: the unit tests build `retriever` as a bare `SimpleNamespace` in `_config()` (`tests/unit/services/orchestrators/test_retrieval_service.py`), which doesn't set `max_partition_concurrency` by default, so a direct access would `AttributeError` on most of the existing tests. Switching would need `max_partition_concurrency=16` added to that fixture alongside the code change.

andyne13:
Thanks for the thorough review @Ahmath-Gadji (and CodeRabbit). Addressed all the points in 0511e0df:

- **Unreachable `partition == "all"` branch** (CodeRabbit + your review) — removed. `_pipeline_groups_for_partitions` expands `"all"` to concrete keys (or returns the legacy fallback) before `_pipeline_for_partition` runs, so its only caller never passes the sentinel. Kept the no-configs guard and added a comment explaining callers only pass concrete names.
- **`getattr(..., 16) or 16` simplification** — done, now `self._config.retriever.max_partition_concurrency`. As you noted, `_BaseRetrieverConfig` defines it `Field(default=16, gt=0)`, so both guards protected a state pydantic already prevents. Added `max_partition_concurrency=16` to the `_config()` test fixture so the direct access holds there too.
- **`RETRIEVER_MAX_PARTITION_CONCURRENCY` docs** — added a row to the Retriever Configuration table in `env_vars.md`, right after `RETRIEVER_ALLOW_FILTERLESS_FALLBACK`, using your suggested wording.
- **N × cap under `retrieve_per_query`** — confirmed intended, not a strict per-request cap. The semaphore is deliberately per-call so concurrent requests aren't coupled; under `retrieve_per_query`'s unbounded gather over N sub-queries the worst case is N × `max_partition_concurrency`. With N small (`k_queries` default 3) this still eliminates the original partition-count-proportional blowup, which was the bug. The `_gather_partition_groups` docstring documents the per-call/compositional semantics; I've left the guarantee as-is intentionally.

16 retrieval-service tests + 368 orchestrator tests pass; ruff + layer-import-guard green.

Ahmath-Gadji (dismissed):
Approving — all three of my review points are resolved in \`0511e0df\`, verified against the code (not just the changelog):

1. **Unreachable \`partition == "all"\` branch** — removed from \`_pipeline_for_partition\`; now \`if not self._partition_configs()\` with a comment explaining that \`_pipeline_groups_for_partitions\` expands the sentinel to concrete keys (or takes the legacy fallback) before this ever runs. Traced both callers — the method only receives concrete partition names now. ✅
2. **\`getattr(..., 16) or 16\` → \`self._config.retriever.max_partition_concurrency\`** — done, and \`max_partition_concurrency=16\` added to the \`_config()\` test fixture so the direct access holds. ✅
3. **Docs** — \`RETRIEVER_MAX_PARTITION_CONCURRENCY\` row added to the Retriever Configuration table in \`env_vars.md\`, right after \`RETRIEVER_ALLOW_FILTERLESS_FALLBACK\`. ✅

Re-ran the changed suite at HEAD: `tests/unit/services/orchestrators/test_retrieval_service.py` — **16 passed**; `ruff check` on the changed source files clean.

⚠️ **Pre-merge, not blocking approval:** the branch currently conflicts with `develop` in `tests/unit/services/orchestrators/test_retrieval_service.py` — `develop` gained `test_pipeline_for_partition_threads_rrf_k` (#707) next to the `_config()` fixture this PR edits. A quick rebase should resolve it (source files auto-merge cleanly).

andyne13:
Heads-up @Ahmath-Gadji — your approval was auto-dismissed because I had to resolve a merge conflict with develop (the branch had drifted behind #740/#737). The conflict was purely in `test_retrieval_service.py` — both sides added independent tests (your #708 fan-out tests + the #707 `test_pipeline_for_partition_threads_rrf_k`); I kept both, no production-code change beyond the earlier review fixes. Full unit suite is green (2042 passed) and CI is green. Could you re-approve so it can merge? Nothing else changed since your 13:03 sign-off.

Ahmath-Gadji (dismissed):
LGTM

andyne13:
Addressed the CodeRabbit doc nit in 83d34434: the `RETRIEVER_MAX_PARTITION_CONCURRENCY` row now says the cap is **per retrieval call**, and explicitly notes a multi-query request reaches `N × this value` (not a flat per-request cap) — consistent with the PR description. Docs-only change; no code touched.

(The prior `api-tests` red was an unrelated CI infra flake — `ECONNRESET` while downloading docker-compose in the setup step, before any test ran; the new run should clear it.)

Ahmath-Gadji (approved):

PR #737, opened 2026-07-21
fix(indexation): default preset follows the global chunking config (CHUNK_SIZE/CHUNKER)

### LINKED ISSUE #709: Global CHUNK_SIZE / CHUNK_OVERLAP_RATE / CHUNKER are silently ignored (seeded preset hardcodes 512/0.2)
## Problem

`CHUNK_SIZE`, `CHUNK_OVERLAP_RATE` and `CHUNKER` load into `cfg.chunker` and are then ignored for every indexed document, because the seeded `default` indexation preset hardcodes chunking and the preset always wins.

Seed — `openrag/services/orchestrators/preset_service.py:47`:
```python
"chunking": {"name": "recursive_splitter", "chunk_size": 512, "chunk_overlap_rate": 0.2},
```
`_finalize_seed` deliberately folds *some* globals into the seed but not the chunking ones — it injects `enable_contextualization` and `enable_image_captioning` only.

Runtime — `openrag/services/workers/pipeline_builder.py:293-296`:
```python
def _select_chunker(self, config):
    if config is not None and self.chunker_factory is not None:
        return self.chunker_factory(config.chunking)
    return self.chunker          # global cfg.chunker — unreachable once seeded
```
`IndexingService._resolve_indexation_dispatch_config` always returns `partition_cfg.indexation`, defaulting to the `default` preset, so `config` is non-`None` for every file in any deployment that has run `seed_defaults()`.

## Impact

An operator sets `CHUNK_SIZE=1024` (documented in `conf/config.yaml` and mapped in `core/config/loader.py:145-146`). It loads, it's visible in config introspection, and every document is still chunked at 512/0.2. Nothing errors.

## Fix

Inject the global chunker settings into the seeded `default` indexation preset in `_finalize_seed`, the same way the contextualization and captioning toggles already are — or document that global chunking settings are superseded by presets and remove the env vars.

andyne13:
Closes #709 (fresh-install scope — see "Scope & known limitation" below).

## Problem

The seeded `default` indexation preset hardcodes chunking (512 / 0.2), and the preset always wins at runtime (`pipeline_builder._select_chunker`), so a deployment's global `CHUNKER` / `CHUNK_SIZE` / `CHUNK_OVERLAP_RATE` (env → `config.chunker`) was silently ignored — the value loaded, showed in config introspection, and never reached the splitter.

## Fix

`_finalize_seed` now overlays the global chunker's `name` / `chunk_size` / `chunk_overlap_rate` onto the `default` preset — mirroring how `enable_contextualization` / `enable_image_captioning` are already folded in. Only the 3 shape fields, only for `default`; named presets (legal/finance) keep their deliberate sizes.

## Hardening (added after an adversarial pre-merge review)

- **`chunk_overlap_rate` bounded to [0, 1)** and `chunk_size > 0` on `ChunkerConfig`. An overlap ≥ size makes the recursive splitter raise at construction, so a bad `CHUNK_OVERLAP_RATE` now fails at **config load** (earliest gate) instead of per-file — closing both the surface this overlay newly exposed on the default preset and the pre-existing eager-global-build crash, at the source.
- **Seed path now runs `_validate_config`** (the same validation the CRUD path uses) as defense-in-depth, so a structurally-invalid overlaid seed is rejected at seeding, not persisted.
- **Unknown chunker *name* now rejected** in `_validate_config` against the chunking registry. Previously the name was a free string that passed structural validation and only failed later at chunker build — once per document, at index time. Since this PR makes `CHUNKER` actually reach the default preset, a typo would now propagate to the seed; it instead fails fast at **seed time (startup)** and on the **admin CRUD path**, listing the available chunkers. (Added after review — same "a configured value is now actually consumed, so validate it" pattern as the bounds above.)

## Scope & known limitations (important)

This makes a **fresh, never-seeded** deployment honour the global chunking config. It does **not** retroactively change an **already-seeded** deployment: `seed_defaults` skips a type that already has rows, so a stored `default` preset keeps its `chunk_size` even if `CHUNK_SIZE` changes. This is **consistent with the existing `_finalize_seed` contract** — the sibling `enable_contextualization` / `enable_image_captioning` overlays behave identically. Existing deployments change chunking by editing the `default` preset (admin UI / `update_preset`).

Reconciliation for existing deployments is tracked in #739 (it can't be done safely without provenance — a stored 512 is indistinguishable from a deliberate choice).

**Upgrade caveat (bounds are global, not fresh-install-only):** the new `chunk_size > 0` / `chunk_overlap_rate ∈ [0, 1)` bounds apply at **global config load for every deployment**. A deployment that had set a genuinely-invalid-but-previously-tolerated value (e.g. `CHUNK_OVERLAP_RATE=1.0`, which the old hardcoded preset ignored) will now fail fast at **startup** with a clear Pydantic error instead of silently ignoring it. This is deliberate (fail loud on invalid config, don't let it reach the splitter per-file), but it is a boot-time behaviour change on upgrade — flagged here so it isn't a surprise.

## Tests (each meaningful; the guard tests fail on revert)

- `default` preset reflects global `CHUNK_SIZE=1024` / overlap `0.3` — **fails against pre-fix code** (stays 512)
- named `finance` keeps its explicit 768 despite global 1024
- overlap out of [0,1) rejected at config load — **fails without the new bound**
- **documents the limitation**: an already-seeded `default` is NOT rewritten by a changed `CHUNK_SIZE`
- seed-path validation rejects a structurally-invalid finalized chunking
- unknown global `CHUNKER` rejected at `seed_defaults` (startup) and unknown chunker name rejected on the CRUD path — both **fail on revert** of the new registry check

Full suite (merged with latest develop): green; ruff check / format / layer-import-guard green.

## Summary by CodeRabbit

* **New Features**
  * Default indexation presets now inherit configured chunking settings.
  * Existing default preset settings are preserved during seeding.
* **Bug Fixes**
  * Invalid chunk sizes, overlap rates, and unknown chunker names are rejected before saving.
  * Preset creation and default seeding now validate configurations consistently.
* **Tests**
  * Added coverage for global chunker settings, preset preservation, fallback values, and validation errors.

### DIFF
--- a/openrag/core/config/chunking.py
+++ b/openrag/core/config/chunking.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from pydantic import Field
+
 from .base import ConfigMixin
 
 
@@ -12,5 +14,8 @@ class ChunkerConfig(ConfigMixin):
     contextual_retrieval: bool = True
     contextualization_timeout: int = 120
     max_concurrent_contextualization: int = 10
-    chunk_size: int = 512
-    chunk_overlap_rate: float = 0.2
+    chunk_size: int = Field(default=512, gt=0)
+    # Bounded to [0, 1): an overlap >= chunk_size makes the recursive splitter
+    # raise at construction ("larger chunk overlap than chunk size"), so a bad
+    # CHUNK_OVERLAP_RATE must fail at config load, not per-file at index time.
+    chunk_overlap_rate: float = Field(default=0.2, ge=0.0, lt=1.0)
--- a/openrag/services/orchestrators/preset_service.py
+++ b/openrag/services/orchestrators/preset_service.py
@@ -309,14 +309,38 @@ class PresetService:
 
     def _validate_config(self, preset_type: str, config: dict[str, Any]) -> None:
         """Instantiate the Pydantic pipeline model to validate the config dict."""
+        indexation_model: IndexationPipelineConfig | None = None
         try:
             match preset_type:
                 case "indexation":
-                    IndexationPipelineConfig(**config)
+                    indexation_model = IndexationPipelineConfig(**config)
                 case "retrieval":
                     RetrievalPipelineConfig(**config)
         except Exception as exc:
             raise ValidationError(f"Invalid {preset_type} preset config: {exc}") from exc
+
+        # Structural validation accepts any chunker *name* (ChunkerConfig.name is a
+        # free string), but only registered strategies can actually be built. Reject
+        # an unknown name here — at seed time (startup) and on the admin CRUD path —
+        # so a typo fails fast instead of letting every document explode at chunker
+        # build during indexing. #709 wired the global CHUNKER into the default
+        # preset, so a bad env value now reaches this seed instead of being ignored.
+        if indexation_model is not None:
+            self._ensure_chunker_registered(indexation_model.chunking.name)
+
+    @staticmethod
+    def _ensure_chunker_registered(name: str) -> None:
+        """Raise ``ValidationError`` if *name* is not a registered chunking strategy."""
+        # Importing the factory triggers chunker registration (import side-effect),
+        # so the registry is populated before the membership check. Mirrors the
+        # trigger used by ``create_chunker`` itself.
+        import core.chunking.factory  # noqa: F401
+        from core.chunking.registry import chunking_registry
+
+        if name not in chunking_registry:
+            raise ValidationError(
+                f"Unknown chunker '{name}'. Available chunkers: {chunking_registry.list_registered()}"
+            )
 
 
 __all__ = ["PresetService"]
--- a/openrag/services/orchestrators/preset_service.py
+++ b/openrag/services/orchestrators/preset_service.py
@@ -149,8 +149,20 @@ class PresetService:
         if preset_type == "retrieval":
             return {**config, "enable_reranker": self._config.reranker.enabled}
         if preset_type == "indexation" and name == "default":
+            chunker = self._config.chunker
             return {
                 **config,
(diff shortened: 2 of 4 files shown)


hedhoud (dismissed):
LGTM. I checked the seed-time behavior and the default preset now follows the global chunking config for fresh installs, while existing DB presets stay admin-managed. Targeted tests and checks are green on my side too.

andyne13:
Addressed the remaining review points:

- **CodeRabbit nit** — dropped the redundant `@pytest.mark.asyncio`/`async def` on the synchronous `test_seed_defaults_rejects_invalid_overlap_in_default_chunking` (it calls the sync `_validate_config`, no `await`).
- **Unknown chunker name now validated** — the earlier version left a bad `CHUNKER` *name* as a free string that only failed at chunker build (per-file, at index time). `_validate_config` now checks the name against the chunking registry, so a typo fails fast at seed time (startup) and on the CRUD path, listing the available chunkers. Two revert-proved tests added (`test_seed_defaults_rejects_unknown_global_chunker`, `test_create_preset_rejects_unknown_chunker`).
- **PR description updated** to flag that the new `ChunkerConfig` bounds are a global config-load gate (a boot-time change on upgrade for deployments that had an invalid-but-tolerated value), not fresh-install-only.

38 preset-service tests + 448 orchestrator/config tests pass; ruff + layer-import-guard green. Re-review appreciated since the new commit dismissed the approval.

hedhoud (approved):
Approved. The default preset now follows the global chunking settings on a fresh install, invalid chunker configuration fails early, and named or already-seeded presets keep their existing ownership behavior. Focused config and preset tests pass.

PR #740, opened 2026-07-21
fix(retrieval): thread per-partition rrf_k into sub-query fusion (incl. websearch path)

### LINKED ISSUE #707: rrf_k is configured and UI-editable but never passed to rrf_reranking (fusion constant hardcoded to 60)

andyne13:
Closes #707.

## Problem
`RetrievalPipelineConfig.rrf_k` (per-partition, UI-editable, validated) was never passed to `rrf_reranking` — the app-level RRF dampening constant was hardcoded to the function default 60, so per-partition tuning did nothing.

## Fix
- Thread `rrf_k` into `RetrieverPipeline` and use it in `get_relevant_docs`, where a partition's multiQuery sub-query rankings are fused (the fusion the per-partition field governs, and the sole fusion for a single-partition multiQuery search). Wired from `pipeline_cfg.rrf_k`.
- **websearch path** (found by adversarial review): the combined RAG+websearch branch fused sub-queries via `retrieve_per_query` + `fuse()` at the hardcoded 60, so enabling websearch silently ignored `rrf_k` — the same preset fused differently based on the toggle. Routed the doc branch through `retrieve_multi` so `rrf_k` applies consistently.

## Scope (documented, not silently narrowed)
- **Cross-partition / doc+web fusion** (`RetrievalService.fuse`) stays at canonical 60: it fuses lists from *different* partitions, so no single partition's `rrf_k` applies. Docstring corrected to match.
- **Legacy pipeline** (`"all"` / no-preset deployments) has no per-partition config, so `rrf_k` stays 60. There is no global `rrf_k` knob outside partition presets — noted as a limitation.
- **Milvus-native dense/BM25 hybrid fusion** (`RRFRanker(RRF_K=100)` in `milvus_store`) is a *separate* per-query fusion layer and is **not** governed by this per-partition field. An operator tuning `rrf_k` will not change the semantic-vs-keyword blend. Tracked as a follow-up (see below).

## Tests
- configured `rrf_k` reaches `rrf_reranking` (spy asserts `k=17`, fails against the hardcoded 60)
- a partition's `rrf_k` reaches its pipeline
- 34 query_service tests pass unchanged after the websearch reroute

Full suite: **1887 passed**; ruff/format/layer-guard green.

## Follow-up
Per-partition `rrf_k` does not reach the Milvus dense/BM25 `RRFRanker` (fixed at 100) — a separate cross-layer plumbing change. Will file a tracked issue.

## Summary by CodeRabbit

* **New Features**
  * Added per-partition configuration for result-fusion dampening, with the existing default preserved.
* **Bug Fixes**
  * Prevented duplicate result fusion in the combined retrieval + web search flow by ensuring fusion happens once using the partition’s configured settings.
* **Tests**
  * Added unit tests validating the default fusion setting and verifying per-partition configuration propagation through the retrieval pipeline and web search path.

### DIFF
--- a/openrag/core/retrieval/pipeline.py
+++ b/openrag/core/retrieval/pipeline.py
@@ -67,11 +67,17 @@ class RetrieverPipeline:
         reranker: Reranker | None = None,
         reranker_top_k: int = 5,
         allow_filterless_fallback: bool = True,
+        rrf_k: int = 60,
     ) -> None:
         self.retriever = retriever
         self.reranker = reranker
         self.reranker_top_k = reranker_top_k
         self.allow_filterless_fallback = allow_filterless_fallback
+        # RRF dampening for fusing this partition's multiQuery sub-query rankings
+        # (see get_relevant_docs). 60 is canonical; a preset can tune it via
+        # RetrievalPipelineConfig.rrf_k. Cross-partition fusion happens a layer
+        # up in RetrievalService.fuse, which is not partition-scoped.
+        self.rrf_k = rrf_k
 
     @property
     def reranker_enabled(self) -> bool:
@@ -146,7 +152,7 @@ class RetrieverPipeline:
             for q in search_queries.query_list
         ]
         ranked_lists = await asyncio.gather(*tasks)
-        fused = rrf_reranking(ranked_lists, key_fn=_chunk_key)
+        fused = rrf_reranking(ranked_lists, key_fn=_chunk_key, k=self.rrf_k)
         if top_k is not None:
             fused = fused[:top_k]
         return fused
--- a/openrag/services/orchestrators/query_service.py
+++ b/openrag/services/orchestrators/query_service.py
@@ -321,8 +321,7 @@ class QueryService:
 
         web_results: list = []
         if partition is not None and use_websearch:
-            doc_lists, web_lists = await self._gather_rag_and_web(queries.query_list, partition, top_k, filter_params)
-            chunks = self._retrieval.fuse(doc_lists, top_k=top_k)
+            chunks, web_lists = await self._gather_rag_and_web(queries, partition, top_k, filter_params)
             web_results = _dedupe_web(web_lists)
         elif partition is not None:
             chunks = await self._retrieval.retrieve_multi(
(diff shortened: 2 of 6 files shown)


hedhoud (dismissed):
LGTM. I checked the retrieval and websearch paths, and the per-partition `rrf_k` now flows through the right fusion layer. CI is green; the only remaining note is a tiny test-style cleanup, not something that should block merge.

andyne13:
Added a unit test (`test_websearch_with_partition_fuses_docs_via_retrieve_multi`) that pins the websearch behaviour this PR fixes. Previously nothing guarded it: `FakeRetrieval` exposes both `retrieve_multi` and `retrieve_per_query`+`fuse`, so a revert of `_gather_rag_and_web` back to the legacy hardcoded-60 fuse passed silently. The new test spies all three and asserts the document branch goes through `retrieve_multi` (with `retrieve_per_query`/`fuse` unused) when a partition and websearch are both enabled — I revert-proved it (fails `assert 0 == 1` on the legacy path, passes on the fix). Only the test file changed; re-review appreciated since the new commit dismissed the approval.

andyne13:
Also addressed the CodeRabbit nitpick: dropped the redundant `@pytest.mark.asyncio`/`async def` on `test_pipeline_for_partition_threads_rrf_k` (it calls the synchronous `_pipeline_for_partition` with no `await`). 12 tests in that file still pass.

hedhoud (changes requested):
The sub-query fusion change is implemented cleanly and the focused tests pass, but one operator-facing gap remains.

The Admin UI says that `rrf_k` controls hybrid dense/BM25 fusion. This PR applies it only when combining generated-query result lists, while the Milvus hybrid ranker remains fixed at 100. With a single generated query, changing this setting still has no effect.

Could we either wire the setting to the hybrid fusion it currently describes, or update the UI wording and keep the remaining hybrid work tracked instead of closing #707? Otherwise users will believe the original setting is fully fixed when its documented behavior is still unchanged.

hedhoud (approved):
Approved. The per-partition sub-query fusion fix is correct and covered by focused tests. The remaining dense/BM25 wording and Milvus fusion alignment can be handled as follow-up work.

PR #743, opened 2026-07-21
fix(chunking): match canonical GFM table delimiters (| --- | --- |)

### LINKED ISSUE #710: Markdown table chunking misses canonical GFM delimiters (| --- | --- |); only tight |---|---| matches
## Problem

`TABLE_RE` fails to match the **canonical GFM delimiter form** (`| --- | --- |`). Only the tight form (`|---|---|`) matches.

`openrag/core/chunking/markdown_utils.py:21-24`:
```python
TABLE_RE = re.compile(
    r"((?:^|\n)\|.*?\|\r?\n\|\s*[:-]+(?:\s*\|[:-]+)*\|\r?\n(?:\|.*?\|\r?\n)+)",
    re.DOTALL | re.MULTILINE,
)
```
The repeat group `(?:\s*\|[:-]+)*` allows whitespace *before* the pipe but not *after* it, so any delimiter cell padded on its left fails.

Reproduced against the real pattern:
```
False  | A | B |\n| --- | --- |\n| 1 | 2 |      <- canonical GFM / prettier / pandoc output
True   | A | B |\n|---|---|\n| 1 | 2 |
False  | A | B |\n| :--- | ---: |\n| 1 | 2 |    <- aligned
True   | A | B |\n|:---|---:|\n| 1 | 2 |
```

## Impact

Spaced-delimiter tables bypass `chunk_table()` entirely: no header replay across chunks, no last-row overlap, and `chunk_type` is mislabelled `text` so any table-aware handling or filtering misses them. A table split across 6 chunks leaves 5 chunks of bare numbers with no column names — retrievable but semantically useless.

Affects hand-written `.md` uploads for certain, and any parser backend that pads delimiters.

## Why tests missed it

Every fixture in `tests/unit/core/chunking/test_markdown_utils.py` and `test_recursive.py` uses the tight form. The one form users actually write is untested.

## Fix

```python
r"...\|\r?\n\|(?:\s*:?-+:?\s*\|)+\r?\n(?:\|.*?\|\r?\n)+..."
```
and add spaced/aligned delimiter variants to the `split_md_elements` test matrix.

andyne13:
Closes #710.

## Problem
`TABLE_RE`'s delimiter row was `\|\s*[:-]+(?:\s*\|[:-]+)*\|` — whitespace was allowed only *before* each pipe. A space *after* a pipe (canonical GFM `| --- | --- |`, plus aligned `| :--- | ---: |` and centered `| :---: |`) failed to match, so those tables fell through to the plain-text splitter: no header replay across chunks, no table-aware handling, `chunk_type` mislabelled as text. Only the tight `|---|---|` form worked.

## Fix
Delimiter row → `\|(?:\s*:?-+:?\s*\|)+`: one or more `:?-+:?` cells with optional surrounding whitespace. Matches spaced / tight / aligned / centered / single-dash delimiters and CRLF, and still rejects non-tables (`-+` is required, so `| a | b |` data rows are not misread as a delimiter).

## Verified (no false positives, no regressions)
Ran old vs new across 9 cases: fixes spaced/aligned/centered/single-dash/CRLF, preserves tight forms, rejects delimiter-less pipe blocks and text-cell false delimiters.

## Tests
Parametrized over all 6 delimiter variants + a false-positive guard + CRLF. Reverting the regex fails the 5 spaced-variant cases.

Full suite: **1892 passed**; ruff/format/layer-guard green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of GitHub Flavored Markdown tables across more pipe-delimiter spacing variants (including aligned/compact forms).
  * Fixed an edge case where whitespace right after a pipe could prevent table recognition.
  * Ensured tables with Windows-style line endings and delimiter rows ending at EOF (no trailing newline) are handled correctly.
* **Tests**
  * Added unit tests for valid and invalid table delimiter patterns, including malformed multi-line cases and trailing whitespace within rows.

### DIFF
--- a/openrag/core/chunking/markdown_utils.py
+++ b/openrag/core/chunking/markdown_utils.py
@@ -17,9 +17,14 @@ from typing import Literal
 
 from core.utils.text import clean_markdown_table_spacing
 
-# Header + delimiter + at least one row.
+# Header + delimiter + at least one row. The delimiter row is one or more cells
+# of ``:?-+:?`` with optional surrounding whitespace — this matches canonical GFM
+# spacing (``| --- | --- |``), alignment colons (``| :--- | ---: |``, ``| :---: |``)
+# and the tight form (``|---|---|``) alike. A previous ``\s*[:-]+(?:\s*\|[:-]+)*``
+# allowed whitespace only *before* each pipe, so a space *after* a pipe (the
+# canonical form) failed to match and the table fell through to plain text (#710).
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|\s*[:-]+(?:\s*\|[:-]+)*\|\r?\n(?:\|.*?\|\r?\n)+)",
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:\s*:?-+:?\s*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
     re.DOTALL | re.MULTILINE,
 )
 
--- a/openrag/core/chunking/markdown_utils.py
+++ b/openrag/core/chunking/markdown_utils.py
@@ -24,10 +24,12 @@ from core.utils.text import clean_markdown_table_spacing
 # allowed whitespace only *before* each pipe, so a space *after* a pipe (the
 # canonical form) failed to match and the table fell through to plain text (#710).
 # In-cell whitespace is ``[ \t]`` (not ``\s``) so a delimiter can't span line
-# breaks and swallow following rows as one multi-line "delimiter".
+# breaks and swallow following rows as one multi-line "delimiter". No re.DOTALL,
+# so a row's ``.*?`` stays on its line (a pipe-looking text line can't be pulled
+# into a later table); the final row may end at a newline OR EOF.
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|(?:[ \t]*:?-+:?[ \t]*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
-    re.DOTALL | re.MULTILINE,
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:[ \t]*:?-+:?[ \t]*\|)+\r?\n(?:\|.*?\|(?:\r?\n|$))+)",
+    re.MULTILINE,
 )
 
 # `<image_description>...</image_description>` block injected by the VLM step.
--- a/openrag/core/chunking/markdown_utils.py
+++ b/openrag/core/chunking/markdown_utils.py
@@ -18,13 +18,15 @@ from typing import Literal
 from core.utils.text import clean_markdown_table_spacing
 
 # Header + delimiter + at least one row. The delimiter row is one or more cells
-# of ``:?-+:?`` with optional surrounding whitespace — this matches canonical GFM
+# of ``:?-+:?`` with optional surrounding spaces/tabs — this matches canonical GFM
 # spacing (``| --- | --- |``), alignment colons (``| :--- | ---: |``, ``| :---: |``)
 # and the tight form (``|---|---|``) alike. A previous ``\s*[:-]+(?:\s*\|[:-]+)*``
 # allowed whitespace only *before* each pipe, so a space *after* a pipe (the
 # canonical form) failed to match and the table fell through to plain text (#710).
+# In-cell whitespace is ``[ \t]`` (not ``\s``) so a delimiter can't span line
+# breaks and swallow following rows as one multi-line "delimiter".
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|(?:\s*:?-+:?\s*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:[ \t]*:?-+:?[ \t]*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
     re.DOTALL | re.MULTILINE,
 )
 
--- a/openrag/core/chunking/markdown_utils.py
+++ b/openrag/core/chunking/markdown_utils.py
@@ -26,9 +26,12 @@ from core.utils.text import clean_markdown_table_spacing
 # In-cell whitespace is ``[ \t]`` (not ``\s``) so a delimiter can't span line
 # breaks and swallow following rows as one multi-line "delimiter". No re.DOTALL,
 # so a row's ``.*?`` stays on its line (a pipe-looking text line can't be pulled
-# into a later table); the final row may end at a newline OR EOF.
+# into a later table); the final row may end at a newline OR EOF. Each row's
+# closing pipe may be followed by ``[ \t]*`` before the line ends — GFM permits
+# trailing line whitespace, and without this a stray space after ``|`` would drop
+# the whole table back to plain text (the same failure class as #710).
 TABLE_RE = re.compile(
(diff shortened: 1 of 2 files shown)


openrag/core/chunking/markdown_utils.py
```diff
@@ -17,9 +17,14 @@
...
-# Header + delimiter + at least one row.
+# Header + delimiter + at least one row. The delimiter row is one or more cells
+# of ``:?-+:?`` with optional surrounding whitespace — this matches canonical GFM
+# spacing (``| --- | --- |``), alignment colons (``| :--- | ---: |``, ``| :---: |``)
+# and the tight form (``|---|---|``) alike. A previous ``\s*[:-]+(?:\s*\|[:-]+)*``
+# allowed whitespace only *before* each pipe, so a space *after* a pipe (the
+# canonical form) failed to match and the table fell through to plain text (#710).
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|\s*[:-]+(?:\s*\|[:-]+)*\|\r?\n(?:\|.*?\|\r?\n)+)",
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:\s*:?-+:?\s*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
```
hedhoud: `\s*` also matches line breaks, so a delimiter split across multiple lines can be detected as a table. Please limit it to spaces and tabs and add a small regression test. After that, LGTM.

hedhoud (commented):
One small parser edge case remains.

andyne13:
Good catch — fixed in 7d93280d. In-cell whitespace is now `[ \t]*` instead of `\s*`, so a delimiter can't span line breaks. Valid spaced/tight/aligned/centered/tab-separated/CRLF delimiters still match; a delimiter split across physical lines no longer does. Added `test_delimiter_cannot_span_line_breaks` covering it.

openrag/core/chunking/markdown_utils.py
```diff
@@ -17,9 +17,16 @@
...
+# of ``:?-+:?`` with optional surrounding spaces/tabs — this matches canonical GFM
+# spacing (``| --- | --- |``), alignment colons (``| :--- | ---: |``, ``| :---: |``)
+# and the tight form (``|---|---|``) alike. A previous ``\s*[:-]+(?:\s*\|[:-]+)*``
+# allowed whitespace only *before* each pipe, so a space *after* a pipe (the
+# canonical form) failed to match and the table fell through to plain text (#710).
+# In-cell whitespace is ``[ \t]`` (not ``\s``) so a delimiter can't span line
+# breaks and swallow following rows as one multi-line "delimiter".
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|\s*[:-]+(?:\s*\|[:-]+)*\|\r?\n(?:\|.*?\|\r?\n)+)",
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:[ \t]*:?-+:?[ \t]*\|)+\r?\n(?:\|.*?\|\r?\n)+)",
```
hedhoud: One edge case is still open here. Because the regex still uses `re.DOTALL`, the `.*?` parts can cross line breaks. That means a pipe-looking text line before a blank line can be swallowed into a later table, and a valid table ending exactly at EOF is still missed because the last row must end with `\n`. A small fix would be to remove `re.DOTALL` and let the final row end with either `\r?\n` or EOF. I’d also add two tests: one table with no trailing newline, and one pipe block separated by a blank line before a real table.

hedhoud (commented):
I left one focused comment on the remaining table-regex edge case.

aditykris (dismissed):
Looks good,No further issues found.

andyne13:
Fixed in 57126b76. Removed `re.DOTALL` (so a row's `.*?` stays on its line — a pipe-looking text line is no longer pulled into a later table) and made the final row end with `\r?\n` or EOF via `(?:\|.*?\|(?:\r?\n|$))+`. Verified all valid variants still match and the split-delimiter/non-table cases stay rejected. Added the two tests you asked for: `test_table_at_eof_without_trailing_newline` and `test_pipe_text_line_before_blank_not_swallowed_into_later_table`. 1895 tests pass.

openrag/core/chunking/markdown_utils.py
```diff
@@ -17,10 +17,19 @@
...
+# allowed whitespace only *before* each pipe, so a space *after* a pipe (the
+# canonical form) failed to match and the table fell through to plain text (#710).
+# In-cell whitespace is ``[ \t]`` (not ``\s``) so a delimiter can't span line
+# breaks and swallow following rows as one multi-line "delimiter". No re.DOTALL,
+# so a row's ``.*?`` stays on its line (a pipe-looking text line can't be pulled
+# into a later table); the final row may end at a newline OR EOF.
 TABLE_RE = re.compile(
-    r"((?:^|\n)\|.*?\|\r?\n\|\s*[:-]+(?:\s*\|[:-]+)*\|\r?\n(?:\|.*?\|\r?\n)+)",
-    re.DOTALL | re.MULTILINE,
+    r"((?:^|\n)\|.*?\|\r?\n\|(?:[ \t]*:?-+:?[ \t]*\|)+\r?\n(?:\|.*?\|(?:\r?\n|$))+)",
```
Ahmath-Gadji: **Trailing whitespace after the final pipe still drops the whole table to plain text**

`openrag/core/chunking/markdown_utils.py:31`

The delimiter fix closes the *in-cell* spacing case (`| --- | --- |`), but a row that ends with whitespace after its closing pipe still fails the match, so the entire block falls through to the plain-text splitter — the same failure mode as #710:

```
"| A | B |  \n| --- | --- |\n| 1 | 2 |\n"   -> text   (trailing spaces after header pipe)
"| A | B |\n| --- | --- |   \n| 1 | 2 |\n"  -> text   (after delimiter pipe)
"| A | B |\n| --- | --- |\n| 1 | 2 |  \n"   -> text   (after data-row pipe)
```

GFM explicitly allows trailing line whitespace, and the markdown reaching `split_md_elements` is not whitespace-normalized first (`clean_markdown_table_spacing` only runs later, inside `chunk_table`, *after* detection). It is not a regression (the old regex missed it too), but since it is the same whitespace class this PR targets, folding it in is worthwhile.

Fix: allow optional `[ \t]*` before each line terminator. It keeps rejecting the non-table cases (missing delimiter, colon-only cells, line-break-split delimiter).

Ahmath-Gadji: Fixed in commit e95e0524 (https://github.com/linagora/openrag/commit/e95e052416625db6e3bb4e950ad997eacfe4783a).

`TABLE_RE` now allows optional `[ \t]*` before each line terminator on the header, delimiter, and data rows — `\|.*?\|[ \t]*\r?\n`, `...\|)+[ \t]*\r?\n`, and `\|.*?\|[ \t]*(?:\r?\n|$)`. All three trailing-whitespace positions (plus trailing whitespace at EOF and trailing tabs) are now detected, and the non-table cases stay rejected (missing delimiter, colon-only cells, line-break-split delimiter). Added a parametrized regression test `test_trailing_line_whitespace_still_detected_as_table`; 33 tests pass.

Ahmath-Gadji (approved):
LGTM. Verified independently:

- All prior review threads resolved — the `[ \t]`-restricted delimiter (hedhoud), dropped `re.DOTALL`, and EOF-terminated final row (CodeRabbit + hedhoud) all confirmed by diffing old vs. new `TABLE_RE` across the edge cases: canonical/tight/aligned/centered/single-dash/CRLF now match, blank-line-separated tables split correctly, and non-tables (missing delimiter, colon-only cells, line-break-split delimiter) stay rejected. No regressions.
- One same-class gap folded in: trailing whitespace after a row pipe was still dropping the whole table to plain text (e95e0524), with a parametrized regression test.

Full unit suite + ruff + layer-guard + api/milvus integration all green.

PR #744, opened 2026-07-21
fix(retrieval): enforce workspace file scoping through to Milvus search (#706)

### LINKED ISSUE #706: Workspace scoping is inert: filter_params is threaded through retrieval and dropped at the vector store
## Problem

`filter_params` is threaded through the whole retrieval stack and then **silently dropped at the storage boundary**, so workspace-scoped search and chat return the entire partition.

`openrag/services/storage/vector_store_searcher.py:61-80`:
```python
async def search(self, query, partition, top_k, filter=None,
                 filter_params: dict | None = None, ...):
    filters: dict[str, Any] = {"partition": partition}
    if filter:
        filters["expr"] = filter
    results = await self._store.search(..., filters=filters, ...)   # filter_params never read
```
`MilvusVectorStore.search` has no `filter_params` parameter at all. `grep -rn ARRAY_CONTAINS openrag/` returns **nothing** — the predicate this needs exists only in `tests/load/workspace/`, i.e. it was benchmarked and never implemented.

Workspace scoping rides exclusively on this value: `api/routers/user/search.py:170`, `:255`, `services/orchestrators/query_service.py:320`.

## Why it looks like it works

The router fully validates the workspace and 404s if it isn't in an accessible partition (`search.py:167-170`), then binds `workspace=...` into the log line (`:160`). Caller and operator both get every signal that scoping was applied.

The codebase already knows — `services/orchestrators/mcp_service.py:225-229` documents it and works around it by inlining `json.dumps(file_id)`. The MCP path got the workaround; `/search` and chat did not.

## Impact

Not cross-tenant (the workspace is verified to belong to an accessible partition), but the documented sharing boundary is inoperative. Anyone using workspaces to scope what a user or agent sees **within** a shared partition has a control that does nothing.

Related: `api/routers/user/search.py:328-337` builds `"file_id == {_file_id}"` with `params={"_file_id": file_id}`. Since params are dropped, Milvus receives the literal `{_file_id}` unbound — that route either errors or loses its file scope.

## Fix

Resolve `workspace_id` → `file_id` set in Postgres and merge into `filters`, or add a Milvus `workspace_ids` ARRAY field and bind `filter_params` through. Until then, fail loudly on an unconsumed non-empty `filter_params` rather than ignoring it.

## Why tests missed it

`tests/unit/services/orchestrators/test_retrieval_service.py:116-125` asserts `call["filter_params"] == {"a": 1}` — that the value was forwarded one layer, never that a row was filtered.

aditykris:
## Summary

Fixes #706 — workspace-scoped search and chat validated the workspace but never actually restricted results to it. `filter_params` was threaded through every layer of retrieval but silently dropped at `VectorStoreSearcher`, the one place that builds the Milvus query — so a "workspace search" quietly ran against the whole partition. Root cause and fix details are in the issue.

- Add a single workspace scope resolver (`WorkspaceService.resolve_scope`) that resolves `workspace_id → (partition, file_ids)` via Postgres and fails closed on missing/inaccessible workspaces
- Fix `VectorStoreSearcher.search`/`multi_query_search` to actually merge `filter_params` into the Milvus query (dense + both legs of hybrid)
- Narrow the effective search partition to the workspace's own partition, so `openrag-all` + workspace can't match a same-named file in another partition
- Scope surrounding/related/ancestor chunk expansion to the same file-id allowlist so it can't reintroduce out-of-workspace documents
- Replace "log a warning and search unscoped" with a proper `WorkspaceNotFoundError` (404 for REST search, structured error for both streaming and non-streaming chat)
- Remove the MCP and by-file-search workarounds that hand-built raw Milvus expressions because `filter_params` didn't work; both now bind through it properly
- Fix a real Milvus bug found via integration testing: the empty-file-list short-circuit emitted a bare `false` literal, which Milvus 2.6 rejects outright — changed to `1 == 0`

## Test plan

- [x] `uv run pytest tests/unit/` — 1872 passed
- [x] `uv run pytest tests/integration/repos/` against a real Postgres + Milvus stack — all workspace/Milvus-related suites pass (2 pre-existing, unrelated failures in `test_partition_repo.py`)
- [x] New end-to-end test (`test_workspace_scoping_e2e.py`): two partitions, a file sharing the same `file_id` across both, a workspace scoped to one — confirms cross-partition search returns only the in-scope file, and that removing the file returns zero results rather than falling back to the full partition
- [x] `uv run ruff check` / `uv run ruff format --check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace-scoped searches now fail closed using an authorization-checked resolved scope, returning consistent 404 behavior when the workspace is unresolved or inaccessible.
  * Related/ancestor/surrounding expansion is now constrained by the same authorized file allowlist, including correct zero results for empty allowlists.
  * File-scoped routes bind `file_id` using parameterized scoping (no hard-coded filter strings), while preserving any caller filter.
  * Milvus empty-scope filtering now uses an always-false predicate to prevent accidental broad matches.
* **Tests**
  * Added end-to-end and unit coverage for workspace scoping, fail-closed empty scopes, and restricted expansion behavior.

### DIFF
--- a/openrag/api/routers/user/search.py
+++ b/openrag/api/routers/user/search.py
@@ -164,10 +164,14 @@ async def search_multiple_partitions(
 
     filter_params = None
     if workspace:
-        ws = await workspaces.get_workspace(workspace)
-        if not ws or ws["partition_name"] not in partitions:
+        scope = await workspaces.resolve_scope(workspace, partitions)
+        if scope is None:
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Workspace not found")
-        filter_params = {"workspace_id": workspace}
+        # A workspace belongs to exactly one partition — narrow the search to
+        # it, otherwise file_id-only filtering could match a same-named file
+        # in another partition the caller also has access to (#706).
+        partitions = [scope.partition]
+        filter_params = {"file_id": scope.file_ids}
 
     results = await service.search(
         text=search_params.text,
@@ -249,10 +253,10 @@ async def search_one_partition(
     )
     filter_params = None
     if workspace:
-        ws = await workspaces.get_workspace(workspace)
-        if not ws or ws["partition_name"] != partition:
+        scope = await workspaces.resolve_scope(workspace, [partition])
+        if scope is None:
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Workspace not found")
-        filter_params = {"workspace_id": workspace}
+        filter_params = {"file_id": scope.file_ids}
 
     results = await service.search(
         text=search_params.text,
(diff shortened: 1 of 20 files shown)


aditykris:
@coderabbit review

hedhoud:
@codex review

hedhoud (approved):
Reviewed the current workspace scoping changes. The search and expansion paths are covered and look good.

Issue #542, opened 2026-06-23
Admin UI should show stage-level indexing progress


hedhoud:
Context
Indexing can be slow for large PDFs or production-like runs with image captioning and contextual retrieval. Admins need to understand where time is being spent.

Problem
The Jobs page currently shows broad task states, but it does not explain whether a job is parsing, chunking, captioning images, contextualizing, embedding, or inserting into the vector store.

Suggested fix
Expose stage-level progress in the job list and job detail page. The first version can be simple: current stage, last stage update time, and a small timeline/history for completed stages.

Acceptance
- A running job shows its current pipeline stage.
- Job detail shows enough stage history to understand where time was spent.
- Failed jobs show the stage where they failed.

GioiaZheng:
This would be really helpful for production indexing. I’d suggest modeling progress as stage events rather than one mutable status string from the start.

Minimum useful fields:

- job_id, stage_name, status, started_at, finished_at, duration_ms
- processed/total when known
- last_error and failed_stage
- optional artifact counters, e.g. pages parsed, chunks created, images captioned, vectors inserted

That gives the UI a timeline now, and later lets admins answer “captioning is slow” vs “vector insertion is slow” without adding a second telemetry system. For long PDFs, an “unknown total” state is also worth supporting because some stages only discover work after parsing.

PR #745, opened 2026-07-21
fix(tasks): remove never-written CHUNKING/INSERTING task states

### LINKED ISSUE #721: Remove the never-written CHUNKING / INSERTING task states (documented and filtered on, but no code writes them)
## Problem

`CHUNKING` and `INSERTING` are documented to API users and enumerated by four consumers, but **no code ever writes them**.

Exhaustive list of `set_state` writers on `develop`:
```
services/workers/dispatcher.py:93          "QUEUED"
services/workers/dispatcher.py:256         "FAILED"
services/workers/task_cancellation.py:206  "CANCELLED"
services/workers/task_cancellation.py:287  "FAILED"
services/workers/indexer_actor.py:76       "SERIALIZING"
services/workers/indexer_actor.py:125      "COMPLETED"
```

Yet they are treated as live states in:
- `core/models/catalog.py:16-17` — `DocumentStatus.CHUNKING` / `.INSERTING`
- `services/orchestrators/job_service.py:20` — `_ACTIVE_STATES`
- `services/orchestrators/mcp_service.py:55` — `_ACTIVE_STATES`
- `services/workers/task_state.py:173` — `pending_states`
- `api/mcp/server.py:325` — MCP tool description
- **`api/routers/admin/jobs.py:79-80`** — public API docs: "`CHUNKING`: Splitting into chunks", "`INSERTING`: Adding to vector database"

Parse, chunk, embed and store all run under one opaque `SERIALIZING`.

## Relationship to #542 / PR #686 — complementary, not overlapping

PR #686 (design note for #542) reaches the same diagnosis:

> "A job that is still marked `SERIALIZING` may already be parsing, captioning, embedding, or waiting on a shared worker pool. Showing fake progress would make production debugging worse."

But it resolves it a **different way**: it adds a new per-stage `stages` event array, and its requirement 4 is explicitly *"Keep broad task states unchanged for compatibility."*

That is the right call — and it means **this issue should not be fixed by emitting the missing states**, which would contradict that design. It should be fixed by removing them, so the two efforts don't collide.

## Proposed fix (revised to align with #686)

Delete the dead states rather than emit them:

- remove `CHUNKING` / `INSERTING` from `DocumentStatus` (`core/models/catalog.py:16-17`)
- drop them from both `_ACTIVE_STATES` sets and from `pending_states`
- correct the MCP tool description (`api/mcp/server.py:325`)
- correct the public API docs (`api/routers/admin/jobs.py:79-80`, and `:34`, `:57`, `:93`)
- update the pipeline description in `CLAUDE.md`, which documents the same non-existent sequence

Real stage visibility then arrives through #542/#686's `stages` field, which is the mechanism actually designed for it.

## Impact if left alone

API consumers filtering `task_status=active` on `CHUNKING`/`INSERTING` match nothing, forever, with no error — and the documented state machine stays a fiction alongside a second, real one.

Ahmath-Gadji:
## Summary

`CHUNKING` and `INSERTING` were documented to API users and enumerated by several consumers, but **no code ever wrote them** — parse, chunk, embed and store all run under a single opaque `SERIALIZING` state. Consumers filtering `task_status=active` on `CHUNKING`/`INSERTING` matched nothing, forever, with no error.

Per #721, and to avoid colliding with the per-stage `stages` field designed in #542 / #686 (whose requirement 4 is explicitly *"keep broad task states unchanged for compatibility"*), this **deletes** the dead states rather than emitting them. Real per-stage visibility will arrive through that separate `stages` mechanism.

## Changes

**Backend**
- `core/models/catalog.py` — remove `CHUNKING` / `INSERTING` from `DocumentStatus`
- `services/orchestrators/job_service.py` — `_ACTIVE_STATES` -> `("QUEUED", "SERIALIZING")` (+ docstring)
- `services/orchestrators/mcp_service.py` — trim `_ACTIVE_STATES`
- `services/workers/task_state.py` — trim `ACTIVE_INDEXING_STATES`
- `services/workers/task_cancellation.py` — trim `_ACTIVE_INDEXING_STATES`
- `api/mcp/server.py` — correct the MCP tool state-sequence description
- `api/routers/admin/jobs.py` — correct the public API docs (4 spots); `SERIALIZING` now described as parsing -> chunking -> embedding -> storing
- `CLAUDE.md` — update the `TaskStateManager` pipeline description

**Admin UI** (same dead-state bug on the frontend)
- `ui/src/lib/api/jobs.ts` — `TaskState` union + `ACTIVE_STATES` poll filter
- `ui/src/components/shared/status-badge.tsx` — status-style map
- `ui/src/mocks/handlers.ts` — msw mock fixtures
- `ui/src/pages/admin/jobs/list.test.tsx` — test fixtures

**Tests** (removed states swapped for the real `SERIALIZING` state)
- `tests/unit/services/orchestrators/test_job_service.py`
- `tests/unit/services/workers/test_task_completion.py`
- `tests/integration/api/test_indexer.py`
- `tests/load/automatic-evaluation-pipeline/test_frames.py`

## Not touched
- `extern/indexer-ui/` — an untracked external submodule (separate repo), out of scope for this repo's PR.

## Verification

- grep over `openrag/`, `tests/`, `ui/`, `CLAUDE.md` for the two states -> no matches
- 125 related backend unit tests pass; `ruff check` clean
- UI changes verified against the jobs-list component logic (`running = active - queued`); local UI deps not installed in this environment

Closes #721

### DIFF
--- a/openrag/services/workers/task_cancellation.py
+++ b/openrag/services/workers/task_cancellation.py
@@ -12,7 +12,7 @@ from services.workers.task_state import PENDING_TASK_DETAILS
 
 logger = get_logger()
 
-_ACTIVE_INDEXING_STATES = frozenset({"QUEUED", "SERIALIZING", "CHUNKING", "INSERTING"})
+_ACTIVE_INDEXING_STATES = frozenset({"QUEUED", "SERIALIZING"})
 _REF_WAIT_INTERVAL = 0.05
 _STALE_REFLESS_TASK_ERROR = (
     "Indexing task never exposed a cancellable worker ref before delete cleanup; marking it failed as stale."
--- a/openrag/services/workers/task_cancellation.py
+++ b/openrag/services/workers/task_cancellation.py
@@ -8,11 +8,9 @@ import ray
 from core.utils.logging import get_logger
 from ray.exceptions import TaskCancelledError
 from services.workers.ray_utils import call_ray_actor_with_timeout
-from services.workers.task_state import PENDING_TASK_DETAILS
+from services.workers.task_state import CANCELLABLE_INDEXING_STATES, PENDING_TASK_DETAILS
 
 logger = get_logger()
-
-_ACTIVE_INDEXING_STATES = frozenset({"QUEUED", "SERIALIZING"})
 _REF_WAIT_INTERVAL = 0.05
 _STALE_REFLESS_TASK_ERROR = (
     "Indexing task never exposed a cancellable worker ref before delete cleanup; marking it failed as stale."
@@ -141,7 +139,7 @@ async def _get_matching_active_task_refs_legacy(
     for task_id, info in all_info.items():
         if not isinstance(info, dict):
             continue
-        if info.get("state") not in _ACTIVE_INDEXING_STATES:
+        if info.get("state") not in CANCELLABLE_INDEXING_STATES:
             continue
         details = info.get("details") or {}
         if not isinstance(details, dict):
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,7 +85,7 @@ The main application entry point is `openrag/api/main.py` which creates a FastAP
 
 **Ray Actors** (distributed components):
 - `Indexer` (`openrag/services/workers/indexer_pool.py`) - Handles document ingestion, chunking, and insertion into vector DB
-- `TaskStateManager` (`openrag/services/workers/task_state.py`) - Tracks async task states: QUEUED → SERIALIZING → CHUNKING → INSERTING → COMPLETED (or FAILED or CANCELLED)
+- `TaskStateManager` (`openrag/services/workers/task_state.py`) - Tracks async task states: QUEUED → SERIALIZING → COMPLETED (or FAILED or CANCELLED)
 - `Vectordb` / `MilvusDB` (`openrag/services/storage/milvus_store.py`) - Vector database operations with hybrid search (dense + BM25 sparse)
 - `DocSerializer` (`openrag/services/workers/parsers/doc_serializer.py`) - Serializes files to Document objects using appropriate loaders
 - `MarkerPool` / `MarkerWorker` (`openrag/services/workers/parsers/marker_workers.py`) - Pool of workers for PDF processing with Marker
--- a/openrag/api/mcp/server.py
+++ b/openrag/api/mcp/server.py
@@ -322,7 +322,7 @@ async def get_chunk_by_id(chunk_id: str) -> dict:
 @server.tool(
     description=(
         "Get the current status and details of an indexation task. "
-        "Task states: QUEUED → SERIALIZING → CHUNKING → INSERTING → COMPLETED (or FAILED). "
+        "Task states: QUEUED → SERIALIZING → COMPLETED (or FAILED). "
         "If the task failed, the error message is included in the response."
     )
 )
--- a/openrag/api/routers/admin/jobs.py
+++ b/openrag/api/routers/admin/jobs.py
@@ -31,7 +31,7 @@ Returns system status including:
 
 **Tasks:**
 - `active`: Total active tasks
-- `active_statuses`: Breakdown by status (QUEUED, SERIALIZING, CHUNKING, INSERTING)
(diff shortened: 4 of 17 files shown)


openrag/services/workers/task_cancellation.py
```diff
@@ -12,7 +12,7 @@
 
 logger = get_logger()
 
-_ACTIVE_INDEXING_STATES = frozenset({"QUEUED", "SERIALIZING", "CHUNKING", "INSERTING"})
+_ACTIVE_INDEXING_STATES = frozenset({"QUEUED", "SERIALIZING"})
```
hedhoud: Could we keep `CHUNKING` and `INSERTING` as legacy active aliases at this cancellation boundary, or normalize them to `SERIALIZING` before filtering? Older detached `Indexer` actors did emit these states, and they can survive an API restart. During that transition, deleting a file or partition would miss the in-flight task, allowing the old worker to write data again after cleanup. A small compatibility path and regression test would make this removal safe without continuing to expose the old states publicly.

hedhoud (commented):
The public cleanup makes sense, but there is one rolling-upgrade compatibility case to address before merge.

openrag/services/workers/task_cancellation.py
Ahmath-Gadji: Good catch — agreed this is a real risk in cluster mode (external RayCluster), where an old detached `Indexer` can outlive the API and keep reporting `CHUNKING`/`INSERTING`. A proper deploy is expected to cycle every actor (and embedded Ray always recreates them on restart), so this is a transition-window edge case rather than the steady state — but missing an in-flight task on delete and letting a stale worker write after cleanup is bad enough to guard against.

Fixed in f8781e02 with a **fencing-only** alias set rather than re-exposing the states:

```python
LEGACY_ACTIVE_INDEXING_STATES = frozenset({"CHUNKING", "INSERTING"})
CANCELLABLE_INDEXING_STATES = ACTIVE_INDEXING_STATES | LEGACY_ACTIVE_INDEXING_STATES
```

`CANCELLABLE_INDEXING_STATES` is used at both fencing boundaries — the actor-side matcher (`_matching_active_task_refs_locked` in `task_state.py`) and the driver-side legacy fallback here in `task_cancellation.py` — so delete/partition cleanup keeps catching such tasks. The legacy states stay out of the public state machine: not in `DocumentStatus`, the `active` counts, the API docs, or the MCP tool description. A regression test (`test_delete_cleanup_still_fences_legacy_indexing_states`) asserts a `CHUNKING`/`INSERTING` task with matching partition/file is still returned for cancellation.

Ahmath-Gadji:
Thanks for the review — the rolling-upgrade compatibility case is now addressed in f8781e02.

The gap was that trimming `CHUNKING`/`INSERTING` from the active-state sets also blinded the delete/cancel **fencing** path: during a rolling deploy against an external Ray cluster, an old detached `Indexer` can outlive the API and keep reporting those states, and a file/partition delete in that window would miss the in-flight task and let the stale worker write back after cleanup.

Fix — a fencing-only alias, not a re-exposure:

```python
LEGACY_ACTIVE_INDEXING_STATES = frozenset({"CHUNKING", "INSERTING"})
CANCELLABLE_INDEXING_STATES = ACTIVE_INDEXING_STATES | LEGACY_ACTIVE_INDEXING_STATES
```

`CANCELLABLE_INDEXING_STATES` is applied at both fencing boundaries — the actor-side matcher (`_matching_active_task_refs_locked`) and the driver-side legacy fallback in `task_cancellation.py` — so cleanup keeps catching such tasks. The states stay out of the public surface: no `DocumentStatus` member, no `active` counts, no API docs, no MCP description. Regression test `test_delete_cleanup_still_fences_legacy_indexing_states` covers it.

Inline thread resolved. Ready for another look.

hedhoud (dismissed):
The rolling-upgrade case is covered now: legacy states remain internal to deletion fencing without returning to the public task model. I also verified the backend and Admin UI changes on current develop. Looks good to merge.

hedhoud (dismissed):
Rechecked against the updated develop branch. The compatibility guard and regression coverage are sound, and the backend plus Admin UI validation remains green. Looks good to merge.

hedhoud (approved):

Issue #712, opened 2026-07-20
batch_size is injected into every LLM/VLM chat request body (400s behind a strict OpenAI-compatible proxy)


andyne13:
## Problem

`batch_size` is an embedder-only setting, but `di/factories.py` passes it for **all four** model kinds, and `VLLMClient` forwards unknown kwargs straight into the request body.

`openrag/di/factories.py:91-97` (one code path for embedder, reranker, llm, vlm):
```python
kwargs = {"endpoint": ..., "model_name": ...,
          "batch_size": model_cfg.batch_size,
          "timeout": model_cfg.timeout, **impl_kwargs}
instance = registry.create(impl, **kwargs)
```
`VLLMClient.__init__` has no `batch_size` parameter, so it lands in `**kwargs` → `self._defaults` (`vllm_client.py:74`) → spread into the JSON payload:
```python
vllm_client.py:113:  payload_kwargs = {**self._defaults, **kwargs}
vllm_client.py:126:  payload = {**self._defaults, **kwargs, "model": model, "prompt": prompt}
vllm_client.py:146:  payload = {**self._chat_payload_kwargs(kwargs), "model": model, "messages": messages, ...}
```

## Impact

Every chat completion and every VLM caption issued through the DB-backed endpoint registry sends a bogus `"batch_size": 32` field. vLLM tolerates unknown body keys; a strict OpenAI-compatible gateway or LiteLLM proxy rejects with 400 — so an endpoint that works in dev fails behind a proxy, for a reason nothing in the config surface explains.

## Fix

Pass `batch_size` only for the embedder kind (move it into `extra_kwargs_fn`), or filter it out in `VLLMClient.__init__` before assigning `self._defaults`.

Verified on `develop` @ `e61479b4`.

PR #746, opened 2026-07-21
fix(models): carry sampling params into DB-backed LLM/VLM endpoints

### LINKED ISSUE #720: temperature and other sampling params never reach LLM/VLM built from the model-endpoint registry
## Problem

`llm.temperature` (and every other sampling parameter) never reaches an LLM or VLM built through the DB-backed model-endpoint registry. Those stages silently run at the provider's default (typically ~0.7–1.0) instead of the configured `0.1`.

`ModelEndpointService._build_default_seeds` (`services/orchestrators/model_endpoint_service.py:118-135`) seeds `extra` with only three keys:
```python
"extra": _with_enable_thinking(
    _with_api_key({"implementation": "vllm"}, s.llm.api_key),
    s.llm.enable_thinking,
),
```
`di/factories.py:89` then splats exactly that: `impl_kwargs = {k: v for k, v in model_cfg.extra.items() if k != "implementation"}`. No `temperature`, no `max_retries`, no `logprobs`.

The VLM factory has the same shape — `indexer_pool._build_vlm_factory:546` splats `model_cfg.extra`, which never carries sampling parameters either.

## Why this is an oversight

The legacy path does it correctly. `di/container.py:501-506` splats the full `settings.llm.model_dump()`, so `temperature` reaches the wire. And `indexer_pool.py:755-764` builds its contextualizer endpoint config explicitly *with* it:
```python
extra = {
    "implementation": "vllm",
    "api_key": getattr(llm_cfg, "api_key", ""),
    "temperature": getattr(llm_cfg, "temperature", 0.1),
    "max_retries": getattr(llm_cfg, "max_retries", 2),
    "logprobs": getattr(llm_cfg, "logprobs", True),
}
```
So two code paths in the same repo disagree about whether sampling params belong in `extra`.

## Impact

The moment a partition selects a *named* LLM or VLM preset, answer generation, chunk contextualization, topic tagging and image captioning all switch to the provider default temperature. Two deployments with identical `config.yaml` behave differently depending on whether a preset points at a named endpoint — and nothing logs the difference. Deterministic-ish RAG answers become noticeably more variable.

## Fix

Include the sampling fields from `LLMParamsConfig` in the seeded `extra` in `_build_default_seeds` (mirroring `indexer_pool.py:755-764`), and do the same for the VLM seed.

Note `indexer_pool.py:763` uses `getattr(llm_cfg, "logprobs", True)` while `LLMParamsConfig.logprobs` defaults to `False` (`core/config/endpoints.py:20`) — whose comment explains that requesting logprobs unsolicited breaks streaming on some backends. That fallback inverts the deliberate default and should be corrected in the same pass.

Ahmath-Gadji:
## Summary

- `ModelEndpointService._build_default_seeds` seeded `extra` with only `{implementation, api_key, enable_thinking}` for `llm`/`vlm`, so a named LLM/VLM endpoint built through the DB-backed registry (`di/factories.py`'s `make_component_factory`, which splats `model_cfg.extra` straight into the client constructor) never received `temperature`, `max_retries`, or `logprobs` — it silently ran at the provider's default sampling params instead of the configured values.
- Added `_with_sampling_params()` to carry `temperature`/`max_retries`/`logprobs` from `LLMParamsConfig` into both the `llm` and `vlm` seed `extra`, mirroring the pattern already used correctly in `indexer_pool._global_llm_endpoint_config`.
- Fixed the same class of bug in `indexer_pool.py`'s global fallback configs:
  - `_global_llm_endpoint_config` defaulted `logprobs` to `True`, inverting `LLMParamsConfig.logprobs`'s deliberate `False` default (requesting logprobs unsolicited breaks streaming on some backends).
  - `_global_vlm_endpoint_config` dropped `temperature`/`max_retries`/`logprobs` entirely — now mirrors its LLM sibling.
- **Follow-up fix (found in review):** carrying `max_retries` through to `VLLMClient`/`VLLMVision`/`OllamaClient` exposed a separate pre-existing bug — those clients absorb unknown constructor kwargs into `self._defaults` and splat them into every outgoing chat/completion/caption request body. `max_retries` is never actually consumed as retry config anywhere (retry attempts are fixed per-method by the `@with_retry` decorator), so it was leaking as a bogus field on the wire — the same class of bug already fixed for `batch_size` in `di/factories.py` (#712). Gave both clients an explicit `max_retries` parameter so it's captured and discarded before reaching `self._defaults`, closing the leak at the client constructor regardless of which factory/call site builds it. Real vLLM (`OpenAIBaseModel`, `extra="allow"`) tolerates unknown fields and just logs them at DEBUG, so this wasn't causing hard failures against the default backend, but it's still not something we want on the wire. Tracked in #747.

Fixes #720

## Test plan

- [x] `uv run pytest tests/unit/services/orchestrators/test_model_endpoint_service.py tests/unit/services/workers/test_indexer_pool.py tests/unit/services/inference/test_vllm_client.py tests/unit/services/inference/test_ollama_client.py -q` — new tests covering seeded sampling params, global-fallback configs, and the max_retries leak
- [x] `uv run pytest tests/unit/ -q` — full unit suite (1891 passed)
- [x] `uv run ruff check` / `uv run ruff format --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved configured LLM/VLM sampling settings (temperature, retry limits, log-probabilities) when seeding and when generating fallback endpoint configs.
  * Ensured log-probabilities default to **disabled** when not explicitly set.
  * Prevented the internal `max_retries` setting from being forwarded into Ollama/vLLM request payloads.
* **Tests**
  * Added unit tests covering sampling-param retention during endpoint seeding/fallback configuration.
  * Added unit tests confirming `max_retries` is not included in outgoing request bodies for both chat and vision requests.

### DIFF
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -25,6 +25,7 @@ if TYPE_CHECKING:
 logger = get_logger()
 
 _VALID_TYPES = frozenset({"embedder", "reranker", "llm", "vlm"})
+_SAMPLING_TYPES = frozenset({"llm", "vlm"})
 
 
 def _slug(model_name: str) -> str:
@@ -46,6 +47,15 @@ def _with_enable_thinking(extra: dict[str, Any], enable_thinking: bool | None) -
     return {**extra, "enable_thinking": enable_thinking}
 
 
+def _sampling_params(llm_cfg: Any) -> dict[str, Any]:
+    """Extract the shared LLM/VLM sampling params (``LLMParamsConfig``) as a dict."""
+    return {
+        "temperature": llm_cfg.temperature,
+        "max_retries": llm_cfg.max_retries,
+        "logprobs": llm_cfg.logprobs,
+    }
+
+
 def _with_sampling_params(extra: dict[str, Any], llm_cfg: Any) -> dict[str, Any]:
     """Add the shared LLM/VLM sampling params (``LLMParamsConfig``) to endpoint extras.
 
@@ -56,12 +66,7 @@ def _with_sampling_params(extra: dict[str, Any], llm_cfg: Any) -> dict[str, Any]
     Mirrors the fallback config built in
     ``indexer_pool._global_llm_endpoint_config``.
     """
-    return {
-        **extra,
-        "temperature": llm_cfg.temperature,
-        "max_retries": llm_cfg.max_retries,
-        "logprobs": llm_cfg.logprobs,
-    }
+    return {**extra, **_sampling_params(llm_cfg)}
 
 
 class ModelEndpointService:
@@ -90,12 +95,20 @@ class ModelEndpointService:
         Seeds are derived from existing Settings / env-var values so that
         existing deployments continue working after the Phase 14 upgrade
         without any admin intervention.
+
+        For ``llm``/``vlm``, a type with existing rows is *not* skipped
+        outright: those rows are backfilled with any sampling params missing
+        from their ``extra`` first (see ``_backfill_sampling_params``), since
+        endpoints created before #720's fix never had them written and would
+        otherwise keep silently running at the provider default forever.
         """
         seeds = self._build_default_seeds()
         now = datetime.now(UTC)
         for model_type, data in seeds.items():
             existing = await self._repo.list_all(model_type=model_type)
             if existing:
+                if model_type in _SAMPLING_TYPES:
+                    await self._backfill_sampling_params(existing, getattr(self._config, model_type))
                 continue
             endpoint: str = data["endpoint"]
             model_name: str = data["model_name"]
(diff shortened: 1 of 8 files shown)


openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -124,7 +142,7 @@ def _build_default_seeds(self) -> dict[str, dict[str, Any]]:
                 "model_name": os.getenv("LLM_MODEL", s.llm.model),
                 "timeout": s.llm.timeout,
                 "extra": _with_enable_thinking(
-                    _with_api_key({"implementation": "vllm"}, s.llm.api_key),
+                    _with_sampling_params(_with_api_key({"implementation": "vllm"}, s.llm.api_key), s.llm),
```
hedhoud, line 158: This fixes newly created seed rows, but not deployments that already have LLM/VLM endpoints in PostgreSQL. On startup, `seed_defaults()` skips a model type as soon as any row exists, and `load_all()` then copies that row's old `extra` unchanged. I reproduced an upgrade with an existing default LLM row: configured temperature/logprobs were still absent after seeding and loading. Could we backfill missing values, or apply the configured defaults while loading old rows without overwriting values explicitly stored on the endpoint? A regression test starting with a pre-fix row would cover the upgrade path.

hedhoud (changes requested):
The new-install path is covered, but existing deployments still keep the old endpoint configuration after upgrading.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/7daef14a11facf857f9fcffb45898440616acb11.

`seed_defaults()` no longer skips a type outright once rows exist for it — for `llm`/`vlm` it now backfills any of `temperature`/`max_retries`/`logprobs` missing from an existing row's `extra` (new `_backfill_sampling_params`, called from `seed_defaults()`). Only keys *absent* from `extra` are filled in, so anything already stored — whether set explicitly by an admin or written by an earlier fix/backfill — is left untouched.

Added three regression tests in `test_model_endpoint_service.py`:
- a pre-fix row (no sampling keys at all) gets fully backfilled
- a partially-fixed row keeps its explicit `temperature` while the still-missing `max_retries`/`logprobs` get filled in
- a fully-configured row is left byte-for-byte as stored even when current Settings would compute different values

Full unit suite (1894 tests) and ruff both pass.

hedhoud (approved):
The upgrade path is covered now: existing LLM and VLM endpoints receive only missing sampling settings, while explicitly stored values remain unchanged. I also verified the client payload behavior and the current develop merge. Looks good to merge.

PR #748, opened 2026-07-21
fix: embedder batch_size not applied + better embedding error diagnostics

ThibautChoppy:
## Context

Indexing a large PDF (1785 chunks) was failing at the embedding stage with an opaque 400 from the inference gateway, with no indication of the actual cause. Investigation happened in stages:

- The log only showed `Embedder API error (400)` — never the response body — so there was no way to know why the request was rejected.
- Once the detail was visible, we found that `EMBEDDER_BATCH_SIZE` (and the other model-endpoint env vars) were never actually applied to the indexing task: the model endpoint registry is seeded into the DB once on first boot, then becomes the source of truth — an env var change on a later rollout silently did nothing without an explicit admin API call.
- That confusion was hiding the real issue: an oversized batch payload was getting truncated by a size limit on the gateway side, producing confusing JSON parse errors instead of a clear cause.

## What changed

**1. Richer embedding error logs** (`openrag/services/inference/vllm_client.py`)
- The `"Embedding failed after N/M batches"` warning now includes `error_detail` — the embedder's actual HTTP response body, which was already captured on the exception but never logged.
- On a 400 response, the failing batch's texts are scanned for a literal, malformed `\u` escape sequence (the shape that trips some downstream JSON parsers even though our own JSON serialization is always valid) — surfaced as `suspect_texts: [{index, snippet}]` to pinpoint the offending input directly from the log line.

**2. Model endpoint config can now be hot-reloaded via a pod rollout** (`core/config/model_endpoints.py`, `core/config/loader.py`, `services/orchestrators/model_endpoint_service.py`)
- New env var `MODEL_ENDPOINT_SYNC_ON_BOOT` (default `false`, existing behavior unchanged).
- When `true`, on every boot the endpoint whose name matches the current env-derived slug (e.g. `bge-m3`) is resynced from Settings/env (`endpoint`, `model_name`, `batch_size`, `timeout`) — so a Helm/env value change plus a rollout is enough, no admin API call required.
- `extra` (API keys, implementation flags) is deliberately never touched by the sync, so a manually-set secret survives.
- Any endpoint created under a different name via the admin API/UI is left untouched — this only affects the one endpoint that mirrors env config.

Why opt-in: the model endpoint registry is meant to be admin-editable after first boot (supports multiple named endpoints per type, hand-tuned independently of env vars). Always resyncing from env on every boot would silently clobber that for anyone managing endpoints through the admin UI. `MODEL_ENDPOINT_SYNC_ON_BOOT=true` is for deployments (like ours) that manage this purely through Helm values + GitOps and want a plain rollout to be enough.

## Test plan

- [x] `uv run ruff check openrag/ tests/`
- [x] `uv run ruff format --check openrag/ tests/`
- [x] `uv run python scripts/check_layer_imports.py`
- [x] `uv run pytest tests/unit/` — 1894 passed
- [x] New unit tests covering: sync left off by default, sync applied when enabled, sync never touches a differently-named (hand-created) endpoint
- [x] Verified in a real cluster: batch_size env change + pod rollout alone now takes effect on the indexing task's embedder client

## Summary by CodeRabbit

* **Configuration**
  * Added an option to control whether model endpoint settings are synchronized from environment values when services start.
  * Defaults to preserving endpoint changes made through the administration interface across restarts.


hedhoud (dismissed):
I found a few issues that should be addressed before merging:

1. Model-name changes are not synchronized. The lookup uses the new model slug, so an endpoint stored under the previous model name is missed; because another endpoint exists, startup then skips the update. With sync enabled, changing the model through Helm still leaves the old database configuration active.

2. This PR now conflicts with develop in the endpoint startup logic. Develop added sampling-parameter backfilling for existing LLM/VLM endpoints. Please rebase and preserve that behavior when combining it with sync-on-boot, with a test covering both paths.

3. The RC workflow and Graphify ignore changes are unrelated and fork-specific. If merged upstream, RC images could be published from a tag on any branch and Docker Hub RC publishing would be removed, which conflicts with the documented release flow. Please revert these changes here or move them to a dedicated release-process PR.

4. The new warning persists the complete upstream response body and raw snippets from indexed documents. That can leak document content into stderr and centralized logs, and the response body is unbounded. Please keep non-sensitive identifiers such as batch index and character position, or gate bounded raw details behind an explicit debug option.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -72,20 +72,47 @@ async def seed_defaults(self) -> None:
...
-            if existing:
-                continue
             endpoint: str = data["endpoint"]
             model_name: str = data["model_name"]
             if not endpoint:
                 logger.info(f"No {model_type} endpoint configured — skipping seed.")
                 continue
+
+            name = _slug(model_name or "")
+            existing_row = await self._repo.get(name, model_type)
```
hedhoud: Thanks for adding the synchronization option. Could we also handle model-name changes? Currently, changing the model name changes the endpoint identifier, so OpenRAG does not find and update the existing endpoint.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -72,20 +72,47 @@ async def seed_defaults(self) -> None:
...
+                        model_type,
+                        endpoint=endpoint,
+                        model_name=model_name or None,
+                        batch_size=data.get("batch_size", 32),
+                        timeout=data.get("timeout", 30.0),
+                    )
+                    logger.info(f"Synced {model_type} endpoint '{name}' from env (MODEL_ENDPOINT_SYNC_ON_BOOT=true).")
+                continue
+
+            if await self._repo.list_all(model_type=model_type):
```
hedhoud: This area now conflicts with the startup logic in develop. When resolving it, could you please preserve the recent LLM/VLM sampling-parameter backfill and add a test covering both behaviors together?

.github/workflows/build_rc.yml
```diff
@@ -1,16 +1,13 @@
 # Build and publish release-candidate images.
 #
-# Git flow: release candidates are cut from a `release/X.Y.Z` branch. Tagging
-# that branch with `vX.Y.Z-rc.N` publishes the api, ray, and admin-ui images
-# under the `vX.Y.Z-rc.N` tag — mirroring the GA pipeline (`build.yml`) but
-# guarded to `release/*` instead of `main`, and NEVER tagging `latest`.
+# Triggered by any tag matching v*-rc.* (e.g. v2.0.0-rc.5), pushed from any
+# branch — mirrors the GA pipeline (build.yml) but never tags `latest` and
```
hedhoud: Could we remove the RC workflow changes from this PR or move them to a separate one? They change the upstream release and publishing process, while this PR is mainly about embedding configuration and diagnostics.

openrag/services/inference/vllm_client.py
```diff
@@ -265,9 +286,15 @@ async def _run(batch: list[str]) -> list[list[float]]:
...
-                "Embedding failed after {d}/{b} batches", d=done, b=len(batches)
-            )
+            # `.extra` (set on OpenRAGError subclasses like EmbeddingAPIError) carries
+            # the embedder's actual response body — repr(exc) alone only has the
+            # generic "Embedder API error (400)" message, not why it was rejected.
+            logger.bind(
+                batches_done=done,
+                n_batches=len(batches),
+                error=repr(exc),
+                error_detail=getattr(exc, "extra", None),
```
hedhoud: The improved diagnostics are useful. However, logging complete response bodies and raw document snippets may expose sensitive content or create very large logs. Could we keep only safe details such as the batch index and character position, or make raw details available only in debug mode?

openrag/core/config/loader.py
```diff
@@ -45,6 +45,8 @@
     ("EMBEDDER_TIMEOUT", "embedder.timeout", float),
     ("EMBEDDER_BATCH_SIZE", "embedder.batch_size", int),
     ("EMBEDDER_CONCURRENCY", "embedder.embed_concurrency", int),
+    # Model endpoint registry
+    ("MODEL_ENDPOINT_SYNC_ON_BOOT", "models.sync_on_boot", bool),
```
Ahmath-Gadji, line 49: **Please document `MODEL_ENDPOINT_SYNC_ON_BOOT` in the docs.**

This new operator-facing toggle isn't documented anywhere — the natural home is `docs/content/docs/documentation/env_vars.md`, which is the canonical env-var reference (it already documents `EMBEDDER_*`, LLM/VLM, and Reranker vars). A short subsection there — near the model-endpoint / Embedding config — would fit.

Worth covering:
- What it does and the `false` default (opt-in re-sync of the env-derived endpoint on every boot).
- **Importantly, the exact scope**: the sync only re-applies `endpoint` / `model_name` / `batch_size` / `timeout`, **not** `extra` — so `api_key`, sampling params like `LLM_TEMPERATURE`, and `enable_thinking` are *not* re-applied by a rollout.

Documenting the scope avoids the surprise of an env change silently not taking effect. `openrag/core/config/loader.py:49`

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Building on this — it might be worth considering whether, when `sync_on_boot` is enabled, the sync should re-apply the **entire** env-derived settings rather than just `endpoint` / `model_name` / `batch_size` / `timeout`.

Right now everything in `extra` is skipped, so changing `LLM_TEMPERATURE`, `LLM_MAX_RETRIES`, `LLM_LOGPROBS`, `enable_thinking`, or `implementation` via env + a rollout silently has no effect — the same "env change doesn't take" confusion this PR set out to fix, just moved to a different set of fields. (There's also a related gap: llm/vlm/reranker seeds carry no `batch_size`, so the sync writes a hardcoded `32` to them — `model_endpoint_service.py:103`.)

If we do broaden it, the one field to keep protecting is `api_key`: re-sync the rest of `extra` from env but preserve a hand-set secret when the env value is unset (rather than blanket-excluding all of `extra`). That would make "managed purely via env + rollout" actually mean the whole config, which is the stated goal of the flag.

andyne13:
Thanks for this — the embedding diagnostics and the "endpoint env vars silently ignored after first boot" find are both genuinely useful.

**+1 on splitting out the CI/build changes.** The `build_rc.yml` edits look fork-specific: they drop the Docker Hub login + publish (`linagoraai/openrag*`) and the `release/*` base-branch guard, and add the lowercase-repo `vars` job (only needed for a mixed-case fork owner — the upstream repo path is already lowercase). Merged upstream, those would stop the RC images publishing to Docker Hub and let RC tags build from any branch, so this is cleaner as its own PR (or dropped here). Same for the `graphify-out/` `.gitignore` entry, which looks local to your tooling.

**One heads-up on develop overlap for the app-code fix** — this area moved recently, which is likely the source of the conflict:
- #728 (merged) touches the same `vllm_client.py` request path — it stopped `batch_size` leaking into the LLM/VLM/reranker request bodies. That's a different facet than the "batch_size not applied" one here, but it's the same file, so a rebase will avoid a collision.
- The env-driven seeding in `model_endpoint_service.py` — the exact "seed once, then the DB is the source of truth" behaviour you're addressing — was established in #560, and develop has since added LLM/VLM sampling-parameter backfill to that same startup path.

Rebasing onto develop and reconciling `seed_defaults`/startup with that backfill (plus a test covering both behaviours, per @hedhoud's note) should clear the conflict and avoid a double-fix. Happy to point at the specific spots if that helps.

ThibautChoppy:
Unecessary changes on ci and gitignore have been removed. PR ready to close ;)

Ahmath-Gadji (approved):
LGTM

PR #749, opened 2026-07-21
fix(streaming): flush answer tail and sources on any stream exit

### LINKED ISSUE #715: Streaming: upstream close without [DONE] drops the answer tail and all extra.sources, with no error
## Problem

The streamed answer's tail flush and the entire `extra.sources` payload are emitted **only** inside the `data: [DONE]` branch, so an upstream stream that ends without `[DONE]` silently loses both.

`openrag/core/utils/source_filtering.py:99-124`:
```python
if line.strip() == "data: [DONE]":
    final_clean, citations = extract_and_strip_sources_block(pending)
    ...
    if chunk_template and len(final_clean) > emitted_len:
        tail_chunk["choices"][0]["delta"] = {"content": final_clean[emitted_len:]}
        tail_chunk["extra"] = filtered_json
        yield ...
    if chunk_template:
        finish_chunk["extra"] = filtered_json
        yield ...
    yield "data: [DONE]\n\n"
```
`VLLMClient.stream_chat` relays `resp.aiter_lines()` verbatim. If the upstream closes without `[DONE]` (timeout, proxy drop, worker restart), the `async for` simply ends and the generator returns — **no exception is raised**, so the router's `except` clauses never fire.

## Impact

The client receives:
- a **truncated answer** missing the trailing `buffer_size` characters, which are withheld by design to catch a split `[Sources: ...]` tag; and
- **no `extra.sources` at all** — a confident-looking answer with zero citations,

with no error envelope and nothing distinguishing it from a normal completion. Invisible in metrics.

Distinct from #390 (buffer too small for long tags), which was about the tag leaking through; this is about the terminal flush never running.

## Fix

Move the flush out of the `[DONE]` branch into a `finally:` / post-loop block guarded by an `emitted_done` flag, so the tail and `extra` are emitted exactly once on any loop exit — and consider signalling truncation to the client when the stream ended without `[DONE]`.

Ahmath-Gadji:
## Summary
- Closes #715: an upstream stream that ends without `data: [DONE]` (timeout, proxy drop, worker restart) silently dropped the buffered answer tail and all of `extra.sources`, with no error surfaced.
- `stream_with_source_filtering` now performs the terminal flush (tail content + `extra.sources`) in a single post-loop block that runs whether the loop exits via `[DONE]` or the upstream connection simply closing, instead of only inside the `data: [DONE]` branch.
- The generator always emits a downstream `data: [DONE]`, even if upstream never sent one, so clients (e.g. the OpenAI SDK) don't hang waiting for stream termination.
- When the stream ended without `[DONE]`, the final chunk's `extra` now includes `"truncated": true` so a silently-truncated answer is distinguishable from a normal completion instead of looking indistinguishable from a clean one.

## Test plan
- [x] Added regression tests in `tests/unit/core/utils/test_source_filtering.py::TestStreamClosedWithoutDone` covering: tail/sources still flushed without `[DONE]`, `truncated` flag set when `[DONE]` is missing, and no `truncated` flag on a normal `[DONE]`-terminated stream.
- [x] `uv run pytest tests/unit/core/utils/test_source_filtering.py` — 42 passed.
- [x] `uv run pytest tests/unit/` — 1888 passed.
- [x] `uv run ruff check` / `uv run ruff format --check` on changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures the streaming terminal “tail flush” runs exactly once across all upstream termination paths (clean `[DONE]`, upstream close without `[DONE]`, and mid-stream exceptions).
  * Preserves buffered answer tail and emits consistent final completion signals so clients don’t hang.
  * Filters and salvages cited sources from buffered output; sets `extra.truncated` and warning behavior based on whether termination was clean.

* **Tests**
  * Expanded SSE regression coverage for combined content+`finish_reason`, role-only preamble closes, missing `[DONE]`, upstream exceptions, and disconnect behavior.

### DIFF
--- a/openrag/core/utils/source_filtering.py
+++ b/openrag/core/utils/source_filtering.py
@@ -187,6 +187,17 @@ async def stream_with_source_filtering(
         # GeneratorExit are BaseException, not Exception, so a downstream client
         # disconnect propagates untouched and correctly skips the flush.
         stream_error = exc
+    finally:
+        # Release the upstream HTTP connection promptly. Breaking on `[DONE]`
+        # leaves the client's stream generator suspended inside its
+        # `async with response` (an `async for` does *not* close its iterator on
+        # break), so the pooled connection stays checked out until GC — which
+        # exhausts the httpx pool under concurrent traffic. Closing it here runs
+        # that cleanup now. Awaiting in a finally is safe (only *yielding* during
+        # teardown is forbidden); the guard tolerates iterables without `aclose`.
+        aclose = getattr(llm_stream, "aclose", None)
+        if aclose is not None:
+            await aclose()
 
     # The finish chunk needs *a* template; prefer the content/finish chunk, but
     # fall back to any earlier chunk so a preamble-only stream still gets flagged.
--- a/openrag/core/utils/source_filtering.py
+++ b/openrag/core/utils/source_filtering.py
@@ -140,10 +140,15 @@ async def stream_with_source_filtering(
             if data.get("choices"):
                 last_chunk = data
 
+            # `content` and `finish_reason` are handled independently: some
+            # OpenAI-compatible providers pack the last token and the terminal
+            # `finish_reason` into the *same* chunk, so gating content on
+            # `elif finish_reason` would silently drop that final token.
             if finish_reason:
                 last_finish_reason = finish_reason
                 chunk_template = data
-            elif content:
+
+            if content:
                 chunk_template = data
                 pending += content
 
@@ -168,7 +173,11 @@ async def stream_with_source_filtering(
                     }
                     yield f"data: {json.dumps(out)}\n\n"
                     emitted_len = safe_end
-            else:
+            elif not finish_reason:
+                # Neither content nor finish_reason (role preamble, usage-only or
+                # keep-alive chunk): pass it through untouched. A finish-only chunk
+                # is intentionally *not* re-emitted here — the terminal flush emits
+                # the finish chunk so it can carry `extra.sources`.
                 data["extra"] = "{}"
                 yield f"data: {json.dumps(data)}\n\n"
     except Exception as exc:
--- a/openrag/core/utils/source_filtering.py
+++ b/openrag/core/utils/source_filtering.py
@@ -84,45 +84,29 @@ async def stream_with_source_filtering(
     model_name: str,
     buffer_size: int | None = None,
 ):
-    """Process an LLM SSE stream, stripping line-terminal source tags."""
+    """Process an LLM SSE stream, stripping line-terminal source tags.
+
+    The terminal flush (tail content + ``extra.sources``) runs exactly once
+    after the loop, regardless of whether it exits via ``data: [DONE]`` or
+    the upstream simply closing the connection without it — otherwise a
+    dropped upstream stream silently loses the answer's tail and all
+    sources with no error surfaced to the caller.
+    """
     if buffer_size is None:
         buffer_size = max(_MIN_STREAM_LOOKAHEAD, _min_sources_tag_buffer_size(len(sources)))
     pending = ""
     emitted_len = 0
     chunk_template = None
     last_finish_reason = None
+    saw_done = False
 
     async for line in llm_stream:
         if not line.startswith("data:"):
             continue
 
         if line.strip() == "data: [DONE]":
-            final_clean, citations = extract_and_strip_sources_block(pending)
-            final_clean = final_clean.rstrip()
-
-            filtered = filter_sources_by_citations(sources, citations)
-            filtered_json = json.dumps({"sources": filtered})
-
-            if chunk_template and len(final_clean) > emitted_len:
-                tail_chunk = copy.deepcopy(chunk_template)
-                tail_chunk["choices"][0]["delta"] = {"content": final_clean[emitted_len:]}
-                # Content chunk must not carry finish_reason (template may be the
-                # finish chunk); clients treat such a chunk as terminal and drop its
-                # delta. The separate finish chunk below emits it with an empty delta.
-                tail_chunk["choices"][0]["finish_reason"] = None
-                tail_chunk["extra"] = filtered_json
-                yield f"data: {json.dumps(tail_chunk)}\n\n"
-
-            if chunk_template:
-                await asyncio.sleep(0.05)
-                finish_chunk = copy.deepcopy(chunk_template)
-                finish_chunk["choices"][0]["delta"] = {}
-                finish_chunk["choices"][0]["finish_reason"] = last_finish_reason or "stop"
-                finish_chunk["extra"] = filtered_json
-                yield f"data: {json.dumps(finish_chunk)}\n\n"
-
-            yield "data: [DONE]\n\n"
-            continue
+            saw_done = True
+            break
 
         data = json.loads(line[len("data: ") :])
         data["model"] = model_name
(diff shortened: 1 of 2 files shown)


aditykris (commented):
1. The truncated flag can silently disappear in a narrow edge case. I reproduced this directly: if the upstream connection ends before any chunk with content or a finish_reason ever arrives (e.g. only a bare role-preamble chunk, then disconnect), chunk_template stays None. Both the tail-chunk and finish-chunk blocks are gated on if chunk_template:, so neither ever fires the client gets data: [DONE] with no extra.truncated anywhere, even though the server logged the warning internally. It's a narrow case (needs a disconnect before the very first content token), and the client at least doesn't hang, but the "distinguishable from a clean completion" guarantee the PR advertises doesn't hold here.

2. More significant — real timeouts/connection drops likely bypass this fix entirely. I checked how llm_stream actually terminates on a network problem: both #[vllm_client.py:161](/openrag/services/inference/vllm_client.py#L161) and [ollama_client.py:112](/openrag/services/inference/ollama_client.py#L112) wrap the streaming loop in try/except httpx.ConnectError / except httpx.TimeoutException, re-raising as InferenceConnectionError/InferenceTimeoutError. I confirmed with a direct repro that when the upstream generator raises mid-iteration (rather than ending cleanly), the exception propagates straight out of stream_with_source_filtering the new post-loop flush code never runs, because Python doesn't execute the code after a for loop when the loop raises. That exception is instead caught by [chat.py](openrag/api/routers/user/chat.py#L327)'s existing except OpenRAGError / except Exception handlers, which yield a generic _make_sse_error(...) chunk the buffered answer tail and sources are lost there too, just as before this PR, only now replaced by an error message instead of a clean tail.

Since the PR explicitly names "timeout" as one of the three root causes of #715, and a mid-stream read timeout is exactly what gets converted to InferenceTimeoutError in the existing client code, this fix likely does not address that particular named cause only the flavor of disconnect where the transport ends with a bare EOF and no distinguishable HTTP-level error (plausible for "proxy drop" / some "worker restart" cases, where the connection just closes without httpx raising). Worth confirming against whatever telemetry/repro originally motivated #715 — if it was specifically an httpx timeout, this PR may not close the loop on it.

Ahmath-Gadji:
@aditykris thanks — both are spot-on. Fixed in 58695292.

**2 (the significant one — mid-stream raise bypasses the flush):** confirmed. A read timeout / connection drop surfaces as the upstream generator *raising* mid-iteration (`InferenceTimeoutError` / `InferenceConnectionError` from the vLLM/Ollama clients), so control unwinds straight past the post-loop flush and the buffered tail + sources were lost — exactly the "timeout" cause named in #715. The loop is now wrapped in `try/except`:
- `asyncio.CancelledError` / `GeneratorExit` still propagate untouched, so the router's client-disconnect handling is unchanged.
- any other error is captured and we **fall through to the same flush**, delivering the partial answer with `extra.truncated = true`.
- if *nothing* was ever streamed (no template to build a chunk from), the error is **re-raised** so `chat.py` surfaces the real failure via `_make_sse_error` instead of a silent, clean-looking empty `[DONE]`.

**1 (truncated flag disappears when `chunk_template is None`):** confirmed too — a preamble-only close left both flush blocks gated out, so `[DONE]` went out with no `truncated`. Now tracking a `last_chunk` fallback (any chunk with a usable `choices[0]`) and templating the flush off `chunk_template or last_chunk`, so a preamble-only stream still emits a finish chunk carrying `truncated: true`. The tail block is naturally a no-op there since there's no buffered content.

Regression tests added for all three paths (`test_preamble_only_close_still_flags_truncated`, `test_upstream_raises_after_content_flushes_tail_and_sources`, `test_upstream_raises_before_any_chunk_reraises`). Full unit suite green (1908 passed).

openrag/core/utils/source_filtering.py
```diff
@@ -84,89 +84,147 @@ async def stream_with_source_filtering(
...
+            content = delta.get("content", "") or ""
+            finish_reason = choice.get("finish_reason")
+
+            # Only chunks with a real `choices[0]` can template the finish chunk;
+            # skip usage-report chunks (`"choices": []`) whose deepcopy would
+            # IndexError in the flush below.
+            if data.get("choices"):
+                last_chunk = data
+
+            if finish_reason:
```
hedhoud, line 147: Could we process `content` before checking `finish_reason` here? Some OpenAI-compatible providers send both in the final chunk. In that case the last text is skipped; for example, `Hello ` followed by a final `world` chunk currently returns only `Hello`. Handling the content and finish reason independently, with a small regression test, should prevent that final piece from being lost.

Ahmath-Gadji: Good catch — fixed in 0dd35e93. `content` and `finish_reason` are now handled independently: the loop always appends `content` to the buffer and records `finish_reason` separately, so a provider that packs the last token + stop reason into one chunk no longer drops it. Your exact example is now a regression test (`test_content_and_finish_reason_in_same_chunk_keeps_last_token`): `Hello ` + a final `world` chunk yields `Hello world`. A finish-only chunk is still deferred to the terminal flush so it can carry `extra.sources`.

openrag/core/utils/source_filtering.py
```diff
@@ -84,89 +84,149 @@ async def stream_with_source_filtering(
...
+                last_finish_reason = finish_reason
+                chunk_template = data
+            elif content:
+                chunk_template = data
+                pending += content
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 0dd35e93 — `content` and `finish_reason` are now handled independently so a chunk carrying both no longer loses its final token. Regression test added (`test_content_and_finish_reason_in_same_chunk_keeps_last_token`).

openrag/core/utils/source_filtering.py
```diff
@@ -84,89 +84,154 @@ async def stream_with_source_filtering(
...
-                    "extra": "{}",
-                }
-                yield f"data: {json.dumps(out)}\n\n"
-                emitted_len = safe_end
-        else:
-            data["extra"] = "{}"
-            yield f"data: {json.dumps(data)}\n\n"
+            if line.strip() == "data: [DONE]":
+                saw_done = True
+                break
```
hedhoud, line 120: Could we close the upstream iterator before breaking here? Both the vLLM and Ollama stream methods keep an HTTP response open inside their async generator. Breaking on `[DONE]` leaves that generator suspended; I reproduced that its cleanup does not run until it is explicitly closed or finalized later. Under concurrent chat traffic, this can keep connections checked out longer than necessary. Please close the iterator when supported, and add a small test confirming the upstream cleanup runs after a normal `[DONE]`.

Ahmath-Gadji: Good catch — and you're right that it's a regression this PR introduced: `develop` handled `[DONE]` with `continue`, which drained `llm_stream` to natural EOF and let the client's `async with response` release the connection; switching to `break` left that generator suspended (an `async for` doesn't close its iterator on break), holding the pooled connection until GC.

Fixed in 90d132bc: the upstream iterator is now closed in a `finally`, so the connection is released promptly on every exit path (awaiting there is safe — only *yielding* during teardown is forbidden), with a `getattr` guard for non-generator iterables:

```python
finally:
    aclose = getattr(llm_stream, "aclose", None)
    if aclose is not None:
        await aclose()
```

Added the test you asked for (`test_upstream_iterator_closed_after_done`): a tracking upstream generator whose `finally` flips a flag, asserted set right after a normal `[DONE]` run. It holds a reference to the generator so a pass can only mean an explicit `aclose()` ran, not GC — I confirmed it fails without the fix.

hedhoud (dismissed):
LGTM. The missing-[DONE], mid-stream error, final-token, and upstream cleanup paths are now covered. The focused tests and CI are green.

openrag/core/utils/source_filtering.py
```diff
@@ -84,89 +84,165 @@ async def stream_with_source_filtering(
...
+                                "delta": {
+                                    **choice.get("delta", {}),
+                                    "content": cleaned[emitted_len:safe_end],
+                                },
+                            }
+                        ],
+                        "extra": "{}",
+                    }
+                    yield f"data: {json.dumps(out)}\n\n"
+                    emitted_len = safe_end
```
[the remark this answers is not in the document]
Ahmath-Gadji: Good catch — and it's a hazard my own `content`/`finish_reason` split exposed. Fixed in 55a27d9f by clearing `finish_reason` on the intermediate emit chunk, mirroring the terminal tail-chunk guard:

```python
"choices": [{
    **choice,
    "delta": {**choice.get("delta", {}), "content": cleaned[emitted_len:safe_end]},
    "finish_reason": None,  # never mark a mid-stream chunk terminal
}],
```

Added a regression test (`test_intermediate_content_chunk_never_marked_terminal`): a >buffer_size answer whose last token arrives packed with finish_reason — asserts no content-bearing chunk carries finish_reason and the full text is delivered. Verified it fails without the fix.

hedhoud (approved):

PR #750, opened 2026-07-21
Restore actor-pool acceptance when retirement is safely aborted

hedhoud:
## Why

The report is confirmed: a successful drain followed by a timeout or status failure left the detached pool alive but permanently unable to accept new indexing submissions.

## Design

Each drain now has its own operation id. A failed retirement restores acceptance only when that same operation started the drain and actor termination has not begun. A second retirement cannot resume another operation’s drain. Once actor removal starts, the pool is intentionally left drained.

## Validation

Added coverage for normal retirement, timeout and status failures, recovery failure, ownership conflicts, retries, and continued submission on the same dispatcher after a failed retirement.

Focused worker tests, the full unit suite, repository Ruff checks, formatting checks, and the package build passed. The project has no configured type-check command.

## Recovery

If restoring acceptance fails, the original retirement error is preserved and the command logs that the pool may remain drained. Stop API traffic and restart the Ray cluster to recreate the generation.


hedhoud:
@codex review

PR #755, opened 2026-07-22
fix(chat): resolve the default chat LLM from the catalog default endpoint

### LINKED ISSUE #754: Default chat LLM ignores the catalog default endpoint — pinned to the .env (settings.llm) client
When a partition has no `chat_llm` preset (the "Default (from retrieval config)" option in the UI), `QueryService._resolve_llm()` returns the static `self._llm` — a client built **once at startup** from `settings.llm` (the `.env` `LLM_ENDPOINT` / `LLM_MODEL`) in `di/container.py` (`query_service` property, `create_llm("vllm", endpoint=llm_cfg["base_url"], …)`). It never consults the model-endpoint catalog.

Meanwhile the catalog's default endpoint (the `is_default=True` row) is exposed by `llm_factory("default")`, kept in sync by `ModelEndpointService.load_all()` and cache-invalidated on every default change.

`services/orchestrators/query_service.py:_resolve_llm` — both the "no partition-scoped preset" and "no single `chat_llm` preset among the partitions" branches `return self._llm`.

**Effect:** promoting a new default LLM endpoint in the catalog (admin UI, or `POST /model-endpoints/llm/{name}/set-default`) has **no effect** on the default chat path — it keeps answering with the `.env` model. The only way to change the model is to set the partition's `chat_llm` explicitly, which routes through `llm_factory(name)` and works. Deleting every catalog endpoint except the intended one also has no effect: the static `self._llm` still points at the `.env` endpoint.

**Repro:**
1. Deploy with `LLM_MODEL=A` (seeds catalog default `A`).
2. Add endpoint `B`, promote it to default.
3. Chat on a partition with `chat_llm` unset → still answered by `A`, not `B`.

Observed logs (partition `test`, `chat_llm=null`, catalog default = `B`):
```
_resolve_llm - Answering with the default LLM (no single chat_llm preset among the partitions) [partitions=['test'] | chat_llm_presets=[]]
```

**Fix:** route the default path through `llm_factory("default")` (the live catalog default), falling back to the static `self._llm` only when no factory is wired (unit tests) or the catalog has no default endpoint yet. Additionally, log the resolved preset name on **every** path — including the default — so "which chat LLM answered?" is answerable from the logs.

Ahmath-Gadji:
Closes #754.

## Problem

When a partition has no `chat_llm` preset ("Default (from retrieval config)"), `QueryService._resolve_llm()` returned the static `self._llm` — a client built once at startup from `settings.llm` (the `.env` `LLM_ENDPOINT`/`LLM_MODEL`) in `di/container.py`. It never consulted the model-endpoint catalog, so **promoting a new default endpoint in the catalog had no effect on the default chat path** — chat kept answering with the `.env` model. Only setting a partition's `chat_llm` explicitly worked (that path goes through `llm_factory(name)`).

## Fix

`services/orchestrators/query_service.py`:

- The default path now resolves the **live catalog default** via `llm_factory("default")` (the `is_default=True` endpoint, kept in sync by `ModelEndpointService.load_all()` and cache-invalidated on every default change). Falls back to the static `self._llm` only when no factory is wired (unit tests) or the catalog has no default yet.
- A stale/unknown partition `chat_llm` now also falls **through to the catalog default** (was the static `.env` client).
- Split resolution into small helpers: `_agreed_partition_chat_llm()`, `_default_llm()`, `_default_llm_name()`.
- **Logging (issue ask):** every path now logs the resolved preset name at debug — including the default (`Answering with the default chat_llm preset · chat_llm=<endpoint>`) — so the answering chat LLM is identifiable from the logs.

Map-reduce relevancy (`_infer_relevancy`) stays deliberately pinned to the static default — pre-existing documented behavior, unchanged here.

## Live verification

Two endpoints in the catalog (same host, distinct `model_name`). Partition `test` left on `chat_llm=null` the whole time; **only the catalog default was flipped**:

| Catalog default | `_resolve_llm` log | Actual answer |
|---|---|---|
| Mistral (`is_default`) | `Answering with the default chat_llm preset · chat_llm=Mistral-Small-3.2-24B-Instruct-2506-FP8 · partitions=['test']` | "Pong" |
| Luciole (after `set-default`) | `Answering with the default chat_llm preset · chat_llm=Luciole · partitions=['test']` | "Je ne comprends pas la demande…" |

The resolved LLM followed the catalog default (Mistral → Luciole) without touching the partition; the two different answers confirm a different endpoint actually served each request. Before this fix, the second row would still have resolved to the `.env` Mistral client.

## Tests

`tests/unit/services/orchestrators/test_query_service.py` — updated the default-path assertions to the corrected contract and added coverage for `_default_llm`/`_default_llm_name`. Suite green (45 in the file, 394 across orchestrators). `ruff check` clean.

### DIFF
--- a/openrag/services/orchestrators/query_service.py
+++ b/openrag/services/orchestrators/query_service.py
@@ -180,50 +180,113 @@ class QueryService:
     def _resolve_llm(self, partition: list[str] | None) -> LLM:
         """Effective LLM for this request — query generation and answering.
 
-        Honors a partition's configured ``chat_llm`` model-endpoint preset
-        (set via the admin API) over the default LLM. Resolved once per
-        request (``chat`` / ``chat_stream`` / ``complete``) and used for
-        both the query-contextualization call and the final answer;
-        map-reduce stays on the default LLM until its post-release
-        refactor. Same partition semantics as
-        ``_resolve_chat_history_depth``: no partition and the ``"all"``
-        sentinel use the default; a multi-partition request uses the
-        preset only when every partition that sets one names the same
-        endpoint — with conflicting presets there is no single owning
-        partition, so the default applies.
+        Resolution order:
+
+        1. A partition's configured ``chat_llm`` model-endpoint preset (set
+           via the admin API) wins when the request scopes to one or more
+           named partitions that agree on a single preset. Resolved once per
+           request (``chat`` / ``chat_stream`` / ``complete``) and used for
+           both the query-contextualization call and the final answer;
+           map-reduce stays on the catalog default until its post-release
+           refactor.
+        2. Otherwise the **catalog default** endpoint — the ``is_default=True``
+           row, exposed by ``llm_factory`` under the ``"default"`` alias and
+           resolved fresh per request so promoting a new default endpoint at
+           runtime takes effect immediately. This covers a direct/web-only
+           request (no partition), the cross-partition ``"all"`` sentinel,
+           partitions that set no preset, and partitions whose presets
+           conflict (no single owning partition). Same partition semantics as
+           ``_resolve_chat_history_depth``.
+        3. The static ``self._llm`` (built from ``settings.llm`` at startup)
+           only as a last resort — no endpoint factory is wired (unit tests)
+           or the catalog has no default endpoint yet.
 
         ``chat_llm`` is validated against the endpoint catalog when it is
         assigned (``PartitionService`` rejects an unknown name at create /
         PATCH time), but a stored name can still go stale afterwards — the
         endpoint may be renamed or deleted after assignment — so an
-        unresolvable name here must not fail the chat request; it falls back
-        to the default LLM with a warning.
+        unresolvable name here must not fail the chat request; it falls
+        through to the catalog default with a warning.
+
+        The resolved preset name is always logged (at debug), including for
+        the default, so "which model answered?" is answerable from the logs.
+        """
+        chat_llm = self._agreed_partition_chat_llm(partition)
+        if chat_llm is not None:
+            try:
+                llm = self._llm_factory(chat_llm)  # factory is not None when chat_llm is set
+            except KeyError:
+                logger.warning(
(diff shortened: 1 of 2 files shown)


openrag/services/orchestrators/query_service.py
```diff
@@ -180,50 +180,113 @@ def _resolve_chat_history_depth(self, partition: list[str] | None) -> int:
...
-        endpoint — with conflicting presets there is no single owning
-        partition, so the default applies.
+        Resolution order:
+
+        1. A partition's configured ``chat_llm`` model-endpoint preset (set
+           via the admin API) wins when the request scopes to one or more
+           named partitions that agree on a single preset. Resolved once per
+           request (``chat`` / ``chat_stream`` / ``complete``) and used for
+           both the query-contextualization call and the final answer;
+           map-reduce stays on the catalog default until its post-release
```
Ahmath-Gadji: **Doc inaccuracy (self-review): map-reduce LLM description is now wrong**

This docstring says *"map-reduce stays on the catalog default"*, but `_infer_relevancy` (`query_service.py:339`) uses the static `self._llm` — the env-built client from `settings.llm`, **not** the catalog default.

Before this PR both the chat-default path and map-reduce shared `self._llm`, so "the default LLM" was unambiguous. This PR moves the chat-default path onto the catalog default (via `llm_factory("default")`) while map-reduce stays on the static client, so the two now **diverge** — both this docstring and the `_infer_relevancy` comment ("pinned to the default LLM") are now misleading. Will correct both to state the divergence explicitly.

Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/c3799d9b — corrected the `_resolve_llm` docstring and the `_infer_relevancy` comment to state that map-reduce stays on the static `self._llm` (env client) and therefore diverges from the resolved catalog default the answer uses, until the deferred map-reduce refactor. No behavior change.

hedhoud (approved):

PR #757, opened 2026-07-22
Feat/display user name in partition

Baboulinet-33:
In the user table for a partition, display name instead of ID

## Summary by CodeRabbit

* **Enhancements**
  * Partition member lists now show users’ display names when available.
  * User IDs are used as a fallback when no display name exists.
  * Removal confirmations now identify users by display name or ID.
* **Bug Fixes**
  * Missing users are handled gracefully without breaking partition member listings.


hedhoud:
Thanks for this work! 
We’ve integrated your initial commit into PR #809 and extended it there. Since the changes will now be delivered through #809, we can close this PR.

PR #758, opened 2026-07-22
feat(models): endpoint sync-on-boot + richer embedder error diagnostics

Ahmath-Gadji:
## What

Cherry-picks the **logging / diagnostics** and **env-sync** work from #748 onto `develop`, plus documentation and review follow-ups. The three cherry-picked commits are authored by @ThibautChoppy (ownership preserved via `git cherry-pick`); the docs and review-fix commits are added on top.

## Commits

| Commit | Author | Summary |
|--------|--------|---------|
| `chore: add error logs on api` | @ThibautChoppy | On an embedding batch failure, also log `error_detail` (the `.extra` payload on `EmbeddingAPIError`) — `repr(exc)` alone only carries the generic `"Embedder API error (400)"` message, not *why* the embedder rejected the request. |
| `chore: add know your chunk` | @ThibautChoppy | On an embedder **400**, scan the batch for literal `\u…` escapes (which some downstream JSON gateways choke on) and attach the offending chunk snippets as `suspect_texts` in the error `extra`, so operators can pinpoint which input text to fix. |
| `chore: add model endpoint sync on boot` | @ThibautChoppy | New opt-in `MODEL_ENDPOINT_SYNC_ON_BOOT` (`models.sync_on_boot`). When enabled, the env-seeded endpoint of each type is re-synced from env on every boot (`endpoint`/`model_name`/`batch_size`/`timeout` — never `extra`). Default `false` keeps the "DB is the source of truth after first boot" behavior. |
| `docs(env): document the model endpoint registry and MODEL_ENDPOINT_SYNC_ON_BOOT` | (this PR) | New **Model Endpoint Registry** section in `env_vars.md` documenting the registry seeding model and the new variable. |
| `fix(inference): bound and redact the embedder error payload` | (this PR) | Review follow-up — see below. |
| `docs(env): state the sync-on-boot model-slug limitation` | (this PR) | Review follow-up — see below. |

## New environment variable

| Variable | Type | Default | Description |
|----------|------|---------|-------------|
| `MODEL_ENDPOINT_SYNC_ON_BOOT` | `bool` | `false` | Re-sync each type's env-seeded endpoint from env on every boot. Opt-in; hand-created endpoints and the `extra` field (API keys) are never touched. |

## Conflict resolution notes

Two commits needed a merge against newer `develop`:

- **`vllm_client.py`** — kept #704's `status_code=exc.response.status_code` propagation (so 429/5xx retry and 4xx fail fast) **and** layered in the new `suspect_texts` 400-diagnostics.
- **`model_endpoint_service.py` `seed_defaults`** — preserved #720's `llm`/`vlm` sampling-param backfill for existing rows **and** the new `sync_on_boot` path (which now also covers `llm`/`vlm`).

## Review follow-ups

**Bounded + redacted error payload** (`fix(inference): …`). `EmbeddingAPIError.extra` is not log-only — `error_handlers.py:115` serializes `exc.to_dict()` into the HTTP error body, so the raw upstream response and the new document snippets were client-visible. Bounded at the source so both paths benefit:

- `error` truncated to 500 chars, the same bound the LLM/VLM/Ollama paths in this file already apply; the embedder path was the only one echoing an upstream body whole.
- `_find_suspect_escapes` stops after 5 findings, so a 512-chunk batch cannot mirror its document text into the payload.
- The batch-failure log binds `redact_secrets(...)` over `.extra` instead of the raw dict.

Snippets were kept rather than reduced to a count — identifying *which* chunk is malformed is the point of the diagnostic, and the 32-char window is mostly the offending escape itself. Covered by `TestEmbedderErrorPayloadIsBounded`.

## Known limitation

`MODEL_ENDPOINT_SYNC_ON_BOOT` matches the endpoint by name, which is derived from the model slug (`org/my-model` → `my-model`). Sync therefore covers **URL, batch size and timeout**, but **changing the model itself does not take effect**: the new slug matches no row, and the seed then declines to create a competing default, so the old endpoint stays in use. Switching models still requires an admin edit.

This is documented in `env_vars.md` and pinned by `test_seed_defaults_sync_on_boot_does_not_follow_a_changed_model_slug` so it cannot change silently. It is **not a regression** — changing the model did nothing before this PR either. Properly following a changed slug needs a persisted env-managed marker (a column + Alembic migration) and would loosen the "only ever touch the slug-matched row" guarantee the feature is built around; that is a design call for @ThibautChoppy rather than something to fold into a cherry-pick. The review thread is left open as the tracking point.

## Testing

- `uv run pytest tests/unit/` → **2063 passed** on the cherry-picks; the affected suites (`test_vllm_client.py`, `test_model_endpoint_service.py`) → **98 passed** after the review follow-ups.
- Lint/format and the full CI matrix run on the branch.

The `.env.example` file is intentionally left unchanged — it is the minimal "required vars" file, and this variable is optional with a safe default, so it is documented in `env_vars.md` only.

## Summary by CodeRabbit

* **New Features**
  * Added `MODEL_ENDPOINT_SYNC_ON_BOOT` to refresh env-owned model endpoint settings on every boot.
  * Improved the model endpoint registry with first-boot env seeding and database-as-source-of-truth behavior.
* **Bug Fixes**
  * Enhanced vLLM embedder error diagnostics for malformed unicode escapes with bounded, redacted details and safer logging.
* **Documentation**
  * Documented registry seeding/sync behavior and API key rotation rules during sync.
* **Tests**
  * Expanded unit coverage for sync-on-boot behavior (preserving admin edits, adoption rules, slug changes, API key handling) and error-payload/log bounds/positions.


openrag/services/inference/vllm_client.py
```diff
@@ -276,9 +297,15 @@ async def _run(batch: list[str]) -> list[list[float]]:
...
-            )
+            # `.extra` (set on OpenRAGError subclasses like EmbeddingAPIError) carries
+            # the embedder's actual response body — repr(exc) alone only has the
+            # generic "Embedder API error (400)" message, not why it was rejected.
+            logger.bind(
+                batches_done=done,
+                n_batches=len(batches),
+                error=repr(exc),
+                error_detail=getattr(exc, "extra", None),
+            ).warning("Embedding failed after {d}/{b} batches", d=done, b=len(batches))
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 09c85027 — and the exposure was wider than the log line: `error_handlers.py:115` serializes `exc.to_dict()` (which includes `extra`) straight into the HTTP error body, so this was client-visible too, not just log noise. Bounded at the source so both paths benefit:

- `error` is truncated to 500 chars — the same bound the LLM/VLM/Ollama paths in this file already apply to an upstream body; the embedder path was the only one echoing it whole.
- `_find_suspect_escapes` now stops after 5 findings, so a 512-chunk batch cannot mirror its document text into the payload.
- The batch-failure log binds `redact_secrets(...)` over `.extra` rather than the raw dict.

I kept the snippets rather than reducing them to a count: identifying *which* chunk is malformed is the entire point of the diagnostic, and a 32-char window around the offending escape is mostly the escape sequence itself. Bounding + redaction addresses the leak without gutting it. Covered by three tests in `TestEmbedderErrorPayloadIsBounded`.

---

**Correction (f03dd041).** The third bullet above is wrong and I have removed that change. `redact_secrets()` masks values under secret-*named* keys; this payload's keys are `model_name` / `base_url` / `error` / `suspect_texts`, so it was a literal no-op and passed the document snippets through unchanged. It described a mitigation that was not happening.

The real exposure was never credentials — it is **document text**, and it matters in *logs* (operators read them across tenants) rather than in the API error (whose audience is the uploader who already owns the document). Per @hedhoud's review, the log now binds a summary — model, status, suspect count and chunk *indices* — and no document text at all. The two bounding changes (`error[:500]`, 5-finding cap) stand on their own merits and are unchanged.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,22 +101,55 @@ async def seed_defaults(self) -> None:
...
+                        timeout=data.get("timeout", 30.0),
+                    )
+                    logger.info(f"Synced {model_type} endpoint '{name}' from env (MODEL_ENDPOINT_SYNC_ON_BOOT=true).")
+                continue
+
+            if existing:
+                # Some other endpoint of this type already exists (hand-created
+                # via the admin API) — don't create a competing default. Reuses
+                # the list fetched above (nothing added/removed a row since).
+                continue
```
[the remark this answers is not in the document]
Ahmath-Gadji: Confirmed as a real gap — verified the exact path: with the row named `old-model` and the env model changed to `new-model`, `_repo.get("new-model", ...)` returns `None`, then the `if existing:` guard (don't create a competing default) short-circuits, so nothing syncs and nothing is created. The old endpoint silently stays in use.

I have **not** implemented the provenance fix here, deliberately. These three commits are a cherry-pick of @ThibautChoppy's work from #748 onto `develop`, kept under his authorship; persisting an env-managed marker means a column + Alembic migration and changes the "never touch anything but the slug-matched row" guarantee the feature is built around. That is a design call for the author, not something to slip into a cherry-pick — and it is not a regression, since changing the model did nothing before this PR either.

What I did instead, in ab52bfc2:
- Scoped the doc claim, which said the endpoint could be managed "purely via env vars". It now states plainly that sync covers URL/batch size/timeout and that **changing the model does not take effect**, with switching models still requiring an admin edit.
- Added `test_seed_defaults_sync_on_boot_does_not_follow_a_changed_model_slug`, pinning the current behaviour as a known limitation so it cannot change silently — and so it fails loudly for whoever implements the provenance fix.

Leaving this thread open as the tracking point for that follow-up.

openrag/services/inference/vllm_client.py
```diff
@@ -276,9 +297,15 @@ async def _run(batch: list[str]) -> list[list[float]]:
...
-                "Embedding failed after {d}/{b} batches", d=done, b=len(batches)
-            )
+            # `.extra` (set on OpenRAGError subclasses like EmbeddingAPIError) carries
+            # the embedder's actual response body — repr(exc) alone only has the
+            # generic "Embedder API error (400)" message, not why it was rejected.
+            logger.bind(
+                batches_done=done,
+                n_batches=len(batches),
+                error=repr(exc),
+                error_detail=getattr(exc, "extra", None),
```
hedhoud: Could we avoid logging `exc.extra` directly here? It contains the provider's raw response and, for a 400, snippets from the indexed document. A failed upload with sensitive text would copy that content into container and centralized logs. Logging the status, model, suspect count, and positions should keep the diagnostic useful without exposing document content.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,22 +101,55 @@ async def seed_defaults(self) -> None:
...
+                await self._backfill_sampling_params(existing, getattr(self._config, model_type))
+
             endpoint: str = data["endpoint"]
             model_name: str = data["model_name"]
             if not endpoint:
                 logger.info(f"No {model_type} endpoint configured — skipping seed.")
                 continue
+
+            name = _slug(model_name or "")
+            existing_row = await self._repo.get(name, model_type)
```
hedhoud: This lookup works only while the configured model keeps the same slug. For example, changing `old-model` to `new-model` in Helm leaves the old database row in place, so the rollout still uses the old model. We need a stable way to mark and find the env-managed row, then rename and update that row when the model changes.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,22 +101,55 @@ async def seed_defaults(self) -> None:
...
+            if existing_row is not None:
+                if sync_on_boot:
+                    await self._repo.update(
+                        name,
+                        model_type,
+                        endpoint=endpoint,
+                        model_name=model_name or None,
+                        batch_size=data.get("batch_size", 32),
+                        timeout=data.get("timeout", 30.0),
```
hedhoud: One more env-sync case: API keys live in `extra`, so rotating `EMBEDDER_API_KEY` or `API_KEY` and rolling the pod keeps the old database key. Once the provider disables the old key, requests will fail even though the new env value is present. Could we sync the key only for rows explicitly marked as env-managed, while preserving extras on admin-managed rows?

hedhoud (commented):
I found three cases that should be addressed before this is merged.

openrag/services/inference/vllm_client.py
Ahmath-Gadji: Agreed, and done in f03dd041 — the log now carries a summary instead of `.extra`:

```
model_name, base_url, status_code, suspect_count, suspect_indices
```

So an operator still learns *which* chunks are malformed (e.g. `suspect_indices=[3, 7]`) and can go inspect them, but no document text reaches the log pipeline. The full detail stays on the `EmbeddingAPIError`, so it still reaches the uploader on the API error — an audience that already owns the document. Pinned by `TestEmbedderFailureLogOmitsDocumentText`, which asserts the fake PII in the snippets never appears in the logged dict.

I also dropped the `redact_secrets()` call I had added earlier in this thread. You were right to push past it: it masks values under secret-*named* keys, and this payload's keys are `model_name` / `base_url` / `error` / `suspect_texts` — so it passed the document snippets straight through unchanged. It read like a mitigation without being one. Verified it was a literal no-op on this payload before removing it.

One deliberate call worth flagging: I took your list literally and dropped the **provider's raw response** from the log too, not just the snippets, since it can echo the rejected input back. The cost is that @ThibautChoppy's original commit here existed precisely to surface *why* the provider rejected the batch, and the log no longer answers that — you get the status, not the message. If you would rather keep the provider message in logs (bounded to 500 chars, as it now is), say so and I will add it back as a follow-up; I did not want to silently trade your requirement away.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Confirmed, and I think this is the more serious of the two — thanks for catching it. Verified against the code: the sync path calls

```python
await self._repo.update(name, model_type, endpoint=..., model_name=..., batch_size=..., timeout=...)
```

and never passes `extra`. The key lives in `extra` (seeded via `_with_api_key(...)`), so rotating `EMBEDDER_API_KEY` / `API_KEY` and rolling the pod leaves the old key in the row. That is worse than the slug gap: the slug gap is a silent no-op, whereas this is a **silent outage** — once the provider disables the old key, embedding fails while the correct value is sitting right there in the environment.

Worth being explicit that the omission was deliberate, not an oversight: never touching `extra` is exactly what stops sync from clobbering a hand-set key on an admin-managed row. So the two requirements are in direct tension, and your suggestion is the way out of it — sync `extra` only for rows explicitly marked env-managed, preserve it everywhere else. That is the same marker this thread's sibling needs to follow a changed slug, so one piece of design covers both.

That means a new column plus an Alembic migration (idempotent, per this repo's `create_all`-then-migrate startup), which is a real change to the shape of @ThibautChoppy's feature rather than something I would fold into a cherry-pick unannounced. I am checking how we want to sequence it — implement it in this PR, or land the diagnostics + opt-in sync here with the limitations pinned and do the marker as a focused follow-up with Thibaut. Either way this should be tracked before `MODEL_ENDPOINT_SYNC_ON_BOOT` is recommended to operators, since the failure mode is an outage. I will not merge this past your review while it is open.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Agreed on the diagnosis — I traced the same path and pinned it in `ab52bfc2` as `test_seed_defaults_sync_on_boot_does_not_follow_a_changed_model_slug`: with the row named `old-model` and the env model changed to `new-model`, `_repo.get("new-model", ...)` returns `None`, then the `if existing:` guard (don't create a competing default) short-circuits, so nothing syncs and nothing is created.

I had documented it as a known limitation rather than fixing it, on the grounds that it is not a regression (changing the model did nothing before this PR either) and that a cherry-pick is the wrong place to redesign the author's feature. Your review changes that calculus, and combined with the API-key point in the sibling thread I no longer think shipping it as a documented limitation is the right call — the two together are what an operator would reasonably expect `MODEL_ENDPOINT_SYNC_ON_BOOT=true` to cover.

Both need the same thing you describe: a persisted env-managed marker on the row, so the seeder can find *its* row independently of the current slug, then rename and update it (and sync `extra` for that row only). Concretely a new column + an idempotent Alembic migration, and a deliberate loosening of the "only ever touch the slug-matched row" guarantee the feature is currently built around — that guarantee is doing real work today, so replacing it with the marker needs to be explicit rather than incidental.

Sequencing is the open question: do it in this PR, or land the diagnostics + opt-in sync here (limitations pinned by tests) and do the marker as a focused follow-up with @ThibautChoppy, who owns the design. I have a slight preference for the follow-up so his feature is not redesigned inside a cherry-pick of it, but I do not feel strongly and I would rather match how you two want it handled. Happy to implement it here if you prefer — say the word and I will.

Ahmath-Gadji: Implemented in adc284f2 — you were right that this needed the marker rather than a documented limitation.

The seeder now stamps `managed_by: "env"` into the endpoint's `extra` and finds its row by that marker instead of by name, so a model change re-points and renames the same row rather than matching nothing. A row seeded before the marker existed is adopted by matching the slug once, so upgraded deployments are not stranded.

One deviation worth calling out: I put the marker in the existing `extra` JSON instead of adding a column, so **there is no Alembic migration**. `extra` already carries a control key (`implementation`) that `factories.py` strips before splatting the rest into a client constructor; I extended that single-key exclusion into a `_CONTROL_EXTRA_KEYS` frozenset so the marker cannot leak in as an unexpected kwarg — the #712 `batch_size` failure mode. If you would rather have this as a first-class column, say so and I will redo it as a migration; the JSON marker was the cheaper way to get the behaviour you asked for.

Guards, since this loosens the old "never touch anything but the slug-matched row" rule:
- Hand-created endpoints carry no marker and are never modified, renamed or deleted.
- The rename is **skipped** (with a warning) if the target name is already taken, so sync can never collide with an endpoint you created.

Covered by `test_seed_defaults_sync_on_boot_follows_a_changed_model_slug`, `..._adopts_a_row_seeded_before_the_marker` and `..._will_not_rename_over_a_hand_created_name`. The test that previously pinned the limitation is gone, replaced by these.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Also implemented in adc284f2, using the same marker. Sync now writes `extra` for the env-managed row and refreshes `api_key` from env, so rotating `EMBEDDER_API_KEY` / `API_KEY` and rolling the pod actually reaches the database. `extra` is **merged, not replaced** — the marker and the key come from env, and every other key an admin put there (sampling params, custom kwargs) is preserved. Rows without the marker still have their `extra` left completely alone, which is the split you suggested.

One thing worth flagging, because it nearly went in wrong: your existing test `test_seed_defaults_syncs_env_named_row_when_sync_on_boot_enabled` caught a real bug in my first attempt. My initial guard was `if env_api_key:` — but the embedder and reranker defaults are the literal string `EMPTY` (`endpoints.py:50`, `config.yaml:52`), which is truthy. Simply switching `sync_on_boot` on would therefore have **overwritten a real hand-set key with the `EMPTY` placeholder**. That is a worse failure than the one this thread is about, and it only surfaced because that assertion existed — so thanks for it.

The guard is now an explicit sentinel set:

```python
_PLACEHOLDER_API_KEYS = frozenset({"", "EMPTY"})
```

so a real rotation applies, while an unset or placeholder `*_API_KEY` can never blank a working credential. I updated that test rather than deleting it: it now asserts the hand-set key still survives when env has no real key to offer, plus the new marker. New coverage in `test_seed_defaults_sync_on_boot_rotates_the_api_key`.

Full unit suite 2072 passing, ruff and the layer-import guard clean.

openrag/services/inference/vllm_client.py
```diff
@@ -332,12 +398,22 @@ async def _embed_batch(self, texts: list[str]) -> list[list[float]]:
...
+            extra: dict = {
+                "model_name": self._model,
+                "base_url": self._endpoint,
+                "error": exc.response.text[:500],
+            }
+            # A 400 usually means the embedder rejected the payload itself;
+            # pinpoint the offending chunk(s) so operators know which input to
+            # inspect (observed: literal `\u…` escapes choking a Go gateway).
+            if exc.response.status_code == 400:
+                suspect_texts = _find_suspect_escapes(texts)
```
hedhoud: Could we make this index document-global? When `embed()` splits a large file into batches, this helper only sees the current batch, so a bad chunk at original index 97 can be reported as 1 with `batch_size=32`. That makes the diagnostic point operators to the wrong chunk. Passing the batch start offset into the suspect-finding path, plus a multi-batch test, would keep this reliable.

hedhoud (commented):
Left one comment about the suspect chunk index diagnostic.

openrag/services/inference/vllm_client.py
Ahmath-Gadji: Good catch, and it mattered more than it looks. Fixed in 95fb04c0.

You are right about the mechanism: `_embed_batch` only ever sees one batch, so `enumerate` gave a batch-local position — chunk 97 reported as 1 under `batch_size=32`, and every batch reported the same 0..31 range.

What makes it worse is my own earlier commit in this review. When I stopped logging the snippets and kept only the positions, those positions became the *only* pointer to the bad chunk — so there was no longer any snippet to cross-check against, and the log just pointed at an innocent chunk with full confidence. A confidently wrong diagnostic is worse than no diagnostic, so this was effectively a bug I introduced the severity of.

The fix costs almost nothing, because the offset was already there and simply discarded:

```python
batches = [texts[i : i + self._batch_size] for i in range(0, len(texts), self._batch_size)]
#                                                        ^ already the document-global offset
```

It is now kept and threaded through `_run` → `_embed_batch(..., offset=)` → `_find_suspect_escapes(..., offset=)`, with `offset=0` defaults so the single-batch path and existing callers are unaffected.

Tests are as you asked — a multi-batch one through `embed()` with the bad chunk at index 97 and `batch_size=32`, plus direct offset tests. I mutation-checked both: removing the offset makes them fail with exactly `1 != 97`, so they genuinely pin the behaviour rather than passing by construction. Full unit suite 2075 passing, ruff and layer guard clean.

One limitation worth stating rather than implying: this is an index into the list handed to `embed()`, not a chunk ID, so it still has to be mapped back to a chunk. Correct and well-defined beats silently wrong, but it is not a durable identifier — happy to look at threading a real chunk ref through if you think the diagnostic warrants it.

Ahmath-Gadji:
### Closing and immediately reopening to force a head resync — not abandoning this PR

GitHub has this PR stuck two commits behind its own branch:

| Source | SHA |
|---|---|
| `git ls-remote refs/heads/feat/embedder-diagnostics-and-endpoint-sync` | `95fb04c0` |
| `compare develop...feat/embedder-diagnostics-and-endpoint-sync` | ahead by **9**, tip `95fb04c0` |
| **this PR's `head.sha`** | **`f03dd041`, `commits=7`** |

So the branch really does have every commit — the PR object just has not advanced. Two consequences worth knowing while it lasts: the diff shown above is stale, and **CI has not run against the last two commits**, because checks trigger off the PR head. Please do not judge the code from the rendered diff until this settles.

The two commits GitHub is not showing are the ones answering @hedhoud's review:

- `adc284f2` — identify the env-managed endpoint by a marker instead of the model slug, so a model change re-points the row and a rotated API key actually reaches the DB.
- `95fb04c0` — report the suspect chunk's document position rather than its batch-local one.

A second push and a PATCH to the PR both failed to shift it, so close/reopen is the remaining lever. Reopening right away; review comments and threads are unaffected.

Apologies for the notification noise, @hedhoud.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,35 +110,120 @@ async def seed_defaults(self) -> None:
...
+
+        await self._repo.update(
+            row.name,
+            model_type,
+            endpoint=data["endpoint"],
+            model_name=data["model_name"] or None,
+            batch_size=data.get("batch_size", 32),
+            timeout=data.get("timeout", 30.0),
+            extra=new_extra,
+        )
```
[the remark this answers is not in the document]
Ahmath-Gadji: Valid — fixed in d28e287d. Verified the mechanism rather than taking it on trust: `_build_default_seeds` supplies a `batch_size` only for `embedder`; the `llm`, `vlm` and `reranker` seed dicts have none, so `data.get("batch_size", 32)` was a constant 32 for those three and every sync pass would have reset an admin-tuned row.

Took the suggested fallback to `row.batch_size`, and applied the same to `timeout`. `timeout` is not currently broken — all four seeds supply one, so that literal was dead code — but it is the identical latent bug the moment a seed stops supplying it, and the asymmetry would have been an easy thing to reintroduce.

The underlying rule this encodes: sync should only write fields the environment actually owns, and fall back to what the row already holds for everything else — never to a literal, which silently asserts a value nobody configured.

Covered by `test_seed_defaults_sync_on_boot_keeps_admin_batch_size_for_non_embedder_types` (an llm row with `batch_size=256`). Mutation-checked: restoring the `32` literal makes it fail, so it pins the behaviour rather than passing by construction. Full unit suite 2076 passing, ruff clean.

Ahmath-Gadji:
### Review threads resolved — two open questions lifted out so they are not buried

All five threads are addressed and resolved. Summary of what landed:

| Finding | Fix |
|---|---|
| @hedhoud — `exc.extra` logged wholesale (provider body + document snippets) | `f03dd041` — log a summary (model, status, suspect count, positions); no document text reaches the log pipeline |
| @hedhoud — sync keyed on the model slug, so a model change did nothing | `adc284f2` — find the row by a `managed_by` marker in `extra`; re-point and rename it |
| @hedhoud — rotated `*_API_KEY` never reached the DB (silent outage) | `adc284f2` — sync `extra` for marked rows only; `EMPTY`/unset never overwrites a hand-set key |
| @hedhoud — suspect index was batch-local, pointing at the wrong chunk | `95fb04c0` — thread the batch offset through so the position is document-global |
| CodeRabbit — `batch_size` fallback clobbered admin values on llm/vlm/reranker | `d28e287d` — fall back to the row's own value, never a literal |

Every fix is pinned by a test, and the three behavioural ones were mutation-checked (reverting the fix makes the test fail), so they pin behaviour rather than passing by construction.

**Two decisions are still genuinely open — resolving the threads should not be read as settling them:**

1. **Provider error message in logs.** Following @hedhoud's list I dropped the provider's raw response from the log, not just the snippets. That removes what @ThibautChoppy's original commit existed to surface: the log now says a batch failed with a 400, but not *why* the provider rejected it. The full text still reaches the uploader on the API error. Happy to put it back, bounded to 500 chars, if you would rather keep that diagnostic.

2. **Marker as JSON key vs. column.** `managed_by` lives in the existing `extra` blob, which is why this needed no Alembic migration. A first-class column would be more explicit and queryable. I went with the cheaper option to keep a cherry-pick from turning into a schema change; say the word if you want it promoted.

Also for the record: the `tests (3.12)` failure on this branch was `test_dispatcher.py::test_delete_file_rechecks_ref_less_task_before_marking_stale`, a 10 ms-timeout test unrelated to any file here (these commits touch nothing under `services/workers/`). It passes 8/8 locally; I re-ran the job.

openrag/di/factories.py
```diff
@@ -86,7 +94,7 @@ def factory(name: str = "default") -> T:
                 raise KeyError(f"Unknown model '{name}'. Available: {list(config_section)}")
             # `implementation` is a control key (which class to build), not a
             # constructor argument, so it is read out before splatting `extra`.
-            impl_kwargs = {k: v for k, v in model_cfg.extra.items() if k != "implementation"}
+            impl_kwargs = {k: v for k, v in model_cfg.extra.items() if k not in _CONTROL_EXTRA_KEYS}
```
hedhoud: This filters `managed_by` in the API factory, but the Ray worker factories still remove only `implementation`. Because every newly seeded endpoint now contains this marker, VLM, contextualization, and topic-tagging requests can send `managed_by: "env"` to the provider. Strict OpenAI-compatible services may reject those indexing requests with a 400. Could we reuse the same control-key filter in the worker factories and add a request-body test?

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,35 +110,124 @@ async def seed_defaults(self) -> None:
...
             if not endpoint:
                 logger.info(f"No {model_type} endpoint configured — skipping seed.")
                 continue
+
+            name = _slug(model_name or "")
+            # The marker survives a model change; the slug does not, so look for
+            # the marker first and fall back to the slug only to adopt a row
+            # seeded before the marker existed.
+            managed_row = next((r for r in existing if r.extra.get(ENV_MANAGED_KEY) == ENV_MANAGED_VALUE), None)
+            existing_row = managed_row or await self._repo.get(name, model_type)
```
hedhoud, line 179: This fallback cannot tell an older seeded row from an endpoint an admin created with the same name. For example, a hand-created `embed-model` row without the marker is adopted here and its URL, model, batch size, and timeout are overwritten on the first synced restart. Could we require an explicit marker or a one-time migration instead of adopting every row that matches the slug?

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,35 +110,124 @@ async def seed_defaults(self) -> None:
...
+            row.name,
+            model_type,
+            endpoint=data["endpoint"],
+            model_name=data["model_name"] or None,
+            # Fall back to what the row already holds, never to a literal:
+            # _build_default_seeds only supplies batch_size for `embedder`, so a
+            # hardcoded default would silently reset an admin-tuned llm/vlm/
+            # reranker row to 32 on every boot. Same reasoning for timeout.
+            batch_size=data.get("batch_size", row.batch_size),
+            timeout=data.get("timeout", row.timeout),
```
hedhoud: The timeout fallback does not currently preserve an admin value because every seed already contains the config default. I reproduced an LLM row with timeout `99` and no timeout env override; after sync it became `60`. Could we track whether the timeout was explicitly supplied by env and only overwrite it in that case? The test should assert the preserved timeout too.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -101,35 +110,124 @@ async def seed_defaults(self) -> None:
...
+        # row named after the previous model. Skipped when the name is taken, so
+        # sync can never collide with a hand-created endpoint.
+        renamed_to = None
+        if target_name and target_name != row.name:
+            if any(r.name == target_name for r in existing):
+                logger.warning(
+                    f"Not renaming {model_type} endpoint '{row.name}' to '{target_name}': name already in use."
+                )
+            else:
+                await self._repo.rename(row.name, model_type, target_name)
```
hedhoud: Renaming this row can leave stored endpoint references behind. For example, a partition using embedder `old-model`, or a preset using that VLM/LLM name, still points to `old-model` after the row becomes `new-model`, so the next indexing job cannot resolve it. Keeping the registry name stable would be the smallest fix; otherwise every persisted reference needs to move atomically with the rename.

hedhoud (commented):
I found a few cases that still need attention before approval. The existing CodeRabbit finding about the sampling backfill is also still valid: sync rebuilds `extra` from the old row and removes the values that were just backfilled.

openrag/di/factories.py
Ahmath-Gadji: Confirmed and fixed in 4b8cd17b — I had only plugged one of two holes. `indexer_pool.py` has **four** such factories (lines 664, 715, 794, 856), each stripping just `implementation`, so VLM, contextualization and topic-tagging would all have forwarded `managed_by: "env"` to the provider. Exactly the #712 failure mode, and the 400 would point at nothing resembling the cause.

Rather than repeat the literal a fifth time, the control-key set now lives in `core/config/model_endpoints.py` as `CONTROL_EXTRA_KEYS`; `di/factories.py` and all four worker factories import it, so this cannot drift again. `core` is importable from both layers, so the layer-import guard stays green.

Request-body test added as you asked: `test_worker_factories_do_not_forward_the_env_managed_marker` builds a topic tagger through the real factory and asserts the marker never reaches the client kwargs while genuine kwargs (`temperature`) still do. Mutation-checked — restoring `!= "implementation"` fails it.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: You are right, and my earlier claim that this was only a latent issue was wrong — thanks for actually reproducing it. Fixed in 4b8cd17b.

The flaw was assuming the seed dict tells you whether env supplied a value. It never did: `_build_default_seeds` always fills `timeout` from `s.<type>.timeout`, so `data.get("timeout", row.timeout)` could not fall back and wrote the config default over the admin value — your 99 → 60. Worse for `llm`, which has no timeout env var at all, so env can never legitimately own that field.

Sync now consults an explicit ownership table:

```python
_ENV_OWNED_TUNABLES = {
    "embedder": {"batch_size": "EMBEDDER_BATCH_SIZE", "timeout": "EMBEDDER_TIMEOUT"},
    "vlm": {"timeout": "VLM_TIMEOUT"},
    "reranker": {"timeout": "RERANKER_TIMEOUT"},
    "llm": {},
}
```

and writes such a field only when that env var is actually set. `endpoint`/`model_name`/`api_key` remain unconditional — those are what env identity means.

`test_seed_defaults_sync_on_boot_keeps_admin_timeout_when_no_env_var_set` asserts the preserved timeout on exactly your scenario (llm row at 99, no override). Mutation-checked.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Agreed — fixed in 4b8cd17b. The slug fallback existed so deployments upgraded from a build without the marker would not be stranded, but as written it could not tell an old seed from an endpoint an admin happened to name after the model, and would overwrite it on the first synced restart.

Rather than require a migration, adoption now additionally requires the row to still match the environment exactly (`endpoint` and `model_name`):

```python
def _is_unmodified_seed(row, data) -> bool:
    return row.endpoint == data["endpoint"] and (row.model_name or "") == (data["model_name"] or "")
```

That makes adoption safe by construction rather than by heuristic: if the row still matches env, taking ownership changes nothing that was not already true; if an admin has touched it at all, it is never adopted and never written. A marked row is always synced regardless, so this check only gates the one-time legacy path.

The trade-off, stated plainly: a genuinely-seeded row that has since drifted (someone edited the URL by hand) will not be auto-adopted either — it needs the marker set once. I preferred that over any chance of silently overwriting an admins endpoint. If you would rather have the explicit migration instead, I am happy to swap it.

`test_seed_defaults_sync_on_boot_does_not_adopt_a_modified_row` covers it; mutation-checked.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Agreed, and this reverses the direction we settled on earlier in the review — you asked for rename-and-update, and you are right that it does not survive contact with stored references. Fixed in 4b8cd17b by taking the smallest option you name: **the row is no longer renamed at all.**

I had checked that the admin rename path has the same gap and concluded it was acceptable precedent. That was the wrong read. The admin case is a deliberate human action on one endpoint; boot-time sync would do it silently, on every rollout, to whichever row env points at — the same defect is materially worse when it is automatic.

So the name is now a stable identifier: after a model change the endpoint keeps its original name and serves the new model. Nothing in `partitions.chat_llm` or presets can dangle, and no atomic multi-table move is needed.

The cost is cosmetic — a row named `old-model` serving `new-model` reads oddly in the admin UI. Documented in `env_vars.md` rather than left as a surprise. If that bothers us later, the clean fix is to stop deriving the seed name from the model at all (a constant like `env-managed`), which sidesteps the mismatch instead of chasing it with renames — happy to do that as a follow-up if you like it.

`test_seed_defaults_sync_on_boot_follows_a_changed_model_slug` now asserts the row is re-pointed and *not* renamed.

hedhoud (dismissed):
Re-checked the current head. The previously raised issues are addressed, focused tests passed locally, and the remaining note is non-blocking. Approving.

tests/unit/services/orchestrators/test_model_endpoint_service.py
```diff
@@ -421,6 +422,360 @@ async def test_seed_defaults_skips_reranker_when_unconfigured(monkeypatch):
...
+        models={"sync_on_boot": True},
+    )
+    svc = _make_service(repo, settings=settings)
+
+    await svc.seed_defaults()
+
+    updated = repo._store[("embed-model", "embedder")]
+    # EMBEDDER_BATCH_SIZE is not set, so env does not own batch_size here and the
+    # admin's 512 stands; endpoint/model_name are always env-owned and do sync.
+    assert updated.batch_size == 512
```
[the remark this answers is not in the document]
Ahmath-Gadji: Valid — fixed in 8c57297c. The assertion claims an admin's `batch_size` survives *because env does not own it*, so it is only meaningful when the variable is genuinely unset; clearing only `LLM_*` left that premise up to whatever the runner happened to export.

Confirmed it rather than taking it on trust: with `EMBEDDER_BATCH_SIZE=64` exported the test fails `assert 64 == 512` before the change and passes after.

Two deviations from the suggestion:

1. **Placement.** The committable suggestion puts the `delenv` next to the assertion, which is after `seed_defaults()` has already run — too late to affect anything. It has to sit with the other `delenv` calls, before the service executes. The diff in your summary had it right; only the auto-generated suggestion block was off.
2. **Also cleared `EMBEDDER_TIMEOUT`.** It is the other env-owned tunable for this type (`_ENV_OWNED_TUNABLES["embedder"]`), so the identical trap is waiting for whoever next asserts a preserved timeout here.

I also swept the rest of the file for the same hazard: the other `batch_size == …` assertions are on rows that are never synced (sync_on_boot off, hand-created, or deliberately not adopted), so they were already hermetic. Verified by running the whole file with `EMBEDDER_BATCH_SIZE`, `EMBEDDER_TIMEOUT`, `VLM_TIMEOUT` and `RERANKER_TIMEOUT` all exported — 51 passed, same as with a clean environment.

(The ast-grep `http://` note on that line is a test fixture URL, not a real endpoint — leaving it.)

hedhoud (approved):
Re-checked the updated head after the merge. The follow-up changes preserve the reviewed behavior, and the focused test set passed locally. Approving.

PR #760, opened 2026-07-23
fix(storage): fail closed on an empty partition scope (cross-partition leak)

### LINKED ISSUE #759: Cross-partition data leak: empty partition scope fails open (searches all partitions)
## Summary

`MilvusVectorStore._build_filter_expr` (`openrag/services/storage/milvus_store.py`) **fails open** on an empty partition scope: `{"partition": []}` produces an empty filter expression (`""`), so the search runs **unfiltered across every partition** instead of matching nothing. This is a cross-tenant data-isolation regression vs. the pre-refactor (v1.1.x) line, which failed closed.

## Impact (high — cross-partition read)

A query that resolves to **no accessible partition** returns **every** partition's rows. Concretely, a user who is a member of zero partitions (or whose scope otherwise resolves to `[]`) hitting `openrag-all` gets other tenants' documents back, and the answer LLM cites them. Observed: a query scoped to `tenant-b` returned a document belonging to `tenant-a`, where pre-v2 the system correctly reported "the documents do not contain the answer."

## Root cause (verified)

`_build_filter_expr` builds the partition clause only when the list is truthy:

```python
partition = filters.pop("partition", None)
if isinstance(partition, (list, tuple)):
    has_wildcard = any(p in self._PARTITION_WILDCARDS for p in partition)
    if has_wildcard and len(partition) > 1:
        raise ValueError("`partition` cannot mix wildcard with explicit values.")
    if not has_wildcard and partition:      # <-- empty list is falsy → clause SKIPPED → unfiltered
        quoted = ", ".join(self._format_value(p) for p in partition)
        parts.append(f"partition in [{quoted}]")
```

An empty `partition` list is falsy, so the clause is dropped and the expression is empty → Milvus searches all partitions.

There is a telling asymmetry: an empty **non-partition** IN-list already short-circuits to match-nothing (`_MATCH_NOTHING_EXPR`, `1 == 0`), but the **security-critical partition** key is the one that fails open.

## Regression vs. v1.1.x

Pre-refactor failed **closed** (`openrag/components/indexer/vectordb/vectordb.py`):

```python
if partition != ["all"]:
    expr_parts.append(f"partition in {partition}")   # [] -> "partition in []" -> matches NOTHING
```

So `partition == []` matched nothing in v1.1.x; in v2.0.0 it matches everything. This is the exact behavioral flip that turns "no results" into "another tenant's results".

The current fail-open behavior is also **enshrined in a test** (`tests/unit/services/storage/test_milvus_store.py`):

```python
def test_empty_partition_list_is_skipped(...):
    # No partitions means no partition clause (not match-nothing).
    assert store._build_filter_expr({"partition": []}) == ""
```

## Proposed fix

Fail **closed**: an empty (non-wildcard) partition list must short-circuit to `_MATCH_NOTHING_EXPR` (match nothing), and dominate any other filter present. Keep the explicit `"all"` / `["all"]` wildcard as the *only* way to run an intentionally unscoped search. Update the enshrined test to assert `1 == 0` (it then revert-proves the fix), and add empty-tuple + "empty-scope-ignores-other-filters" cases.

Defense-in-depth follow-ups to consider separately:
- Harden the resolution layer so an empty `user_partitions` on `openrag-all` never reaches the store as an unscoped search.
- Separately: `vector_store_searcher.search`/`multi_query_search` accept but never read `filter_params`, so workspace file-scoping (#706) does not reach Milvus — within-partition over-retrieval, not this cross-partition leak, but worth its own issue.

## Severity
High — cross-partition (cross-tenant) read of document content.

andyne13:
Closes #759.

## Problem

`MilvusVectorStore._build_filter_expr` **failed open** on an empty partition scope: `{"partition": []}` dropped the partition clause and returned an empty expression, so the search ran **unfiltered across every partition** — a cross-tenant read. A scope that resolves to no accessible partition (e.g. a user with zero memberships hitting `openrag-all`) got other tenants' documents back, where the pre-refactor line failed **closed** (`partition in []` matched nothing). The fail-open was even enshrined in a test.

## Fix

An empty (non-wildcard) partition list/tuple now short-circuits to `_MATCH_NOTHING_EXPR` (`1 == 0`) — fail closed — and dominates any co-filter (early return before `expr`/other keys are read). The explicit `"all"` wildcard remains the only way to run an intentionally unscoped search.

- One shared chokepoint: `_build_filter_expr` is the sole filter builder for **all** read/delete paths (dense + hybrid search, scalar query, delete), so the guarantee holds everywhere. Verified the hybrid path applies the expr to **both** the dense and BM25 sparse legs.
- `None`/missing partition is deliberately left unscoped: the search routers already 403 an empty scope upstream (`get_partition_name`, `search.py`), and the partition-less path is required by the authz-gated global-`_id` lookup in `ConversionService.get_chunk` (extractText/download). Closing it there would break that flow with no isolation benefit.
- No regression: the "everything" wildcard is `["all"]`, never `[]`; no caller passes an empty partition list meaning all. `delete_by_filter({"partition": []})` now safely no-ops instead of erroring.

## Tests

Rewrote the test that had enshrined the fail-open into a revert-prove (`test_empty_partition_list_matches_nothing` asserts `1 == 0`; reverting the fix yields `""` and fails). Added empty-tuple and two co-filter-dominance cases (`expr` and `file_id`).

- Revert-prove confirmed (3 tests fail on the reverted code).
- `uv run pytest tests/unit/` green; full `test_milvus_store.py` (78) green; ruff check/format + layer-import-guard clean.

## Scope / follow-ups (separate, noted in #759)

- Defense-in-depth: reject an empty `user_partitions` scope even earlier at the resolution layer.
- Unrelated but real: `vector_store_searcher.search`/`multi_query_search` accept but never read `filter_params`, so #706 workspace file-scoping doesn't reach Milvus (within-partition over-retrieval, not this cross-partition leak).

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty partition filters now safely return no results instead of unintentionally querying across all partitions.
  * This behavior is enforced even when additional search conditions are provided.
* **Documentation**
  * Clarified how wildcard and explicitly specified partitions are handled.

### DIFF
--- a/openrag/services/storage/milvus_store.py
+++ b/openrag/services/storage/milvus_store.py
@@ -469,11 +469,14 @@ class MilvusVectorStore(VectorStore):
 
         Rules:
             * ``filters['partition']`` builds the partition_key clause.
-              Value ``"all"`` (or ``["all"]``) skips the clause. List/tuple
-              values become ``partition in [...]``. Mixing a wildcard with
-              explicit partitions in the same list raises ``ValueError`` —
-              that combination is rejected rather than silently widened to
-              every partition.
+              Value ``"all"`` (or ``["all"]``) is the explicit wildcard and
+              skips the clause. A non-empty list/tuple becomes
+              ``partition in [...]``. An **empty** list/tuple means "no
+              accessible partition" and short-circuits to
+              :data:`_MATCH_NOTHING_EXPR` (fail closed) — it must never widen
+              to every partition. Mixing a wildcard with explicit partitions
+              in the same list raises ``ValueError`` — that combination is
+              rejected rather than silently widened to every partition.
             * ``filters['expr']`` is appended verbatim as an escape hatch
               for callers that need operators the dict form cannot express.
             * Any other key with a scalar value becomes ``key == <literal>``.
@@ -494,7 +497,18 @@ class MilvusVectorStore(VectorStore):
             has_wildcard = any(p in self._PARTITION_WILDCARDS for p in partition)
             if has_wildcard and len(partition) > 1:
                 raise ValueError("`partition` cannot mix wildcard with explicit values.")
-            if not has_wildcard and partition:
+            if has_wildcard:
+                pass  # explicit "all" wildcard → intentionally unscoped, no clause
+            elif not partition:
+                # SECURITY (fail closed): an empty partition list means the
+                # caller resolved to *no* accessible partition (e.g. a user
+                # with zero memberships hitting `openrag-all`). It must match
+                # NOTHING, never every partition. Failing open here dropped the
+                # clause entirely and leaked cross-tenant rows — a query scoped
+                # to one partition returning another tenant's chunks. Restore
+                # the pre-refactor behaviour (`partition in []` matched nothing).
+                return self._MATCH_NOTHING_EXPR
+            else:
                 quoted = ", ".join(self._format_value(p) for p in partition)
                 parts.append(f"partition in [{quoted}]")
         elif partition is not None and partition not in self._PARTITION_WILDCARDS:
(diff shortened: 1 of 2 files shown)


hedhoud (approved):
LGTM. The empty partition scope now fails closed, and the focused regression coverage looks good.

PR #761, opened 2026-07-23
allowing scope retrieval via a list of file_ids

ewan102:
lists sent by a POST /v1/chat/completions, used a lot of the workspace filter mechanism, adding a lists of the file_id effectively indexed on the extra.attachments, in order to know which one of the file_ids' lists were not indexed in RAG

## Summary by CodeRabbit

* **New Features**
  * Added support for Cozy-style chat attachments: attachments are resolved to validated, indexed file IDs and applied to retrieval.
  * Chat responses and streamed SSE metadata now include attachment usage when applicable.
  * Stream payloads can now include additional client-facing extra fields alongside source/truncation info.
* **Bug Fixes**
  * Malformed, empty, or unindexed attachments safely fall back to unscoped retrieval; workspace scoping takes precedence.
  * `partitions=["all"]` uses union-style resolution, while mixing `"all"` with explicit partitions raises an error.
* **Tests**
  * Expanded unit and integration coverage for attachment parsing, scoping, precedence, de-duplication, and empty-input behavior.


openrag/services/orchestrators/query_service.py
```diff
@@ -645,6 +693,20 @@ def _summary_doc(chunk, summary: str):
     return chunk.__class__(page_content=summary, metadata=chunk.metadata)
 
 
+def _extract_attachment_ids(metadata: dict) -> list[str]:
+    """Extract the file_id allowlist from a chat request's attachments.
+
+    Cozy (via cozy-stack) sends ``metadata.attachments = [{"id": "<file_id>"},
```
paultranvan: We could have other clients, there is no dependency to cozy or any client

openrag/services/orchestrators/query_service.py
```diff
@@ -411,6 +413,16 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
             # partition the caller also has access to (#706).
             partition = [scope.partition]
             filter_params = {"file_id": scope.file_ids}
+        elif attachment_ids and partition:
+            # Client-scoped chat on specific files (Cozy attachments). Mirror the
```
paultranvan: Please don't mention Cozy, we could have other clients

openrag/services/orchestrators/query_service.py
```diff
@@ -411,6 +413,16 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
...
+            # workspace flow: a Postgres catalog lookup resolves the real indexed
+            # subset, which both drives the filter and is reported back to the
+            # client in `extra` (see chat/chat_stream). A non-indexed id is simply
+            # left out. The file_id filter is ANDed with the partition filter
+            # downstream and the partition is fixed server-side, so no DB-side
+            # security check is needed — this lookup is purely resolution/reporting.
```
paultranvan: Tihs long comment is way too much focusing on the "what", which we don't really care.

openrag/services/orchestrators/query_service.py
```diff
@@ -424,8 +436,19 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
...
+        if attachment_ids and not chunks:
+            # Diagnostic: attachments were requested but the file_id filter matched
+            # nothing. Usually a file_id/partition mismatch or an eventual-consistency
+            # race (file attached before its indexing reached COMPLETED). The chat
+            # still proceeds without context.
+            logger.warning(
+                "Attachments specified but no chunks matched — check file_id/partition mapping",
+                file_ids=attachment_ids,
+                partition=partition,
+            )
```
paultranvan: We should do nothing for such case: we found no chunk, so the query was not relevant for this set of attachments, it's not different from the case where we searched for the whole partition and found nothing

openrag/services/orchestrators/query_service.py
```diff
@@ -580,7 +620,12 @@ async def chat(
         content = chunk.get("choices", [{}])[0].get("message", {}).get("content", "") or ""
         clean, citations = extract_and_strip_sources_block(content)
         chunk["choices"][0]["message"]["content"] = clean
-        chunk["extra"] = json.dumps({"sources": filter_sources_by_citations(sources, citations)})
+        extra = {"sources": filter_sources_by_citations(sources, citations)}
+        if metadata.get("attachments"):
+            # Report the attachment file_ids actually indexed/used so the client
+            # can tell which of its attachments were leveraged (vs not yet indexed).
```
paultranvan: Simpler comment proposition: `Indicate which attachments were actually searched to generate the answer`

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Addressed in 867105b8 — the docstring is now client-agnostic:

```python
"""file_ids from ``metadata.attachments = [{"id": ...}, ...]``; malformed payloads dropped."""
```

No client name is referenced anywhere in the feature now (`query_service.py:673`).

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Addressed in 867105b8 — the Cozy mention is gone from the branch (`query_service.py:416-420`). The feature is keyed purely on `metadata.attachments`, so any client can use it.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Addressed in 867105b8 — the 7-line block is now two lines stating only the *why* (`query_service.py:417-418`):

```python
# No security check on the ids: file_id is ANDed with the
# server-fixed partition, so a foreign id can never match.
```

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Agreed, and addressed in 867105b8 — the `if attachment_ids and not chunks:` warning block was removed entirely. An attachment-scoped query that matches nothing now behaves exactly like a partition-scoped query that matches nothing: no log, no special case.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Still reads "leveraged" at `query_service.py:604`. Worth taking your exact wording — "searched" is the more accurate one too: the list is what retrieval was scoped to, not what the answer actually drew on (that is `extra.sources`).

openrag/services/orchestrators/query_service.py
```diff
@@ -411,6 +413,11 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
             # partition the caller also has access to (#706).
             partition = [scope.partition]
             filter_params = {"file_id": scope.file_ids}
+        elif attachment_ids and partition:
+            # No security check on the ids: file_id is ANDed with the
+            # server-fixed partition, so a foreign id can never match.
+            indexed_attachment_ids = await self._existing_file_ids(attachment_ids, partition)
```
Ahmath-Gadji, line 422: **`openrag-all` blocks attachments for admins**

With `SUPER_ADMIN_MODE`, `openrag-all` arrives here as `partition == ["all"]` (`api/dependencies/llm.py:40-47`). `_existing_file_ids` then queries `partition_name = 'all'`, which matches no row → `filter_params = {"file_id": []}` → Milvus `_MATCH_NOTHING_EXPR` (`milvus_store.py:509`).

So an admin chatting with attachments over `openrag-all` retrieves **zero chunks**, and `extra.attachments: []` reports the files as not indexed even when they are.

The workspace path already solves this: `WorkspaceService.resolve_scope` (`workspace_service.py:103-117`) accepts `"all"` and resolves from any partition. Suggest the same here — expand `"all"` to the concrete partitions before the lookup, rather than fail-closing.

openrag/services/orchestrators/query_service.py
```diff
@@ -470,7 +477,19 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
...
+    async def _existing_file_ids(self, file_ids: list[str], partitions: list[str]) -> list[str]:
+        """Order-preserving subset of ``file_ids`` indexed in ``partitions``.
+
+        Each lookup is partition-scoped: ``(file_id, partition)`` is the catalog's
+        unique key, so file_id alone is not. A wildcard like ``"all"`` matches no
+        real ``partition_name`` and so contributes nothing (fail-closed).
+        """
+        found: set[str] = set()
+        for p in partitions:
+            found.update(await self._workspace.get_existing_file_ids(p, file_ids))
```
Ahmath-Gadji: **Sequential round-trip per partition**

The `await` is inside the loop. For a regular user, `openrag-all` resolves to their whole membership list, so N partitions means N sequential Postgres round-trips before retrieval even starts.

```python
results = await asyncio.gather(*(self._workspace.get_existing_file_ids(p, file_ids) for p in partitions))
found = {fid for r in results for fid in r}
```

openrag/services/orchestrators/query_service.py
```diff
@@ -470,7 +477,19 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
...
+        return [fid for fid in file_ids if fid in found]
```
Ahmath-Gadji: **Client duplicates survive**

`file_ids` comes straight from the client, so `attachments: [{"id": "a"}, {"id": "a"}]` yields `filter_params {"file_id": ["a", "a"]}` and `extra.attachments: ["a", "a"]`. Harmless for Milvus, but the list reported back is wrong.

Order-preserving dedupe: `return [fid for fid in dict.fromkeys(file_ids) if fid in found]`.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Just an idea, not a prescription — take it or leave it if you see a simpler path.

Probably no need to enumerate partitions. `"all"` reaches here as *exactly* `["all"]` and only for a `SUPER_ADMIN_MODE` admin (`api/dependencies/llm.py:40-47`); a regular user's list can't contain it either, since `"all"` is a reserved partition name (`partition_service.py:307`). That caller already reads every partition, so one unscoped catalog lookup would be enough:

```sql
SELECT DISTINCT file_id FROM files WHERE file_id = ANY($1::text[])
```

`DISTINCT` because the unique key is `(file_id, partition_name)`. Served by the existing index on `files.file_id`.

If you go that way, two things I'd suggest getting right:

- **A separate repo method** (e.g. `get_existing_file_ids_any_partition`) rather than making `partition` nullable on the existing one — a nullable param means an accidental `None` silently drops tenant scoping for every caller.
- **Require exactly `["all"]`**, raising on a mixed list. Same rule Milvus already enforces at `milvus_store.py:495-496`.

```python
if PARTITION_WILDCARD in partitions:
    if len(partitions) > 1:
        raise ValueError("`partitions` cannot mix the wildcard with explicit values.")
    # only a SUPER_ADMIN_MODE admin gets here (api/dependencies/llm.py:40-47)
    found = await self._workspace.get_existing_file_ids_any_partition(file_ids)
else:
    results = await asyncio.gather(*(self._workspace.get_existing_file_ids(p, file_ids) for p in partitions))
    found = {fid for r in results for fid in r}
return [fid for fid in dict.fromkeys(file_ids) if fid in found]
```

That would also fold in the two other threads (concurrent lookups, dedupe).

Whichever way you go: the docstring at `query_service.py:486-487` and `test_chat_attachments_all_partition_is_fail_closed` both assert the current fail-closed behaviour, so they'd need inverting rather than extending.

openrag/services/persistence/workspace_repo.py
```diff
@@ -250,6 +250,22 @@ async def get_existing_file_ids(
         )
         return {r["file_id"] for r in rows}
 
+    async def get_existing_file_ids_any_partition(self, file_ids: list[str]) -> set[str]:
```
Ahmath-Gadji, line 253: **Missing from the port**

This lands on `PgWorkspaceRepository` but isn't declared on `WorkspaceRepository` (`core/ports/workspace_repo.py`), where every other repo method is an `@abstractmethod`. `WorkspaceService` is typed against the port, so it now calls a method the port doesn't define.

Nothing breaks today — there's a single implementation, and CI won't flag it since `lint` is ruff rather than a type checker. But any future implementation or ABC-based test double would silently lack it. Worth adding the abstract declaration next to `get_existing_file_ids` at `core/ports/workspace_repo.py:52`.

Rest of the commit looks right to me — the separate method, the `SELECT DISTINCT`, the mixed-list guard, and inverting the fail-closed test/docstring rather than working around them.

Ahmath-Gadji:
@coderabbit review

openrag/services/persistence/workspace_repo.py
EnjoyBacon7: Thanks — `e7a08759` covers it.

I pushed four follow-ups on top (`081cf6f0`..`f104591b`), all doc/test only, no behaviour change:

- **`0f32bab8`** — integration coverage for both catalog lookups. `get_existing_file_ids_any_partition` was only ever reached through a fake, so its SQL never ran in CI; the repos suite runs against a real database on every PR. Mutation-checked: dropping the partition clause from the scoped query, or adding one to the unscoped query, each fails a test.
- **`081cf6f0`** — the comment at `query_service.py:417` justified skipping an ownership check with "file_id is ANDed with the server-fixed partition". That stopped covering the `"all"` path once its lookup became unscoped, where the safety argument is the `SUPER_ADMIN_MODE` invariant instead. Both are now spelled out.
- **`5d21e437`** — the "leveraged" → "searched" wording from the other thread.
- **`f104591b`** — docstring now mentions the dedupe.

Revert any of them freely if you'd rather word it your own way.

openrag/services/orchestrators/query_service.py
EnjoyBacon7: Applied in 5d21e437 — now reads "actually searched", per your wording.

EnjoyBacon7 (dismissed):
Approving. Scoping retrieval by a client-supplied attachment list, with the indexed subset reported back in `extra.attachments`.

**Review trail** — all raised issues resolved:

- `openrag-all` no longer blocks attachments for `SUPER_ADMIN_MODE` admins. The wildcard takes a single unscoped catalog lookup rather than fail-closing to zero results, guarded against a mixed `["all", ...]` list.
- Per-partition lookups run concurrently instead of serially.
- Client-supplied duplicate ids are deduplicated before they reach the filter and the response.
- `get_existing_file_ids_any_partition` is declared on the `WorkspaceRepository` port, not just the Postgres implementation.

**Verified beyond the diff**

- The filterless fallback in `RetrieverPipeline` drops only the temporal filter and preserves `filter_params`, so an attachment-scoped query that matches nothing can never silently widen to the whole partition.
- `["all"]` reaches this layer solely for a `SUPER_ADMIN_MODE` admin — `get_partition_name` expands it to memberships for everyone else, and `"all"` is a reserved partition name, so it cannot arrive via a membership list.
- Both scoping queries are mutation-checked: changing either partition clause fails a test.

**Left as follow-ups, not blockers**

- A present-but-entirely-malformed `attachments` payload (e.g. `[{"file_id": ...}]` instead of `[{"id": ...}]`) degrades to an unscoped partition search rather than an error. Deliberate and tested, but the absent case and the all-invalid case are distinguishable if you later want them to behave differently.
- The attachment id list is unbounded and client-controlled; a cap would need a product call on truncate-vs-400.

Disclosure: four doc/test commits here are mine. The substantive code is @ewan102's, and I reviewed it before contributing.

Ahmath-Gadji:
## Suggested Improvements

I've reviewed the PR and identified two improvements worth considering:

### 1️⃣ Defensive Guard for Empty Partitions

Add an early return when the partitions list is empty to avoid unnecessary database queries:

```python
async def _existing_file_ids(self, file_ids: Sequence[str], partitions: Sequence[str]) -> list[str]:
    """Order-preserving, deduplicated subset of ``file_ids`` indexed in ``partitions``.

    ``file_ids`` is client-supplied, so repeats are dropped rather than
    echoed back into the filter and into ``extra.attachments``.

    ``"all"`` (``SUPER_ADMIN_MODE`` wildcard) takes an unscoped lookup instead
    of a per-partition one.
    """
    if not partitions:
        return []
    if "all" in partitions:
        if len(partitions) > 1:
            raise ValueError("`partitions` cannot mix the wildcard with explicit values.")
        found = set(await self._workspace.get_existing_file_ids_any_partition(file_ids))
    else:
        results = await asyncio.gather(*(self._workspace.get_existing_file_ids(p, file_ids) for p in partitions))
        found = {fid for r in results for fid in r}
    return [fid for fid in dict.fromkeys(file_ids) if fid in found]
```

**Why:** When no partitions are accessible, the current code still tries to look up files. This guard short-circuits the method cleanly.

---

### 2️⃣ API Documentation for Attachments

Add a row to the metadata table in `docs/content/docs/documentation/API.mdx` documenting the new `attachments` field:

| Parameter | Type | Description |
|-----------|------|-------------|
| `attachments` | `List[{"id": string}]` | Scope RAG retrieval to a list of indexed file IDs. Unindexed files are silently excluded. Useful for multi-file search within a partition. Only files accessible in the target partition(s) are retrieved. |

**Example request:**
```bash
curl -X POST http://localhost:8080/v1/chat/completions \\
  -H "Authorization: Bearer $AUTH_TOKEN" \\
  -H "Content-Type: application/json" \\
  -d '{
    "model": "openrag-partition-name",
    "messages": [{"role": "user", "content": "summarize these files"}],
    "metadata": {
      "attachments": [
        {"id": "file-001"},
        {"id": "file-002"},
        {"id": "file-003"}
      ]
    }
  }'
```

**Response:** The `extra.indexed_attachment_ids` field shows which attachment IDs were actually indexed (deduplicated, order-preserved). Unknown or unindexed IDs are silently filtered out.

---

Both changes are optional polish — the PR implementation is solid as-is. ✅

Ahmath-Gadji (approved):

PR #765, opened 2026-07-23
chore: back-merge v2.0.1 into develop

andyne13:
Standard git-flow back-merge after the v2.0.1 release. Brings two things onto `develop`:

1. **Version bumps** (`pyproject`/`uv.lock` → 2.0.1, chart 0.5.2, compose + chart image tags → v2.0.1) — so the next cycle does not start from 2.0.0.
2. **The build.yml guard hotfix (#764)** — the `verify-tag` gate that replaced the unusable `base_ref` check. **Without this, the next release from develop re-breaks identically** (silent skip of all image builds under protected main).

No new features; pure release reconciliation. `develop` has not moved since `release/2.0.1` was cut, so this is conflict-free.

## Summary by CodeRabbit

- **New Release**
  - OpenRAG updated to version **2.0.1**.
  - Updated container images for the main app, Ray, and Admin UI are now used by Helm charts and Docker Compose.

- **Reliability**
  - Release publishing now validates that **GA tags** match the expected version pattern and that the tagged commit is contained in the main branch before images are built/pushed.


hedhoud (changes requested):
The release/version back-merge itself looks consistent, but I would not merge the GA publishing guard in this shape.

The workflow still accepts broad `v*` tags, so a typo or non-GA prerelease tag that is reachable from `main` can publish release images and `latest`. The tag name is also interpolated directly inside the shell script for a privileged publishing workflow; Git accepts metacharacters in ref names, so this should be treated as untrusted data.

Please tighten the GA tag validation to exact release tags and pass the tag name as data in the shell step. After that, the rest of the PR looks fine to me.

.github/workflows/build.yml
```diff
@@ -11,9 +11,34 @@ env:
...
 jobs:
+  # GA images build for release tags (vX.Y.Z), never for RC tags (build_rc.yml
+  # owns those). This job also enforces "the tag is on main": we can't use
+  # github.event.base_ref for that — it is empty for a tag pushed to a
+  # branch-protected main (which only receives PR-merge commits, not client
+  # pushes), so the old base_ref guard silently skipped every release build.
+  # Instead we verify reachability from origin/main explicitly and fail loud.
+  verify-tag:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-rc.')
```
hedhoud, line 22: This still treats any `v*` tag that is not `-rc.` as a GA release. A typo or prerelease-style tag reachable from `main` could publish release images and `latest`. Please constrain GA publishing to exact release tags only.

.github/workflows/build.yml
```diff
@@ -11,9 +11,34 @@ env:
...
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Fail if the tagged commit is not on main
+        run: |
+          git fetch --no-tags origin main
+          if git merge-base --is-ancestor "${{ github.sha }}" FETCH_HEAD; then
+            echo "OK: ${{ github.ref_name }} (${{ github.sha }}) is on main"
```
hedhoud: This interpolates the tag name directly into the shell script. Tag names are ref data and can contain shell metacharacters, so this publishing workflow should pass the value through the step environment and print that variable instead.

hedhoud (commented):
Adding precise anchors for the requested changes.

andyne13:
Thanks @hedhoud — both findings confirmed valid, and I verified them against the merged workflow rather than just the diff.

To be precise on the first one: the `-rc1` variant is the sharpest case. `build_rc.yml` triggers on `v*-rc.*`, which **requires** the dot — so `v2.0.1-rc1` matched neither workflow's intent and would have published a release candidate as GA and moved `:latest`. A one-character typo. Agreed this should never have shipped in that shape.

On the second, agreed it should be treated as untrusted data: `git check-ref-format` permits `;`, `$`, backtick, `"`, `|` and `&` in ref names, and this job holds `packages: write` plus Docker Hub credentials.

zizmor also flagged a third item in the same step — the checkout did not set `persist-credentials: false`, inconsistent with `build_rc.yml`, which already does on all three checkouts (`95fd86fc`). Fixed alongside.

**Fix is in #767, targeting `main`, not this branch.** The affected guard is already merged on `main` (it published v2.0.1), so pushing the fix here would leave `main` exposed while only `develop` received it.

Sequence: merge #767 into `main` → re-sync this PR from the new `main` tip so the back-merge carries the fix → merge this. Leaving #765 open and unmerged until then.

andyne13:
@hedhoud the guard fix has landed — #767 is merged into `main`, and I have advanced this back-merge branch to the new `main` tip (`44833942`), so this PR now carries it.

Both of your findings are addressed here:
- GA tags validated against `^v[0-9]+\.[0-9]+\.[0-9]+$`, so `v1.2.3-rc1`, `-beta`, `vfoo` fail loudly instead of publishing GA images and moving `:latest`. Well-formed `-rc.` tags still route to `build_rc.yml`.
- Tag name and SHA passed via `env:` and referenced as shell variables — no `${{ }}` left in the `run` body.
- Plus `persist-credentials: false`, matching `build_rc.yml`.

Diff into `develop` is now the 6 expected files: the version bumps plus the hardened `build.yml`.

Could you clear the `CHANGES_REQUESTED` when you get a moment? Worth noting `develop` is still on `2.0.0` **and** still has the old broken guard until this merges — so the next release cut from it would both carry the wrong version and silently skip its image builds.

hedhoud (approved):
Approved. The back-merge now carries the hardened GA publish guard from #767: exact GA tag validation is in place, tag/SHA values are passed through `env` rather than interpolated into the shell body, and the publish jobs remain gated behind `verify-tag`.

A few CI jobs were still pending when I checked; branch protection should keep the merge blocked until those complete.

andyne13:
Thanks @hedhoud. Resolving your `build.yml:22` thread — noting why here so it is not a silent dismissal.

The comment still anchors to live code because line 22 (the job-level `if`) genuinely did not change: it still admits any `v*` tag that is not `-rc.`. The fix went into the **step** below it, which now rejects anything that is not an exact `^v[0-9]+\.[0-9]+\.[0-9]+$` and fails loudly — so `v1.2.3-rc1`, `-beta`, `vfoo` no longer publish GA images or move `:latest`. That is the change you approved in #767, and this branch now carries it.

PR #767, opened 2026-07-24
fix(ci): harden GA publish guard — exact tag format, no shell interpolation

andyne13:
Addresses the review findings on the `verify-tag` gate added in #764. Raised by @hedhoud on #765 and independently by CodeRabbit/zizmor. All three verified against the merged workflow.

Targets `main` rather than the back-merge branch (#765), because the affected guard is already **on main** — fixing it in the back-merge would leave `main` exposed while only `develop` got the fix.

## 1. GA tag format was too loose

The guard only rejected `-rc.`, so these all passed and would **publish GA images and move `:latest`**:

| Tag | Before | After |
|---|---|---|
| `v2.0.1` | builds | builds |
| `v2.0.1-rc.1` | skips → `build_rc.yml` | skips → `build_rc.yml` |
| **`v2.0.1-rc1`** | **publishes as GA** | fails loud |
| `v2.0.1-beta`, `v2.0.1-alpha.1` | publishes as GA | fails loud |
| `vfoo`, `v-test` | publishes as GA | fails loud |

The `-rc1` case is the sharp one: `build_rc.yml` triggers on `v*-rc.*`, which **requires** the dot — so a one-character typo matched neither workflow's intent and shipped a release candidate as GA.

Now validated against `^v[0-9]+\.[0-9]+\.[0-9]+$`.

## 2. Shell injection via `github.ref_name`

`${{ github.ref_name }}` was expanded into the `run` body before the shell executed. `git check-ref-format` permits `;`, `$`, backtick, `"`, `|`, `&` in ref names, so a crafted tag could execute arbitrary code — in a job holding `packages: write` and Docker Hub credentials.

Tag name and SHA are now passed via `env:` and referenced as `"$TAG_NAME"` / `"$TAG_SHA"`.

## 3. Checkout persisted credentials

`build_rc.yml` already sets `persist-credentials: false` on all three checkouts (hardened in `95fd86fc`); the new `verify-tag` checkout was inconsistent. Added. The repo is public, so the `origin/main` fetch still works without them.

## Validation

Regex checked against every case above (plus `v10.20.30`, `v2.0`, `v2.0.1.2`, `2.0.1`, and an injection-shaped tag — only exact `vMAJOR.MINOR.PATCH` passes). YAML re-parsed; confirmed no `${{ }}` remains in the `run` body and all three build jobs are still gated by `needs: verify-tag`.

Once merged, #765 will be re-synced from `main` so the back-merge carries this too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened release validation for GA publishing.
  * Releases now require an exact semantic version tag format (`vMAJOR.MINOR.PATCH`).
  * Added verification that tagged commits are included in the main development branch.
  * Improved release-check error and success messaging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,13 +26,29 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Fail if the tagged commit is not on main
+          persist-credentials: false
+      # Tag name and SHA are passed as env data, never interpolated into the
+      # script body: ${{ }} expands before the shell runs, and git permits
+      # ; $ ` " | & in ref names — so a crafted tag could otherwise execute
+      # arbitrary code in this privileged publishing workflow.
+      - name: Verify this is an exact GA release tag on main
+        env:
+          TAG_NAME: ${{ github.ref_name }}
+          TAG_SHA: ${{ github.sha }}
         run: |
+          # The job-level `if` only filters well-formed `-rc.` tags (owned by
+          # build_rc.yml). Anything else reaching here must be an exact GA tag:
+          # a near-miss like `v1.2.3-rc1` matches neither workflow's intent and
+          # would otherwise publish a prerelease as GA and move `latest`.
+          if ! printf '%s' "$TAG_NAME" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::$TAG_NAME is not an exact GA release tag (vMAJOR.MINOR.PATCH) — refusing to publish GA images."
+            exit 1
+          fi
           git fetch --no-tags origin main
-          if git merge-base --is-ancestor "${{ github.sha }}" FETCH_HEAD; then
-            echo "OK: ${{ github.ref_name }} (${{ github.sha }}) is on main"
+          if git merge-base --is-ancestor "$TAG_SHA" FETCH_HEAD; then
+            echo "OK: $TAG_NAME ($TAG_SHA) is an exact GA tag on main"
           else
-            echo "::error::Tag ${{ github.ref_name }} is not on main — refusing to build GA images."
+            echo "::error::$TAG_NAME is not on main — refusing to publish GA images."
             exit 1
           fi
 


andyne13:
@hedhoud this is the fix for both findings you raised on #765 — could you take a look when you have a moment?

Short version of what changed in the `verify-tag` step:
- GA tags now validated against `^v[0-9]+\.[0-9]+\.[0-9]+$`, so `v2.0.1-rc1` / `-beta` / `vfoo` fail loudly instead of publishing GA images and moving `:latest`. Well-formed `-rc.` tags still skip to `build_rc.yml` as before.
- Tag name and SHA passed through `env:` and referenced as `"$TAG_NAME"` / `"$TAG_SHA"` — no `${{ }}` left in the `run` body.
- Added `persist-credentials: false`, matching what `build_rc.yml` already does.

Worth flagging for prioritisation: the loose guard is what v2.0.1 shipped with and it is live on `main` until this merges. It needs push access to exploit, so not an emergency, but I would rather not leave it sitting.

Once this lands I will re-sync #765 from `main` so the back-merge carries it, then merge that.

hedhoud (approved):
Approved. This addresses the two blocking findings from #765: malformed `v*` prerelease/typo tags are rejected before any publish job can run, and the tag/SHA values are now passed through `env` instead of being interpolated into the shell script.

I do not consider the remaining CodeRabbit comment about non-`v` tags blocking here: this workflow is configured to trigger only on `v*` tag pushes, so tags like `1.2.3` or `release-1` do not enter this publishing workflow and cannot publish GA images.

andyne13:
Re the remaining CodeRabbit thread (`build.yml:45`, "do not bypass validation for non-`v` tags") — **declining this one, false positive.**

The premise is that tags like `1.2.3` or `release-1` reach the `verify-tag` step and escape validation. They do not: the workflow trigger is

```yaml
on:
  push:
    tags: ["v*"]
```

so a non-`v` tag never triggers `build.yml` in the first place — the job-level `if` is never evaluated and nothing can be published. The `startsWith(github.ref, 'refs/tags/v')` clause is defensive redundancy with that filter (and correctly excludes `workflow_dispatch`, where `github.ref` is a branch).

Adopting the suggestion would mean broadening `on:` to all tag refs so malformed ones fail loudly. That is worse for this repo: `build.yml` would then run on every tag, including non-release ones such as `archive/dev-github-flow`, producing red failures on tags that were never meant to publish.

The forgot-the-`v` case (`2.0.2` instead of `v2.0.2`) does silently do nothing — but it publishes nothing either, and Phase 6 of the release runbook already requires confirming the build actually started before moving on. The asymmetry is deliberate: publishing the wrong thing is the dangerous failure and is now blocked; not publishing is immediately visible.

Resolving the thread on that basis.

PR #769, opened 2026-07-24
fix(models): reject '/' in model-endpoint names

### LINKED ISSUE #768: Model endpoint names containing '/' become permanently unreachable (404 on delete/set-default/update)
Every single-endpoint model-endpoint route is shaped `/model-endpoints/{model_type}/{name}` (`GET`/`PUT`/`DELETE`, plus `.../set-default`, `.../reveal-api-key`, `.../validate`). FastAPI/Starlette treats `{name}` as **one path segment** — it cannot contain a `/`.

`CreateModelEndpointRequest`/`UpdateModelEndpointRequest`'s `_normalize_name` (`api/schemas/admin/model_endpoint_schemas.py`) only trims whitespace and rejects blank values, so a name like `"LocalReranker/re"` (e.g. pasted in after a `owner/model`-style slug) is accepted on create.

**Effect:** the row is created fine and shows up in `GET /model-endpoints/` (list has no per-item routing), but every single-item action 404s — even URL-encoded (`%2F`), since uvicorn decodes it back to `/` in `scope["path"]` before Starlette's router sees it:

```
DELETE /model-endpoints/reranker/LocalReranker%2Fre        → 404 {"detail":"Not Found"}
POST   /model-endpoints/reranker/LocalReranker%2Fre/set-default → 404 {"detail":"Not Found"}
```

This is Starlette's generic 404 (route didn't match), not the app's `NotFoundError` (which returns `{"detail":"[NOT_FOUND]: Endpoint '...' not found.", "extra": {...}}`) — the row genuinely exists in the DB, it's just unreachable. In the admin UI this reads as "model not found" when trying to delete or set-default an endpoint that is visibly right there in the list.

**Repro:**
1. `POST /model-endpoints/` with `name: "owner/model"`, `model_type: "reranker"` → 201, row created.
2. `POST /model-endpoints/reranker/owner%2Fmodel/set-default` → 404, even though the row exists.
3. Same for `DELETE`/`PUT`/`GET`/`reveal-api-key`/`validate`.

**Fix:**
- Backend: reject `/` in `name` in `_normalize_name` (both create and rename), so no new row can end up in this state.
- Admin UI: the Name field (`ui/src/pages/admin/models.tsx`) has no client-side guard, and the resulting 422 currently surfaces as an unhelpful generic toast (`ApiError` in `ui/src/lib/api/client.ts` only unwraps `detail` when it's a string; FastAPI's validation-error `detail` is a list, so today the toast just reads "HTTP 422"). Add an inline guard on the Name input so the mistake is caught immediately with a clear message.

Existing rows already created with a `/` in their name (pre-fix) stay stuck and need a direct DB rename/delete — the API can never reach them.

### LINKED ISSUE #770: Model endpoint with a '/' in its name is unmanageable: every {name} route 404s
## Summary

A model endpoint whose registry **name** contains a `/` becomes permanently unmanageable through the API: every route that addresses it by name returns `404 {"detail": "Not Found"}`. The endpoint cannot be promoted to default, edited, renamed, or even deleted.

Endpoint names are not validated, so the API accepts a name it can never address again.

## Reproduce

1. Register a reranker endpoint named `Alibaba-NLP/gte-multilingual-reranker-base` (any type works; any name containing `/`).
2. Register a second endpoint of the same type so there is something to switch between.
3. In the admin UI, click **Set Default** on the slash-named endpoint.

**Expected:** it becomes the default.
**Actual:** a popup reading `Not Found`.

The same 404 occurs for **GET / PUT / DELETE / set-default / reveal-api-key / validate** on that endpoint.

## Root cause

`POST /model-endpoints/{model_type}/{name}/set-default` (`openrag/api/routers/admin/model_endpoints.py:97`) declares `name: str` — a plain path parameter, whose Starlette converter matches `[^/]+`.

The UI correctly `encodeURIComponent`s the name (`ui/src/lib/api/models.ts:215,217`), but ASGI servers decode `%2F` back to `/` when populating the request path, so the routed path gains an extra segment and matches no route. FastAPI then returns its **default** 404 body — which is why the popup shows the bare `Not Found` rather than the service's own message (`Endpoint '<name>' of type '<type>' not found.`).

`encodeURIComponent` cannot protect a `/` inside a path segment. This is not fixable in the frontend.

Minimal reproduction against the same route shape:

```python
from fastapi import FastAPI
from fastapi.testclient import TestClient
from urllib.parse import quote

app = FastAPI()

@app.post("/model-endpoints/{model_type}/{name}/set-default")
def set_default(model_type: str, name: str):
    return {"model_type": model_type, "name": name}

c = TestClient(app)
for name in ["gte-multilingual-reranker-base", "Alibaba-NLP/gte-multilingual-reranker-base"]:
    r = c.post(f"/model-endpoints/reranker/{quote(name, safe='')}/set-default")
    print(name, "->", r.status_code, r.json())
```
```
gte-multilingual-reranker-base              -> 200 {'model_type': 'reranker', 'name': 'gte-multilingual-reranker-base'}
Alibaba-NLP/gte-multilingual-reranker-base  -> 404 {'detail': 'Not Found'}
```

**Why the name is unvalidated:** `_normalize_name` (`openrag/api/schemas/admin/model_endpoint_schemas.py:14-19`) only trims and rejects blank. There is no character restriction, so `/`, `?`, `#`, `%` are all accepted at creation.

**Not reranker-specific** — any `model_type` is affected. It surfaces on reranker/LLM/embedder because HuggingFace-style model IDs carry a vendor prefix, and `openrag/core/config/retrieval.py:17` ships `model_name = "Alibaba-NLP/gte-multilingual-reranker-base"` as the project's own default reranker. Pasting that string as the registry name is the natural thing for an operator to do.

## Impact

The row is **stranded**: unreachable by every `{name}` route, so it cannot be renamed or deleted via the API. Recovery currently requires direct database access.

Seeded endpoints are unaffected — `seed_defaults` runs the model name through `_slug()` (`model_endpoint_service.py:88`). Only hand-created endpoints can carry a raw slash.

## Workaround (no DB access required)

`POST /model-endpoints/` takes the name in the **body**, not the path, so it still works. Re-register the endpoint under a slash-free name with `is_default: true`; the repository demotes the previous default atomically in the same transaction:

```json
{
  "name": "gte-multilingual-reranker-base",
  "model_type": "reranker",
  "endpoint": "<reranker url>",
  "model_name": "Alibaba-NLP/gte-multilingual-reranker-base",
  "is_default": true
}
```

Only the registry **name** must be path-safe — `model_name` can keep the real vendor-prefixed ID. The stranded row remains in the database but is no longer default.

## Fix considerations

**1. Prevent (straightforward).** Add a character validator to `_normalize_name` rejecting `/` (and `?`, `#`, `%`) with an actionable message, e.g. *"name may not contain '/' — use a path-safe name; put the vendor-prefixed model id in model_name instead."*

**2. Rescue existing rows (needs a decision).** Validation alone leaves already-created rows stranded. A migration that slugifies offending names is **not** safe as-is, because endpoint names are referenced by name elsewhere and nothing cascades:

- `partitions.embedder` — `String`, references an embedder endpoint name (`schema.py:101`)
- `partitions.chat_llm` — `String`, references an LLM endpoint name (`schema.py:107`)
- `pipeline_presets.config` — `JSONB`, may embed endpoint names (`schema.py:69-74`)
- `PgModelEndpointRepository.rename()` (`model_endpoint_repo.py:126-132`) updates **only** the `model_endpoints` table

So any rename-based rescue must update those references in the same transaction.

Alternatives worth weighing: accept the name as a query parameter or in the body for mutation routes (encoding-safe, but an API contract change), or add a dedicated admin route to repair/rename by id.

### Related defect found while investigating

**`rename()` does not cascade to referencing rows.** Renaming an endpoint that a partition points at via `embedder` or `chat_llm` leaves a dangling reference. This is independent of the slash bug and reproducible today with a perfectly path-safe name. Worth splitting into its own issue if confirmed.

## Tasks

- [ ] Reject path-breaking characters in endpoint names at creation (and on rename)
- [ ] Decide the rescue path for existing stranded rows
- [ ] Confirm and split out the `rename()` non-cascading reference defect

Ahmath-Gadji:
## Summary
- Reject `/` in model-endpoint `name` on create/rename (`api/schemas/admin/model_endpoint_schemas.py`) — a slash splits across the `/{model_type}/{name}` path segment used by every single-endpoint route, so the row stays visible in the list but every get/update/delete/set-default/reveal-api-key/validate call 404s, even URL-encoded.
- Add the matching client-side guard on the admin UI's Name field (`ui/src/pages/admin/models.tsx`): inline error + disabled submit, instead of letting the mistake reach the backend and surface as an unreadable "HTTP 422" toast (`ApiError` only unwraps string-shaped `detail`; FastAPI's validation errors are a list).
- `RetrievalService` now logs (debug) which reranker resolved for a partition — its own preset, the catalog default, or the static startup reranker — mirroring `QueryService._resolve_llm`/`_default_llm` (#755), so "which reranker ran?" is answerable from logs the way chat_llm already is.
- Along the way, closed the same gap #755 fixed for chat_llm, on the reranker side: a partition's `reranker` preset has no create/PATCH-time validation, so a renamed/deleted endpoint reaching `_reranker_factory` raised an unhandled `KeyError` instead of falling back to the catalog default (and then the static reranker if that's missing too).
- `PgModelEndpointRepository.rename()` now cascades the new name to every stored reference — `partitions.embedder` / `partitions.chat_llm`, and the endpoint-name fields embedded in `pipeline_presets.config` (JSONB) — in the same transaction as the rename. This is the defect @andyne13 flagged on #768 (originally filed independently as #770): reproducible with *any* path-safe name, and a naive rescue rename of a stuck slash-named row would have hit it too. `ModelEndpointService.update_model_endpoint` now reloads `PresetService` then `PartitionService`'s in-memory caches after a rename, mirroring `PresetService.update_preset`'s existing pattern — and finally puts the until-now-unused `partition_service` constructor arg to use.
- New tests throughout: schema rejection, reranker fallback paths, rename cascade SQL, and the post-rename cache-reload wiring.

Closes #768

## Test plan
- [x] `uv run pytest tests/unit/` — 2220 passed
- [x] `uv run ruff check` / `ruff format` — clean (pre-commit hook)
- [x] `npx tsc --noEmit` on `ui/` — clean
- [x] Reproduced the original bug against a live local instance (curl `DELETE`/`set-default` on a `reranker` endpoint named `LocalReranker/re` → 404 even with `%2F`), confirmed the row existed in Postgres, then verified the new validator rejects the same name at the API layer.

Note: pre-existing rows created before this fix (with a `/` already in their name) stay stuck and need a direct DB rename/delete — the API can never reach them.

## Summary by CodeRabbit

* **New Features**
  * Strengthened model endpoint `name` validation to a single safe URL path segment (allowlist + max length), consistently enforced in the backend and admin UI.
  * Improved per-partition reranker resolution: partition preset first, then catalog default, then legacy fallback.

* **Bug Fixes**
  * Model endpoint renames now reliably propagate to related partition and pipeline preset references, with safer behavior during rename races and partial reload failures.

* **Tests**
  * Expanded schema validation cases, reranker fallback scenarios, and rename/cascade coverage (including rename + other field changes and cache eviction).

### DIFF
--- a/openrag/api/schemas/admin/model_endpoint_schemas.py
+++ b/openrag/api/schemas/admin/model_endpoint_schemas.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 from datetime import datetime
 from typing import Any, Literal
 
@@ -15,12 +16,43 @@ ModelEndpointType = Literal["embedder", "reranker", "llm", "vlm"]
 # ``extra`` — validated here so a typo can't persist a nonsensical value.
 _LLM_TOKEN_EXTRA_KEYS = (LLM_CONTEXT_SIZE_KEY, LLM_OUTPUT_TOKENS_KEY)
 
+# Allowlist, not a denylist: `name` is a single path segment in every
+# single-endpoint route (see `_normalize_name`), and enumerating unsafe values
+# one at a time as they're discovered — first `/` (#768), then the RFC 3986
+# dot-segments `.`/`..` — never closes the class. Anchoring both ends on
+# alphanumeric rules out `/`, `.`, `..`, and any leading/trailing separator by
+# construction, while `.`/`_`/`-` stay available in the middle for realistic
+# names like `gpt-4.1` or `jina_v3`.
+_NAME_PATTERN = re.compile(r"[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?")
+_NAME_MAX_LENGTH = 128
+
 
 def _normalize_name(value: str) -> str:
-    """Trim a user-facing registry name and reject blank values."""
+    """Trim a user-facing registry name and reject any value unsafe as a URL path segment.
+
+    ``name`` is embedded as a single path segment in every single-endpoint route
+    (``GET/PUT/DELETE /model-endpoints/{model_type}/{name}``, ``.../set-default``,
+    ``.../reveal-api-key``, ``.../validate``). A value outside ``_NAME_PATTERN``
+    — a ``/`` (splits across path segments), the exact values ``.``/``..``
+    (RFC 3986 dot-segments: browsers and HTTP clients normalize these out of
+    the URL before the request is even sent, resolving to the collection route
+    or dropping the ``model_type`` segment entirely), or anything else that
+    doesn't start/end alphanumeric — would leave the row visible in the list
+    endpoint but permanently unreachable by get/update/delete/set-default,
+    surfacing as a spurious "not found". Percent-encoding never helps: ASGI
+    servers decode ``%2F``/dot-segment escapes before Starlette's router sees
+    the path.
+    """
     value = value.strip()
     if not value:
         raise ValueError("name must be non-empty")
+    if len(value) > _NAME_MAX_LENGTH:
+        raise ValueError(f"name must be at most {_NAME_MAX_LENGTH} characters")
+    if not _NAME_PATTERN.fullmatch(value):
+        raise ValueError(
+            "name must start and end with a letter or digit, and contain only "
+            "letters, digits, '.', '_', or '-' (it is used as a URL path segment)"
+        )
     return value
 
 
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -114,11 +114,13 @@ class ModelEndpointService:
         model_endpoint_repo: ModelEndpointRepository,
         config: Settings,
         partition_service: Any = None,
+        preset_service: Any = None,
         client_caches: dict[str, dict[str, Any]] | None = None,
     ) -> None:
         self._repo = model_endpoint_repo
         self._config = config
         self._partition_service = partition_service
+        self._preset_service = preset_service
         self._client_caches: dict[str, dict[str, Any]] = client_caches or {}
 
     # ------------------------------------------------------------------
@@ -396,6 +398,29 @@ class ModelEndpointService:
         Pass ``new_name=`` to rename. After any change the in-memory config is
         reloaded and the stale cached client instance is evicted so the next
         request builds a fresh client against the updated config.
+
+        A rename also cascades to every stored reference — ``partitions.embedder``
+        / ``partitions.chat_llm`` and endpoint-name fields embedded in
+        ``pipeline_presets.config`` — inside the repo's own rename transaction
+        (see ``PgModelEndpointRepository.rename``, #770). Those writes are
+        invisible until the referencing services reload their in-memory caches,
+        which is why a rename also refreshes presets then partitions here —
+        the same order ``PresetService.update_preset`` uses, since partition
+        resolution reads the presets dict.
+
+        Both reload calls ``await``, so a concurrent request can run between
+        them — and the DB rename has *already* committed by that point. Without
+        ``_alias_renamed_name``, a request landing in that window could resolve
+        a partition/preset that the cascade already repointed at ``new_name``
+        against a registry that (until the final ``load_all()`` below) still
+        only knows ``name`` — a bare ``KeyError``. The alias makes both ``name``
+        and ``new_name`` resolve immediately, built from the row this call just
+        wrote — not whatever the in-memory bucket held before it — so a rename
+        combined with a field change (e.g. a new ``endpoint``) aliases the
+        *updated* config, not a stale pre-update one. That also covers a reload
+        call above raising: the registry stays queryable under both names,
+        correctly, instead of the update's failure leaving it stuck on a stale
+        config until process restart.
         """
         existing = await self._repo.get(name, model_type)
         if existing is None:
@@ -423,6 +448,21 @@ class ModelEndpointService:
             await self._repo.rename(name, model_type, new_name)
             effective_name = new_name
             renamed_from = name
+            self._alias_renamed_name(model_type, name, new_name, updated or existing)
+            # A cached *client instance* under either name would otherwise survive
+            # this alias — the factory checks its cache before consulting the
+            # config registry, so a stale pre-rename/pre-update client would keep
+            # serving until the eviction at the end of this method, which a
+            # reload call below raising would skip entirely. The config alias
+            # above is already fresh, so evicting now is safe: anything rebuilt
+            # from either name resolves through the up-to-date config, not stale
+            # cached state.
+            self._invalidate_client_cache(model_type, name)
+            self._invalidate_client_cache(model_type, new_name)
+            if self._preset_service is not None:
+                await self._preset_service.load_all()
+            if self._partition_service is not None:
+                await self._partition_service.load_partitions()
 
         if promote_to_default:
             # Clears any prior default and sets this row in one transaction, then
@@ -539,6 +579,39 @@ class ModelEndpointService:
     # ------------------------------------------------------------------
     # Internals
     # ------------------------------------------------------------------
+
+    def _alias_renamed_name(self, model_type: str, old_name: str, new_name: str, row: ModelEndpointRow) -> None:
+        """Make both ``old_name`` and ``new_name`` resolve to *row* before any reload runs.
+
+        Runs synchronously right after the rename ``await`` returns — no
+        further ``await`` happens before this executes, so no concurrent
+        request can observe the DB already renamed while the registry still
+        only answers to ``old_name``.
+
+        Built from ``row`` — the just-written DB state — rather than copying
+        whatever the in-memory bucket currently holds under ``old_name``: a
+        rename can land in the same call as a field update (e.g. a new
+        ``endpoint`` URL), applied to the DB *before* this runs, so the stale
+        in-memory entry would alias both names to the pre-update config. If a
+        reload below then raises, that staleness would never get corrected
+        by the final ``load_all()`` this call never reaches — the registry
+        would keep serving the old endpoint under the new (DB-authoritative)
+        name until process restart. The next full ``load_all()`` (below, or
+        from any later CRUD call) rebuilds the bucket straight from DB and
+        drops the ``old_name`` entry on its own.
+        """
+        bucket: dict[str, Any] | None = getattr(self._config.models, model_type, None)
+        if bucket is None:
+            return
+        cfg = ModelEndpointConfig(
+            endpoint=row.endpoint,
+            model_name=row.model_name,
+            batch_size=row.batch_size,
+            timeout=row.timeout,
+            extra=row.extra,
+        )
+        bucket[old_name] = cfg
+        bucket[new_name] = cfg
 
     def _invalidate_client_cache(self, model_type: str, name: str) -> None:
         """Evict ``name`` from the component-factory cache for ``model_type``."""
--- a/openrag/services/persistence/model_endpoint_repo.py
+++ b/openrag/services/persistence/model_endpoint_repo.py
@@ -27,6 +27,20 @@ if TYPE_CHECKING:
 # transaction; ModelEndpointService.update_model_endpoint routes is_default there.
 _ALLOWED_UPDATE_FIELDS = frozenset({"endpoint", "model_name", "batch_size", "timeout", "extra"})
 
+# Endpoint names are referenced by value elsewhere, and nothing updates those
+# references when an endpoint is renamed (#770) — so ``rename()`` cascades to
+# every known reference in the same transaction as the name change. Direct
+# endpoint-name columns on ``partitions``, keyed by the model_type they hold:
+_PARTITION_COLUMN_BY_TYPE = {"embedder": "embedder", "llm": "chat_llm"}
+# Endpoint-name keys embedded in ``pipeline_presets.config`` (JSONB), by the
+# preset_type that carries them — see core/config/retrieval_pipeline.py and
+# core/config/indexation_pipeline.py for the field definitions.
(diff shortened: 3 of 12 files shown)


openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -423,6 +434,10 @@ async def update_model_endpoint(self, name: str, model_type: str, **fields: obje
             await self._repo.rename(name, model_type, new_name)
             effective_name = new_name
             renamed_from = name
+            if self._preset_service is not None:
```
hedhoud, line 462: Could we make this refresh consistent from the running app's point of view? The rename is already committed before these awaited reloads run. After partitions switch to the new endpoint name but before the model registry reloads, a concurrent request can resolve `new-name` against a registry that still only has `old-name` and fail with a `KeyError`. A transient reload failure is worse: the API returns 500, but the rename stays committed and the process keeps the old registry until restart. Please stage the refreshed state before swapping it, or keep both names resolvable until every cache has refreshed, and add coverage for a request during the reload and for a failed reload.

Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/7c0fb84d714a0b87c0d4e72b74bb4e565727a70b.

You're right — the DB rename (and its cascade) commits before either `await` below runs, so a request racing that window could resolve a name the cascade already wrote against a registry the reload hasn't reached yet, and a failed reload would leave it wedged there. Fixed by aliasing `new_name` to the (unchanged) config synchronously right after the rename commits — no `await` in between, so there's no window where it's unresolvable — and leaving `old_name` aliased too so a failed `load_all()`/`load_partitions()` doesn't strand the registry on the pre-rename name. Both stay resolvable to the same underlying client until the next successful `load_all()` naturally drops the old one. Added coverage for both scenarios: a request racing the reload, and a reload that raises.

Ahmath-Gadji:
Addressed the two review items:

- **@hedhoud**'s concern on the rename cache-reload race (`model_endpoint_service.py`): fixed in [7c0fb84](https://github.com/linagora/openrag/commit/7c0fb84d714a0b87c0d4e72b74bb4e565727a70b), replied inline on the thread.
- **CodeRabbit**'s nitpick on `retrieval_service.py:193-198` (unbound logger kwargs — actually silently dropped the context, not just style): fixed in [e2467ee](https://github.com/linagora/openrag/commit/e2467ee7dbec72f558d8fe6d0b2d5ac726b3259c).

Full unit suite (2223 tests) and ruff both green locally after the changes.

@coderabbitai review

Ahmath-Gadji:
Fixed the annotation nitpick in e09d484d.

@coderabbitai review

openrag/services/orchestrators/model_endpoint_service.py
hedhoud: Thanks, this fixes the pure rename case, but a rename can also change the endpoint settings in the same request. If a reload fails, the DB has the new URL while this alias still points to the old config. I reproduced that with a rename from old-name to new-name plus an endpoint URL change. Could we build the alias from the updated row and invalidate the affected cached clients before the reloads, then add this combined case to the failure test?

openrag/api/schemas/admin/model_endpoint_schemas.py
```diff
@@ -17,10 +17,22 @@
...
+    segments and never matches those routes again, even percent-encoded (ASGI
+    servers decode ``%2F`` before Starlette's router sees the path). Such a row
+    stays visible in the list endpoint but is permanently unreachable by
+    get/update/delete/set-default, surfacing as a spurious "not found".
+    """
     value = value.strip()
     if not value:
         raise ValueError("name must be non-empty")
+    if "/" in value:
+        raise ValueError("name must not contain '/' (it is used as a URL path segment)")
```
hedhoud: The slash check fixes the reported case, but exact names like '.' and '..' can still become unreachable. The browser keeps those values unchanged after encodeURIComponent and then normalizes them out of the URL, so an endpoint can appear in the list while update or delete targets a different route. Could we reject these two dot-segment names here and in the UI as well?

openrag/services/persistence/model_endpoint_repo.py
```diff
@@ -124,12 +138,64 @@ async def update(self, name: str, model_type: str, **fields: object) -> ModelEnd
...
+                    name,
+                    model_type,
+                    new_name,
+                )
+                if renamed is None:
+                    raise NotFoundError(f"Endpoint '{name}' of type '{model_type}' not found.")
+
+                partition_col = _PARTITION_COLUMN_BY_TYPE.get(model_type)
+                if partition_col:
+                    await conn.execute(
```
hedhoud, line 192: Could we also serialize this cascade with writes that assign endpoint names? A partition update can validate old-name, wait behind this UPDATE, then resume after the rename commits and write old-name back. Once the temporary alias is removed, that partition points to a missing endpoint again. A shared lock plus a DB-authoritative endpoint check on partition and preset writes would make the ordering safe.

openrag/api/schemas/admin/model_endpoint_schemas.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/a4f34a6d35dccbab59269ad509957f31c52a865e.

Good catch — you're right that denylisting each unreachable value one at a time will keep missing cases. Replaced both the `/` check and this with a single allowlist: name must start and end with an alphanumeric character, with `.`/`_`/`-` allowed in between. That rules out `/`, `.`, `..`, and any leading/trailing separator by construction, while still accepting realistic names like `gpt-4.1` or `jina_v3`. Also added a 128-char cap (the column has none today) and mirrored the same regex in the admin UI's client-side guard. Tests cover the rejected cases (including `.`/`..`) and that realistic names still pass.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/93351f6a0fdeb0fc7b4991b46270a45d471f0c52.

You're right — the alias was copying whatever the in-memory bucket held under `old_name` before this call ran, which is stale the moment a rename is combined with a field update. `_alias_renamed_name` now builds the aliased `ModelEndpointConfig` from the row this call just wrote (the `update()` result when fields changed, else the pre-fetched row), so both `old_name`/`new_name` alias to the correct, up-to-date config regardless of what the registry held before. Added the combined rename+endpoint-change case to the failure test, per your repro.

Ahmath-Gadji:
Addressed CodeRabbit's latest pass:

- **Major**: cached client instances under old_name/new_name could survive a failed reload (the factory checks its client cache before the config registry, so the config alias alone wasn't enough) — fixed in d007cd6, both names' cached clients are now evicted right after aliasing, before either reload await.
- **Nitpick**: update-schema invalid-name test coverage was a subset of create's — fixed in 97c27c4, both tests now share one `_UNSAFE_NAMES` list.

@coderabbitai review

openrag/services/persistence/model_endpoint_repo.py
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/7afc6aa131f8b28404eea55b7152313b43426914.

You're right, that's a real lost-update. rename() now `LOCK`s `partitions` `IN SHARE MODE` before touching anything — the same lock `PgPresetRepository.delete` already takes for the equivalent preset case. `PgPartitionRepository.update_partition` gets a matching DB-authoritative re-check when `chat_llm` is assigned: the write and a `model_endpoints` existence check now run in one transaction that touches `partitions` before `model_endpoints` — same order the rename cascade uses, so the two can only block on each other, never deadlock. A write that loses the race rolls back with `MODEL_ENDPOINT_NOT_FOUND` instead of silently persisting a stale name.

`embedder` isn't covered — it has no assignment-time validation at all today (a separate, pre-existing gap), so there's nothing yet for a rename to race against on that column.

Added repo-level tests for both orderings (endpoint renamed-away vs. still present) plus the lock ordering itself.

Ahmath-Gadji:
Addressed the DB-level TOCTOU race in 7afc6aa1.

@coderabbitai review

Ahmath-Gadji:
@hedhoud all three of your review threads are addressed (fix commits linked inline on each), CodeRabbit's findings are resolved too, and CI is green on the rebased branch. Could you take another look and approve if it looks good?

ui/src/pages/admin/models.tsx
```diff
@@ -316,6 +324,15 @@ function EndpointDialog({
...
   const [extraJson, setExtraJson] = useState("{}");
 
+  // Checked client-side too because the resulting 422 has no readable message —
+  // FastAPI's validation `detail` is a list, and ApiError only unwraps string detail.
+  const nameError =
+    name !== "" && name.length > NAME_MAX_LENGTH
+      ? `Name must be at most ${NAME_MAX_LENGTH} characters.`
+      : name !== "" && !NAME_PATTERN.test(name)
+        ? "Name must start/end with a letter or digit, and contain only letters, digits, '.', '_', or '-' — it's used in the endpoint's URL path."
+        : null;
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in commit https://github.com/linagora/openrag/commit/46c8b3de023de99af9f573cfe65a763c263fea59.

Good catch — validation and submission now both use `name.trim()` consistently (inline error, create/update payloads, and the rename comparison).

hedhoud (approved):
Rechecked the latest update and the earlier findings are addressed. The affected backend tests, lint, UI build, and CI are all clean. Approved from my side.

PR #763, opened 2026-07-23
release: v2.0.1

andyne13:
Release v2.0.1. Cut from `develop` @ `ecda8de7`. Version bumps only — no code changes.

- `pyproject.toml` + `uv.lock`: `2.0.0` → `2.0.1`
- compose + chart image pins → `v2.0.1`
- chart `0.5.1` → `0.5.2`, `appVersion` → `2.0.1`
- Also fixes the chart's api/ray pins, which pointed at a non-existent `1.1.13`
  tag since `81c039e5` — published tags carry the `v` prefix

55 PRs, milestone `v2.0.1`. Five are labelled `breaking-change`; upgrade actions
are in the release notes.

PR #771, opened 2026-07-24
chore(release): bump version to 2.0.2

Ahmath-Gadji:
## Summary
- pyproject.toml + uv.lock: 2.0.1 -> 2.0.2
- compose: openrag and openrag-admin-ui pins -> v2.0.2
- chart: version 0.5.2 -> 0.5.3, appVersion -> 2.0.2
- chart values: ray, openrag, adminUi image tags -> v2.0.2

## Test plan
- [ ] CI passes

## Summary by CodeRabbit

* **Release**
  * Updated the application to version 2.0.2.
  * Updated container images across supported deployment options to version 2.0.2.
  * Updated deployment metadata to reflect the new release version.

### DIFF
--- a/infra/charts/openrag-stack/values.yaml
+++ b/infra/charts/openrag-stack/values.yaml
@@ -33,7 +33,7 @@ ray:
   image:
     repository: ghcr.io/linagora/openrag-ray
     # Pin to a release tag (ideally a digest) for reproducible deploys.
-    tag: "v2.0.1"
+    tag: "v2.0.2"
 
 # === PostgreSQL (bitnami) ===
 postgresql:
@@ -262,7 +262,7 @@ adminUi:
     repository: linagoraai/openrag-admin-ui
     # Pin to a release tag (ideally a digest) for reproducible deploys. Must be a
     # build from infra/docker/ui.Dockerfile (nginx-unprivileged, listens :8080).
-    tag: "v2.0.1"
+    tag: "v2.0.2"
   imagePullPolicy: IfNotPresent
   replicaCount: 1
   service:
@@ -276,7 +276,7 @@ openrag:
   image:
     repository: linagoraai/openrag
     # Pin to a release tag (ideally a digest) for reproducible deploys.
-    tag: "v2.0.1"
+    tag: "v2.0.2"
   service:
     type: ClusterIP
     port: 8080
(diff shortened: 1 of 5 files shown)


infra/charts/openrag-stack/values.yaml
```diff
@@ -33,7 +33,7 @@ ray:
   image:
     repository: ghcr.io/linagora/openrag-ray
     # Pin to a release tag (ideally a digest) for reproducible deploys.
-    tag: "v2.0.1"
+    tag: "v2.0.2"
```
hedhoud, line 36: These image tags are not available yet. Because this PR targets `develop`, merging it will immediately publish chart `0.5.3-dev` with defaults that cannot be pulled, leaving deployments in `ImagePullBackOff`. Please keep the release-only pins on the release branch until `v2.0.2` is published, then back-merge them as we did for `v2.0.1`.

hedhoud (changes requested):
This branch is now 20 commits behind `develop`, so its successful checks do not validate the code that would actually ship. Please cut or refresh the `2.0.2` release from the current `develop`, rerun CI, and keep the image publication order consistent with the `v2.0.1` release flow.

Ahmath-Gadji:
Closing this in favor of doing the version bump as part of the actual release cut, not directly on `develop`.

**Why:**
- This PR targets `develop` and pins `v2.0.2` image tags/versions that don't exist yet — the latest published tag is still `v2.0.1`. Merging now would leave `develop` (and the `0.5.3-dev` chart it drives) referencing images that can't be pulled, per @hedhoud's review.
- The branch is also 20 commits behind `develop` at this point.
- Looking at how `v2.0.1` actually shipped, the bump belongs on the release side of the flow, not the develop side:
  - #763 (`release/2.0.1` → `main`) did the version bump and merged into `main`, where the `v2.0.1` tag was then cut — `build.yml`'s `verify-tag` gate only publishes GA images for tags reachable from `main`.
  - #765 (`backmerge/v2.0.1-into-develop` → `develop`) brought that same bump back into `develop` only afterward, once the images already existed.

**Correct order for 2.0.2:** cut a `release/2.0.2` branch, do the version bump there, PR it into `main`, tag `v2.0.2` on `main` so CI publishes real images, then sync `main` back into `develop` (backmerge PR) with that same bump — mirroring #763 → #765.

Will reopen following that flow once the 2.0.2 release is actually cut.

PR #808, opened 2026-07-27
Make the Grafana metrics link runtime configurable

### LINKED ISSUE #779: Admin UI: add a direct link to the Grafana metrics dashboard
Context

Operators use the Admin UI to inspect OpenRAG and Grafana to investigate runtime metrics.

Problem

There is no direct path between these tools, so operators must know and navigate to the Grafana address separately.

Expected behavior

- The Admin UI provides a clear metrics action that opens the configured Grafana dashboard.
- Deployments can configure the destination without changing the UI.
- The action is hidden or disabled when no dashboard is configured.

hedhoud:
Closes #779

## Context

Operators need a clear path from OpenRAG raw metrics to the deployment dashboard without knowing the Grafana address or rebuilding the Admin UI.

## Problem

A build-time-only link does not work reliably with prebuilt images. Hiding the action when runtime configuration is missing also makes the Grafana capability impossible to discover from the Metrics screen.

## Solution

System → Metrics now always shows one Open in Grafana action. When the deployment provides a valid dashboard destination, it opens in a new tab. When configuration is missing, the same action explains how to enable it. Compose and Helm deployments can provide the destination through `GRAFANA_URL`, while the previous build-time setting remains a compatibility fallback.

## Validation

All 160 Admin UI tests pass, the production UI build succeeds, and the backend runtime-configuration tests pass. Linting completes without new errors.

## Summary by CodeRabbit

* **New Features**
  * Added runtime-configured Grafana access from the Admin UI, with a build-time fallback.
  * Enabled Grafana access under the admin origin through `/grafana/`.
  * Added GPU monitoring panels to the system overview dashboard.
* **Bug Fixes**
  * Improved Grafana URL validation and Prometheus service connectivity.
  * Restricted direct Grafana and metrics-service port exposure.
* **Documentation**
  * Expanded Grafana setup guidance, credentials, and sub-path configuration.
* **Tests**
  * Added coverage for Grafana URL validation and Admin UI behavior.


aditykris (approved):
Looks good.

openrag/api/runtime_ui.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import annotations
+
+import os
+from urllib.parse import urlparse
+
+
+def get_grafana_url() -> str | None:
+    """Return a safe Grafana destination configured for this deployment."""
+    value = os.getenv("GRAFANA_URL", "").strip()
+    if not value or "\\" in value:
```
Ahmath-Gadji: Browsers strip Tab/LF/CR from URLs before resolving, so this guard misses them: `GRAFANA_URL="/<TAB>/evil.example/x"` passes validation and Chromium resolves it to `http://evil.example/x` (verified).

Low severity — the value is operator-set and the anchor is `rel="noopener noreferrer"` — but it's the exact case this check exists for, and it matters when `GRAFANA_URL` comes from a Helm values file or CI variable. Rejecting ASCII control characters alongside `\\` covers it. The tests cover `//`, `\\`, `javascript:` and bad ports, but not control characters.

ui/src/pages/admin/system.tsx
```diff
@@ -27,21 +36,21 @@ function actorStateColor(state: string): string {
...
 export default function SystemPage() {
+  const { data: config } = useQuery({
+    queryKey: ["system-config"],
+    queryFn: getConfig,
+    staleTime: Infinity,
+    refetchOnMount: "always",
+  });
+  const runtimeGrafanaUrl =
+    typeof config?.grafana_url === "string" ? config.grafana_url.trim() : "";
+  const grafanaUrl = runtimeGrafanaUrl || BUILD_TIME_GRAFANA_URL;
```
Ahmath-Gadji: When the runtime value fails validation the API returns `null`, and this falls through to the baked-in `VITE_GRAFANA_URL` rather than the not-configured dialog. So on an image built the pre-PR way, an invalid `GRAFANA_URL` silently opens the stale URL — and the link can't be switched off at runtime at all.

`env_vars.md` claims the opposite: "When this is empty or invalid, the action explains how to configure the dashboard instead of opening it." Either make the runtime value authoritative or adjust the doc. The new tests always supply `grafana_url`, so this branch is uncovered.

Ahmath-Gadji (commented):
Tried this on a live stack before reviewing — the runtime link, the nginx proxy, all four scrape targets and both dashboards work end to end. The parts most likely to break check out: `nvidia_smi_utilization_gpu_ratio` really is 0–1, so `* 100` with unit `percent` is correct and no panel double-scales; the 1.13.1 metric names and units match `nvidia-smi` exactly; `/config` is admin-gated; and `proxy_pass` with a variable correctly preserves the `/grafana/` prefix.

Two comments below, both low severity — nothing blocking.

infra/compose/grafana/dashboards/openrag-http.json
```diff
@@ -1,479 +1,105 @@
...
-        "tooltip": { "mode": "multi", "sort": "desc" }
-      }
+      "datasource": {"type": "prometheus", "uid": "prometheus"}, "description": "Routes with the highest percentage of failed requests.",
+      "fieldConfig": {"defaults": {"color": {"mode": "thresholds"}, "decimals": 1, "noValue": "No failures", "thresholds": {"mode": "absolute", "steps": [{"color": "green", "value": null}, {"color": "yellow", "value": 1}, {"color": "red", "value": 5}]}, "unit": "percent"}, "overrides": []},
+      "gridPos": {"h": 8, "w": 12, "x": 0, "y": 17}, "id": 11, "options": {"displayMode": "gradient", "orientation": "horizontal", "reduceOptions": {"calcs": ["lastNotNull"], "fields": "", "values": false}, "showUnfilled": true},
+      "targets": [{"editorMode": "code", "expr": "topk(5, 100 * sum(rate(openrag_http_request_failures_total[$__rate_interval])) by (method, endpoint) / clamp_min(sum(rate(openrag_http_requests_total[$__rate_interval])) by (method, endpoint), 1e-9))", "instant": true, "legendFormat": "{{method}} {{endpoint}}", "range": false, "refId": "A"}], "title": "Top 5 failing routes", "type": "bargauge"
```
Ahmath-Gadji, line 62: This panel reads `GET /-unresolved-` at 100% on healthy traffic, with 5xx flat at zero.

Not caused by this PR — it reproduces on develop. `scope["route"]` is only set by FastAPI's `APIRoute`, so Socket.IO long-polling, 404s, slash redirects and pre-routing auth rejections all collapse into that one label; and the failure counter includes ordinary browser 401/403 probes.

Root-caused and tracked in #905, which is a draft and will be reworked after this merges, since it touches this same dashboard.

infra/compose/docker-compose.yaml
```diff
@@ -125,7 +125,7 @@ services:
         # to CORS_EXTRA_ORIGINS on the API.
         VITE_API_BASE_URL: ${VITE_API_BASE_URL:-}
         VITE_BASE_PATH: "/app/"
-        # Optional Grafana dashboard link on the admin "System" page; branding name.
+        # Build-time Grafana fallback; prefer runtime GRAFANA_URL in .env.
```
Ahmath-Gadji: *(Edited — my first version of this comment claimed the three docs disagreed on why `VITE_GRAFANA_URL` exists. That was wrong: `ui/.env.example`, `ui/README.md` and `env_vars.md` all point production at `GRAFANA_URL`, and `env_vars.md` already tells new deployments to use it. I'd misread the README by reading only the table cell and skipping the paragraph under it.)*

What still seems worth a look: the scenario named in `ui/.env.example` — "Build-time fallback for frontend-only development" — is obsolete as of this PR, since `grafana_url` is now in the MSW handlers, so `npm run dev` already gets it from the mocked `/config`. Version skew, the reason `env_vars.md` gives, is the one that actually holds.

Separately: `VITE_APP_NAME` is display config that `/config` could carry the same way `grafana_url` now does — not for this PR, just noting the pattern generalises.

infra/compose/prometheus/prometheus.yml
```diff
@@ -21,8 +21,4 @@ scrape_configs:
       type: Bearer
       credentials_file: /etc/prometheus/openrag_token
     static_configs:
-      - targets: ["host.docker.internal:8000"]
-
-  - job_name: "ray"
-    static_configs:
-      - targets: ["host.docker.internal:8080"]
+      - targets: ["openrag:8080"]
```
Ahmath-Gadji, line 24: Follow-up idea, not for this PR: sub-service health (Milvus, Postgres, etc.) is nearly free here. Prometheus synthesises `up{job="..."}` per target, so a health tile is a by-product of adding a scrape job — no application code needed.

Milvus and etcd need nothing beyond this. I checked on a running stack: both already serve Prometheus on the compose network, unauthenticated (12.8k and 1.5k series respectively).

```yaml
  - job_name: "milvus"
    static_configs:
      - targets: ["milvus:9091"]

  - job_name: "etcd"
    static_configs:
      - targets: ["etcd:2379"]
```

MinIO and Postgres cost a bit more — MinIO returns 403 without a scrape token, and Postgres has no HTTP endpoint so it needs an exporter sidecar. Both doable, just not one-liners.

hedhoud: @Ahmath-Gadji, This is a useful follow-up. Let’s handle Milvus and etcd health scraping in a separate PR so this PR stays focused on the Grafana link and HTTP dashboard changes.

Ahmath-Gadji (approved):
Both points from my earlier review are addressed in `a53b63ee`. Re-verified rather than read:

**Control characters in `GRAFANA_URL`** — the exact vector I reported, `/<TAB>/evil.example/x`, now returns `None`, as do LF, CR, VT and 0x1F, while `/grafana/d/...`, `http://localhost:3000/d/x` and whitespace-padded values still resolve and `http://[` / bad ports still reject. `< 0x20` is the right bound: WHATWG URL parsing strips only U+0009/000A/000D, so C0 is a superset of what a browser can be tricked with. DEL (0x7F) is correctly *not* rejected — it percent-encodes into the path instead of being stripped.

**Build-time fallback** — keying `hasRuntimeGrafanaConfig` off property presence rather than truthiness makes the runtime value authoritative, since `/config` emits `grafana_url` unconditionally. An invalid `GRAFANA_URL` now reaches the not-configured dialog instead of opening the stale baked-in URL, and the build-time value survives only where it was meant to: an older API that omits the key, or a failed `/config`. That resolves the contradiction with `env_vars.md`, which is now accurate as written.

One non-blocking note for later, not worth holding this up: `does not fall back to the build-time URL when runtime Grafana is invalid` cannot currently fail. `ui/` has no `.env` and `setupFiles` is empty, so `VITE_GRAFANA_URL` is undefined under vitest and `BUILD_TIME_GRAFANA_URL` is `""` — both branches yield the dialog. Reverting only `system.tsx` to the pre-fix version still gives 5 passed. Adding `env: { VITE_GRAFANA_URL: "http://build-time.example/d/stale" }` to the `test` block in `ui/vite.config.ts` turns it into a real regression test: fixed → 5 passed, pre-fix → 2 failed. The shipped behaviour is correct either way; this only affects whether a future refactor can silently undo it.

The two remaining threads were follow-ups by construction — sub-service health scraping is agreed for its own PR, and the `/-unresolved-` dashboard label is tracked in #905. One correction there: #905 is no longer a draft stacked on this branch. It is now a code-only PR off `develop` and deliberately does not touch `openrag-http.json`; I am holding the dashboard edit back until this merges.

LGTM.

Issue #12, opened 2025-07-07
Asyncfy SQLAlchemy calls + Postgres


andyne13:
It looks like asynchronous operations are supported in SQLAlchemy through the following extension:

```python
from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, create_async_engine
```

You can also switch to PostgreSQL or another database by simply updating the database URL in your SQLAlchemy configuration. For example, in the Ragondin code:

```python
# Code in RAGondin (PartitionFileManager uses SQLAlchemy under the hood)
self.partition_file_manager = PartitionFileManager(
    database_url=f"sqlite:///{self.db_dir}/partitions_for_collection_{name}.db",
    logger=self.logger,
)
```

Replacing the SQLite URL with a PostgreSQL one (e.g., `postgresql+asyncpg://...`) would allow you to use a different backend with minimal changes.

dodekapod:
SQLite is replaced by PostgreSQL in PR https://github.com/linagora/openrag/pull/17 .

Calls are still synchronous because we need the actions to be completed sequentially. Parallelism is achieved at the Ray level.

PR #809, opened 2026-07-27
Add search to user management

### LINKED ISSUE #776: Admin UI: add search to user management
Context

Administrators need to find a specific account quickly as the user list grows.

Problem

The current user view requires manually scanning or paging through the list, which makes routine account management slow and error-prone.

Expected behavior

- Administrators can search users by name and other visible identifiers.
- Search works with the existing pagination and permissions.
- Empty and no-result states are clear.

hedhoud:
Closes #776

## Context

Administrators need to find an account quickly as the user directory grows. Scanning pages manually slows routine account management and makes selecting the wrong account more likely.

## Change

The user-management view now filters the complete authorized user list by display name, email, external identity, or displayed user number. Search works across the existing pagination, reports the result count, can be cleared in one action, and distinguishes an empty directory from a search with no matches.

The existing admin-only permissions and user API remain unchanged.

## Validation

All 161 Admin UI tests pass, including coverage for searches across paginated rows and the empty states. The production UI build succeeds, and linting reports no new warnings or errors.

## Summary by CodeRabbit

* **New Features**
  * Added client-side, case-insensitive search to the admin users list with live counts, clear-search, and distinct “no users yet” vs “no matches” messaging.
  * Partition member listings now show user display name and email, including “Not available” handling when missing.
  * Data tables can now show custom empty-state text and optionally reset pagination.
* **Bug Fixes**
  * Legacy user listings now include email information.
* **Tests**
  * Expanded UI and backend test coverage for search, empty/error states, pagination reset, batched user fetching, and identity enrichment.
* **Documentation**
  * Updated API and TypeScript descriptions for partition member fields and endpoint responses.

### DIFF
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -674,7 +674,10 @@ class PartitionService:
 
     async def list_members(self, partition: str) -> list[dict]:
         await self._ensure_partition(partition)
-        members = await self._membership_repo.list_partition_members(partition)
+        return await self._membership_repo.list_partition_members(partition)
+
+    async def list_members_with_identities(self, partition: str) -> list[dict]:
+        members = await self.list_members(partition)
         users = {
             user.id: user for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
         }
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -676,8 +676,7 @@ class PartitionService:
         await self._ensure_partition(partition)
         members = await self._membership_repo.list_partition_members(partition)
         users = {
-            user.id: user
-            for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
+            user.id: user for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
         }
         for member in members:
             user = users.get(member["user_id"])
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -674,7 +674,11 @@ class PartitionService:
 
     async def list_members(self, partition: str) -> list[dict]:
         await self._ensure_partition(partition)
-        return await self._membership_repo.list_partition_members(partition)
+        members = await self._membership_repo.list_partition_members(partition)
+        users = await asyncio.gather(*(self._user_repo.get_user(m["user_id"]) for m in members))
+        for member, user in zip(members, users, strict=True):
+            member["display_name"] = user.display_name if user else None
+        return members
 
     async def add_member(self, partition: str, user_id: int, role: str) -> None:
         await self._ensure_partition(partition)
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -675,9 +675,14 @@ class PartitionService:
     async def list_members(self, partition: str) -> list[dict]:
         await self._ensure_partition(partition)
         members = await self._membership_repo.list_partition_members(partition)
-        users = await asyncio.gather(*(self._user_repo.get_user(m["user_id"]) for m in members))
-        for member, user in zip(members, users, strict=True):
+        users = {
+            user.id: user
+            for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
+        }
+        for member in members:
+            user = users.get(member["user_id"])
             member["display_name"] = user.display_name if user else None
+            member["email"] = user.email if user else None
         return members
 
     async def add_member(self, partition: str, user_id: int, role: str) -> None:
--- a/openrag/services/orchestrators/partition_service.py
+++ b/openrag/services/orchestrators/partition_service.py
@@ -673,10 +673,12 @@ class PartitionService:
     # ------------------------------------------------------------------
 
     async def list_members(self, partition: str) -> list[dict]:
+        """Return role data without identity lookups for authorization callers."""
         await self._ensure_partition(partition)
         return await self._membership_repo.list_partition_members(partition)
 
     async def list_members_with_identities(self, partition: str) -> list[dict]:
+        """Enrich the admin-facing member list with one bulk user lookup."""
         members = await self.list_members(partition)
         users = {
             user.id: user for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
--- a/openrag/services/persistence/user_repo.py
+++ b/openrag/services/persistence/user_repo.py
@@ -332,6 +332,7 @@ class PgUserRepository(UserRepository):
                 "id": r["id"],
                 "display_name": r["display_name"],
                 "external_user_id": r["external_user_id"],
+                "email": r["email"],
                 "is_admin": r["is_admin"],
                 "file_quota": r["file_quota"],
                 "file_count": r["file_count"],
--- a/openrag/services/persistence/user_repo.py
+++ b/openrag/services/persistence/user_repo.py
@@ -102,6 +102,15 @@ class PgUserRepository(UserRepository):
         memberships = await self._fetch_memberships(user_id)
         return self._row_to_user(row, memberships)
 
+    async def get_users_by_ids(self, user_ids: list[int]) -> list[User]:
+        if not user_ids:
+            return []
+        rows = await self.pool.fetch(
+            "SELECT * FROM users WHERE id = ANY($1::int[])",
+            user_ids,
+        )
+        return [self._row_to_user(row) for row in rows]
+
     async def get_user_by_email(self, email: str) -> User | None:
         row = await self.pool.fetchrow(
             "SELECT * FROM users WHERE email = $1",
--- a/tests/unit/services/persistence/test_user_repo_external_id.py
+++ b/tests/unit/services/persistence/test_user_repo_external_id.py
@@ -53,6 +53,8 @@ class _FakePool:
     async def fetch(self, query: str, *params):
         self.last_query = query
         self.last_params = params
+        if "partition_memberships" in query:
+            return []
         return self._rows
 
 
@@ -204,6 +206,6 @@ async def test_get_users_by_ids_fetches_all_users_in_one_query():
 
     users = await repo.get_users_by_ids([42, 84])
 
-    assert [user.id for user in users] == [42, 84]
+    assert {user.id for user in users} == {42, 84}
     assert pool.last_params == ([42, 84],)
     assert "ANY($1::int[])" in (pool.last_query or "")
--- a/tests/unit/services/persistence/test_user_repo_external_id.py
+++ b/tests/unit/services/persistence/test_user_repo_external_id.py
@@ -30,9 +30,13 @@ class _FakePool:
         self.last_query: str | None = None
         self.last_params: tuple = ()
         self._next_row: _FakeRow | None = None
+        self._rows: list[_FakeRow] = []
 
     def set_next_row(self, **fields):
         self._next_row = _FakeRow(fields)
+
+    def set_rows(self, *rows: _FakeRow):
+        self._rows = list(rows)
 
     async def fetchrow(self, query: str, *params):
         self.last_query = query
@@ -49,7 +53,7 @@ class _FakePool:
     async def fetch(self, query: str, *params):
         self.last_query = query
         self.last_params = params
-        return []
+        return self._rows
 
 
 def _make_user_with_ext(ext: str | None):
@@ -141,3 +145,27 @@ async def test_create_legacy_user_coerces_empty_external_id_to_none():
     )
     # Same column position (display_name, external_user_id, ...)
     assert pool.last_params[1] is None
+
+
+@pytest.mark.asyncio
+async def test_list_users_dict_includes_email():
+    from services.persistence.user_repo import PgUserRepository
+
+    pool = _FakePool()
+    pool.set_rows(
+        _FakeRow(
+            id=42,
+            display_name="Alice",
+            external_user_id="kc-alice",
+            email="alice@example.com",
+            is_admin=False,
+            file_quota=None,
+            file_count=0,
+            created_at=__import__("datetime").datetime(2026, 1, 1),
+        )
+    )
+    repo = PgUserRepository(pool_getter=lambda: pool)
+
+    users = await repo.list_users_dict()
+
+    assert users[0]["email"] == "alice@example.com"
--- a/tests/unit/services/persistence/test_user_repo_external_id.py
+++ b/tests/unit/services/persistence/test_user_repo_external_id.py
@@ -169,3 +169,41 @@ async def test_list_users_dict_includes_email():
     users = await repo.list_users_dict()
 
     assert users[0]["email"] == "alice@example.com"
+
+
+@pytest.mark.asyncio
+async def test_get_users_by_ids_fetches_all_users_in_one_query():
+    from services.persistence.user_repo import PgUserRepository
+
+    pool = _FakePool()
+    pool.set_rows(
+        _FakeRow(
+            id=42,
+            display_name="Alice",
+            external_user_id="kc-alice",
+            email="alice@example.com",
+            token=None,
+            is_admin=False,
+            file_quota=None,
+            file_count=0,
+            created_at=__import__("datetime").datetime(2026, 1, 1),
+        ),
+        _FakeRow(
+            id=84,
+            display_name="Bob",
+            external_user_id="kc-bob",
+            email="bob@example.com",
+            token=None,
+            is_admin=False,
+            file_quota=None,
+            file_count=0,
+            created_at=__import__("datetime").datetime(2026, 1, 2),
+        ),
+    )
+    repo = PgUserRepository(pool_getter=lambda: pool)
+
+    users = await repo.get_users_by_ids([42, 84])
+
+    assert [user.id for user in users] == [42, 84]
+    assert pool.last_params == ([42, 84],)
+    assert "ANY($1::int[])" in (pool.last_query or "")
--- a/ui/src/pages/admin/users/list.tsx
+++ b/ui/src/pages/admin/users/list.tsx
@@ -1,9 +1,9 @@
-import { useState } from "react";
+import { useMemo, useState } from "react";
 import { Link } from "react-router-dom";
 import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
 import { toast } from "sonner";
 import type { ColumnDef } from "@tanstack/react-table";
-import { Trash2, Eye, Plus, Copy } from "lucide-react";
+import { Trash2, Eye, Plus, Copy, Search, X } from "lucide-react";
 import { listUsers, deleteUser, createUser, effectiveQuota } from "@/lib/api/users";
 import type { UserResponse, UserWithToken } from "@/lib/api/users";
 import { getConfig } from "@/lib/api/system";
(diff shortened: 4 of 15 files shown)


hedhoud:
@codex review

ui/src/pages/admin/users/list.tsx
```diff
@@ -55,6 +58,30 @@ export default function UserListPage() {
...
     (config?.rdb as { default_file_quota?: number } | undefined)?.default_file_quota ?? null;
+  const users = data?.users ?? EMPTY_USERS;
+  const normalizedSearch = search.trim().toLowerCase();
+  const filteredUsers = useMemo(() => {
+    if (!normalizedSearch) return users;
+    return users.filter((user) =>
+      [
+        user.display_name,
+        user.external_user_id,
+        user.email,
```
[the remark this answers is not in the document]
hedhoud: Good catch. The UI contract already allowed an email value, but the production list serializer dropped it, so email search only worked against mocked data. I added email to the authorized user-list response and covered that response shape with a repository regression test. This keeps the advertised search behavior aligned with production data. Fixed in 653809f9.

ui/src/pages/admin/users/list.tsx
```diff
@@ -55,6 +58,30 @@ export default function UserListPage() {
...
+    );
+  }, [normalizedSearch, users]);
+  const totalUsersLabel = `${users.length} ${users.length === 1 ? "user" : "users"}`;
+  const resultSummary = normalizedSearch
+    ? `${filteredUsers.length} of ${totalUsersLabel}`
+    : totalUsersLabel;
+  const emptyMessage =
+    users.length === 0
+      ? "No users have been created yet."
+      : `No users match “${search.trim()}”.`;
```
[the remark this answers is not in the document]
hedhoud: Agreed. An undefined query result was being treated as a successful empty response, which could mislead an administrator during an outage. The page now renders a distinct error alert with the underlying message and a retry action; the empty-directory message is reachable only after a successful response. A regression test covers failure, avoids the false empty state, and verifies recovery after retry. Fixed in 653809f9.

ui/src/pages/admin/users/list.tsx
```diff
@@ -167,7 +194,48 @@ export default function UserListPage() {
...
+                  variant="ghost"
+                  size="icon-sm"
+                  onClick={() => setSearch("")}
+                  className="absolute right-0.5 top-1/2 -translate-y-1/2"
+                  aria-label="Clear user search"
+                  title="Clear search"
+                >
+                  <X className="h-4 w-4" />
+                </Button>
+              )}
```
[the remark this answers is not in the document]
hedhoud: Valid cross-browser issue. WebKit/Blink can render its native cancel control in the same area as our custom clear button, producing overlapping actions. The native search-cancel pseudo-element is now hidden, leaving one consistently positioned, accessible clear control. The production UI build confirms the Tailwind selector is accepted. Fixed in 653809f9.

hedhoud:
Addressed the remaining non-inline review request in 653809f9.

The displayed user number was a supported search key but had no regression coverage, so that contract could have been removed accidentally while the name/email/external-ID tests still passed. The existing cross-page search test now searches for `User #12` and verifies both the matching user and the `1 of 11 users` result summary.

Validation after all review fixes: 162 Admin UI tests passed, the production UI build passed, 32 focused backend tests passed, and Python lint passed. UI lint reports only the existing unrelated hook warning in the models page.

ui/src/pages/admin/users/list.tsx
```diff
@@ -164,10 +192,76 @@ export default function UserListPage() {
         }
       />
 
-      {isLoading ? (
+      {usersQuery.isLoading ? (
         <Skeleton className="h-64" />
+      ) : usersQuery.isError ? (
```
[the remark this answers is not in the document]
hedhoud: Valid finding. `isError` covers both initial-load and refetch failures, and TanStack Query intentionally preserves successful cached data in the latter case. I now use `isLoadingError` for the blocking state and `isRefetchError` for a non-blocking warning with retry, so the cached directory and search controls remain usable. A regression test seeds the query cache, fails the background request, and verifies that both the warning and cached user row stay visible. Fixed in 15164151; all 163 Admin UI tests and the production build pass.

hedhoud:
Integrated the partition-member identity work from #757 and extended it with email while keeping the stable user ID visible. The original per-member concurrent lookup was replaced with one bulk lookup so large partitions do not put unnecessary pressure on the database connection pool. Missing profile data still falls back safely to the user ID.\n\nValidated with the focused Python tests, the complete UI test suite, the production UI build, Ruff, and ESLint.

openrag/services/orchestrators/partition_service.py
```diff
@@ -674,7 +674,16 @@ def _meta(row: dict[str, Any]) -> dict[str, Any]:
 
     async def list_members(self, partition: str) -> list[dict]:
         await self._ensure_partition(partition)
-        return await self._membership_repo.list_partition_members(partition)
+        members = await self._membership_repo.list_partition_members(partition)
+        users = {
+            user.id: user
+            for user in await self._user_repo.get_users_by_ids([member["user_id"] for member in members])
```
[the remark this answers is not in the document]
Ahmath-Gadji: Confirmed — this is a real regression, and the fix should be a split rather than a tweak.

`list_members()` has three callers, and only one of them wants identities:

- `openrag/api/routers/admin/partitions.py:406` — the admin UI, needs `display_name`/`email`
- `openrag/services/orchestrators/mcp_service.py:156` (`_enforce_editor_access`, reached from 5 MCP write paths: delete, metadata update, copy, URL index) — reads `role` only
- `openrag/services/orchestrators/indexing_service.py:158` (`_ensure_editor_access_after_create_race`) — reads `role` only

So every non-admin MCP write now pays an extra `SELECT * FROM users WHERE id = ANY(...)` plus N `User` model constructions purely to throw them away. (Minor correction to the report: `_row_to_user` never reads the `token` column, so nothing sensitive is materialised — the cost is the round trip and the model building, not a leak.)

Fixing by keeping `list_members()` as the single-query membership read it was, and adding a separate `list_members_with_identities()` that the partitions router calls. That leaves both authorization paths at their original cost and keeps enrichment where it is actually consumed.

ui/src/pages/admin/users/list.tsx
```diff
@@ -164,10 +228,70 @@ export default function UserListPage() {
...
+            </div>
+            <p
+              role="status"
+              aria-live="polite"
+              className="text-sm text-muted-foreground tabular-nums"
+            >
+              {resultSummary}
+            </p>
+          </div>
+          <DataTable columns={columns} data={filteredUsers} emptyMessage={emptyMessage} />
```
Ahmath-Gadji: **Bug: searching from a non-first page leaves the table on that page of the results**

`DataTable` sets `autoResetPageIndex: false` (`ui/src/components/shared/data-table.tsx:123`) so that deleting a row does not bounce the admin back to page 1. Because the search filters `data` in the parent rather than going through the table filter, that opt-out now also applies to the search: changing the query swaps the row set underneath a stale `pageIndex`.

Verified against this branch with a throwaway test — 100 users, navigate to page 5, then type a query that matches all of them:

```
page:      "Page 5 of 10"
first row: "Match 41"
```

The admin types a search and lands in the middle of the matches. When the query narrows the set instead, the clamp effect at `data-table.tsx:131-135` pins them to the *last* page of results, so `status` reads e.g. `12 of 100 users` while the table shows only the final two. Either way the first match is not what they see, which reads as "my search did not work" — exactly the scan-the-list problem #776 set out to remove.

The row set changing because the query changed is a different event from the row set changing because a row was deleted, and only the first should reset the page. Suggest an opt-in prop on `DataTable` (e.g. `resetPageOnChange`) that the users page feeds the normalized query, leaving the delete behaviour untouched.

`ui/src/pages/admin/users/list.tsx:293`

openrag/services/persistence/user_repo.py
```diff
@@ -332,6 +341,7 @@ async def list_users_dict(self) -> list[dict]:
                 "id": r["id"],
                 "display_name": r["display_name"],
                 "external_user_id": r["external_user_id"],
+                "email": r["email"],
```
Ahmath-Gadji, line 344: **Nit: the `GET /users/` contract gained a field but its documented response did not**

Adding `email` here changes the public payload of `GET /users/`, whose OpenAPI description (`openrag/api/routers/admin/users.py:36-44`) still enumerates only `id`, `display_name`, `external_user_id`, `is_admin`, `created_at`. `GET /users/{user_id}` has the same list (`users.py:139-147`) and its serializer `get_user_dict_by_id` already returned `email` before this PR, so that one was stale on arrival.

Worth closing here because this PR set the opposite precedent two files over: `partitions.py` was updated in the same change to document the new `display_name`/`email` on the members response. Consumers reading the generated docs should not have to discover `email` by calling the endpoint.

`openrag/services/persistence/user_repo.py:344`

openrag/services/orchestrators/partition_service.py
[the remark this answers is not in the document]
Ahmath-Gadji: Leaving this as a remark for you to action rather than pushing a change.

The shape I'd suggest is a split rather than a tweak: `list_members()` goes back to the single membership query it was before this PR, and a new `list_members_with_identities()` — sole caller `GET /partition/{p}/users` — does the bulk lookup and enrichment. That leaves the five MCP write paths and the indexing create-race check at their original cost while the admin UI still gets the identities.

Worth pinning the property with a test that asserts the user repo recorded no lookup when `list_members()` is called, so a future caller cannot quietly reintroduce the query on the authorization path.

_(I had this pushed briefly as `3e7a7e67`; it has been dropped from the branch — the call is yours.)_

ui/src/pages/admin/users/list.tsx
Ahmath-Gadji: Leaving this as a remark for you to action rather than pushing a change.

What I'd suggest: an opt-in prop on `DataTable` — e.g. `resetPageOnChange` — that the users page feeds `normalizedSearch`. A query change then returns to the first page while a delete still leaves the admin where they were, so `autoResetPageIndex: false` stays as-is and no other table changes behaviour.

Two things worth covering if you take it: clearing the search is a query change too and needs the same reset, and the regression test should be checked against the current commit first — it has to fail on `Page 3 of 3` before it means anything.

_(I had this pushed briefly as `2ee700eb`; it has been dropped from the branch — the call is yours.)_

openrag/services/persistence/user_repo.py
Ahmath-Gadji: Fixed in https://github.com/linagora/openrag/commit/bd9e494e.

`email` is now listed in the documented response for both `GET /users/` and `GET /users/{user_id}`, matching what this PR already did for the partition-members endpoint.

Ahmath-Gadji:
### Review pass — 1 fix, 2 remarks, 1 observation

Reviewed the full diff and ran the suites locally against the branch: 2215 backend unit tests, 168 UI tests, `tsc -b`, ESLint and Ruff all pass.

One commit pushed — [`bd9e494e`](https://github.com/linagora/openrag/commit/bd9e494e), which adds `email` to the documented response of `GET /users/` and `GET /users/{user_id}`. The payload gained the field here but the OpenAPI descriptions still listed only `id`, `display_name`, `external_user_id`, `is_admin`, `created_at`, and this PR had already set the opposite precedent by updating the partition-members description in the same change.

The two substantive findings are left as remarks in their threads rather than as pushed changes — both are design calls that belong to you:

1. **`list_members()` charges every authorization check for identity data it never reads.** Three callers share the method; only `GET /partition/{p}/users` displays members. `_enforce_editor_access` (5 MCP write paths) and the indexing create-race check read `role` alone, and now pay an extra `SELECT * FROM users WHERE id = ANY(...)` plus N discarded model constructions per request.
2. **Searching from a later page leaves the table on that page of the results.** `autoResetPageIndex: false` is deliberate for deletes, but the search filters `data` in the parent, so it applies there too. Verified on the branch: 100 users, navigate to page 5, type a matching query, and the first row is `Match 41` under `Page 5 of 10`. When the query narrows instead, the clamp at `data-table.tsx:131-135` pins you to the *last* page, so the status reads `12 of 100 users` while two rows show. Either way the first match is off screen.

The rest of the change reads well. The bulk `get_users_by_ids` lookup is the right shape for the members list, `_row_to_user` never touches the `token` column so nothing sensitive is materialised, and the missing-user fallback to `None` identities is handled on both the service and component sides.

**Observation, not a blocker and not in scope here:** no workflow runs the `ui/` suite — `astro.yml` is the docs site, and the other nine are Python. The 168 vitest tests, including the ones this PR adds, are only ever green because a human ran them. Worth a follow-up issue.

openrag/services/orchestrators/partition_service.py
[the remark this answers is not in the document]
Ahmath-Gadji: Verified on `0d975fc7` — this is resolved.

`list_members()` is back to the single membership query, and the call graph now separates cleanly:

```
partitions.py:406        → list_members_with_identities   (admin UI)
mcp_service.py:156       → list_members                   (role only)
indexing_service.py:158  → list_members                   (role only)
```

I checked the guard test is not vacuous: collapsing the split back into one method makes `test_list_members_does_not_lookup_user_identities` fail. `requested_user_id_batches` is a better fake than the single-assignment version I had suggested — recording every batch and asserting `len(...) == 1` against a two-member partition catches an N+1 regression, not just the presence or absence of a lookup.

One small thing left, non-blocking: neither method carries a docstring, so nothing in the code says *why* they are separate. The next person adding a field to the members response could reasonably put it back into `list_members`. The test does guard it, which is arguably stronger than a comment — but a one-liner naming the authorization callers would put the constraint where someone is tempted to break it.

ui/src/pages/admin/users/list.tsx
Ahmath-Gadji: Verified on `0d975fc7` — this is resolved.

`pageResetKey` does the job and leaves `autoResetPageIndex: false` untouched, so delete behaviour is unchanged. Confirmed the test is not vacuous: removing `pageResetKey={normalizedSearch}` from the users page makes `returns to the first page when search changes or is cleared` fail. Good that it covers clearing as well as typing — that was the case most likely to be missed.

tests/unit/services/persistence/test_user_repo_external_id.py
```diff
@@ -141,3 +145,65 @@ async def test_create_legacy_user_coerces_empty_external_id_to_none():
...
+            file_quota=None,
+            file_count=0,
+            created_at=__import__("datetime").datetime(2026, 1, 2),
+        ),
+    )
+    repo = PgUserRepository(pool_getter=lambda: pool)
+
+    users = await repo.get_users_by_ids([42, 84])
+
+    assert [user.id for user in users] == [42, 84]
```
Ahmath-Gadji: **Nit: this asserts an ordering the query does not guarantee**

`get_users_by_ids` runs `SELECT * FROM users WHERE id = ANY($1::int[])` with no `ORDER BY`, so Postgres is free to return these rows in any order — `ANY` does not preserve the array's order. What passes here is `_FakePool`'s insertion order, not a property of the code under test.

Harmless in production, since `list_members_with_identities` immediately keys the result by `user.id`. The cost is in the test: it would go red on a harmless plan change and stay green through a real reordering bug. `assert {user.id for user in users} == {42, 84}` says what the method actually promises.

`tests/unit/services/persistence/test_user_repo_external_id.py:207`

tests/unit/services/persistence/test_user_repo_external_id.py
```diff
@@ -49,7 +53,7 @@ async def execute(self, query: str, *params):
     async def fetch(self, query: str, *params):
         self.last_query = query
         self.last_params = params
-        return []
+        return self._rows
```
Ahmath-Gadji, line 58: **Nit: `fetch` now answers every query with the same rows**

This used to return `[]` unconditionally, so it was inert. Now any `fetch` on the fake replays whatever `set_rows` was given — including calls the test never intended to stub.

`get_user` and `get_user_by_email` both reach `_fetch_memberships`, which goes through `fetch`. A future test that calls `set_rows` and then exercises one of those would get user rows back as *membership* rows, and `_row_to_user_partition` would fail on the missing `partition_name` — from a fixture that looks unrelated to the assertion.

Not reachable from the tests here; flagging because the fake is shared and the failure would be confusing when it lands. Keying on the query (`"partition_memberships" in query`) or resetting `_rows` after a read would keep it honest.

`tests/unit/services/persistence/test_user_repo_external_id.py:56`

Ahmath-Gadji:
### Re-review of `0d975fc7` — both findings resolved

Both remarks are addressed and verified. Details are in their threads; in short:

- **`list_members()`** is back to the single membership query, with `list_members_with_identities()` serving the admin UI. The two authorization callers are at their original cost.
- **`pageResetKey`** returns the table to the first page on a query change, including when the search is cleared, and leaves `autoResetPageIndex: false` untouched so delete behaviour is unchanged.

I checked both new tests actually catch their bug rather than passing vacuously — collapsing the service split, and removing `pageResetKey` from the users page, each turn the corresponding test red.

Full suites on this head: 2214 backend unit tests, 168 UI tests, `tsc -b` and ESLint pass (the one ESLint warning is the pre-existing `models.tsx` hook dependency). All six CI checks are green.

Two small nits posted inline on the test fake, neither blocking. Nothing else outstanding from my side.

**One thing worth a decision before merge:** `3ef20eb3` is titled "init". This repo merges rather than squashes — the five most recent commits on `develop` are all `Merge pull request #…` — so that message lands in the permanent history as-is. Either reword it or squash-merge this PR.

tests/unit/services/persistence/test_user_repo_external_id.py
Ahmath-Gadji: Verified on `251586f5` — resolved. The set comparison now states what the method promises rather than what the fake happens to return.

tests/unit/services/persistence/test_user_repo_external_id.py
Ahmath-Gadji: Verified on `251586f5` — resolved, and the guard is doing real work.

I checked the scenario directly: a throwaway test that calls `set_rows` and then `get_user(42)` passes with the guard in place, and without it fails with `KeyError: 'user_id'` inside `_row_to_user_partition` — the user row being replayed as a membership row, exactly the confusing failure this prevents. (I guessed `partition_name` above; `user_id` is simply read first.)

Ahmath-Gadji (approved):
Approving. Every finding I raised is addressed and verified on `251586f5`, and all six CI checks are green.

For the record, what I checked rather than assumed: both fix commits were tested for vacuity — collapsing the `list_members` split, removing `pageResetKey`, and dropping the `partition_memberships` guard each turn the corresponding test red, so none of them pass by accident. Locally on this head: 2214 backend unit tests, 168 UI tests, `tsc -b` and ESLint pass.

---

### One thing to settle before merge: the `init` commit

`3ef20eb3` is titled "init". It is not a scratch commit — it is the original partition-member display-name work, authored by Baptiste (`bnoleau@ik.me`), and it is the seed of the identity feature this PR builds on. It still exists independently as **#757**, which is open.

That makes the obvious fix the wrong one. This repo merges rather than squashes, so the message lands in `develop` verbatim — but **squash-merging to hide it would collapse Baptiste's authorship into a single commit attributed to this PR's author**. The commit is worth attributing; only its message is the problem.

Suggested instead:

1. Reword `3ef20eb3` in place (`git rebase`, keeping the `Author:` line intact) to something like *"Show partition member display name in the partitions UI"*. Authorship is preserved, history stays readable.
2. Close **#757** with a note pointing at this PR, so it does not sit stale after the content merges here.

Neither is blocking from my side — flagging because both are easy to miss and awkward to undo after the merge.

PR #810, opened 2026-07-27
Select partition members by user name

### LINKED ISSUE #777: Admin UI: select partition members by user name
Context

Partition access is assigned to people, but administrators usually know a person's name rather than their internal user ID.

Problem

Entering or reading raw user IDs makes membership management difficult and increases the risk of granting access to the wrong account.

Expected behavior

- Partition membership uses a searchable checkbox or multi-select list of users.
- Each option shows the user's name and enough identity information to distinguish duplicate names.
- Existing roles and permission checks remain unchanged.

hedhoud:
## Why

Partition access is granted to people, but partition owners usually know a name rather than an internal numeric ID. Requiring raw IDs makes the workflow difficult to verify and increases the risk of granting access to the wrong account.

## What changed

The add-member dialog now supports targeted search by display-name prefix and matching ASCII user ID. Results include email when available so duplicate names can be distinguished, selected identities remain visible, and one role can be assigned to multiple people.

Candidate search is paginated and remains stable while memberships change. Concurrent additions cannot overwrite an existing role, and refresh, paging, and per-user failures remain visible and retryable.

## Compatibility

Adding a member is now insert-only. Posting an existing member returns `409` and preserves their current role; callers must use the existing role-update endpoint to change it.

## Trust model

Search gating and pagination prevent the UI from loading the complete user directory at once. They are load and usability controls, not a privacy boundary. Partition owners are trusted to discover the identities needed to manage access.

## Validation

- Admin UI: 179 tests passed
- Backend unit suite: 2234 tests passed
- Production Admin UI build
- Python and Admin UI lint
- Alembic migration graph has one head

Closes #777

Related to #757 and #809.

### DIFF
--- a/docs/content/docs/documentation/API.mdx
+++ b/docs/content/docs/documentation/API.mdx
@@ -383,9 +383,12 @@ Manage who can access a partition and with which role — all **owner** only.
 | Endpoint | Method | Description |
 |----------|--------|-------------|
 | `/partition/{partition}/users` | GET | List members → `{ "members": [...] }` |
-| `/partition/{partition}/users` | POST | Add a member — form fields `user_id` (int), `role` (default `viewer`) → `201` |
+| `/partition/{partition}/users/candidates` | GET | Search non-members by display-name prefix or exact user ID → paginated identities |
+| `/partition/{partition}/users` | POST | Add a new member → `201`; returns `409` if already present |
 | `/partition/{partition}/users/{user_id}` | PATCH | Update a member's role — form field `role` → `200` |
 | `/partition/{partition}/users/{user_id}` | DELETE | Remove a member → `204` |
+
+`POST` no longer updates an existing member's role. Use the `PATCH` endpoint when changing a role.
 
 #### Document Relationships
 
--- a/openrag/api/routers/admin/partitions.py
+++ b/openrag/api/routers/admin/partitions.py
@@ -418,7 +418,8 @@ async def list_partition_users(
 - `limit`: Page size (maximum 100)
 
 **Response:**
-Returns a bounded page of non-member users and continuation metadata.
+Returns a bounded page of non-member users with their display name and email,
+plus continuation metadata.
 
 **Permissions:**
 - Requires partition owner role
--- a/openrag/api/routers/admin/partitions.py
+++ b/openrag/api/routers/admin/partitions.py
@@ -405,6 +405,30 @@ async def list_partition_users(
     return JSONResponse(status_code=status.HTTP_200_OK, content={"members": members})
 
 
+@router.get(
+    "/{partition}/users/candidates",
+    description="""List users who can be added to a partition.
+
+**Parameters:**
+- `partition`: The partition name
+
+**Response:**
+Returns non-member users with their display name and stable user ID.
+
+**Permissions:**
+- Requires partition owner role
+""",
+)
+async def list_partition_user_candidates(
+    partition: str,
+    partition_owner=Depends(require_partition_owner),
+    service=Depends(get_partition_service),
+):
+    """List users who are not already members of the given partition."""
+    candidates = await service.list_member_candidates(partition=partition)
+    return JSONResponse(status_code=status.HTTP_200_OK, content={"candidates": candidates})
+
+
 @router.post(
     "/{partition}/users",
     description="""Add a user to a partition with a specific role.
--- a/openrag/api/routers/admin/partitions.py
+++ b/openrag/api/routers/admin/partitions.py
@@ -411,9 +411,12 @@ async def list_partition_users(
 
 **Parameters:**
 - `partition`: The partition name
+- `search`: Optional display name or user ID fragment
+- `offset`: Page offset
+- `limit`: Page size (maximum 100)
 
 **Response:**
-Returns non-member users with their display name and stable user ID.
+Returns a bounded page of non-member users and continuation metadata.
 
 **Permissions:**
 - Requires partition owner role
@@ -421,12 +424,20 @@ Returns non-member users with their display name and stable user ID.
 )
 async def list_partition_user_candidates(
     partition: str,
+    search: str | None = Query(default=None, max_length=200),
+    offset: int = Query(default=0, ge=0),
+    limit: int = Query(default=25, ge=1, le=100),
     partition_owner=Depends(require_partition_owner),
     service=Depends(get_partition_service),
 ):
-    """List users who are not already members of the given partition."""
-    candidates = await service.list_member_candidates(partition=partition)
-    return JSONResponse(status_code=status.HTTP_200_OK, content={"candidates": candidates})
+    """List a searchable page of users who are not partition members."""
+    page = await service.list_member_candidates(
+        partition=partition,
+        search=search,
+        offset=offset,
+        limit=limit,
+    )
+    return JSONResponse(status_code=status.HTTP_200_OK, content=page)
 
 
 @router.post(
--- a/openrag/api/routers/admin/partitions.py
+++ b/openrag/api/routers/admin/partitions.py
@@ -411,8 +411,8 @@ async def list_partition_users(
 
 **Parameters:**
 - `partition`: The partition name
-- `search`: Optional display name or user ID fragment
-- `offset`: Page offset
+- `search`: Display-name prefix (at least 3 characters) or exact user ID
+- `cursor`: Last user ID from the previous page
 - `limit`: Page size (maximum 100)
 
 **Response:**
@@ -424,8 +424,8 @@ Returns a bounded page of non-member users and continuation metadata.
 )
 async def list_partition_user_candidates(
     partition: str,
-    search: str | None = Query(default=None, max_length=200),
-    offset: int = Query(default=0, ge=0),
+    search: str = Query(..., max_length=200),
+    cursor: int | None = Query(default=None, ge=0, le=2_147_483_647),
     limit: int = Query(default=25, ge=1, le=100),
     partition_owner=Depends(require_partition_owner),
     service=Depends(get_partition_service),
@@ -434,7 +434,7 @@ async def list_partition_user_candidates(
     page = await service.list_member_candidates(
         partition=partition,
         search=search,
-        offset=offset,
+        cursor=cursor,
         limit=limit,
     )
     return JSONResponse(status_code=status.HTTP_200_OK, content=page)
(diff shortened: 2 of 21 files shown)


andyne13:
## Review — verified against `a5b5c55d`

Reviewed the delta against the merge-base `db3482c3` (i.e. excluding #809, which is already on `develop`). I stood up a throwaway PostgreSQL 16 with 200k users to check the query plans rather than reason about them, and compiled the declarative index DDL to compare it against the migration.

**Verdict: the backend design is solid and several things I expected to be wrong turned out to be right.** The blocking concerns are a breaking API change that isn't documented, a front-end/back-end validation mismatch, and an overclaimed security property.

---

## Verified correct

I want to be explicit about what I checked and found sound, because most of it is the hard part:

**The index actually works.** I assumed `LIKE LOWER($3) || '%'` would defeat the prefix optimisation (the pattern isn't a `Const`, and generic plans can't fold `Param`s). It does not — PG16 produces a genuine prefix range scan in *both* plan modes:

```
Bitmap Index Scan on ix_users_lower_display_name_pattern
  Index Cond: ((lower(display_name) ~>=~ 'alice') AND (lower(display_name) ~<~ 'alicf'))
```

Identical under `force_custom_plan` and `force_generic_plan`, so it survives asyncpg's prepared-statement caching. `text_pattern_ops` is the right call under a non-C collation (`en_US.utf8` here). I also probed the pathological case — a 3-char prefix matching ~200k rows, with a deep cursor — and the planner switches to a `users_pkey` scan with a filter; sub-millisecond execution, no seq scan, no sort blow-up. No performance finding.

**Migration hygiene is right.** `metadata.create_all()` runs at startup, so I compiled the new declarative `Index`:

```
CREATE INDEX ix_users_lower_display_name_pattern ON users (lower(display_name) text_pattern_ops)
```

— byte-for-byte the same shape as the migration, and the `index_exists()` guard makes re-application a no-op, per the idempotency rule. Walking every `down_revision` in the tree confirms a single head (`e5f6a7b8c9d0` ← `d4e5f6a7b8c9`, no sibling).

**The conflict handling is genuinely atomic.** `ON CONFLICT ... DO NOTHING` + inspecting the row count is decided by the database, not by a read-then-write in the service, so two concurrent adds of the same user cannot both win and cannot silently overwrite a role. That is the correct shape and the integration test pins the role-preservation behaviour.

**Keyset pagination is correct** — `u.id > $4` / `ORDER BY u.id` / no `OFFSET`, so pages stay stable while memberships change underneath. The `LIKE` escaping (backslash first, then `%` and `_`, with `ESCAPE '\'`) is in the right order and the unit test asserts it. No route-ordering conflict either: there is no `GET /{partition}/users/{user_id}` for `candidates` to shadow.

---

## 🔴 Breaking API change, not reflected in the public docs

`add_partition_member` changes from `DO UPDATE SET role = EXCLUDED.role` to `DO NOTHING`, so `POST /{partition}/users` goes from **upsert** to **insert-only + 409**. Previously, posting for an existing member changed their role and returned 201; now it fails and preserves the old role.

The route docstring is updated, but `docs/content/docs/documentation/API.mdx:386` still reads:

> `/partition/{partition}/users` | POST | Add a member — form fields `user_id` (int), `role` (default `viewer`) → `201`

Any caller currently using POST as set-or-update — scripts, provisioning, IdP sync — stops working. It does surface as a 409, so a client that checks status codes sees an error; one that doesn't just observes that the role never changed. This needs the API docs updated and a release note. I'd also mention the `PATCH` endpoint explicitly in the 409 body so callers know the migration path.

I checked: `add_partition_member` has exactly one caller (`PartitionService.add_member`), so nothing inside the codebase depends on the old upsert.

## 🟠 The front-end and back-end disagree on "is this a user ID?"

Front-end (`detail.tsx`): `/^\d+$/` — ASCII digits only.
Back-end (`partition_service.py`): `normalized_search.isdecimal()` — **any** Unicode decimal — then `int()`.

Verified:

```python
'١٢٣'.isdecimal()   # True
int('١٢٣')          # 123
# JS /^\d+$/ rejects it
```

So typing `١٢٣` (Arabic-Indic digits): the UI classifies it as a 3-character *name* prefix and renders it as a name search, while the back-end performs an exact lookup of **user ID 123** and returns an unrelated account. Silent wrong-account risk — which is precisely what this PR sets out to reduce.

Suggested fix, keeping the two sides aligned:

```python
if normalized_search.isascii() and normalized_search.isdecimal():
```

Related: because *any* all-digit search routes to the ID branch, a user whose `display_name` is numeric can never be found by name. Worth a deliberate decision rather than a side effect.

## 🟠 "No longer exposes an unfiltered user directory" doesn't hold

The PR body claims candidate discovery no longer exposes the directory. The 3-character minimum and the membership exclusion are real, but:

- the exact-ID branch returns **any** user by ID, and `users.id` is a sequential serial;
- nothing throttles a *successfully authenticated* caller. The only limiter in the stack is `AuthFailureRateLimiter` (`api/middleware/auth.py:114`), and its counter is incremented solely by `record_failure`, which is called only from `_auth_failure` (`:193`) — so an owner enumerating with a valid token never accrues a hit and is never limited;
- the prefix floor is trivially walked (`aaa`, `aab`, …).

So any partition owner can enumerate the full directory — `display_name` included — in *N* requests with a five-line script. The control is friction, not a boundary.

That may well be acceptable if partition owners are considered trusted to see who exists. But then the claim in the description is the thing to fix, and the constraint mostly costs UX. Worth stating the intended threat model explicitly, because right now the code implies a boundary it doesn't enforce.

## 🟡 Candidates omit email, which undercuts the PR's own goal

The stated motivation is that owners "know a name rather than an internal numeric ID" and that raw IDs increase "the risk of granting access to the wrong account". But the candidate payload is only `{user_id, display_name}`, and `member-picker.tsx` disambiguates duplicates with `#{user_id}` — the very identifier the PR is trying to get away from. Two people called "Jean Dupont" are still indistinguishable to an owner who doesn't know their internal IDs.

This is also asymmetric with #809, which added email to the *member* list: you can see an email after adding someone, but not at the moment you decide. Either include email in candidates or note why it's deliberately withheld.

## 🟡 `CREATE INDEX CONCURRENTLY` failure is silently unrecoverable

A failed concurrent build leaves an **invalid** index behind. `index_exists()` (`schema_helpers.py:22`) matches on name only, and SQLAlchemy's PG reflection does not filter `pg_index.indisvalid` (it's declared in `pg_catalog.py:183` but unused in the index query). So on re-run the guard sees the name, skips, and reports success — leaving an unusable index and a search path that silently degrades with no error anywhere.

Worth checking validity rather than existence, and dropping/rebuilding when invalid:

```sql
SELECT 1 FROM pg_class c JOIN pg_index i ON i.indexrelid = c.oid
WHERE c.relname = 'ix_users_lower_display_name_pattern' AND i.indisvalid
```

## 🟡 Status-string parsing in the repository

```python
return result.endswith(" 1")
```

This works (`INSERT 0 1` vs `INSERT 0 0`), but it depends on an undocumented asyncpg command-tag format to carry a correctness-critical signal — whether a membership was created. `RETURNING` says the same thing structurally:

```python
created = await conn.fetchval(
    """INSERT INTO partition_memberships (partition_name, user_id, role, added_at)
       VALUES ($1, $2, $3, NOW())
       ON CONFLICT (partition_name, user_id) DO NOTHING
       RETURNING 1""",
    partition, user_id, role,
)
...
return created is not None
```

## 🟡 A batch UX over a single-item API

`member-batch.ts` awaits one `addPartitionMember` per candidate in a sequential `for` loop. Fifty selected users is fifty serialized round trips, and the operation is non-atomic by construction — partial success is the normal case, not an edge case. The failure reporting handles it well and the 401/403 short-circuit is a nice touch, but it's worth acknowledging this is a client-side loop rather than a batch operation, and considering a real batch endpoint (or at least bounded concurrency) before the selection limit grows.

---

## Nits

- `MemberAddFailure.attempted` is set and tested but never read by any component — the attempted/not-attempted distinction reaches the user only through the message string. Either render it or drop the field.
- Validation of the same rules lives in three places: the route (`max_length`, `ge`/`le`), the service (limit/cursor bounds, the 3-char rule), and the UI regex. The route has no `min_length`, so `?search=` reaches the service to be rejected there. Harmless, but the ID/prefix rule in particular now has to be kept in sync across three files — and per the finding above, it already drifted.
- `_MAX_POSTGRES_INTEGER` is defined in the service and repeated as the literal `2_147_483_647` in the route's `Query(le=...)`.

---

Happy to re-review once the API-doc/breaking-change note and the `isdecimal` mismatch are addressed. The data-access layer here is the strongest part of the change — the index, the keyset pagination, and the atomic conflict handling all hold up under scrutiny.



---

*Edited after posting, for accuracy — no findings added or removed, both conclusions unchanged:*
- *the rate-limiting bullet: an auth-failure limiter does exist (`AuthFailureRateLimiter`); it simply never throttles a successfully authenticated caller. The original wording said there was none.*
- *the breaking-change paragraph: the failure does surface as a 409, so "silently stops working" was imprecise.*

aditykris (approved):
No issues found,

andyne13 (approved):
## Approving — every review finding verified as addressed on `9774c083`

- **Breaking change** — `API.mdx` now documents the `409` and the "use `PATCH` to change a role" note; the error message names the endpoint.
- **ID vs prefix mismatch** — `isascii() and isdecimal()`. I re-ran the full matrix: UI and backend now agree on every shape, including `١٢٣` (prefix search, no longer an exact lookup of user 123). Numeric display names became searchable as a bonus.
- **Directory exposure** — resolved as an explicit decision rather than code: the new "Trust model" section states the gating is a load/usability control, not a privacy boundary. That's the honest framing.
- **Candidate identity** — `email` is selected and rendered through `candidateSecondaryLabel()` with an ID fallback, so duplicate names are distinguishable without requiring email to exist.
- **Invalid index** — `_index_validity()` drops and rebuilds. Verified against a real PostgreSQL 16, not just the mock test: fresh → valid; re-run → no rebuild; forced `indisvalid=false` → dropped, rebuilt, valid; and the rebuilt index is used by the candidate query at 200k rows.
- **Status-string parsing** — replaced with `RETURNING 1` / `created is not None`.
- **`attempted`** — removed.

Gates run locally: 170 backend unit tests, 179 UI tests, `ruff check`, `ruff format --check`, and the layer-import guard — all green.

Two things left that I'm happy to see ship as-is: the batch add is still a sequential per-user loop, and `2_147_483_647` remains a literal in the route alongside `_MAX_POSTGRES_INTEGER`.

LGTM.

PR #815, opened 2026-07-27
feat(evaluation) 4/15: fold promptfoo output into retrieval and answer metrics

Ahmath-Gadji:
Part **4 of 15** of the split of #811. Targets `eval/03-indexing-metrics` (#814).

## What

The second half of `metrics.py`. `summarize` turns the two promptfoo `results.json` files into:

| | Metrics |
|---|---|
| **Retrieval** | hit rate, MRR, recall, context relevance |
| **Answer** | pass rate, factuality, rubric score |

plus the per-question `EvalCaseResult` detail the run page tabulates.

Ranking definitions follow the write-up already in `tests/load/automatic-evaluation-pipeline/README.md`, so these numbers mean the same thing as the offline pipeline's.

## Notable

- **Rows without `expected_file_ids` are skipped, not failed.** They are reported in `skipped_cases` and left out of hit rate / MRR / recall. Scoring them as misses would make a sparsely-annotated test set look like a broken retriever, and the count is surfaced so a near-empty ground truth cannot masquerade as a perfect score either.
- **Matching is identifier-agnostic and sanitisation-symmetric.** A document matches on either `metadata.source` or `metadata.file_id`, and *both* sides run through `sanitize_file_id` (part 2) — a test set naming `A B.pdf` matches the `A_B.pdf` the indexer stored. The document is *displayed* by `file_id`, because `source` is a server-side storage path and means nothing to the admin reading the table.
- **promptfoo's envelope is treated as unstable.** `extract_results` accepts `{"results": {"results": [...]}}` or a bare list, and every field read off a row is optional. promptfoo can drop a row on a provider error; that leaves the case unscored rather than raising mid-run.

## Review follow-up

Two review nitpicks taken, both in `metrics.py` / its tests:

- **Recall folds over a single union** rather than scanning `documents` once per expected id. Rank-sensitive hit/MRR still walks the list in order; recall doesn't care about rank, so `len(expected & retrieved) / len(expected)` states the definition more directly than the nested `any(...)` did.
- **`test_matching_survives_file_id_sanitisation_on_either_side` is parametrized.** The three `metadata` shapes are independent claims, and the manual loop reported only the first failure — hiding whether the others also broke. 34 → 36 tests in the file.

## Testing

15 further unit tests (34 in the file): first-matching-rank MRR, misses, skipped cases, fractional recall, context-relevance averaging and its `None` case, answer component scores, missing rows, and sanitisation matching from either side. `ruff`, format check and the layer-import guard pass.

## Summary by CodeRabbit

- **New Features**
  - Added evaluation result parsing for both nested and list-based result formats.
  - Added summarized retrieval metrics, including hit rate, MRR, recall, and context relevance.
  - Added answer metrics, including pass rate, factuality, and rubric scores.
  - Added per-query evaluation details and retrieved file reporting.
  - Improved handling of missing data, unmatched sources, and sanitized file identifiers.

- **Tests**
  - Expanded coverage for evaluation parsing, scoring, edge cases, and file matching.

### DIFF
--- a/openrag/core/evaluation/__init__.py
+++ b/openrag/core/evaluation/__init__.py
@@ -1,11 +1,13 @@
 """Pure evaluation logic: test-set parsing, promptfoo config, metric math."""
 
 from core.evaluation.identity import sanitize_file_id
-from core.evaluation.metrics import indexing_metrics
+from core.evaluation.metrics import extract_results, indexing_metrics, summarize
 from core.evaluation.testset import parse_testset
 
 __all__ = [
+    "extract_results",
     "indexing_metrics",
     "parse_testset",
     "sanitize_file_id",
+    "summarize",
 ]
--- a/openrag/core/evaluation/metrics.py
+++ b/openrag/core/evaluation/metrics.py
@@ -1,21 +1,42 @@
 """Metric computation for an evaluation run.
 
-Aggregates the per-file indexing timings the worker collected. Pure: the
-worker measures, this decides what the measurements mean.
+Two jobs live here, both pure:
+
+* aggregate the per-file indexing timings the worker collected;
+* turn promptfoo's ``results.json`` into ranking and answer-quality numbers.
+
+The ranking definitions (hit rate, MRR, recall) follow the write-up in
+``tests/load/automatic-evaluation-pipeline/README.md`` so the numbers this
+page reports mean the same thing as the ones the offline pipeline produced.
+
+promptfoo's output envelope varies by release, so :func:`extract_results`
+accepts either the ``{"results": {"results": [...]}}`` shape or a bare list, and
+every field read from a row is treated as optional.
 """
 
 from __future__ import annotations
 
 import math
-from collections.abc import Sequence
+import statistics
+from collections.abc import Iterable, Mapping, Sequence
 from pathlib import Path
+from typing import Any
 
+from core.evaluation.identity import sanitize_file_id
 from core.models.evaluation import (
+    AnswerMetrics,
+    EvalCaseResult,
+    EvalTestCase,
     FileIndexingSample,
     IndexingMetrics,
+    RetrievalMetrics,
 )
 
 _BYTES_PER_MB = 1024 * 1024
+
+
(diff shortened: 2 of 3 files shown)


Ahmath-Gadji:
@coderabbitai full review

Ahmath-Gadji:
Both nitpicks taken.

**Recall now folds over a single union.** `matched` (rank-sensitive, for hit/MRR) still scans `documents` in order, but recall doesn't care about rank, so it no longer needs a scan per expected id:

```python
retrieved = {identifier for _, identifiers in documents for identifier in identifiers}
recalls.append(len(expected & retrieved) / len(expected))
```

Equivalent to the old `any(...)` form and easier to read as "how much of the ground truth came back at all". Agreed it isn't a hot path — `top_k` is 5 by default — but the set expression states the definition more directly than the nested generator did.

**`test_matching_survives_file_id_sanitisation_on_either_side` is now parametrized.** Your reasoning is the right one: the three `metadata` shapes are independent claims, and a manual loop reports only the first failure, hiding whether the others also broke. Each variant is now its own test id, so a regression in the `source`-only path is visible even when the `file_id`-only path is already red. Test count in the file goes 34 → 36.

`ruff`, `ruff format --check` and the suite pass.

hedhoud (changes requested):
The metric calculation looks clear overall, but I found two cases that can make the results misleading.

Provider errors are currently counted as quality failures. For example, if a retrieval request returns an error or promptfoo drops its row, the case is counted as a retrieval miss. An answer error is also counted as a failed answer. This makes an API outage look like poor model quality. Please keep real empty results as misses, but skip or clearly report rows that failed to run. If no valid rows remain, the evaluation should fail instead of completing with misleading scores.

Results are also matched only by the question text. When the test set contains the same question twice with different expected answers, the second promptfoo result is discarded and both cases reuse the first result. Matching by a unique test ID, or rejecting duplicate questions earlier, would prevent incorrect pass rates.

Please add focused tests for provider errors, missing rows, and duplicate questions. Once those cases are covered, the metric logic should be in good shape.

PR #816, opened 2026-07-27
feat(evaluation) 5/15: generate the two promptfoo configs a run executes

Ahmath-Gadji:
Part **5 of 15** of the split of #811. Targets `eval/04-promptfoo-result-metrics` (#815). Completes `core/evaluation/` — everything after this is wiring.

## What

`build_retrieval_config` and `build_answer_config`. Pure: they return plain dicts. Serialisation and execution live in the worker (part 9).

## Notable

- **Two configs, not one.** The two questions need different endpoints: retrieval hits `GET /search/partition/{p}`, whose documents carry the chunk `content` that `context-relevance` grades and the `metadata.file_id` that feeds the ranking metrics; answers hit `POST /v1/chat/completions`, whose message content is what `factuality` and `llm-rubric` grade. Splitting them means every assertion in a config applies to that config's single provider — no assertion can run against an output shape it cannot read.
- **`transformResponse` must be a single JavaScript *expression*.** A statement or an IIFE makes promptfoo error every row before grading. This was found by running the feature end to end, not by reading the docs. The chat transform therefore extracts the message content and nothing more; retrieved sources come from the retrieval pass instead.
- **The grader is OpenRAG's own configured LLM** (`llm.base_url` / `llm.model`), so model-graded assertions need no third-party credentials. An `apiKey` is always set because the OpenAI client refuses to send without one, even though vLLM ignores it.
- **Questions go through Nunjucks' `urlencode`**, so a question containing `&` or `?` cannot corrupt the search query string.
- **Assertions are copied per test.** A shared list serialises as a YAML anchor plus aliases, which is valid but unreadable in a config an operator may need to inspect.
- **`expected_file_ids` is deliberately absent from `vars`** — no assertion reads it. The ranking metrics are computed from the retrieved ids in `summarize` (part 4), not by promptfoo.

## Testing

10 unit tests asserting the generated structure: one provider per config, the endpoints and methods, bearer headers, per-test assertion copies, `urlencode` templating, `top_k` propagation, and the grader provider. The configs are asserted structurally rather than executed — promptfoo is not available in CI.

### DIFF
--- a/openrag/core/evaluation/promptfoo_config.py
+++ b/openrag/core/evaluation/promptfoo_config.py
@@ -1,0 +1,171 @@
+"""Generation of the promptfoo configs a run executes.
+
+A run produces two configs rather than one, because the two questions need
+different endpoints:
+
+* **retrieval** hits ``GET /search/partition/{partition}``, whose documents
+  carry the chunk ``content`` — the text that ``context-relevance`` grades and
+  whose ``metadata.file_id`` feeds hit rate / MRR / recall.
+* **answer** hits ``POST /v1/chat/completions``, whose ``extra.sources`` carry
+  source metadata but no chunk text, and whose message content is what
+  ``factuality`` and ``llm-rubric`` grade.
+
+Keeping them separate means every assertion in a config applies to that
+config's single provider, so no assertion ever runs against an output shape it
+cannot read.
+
+This module is pure: it returns plain dicts. Serialisation and execution live
+in the worker.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+from core.models.evaluation import EvalTestCase
+
+#: promptfoo templates with Nunjucks; ``urlencode`` keeps a question
+#: containing ``&`` or ``?`` from corrupting the search query string.
+_QUERY_TEMPLATE = "{{ query | urlencode }}"
+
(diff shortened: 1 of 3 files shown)


Ahmath-Gadji:
@coderabbitai full review

openrag/core/evaluation/promptfoo_config.py
```diff
@@ -0,0 +1,171 @@
...
+)
+
+
+def _grader(model: str, base_url: str, api_key: str | None) -> dict[str, Any]:
+    """The provider promptfoo uses for model-graded assertions.
+
+    Points at OpenRAG's own OpenAI-compatible LLM endpoint so an eval needs no
+    third-party credentials.
+    """
+    config: dict[str, Any] = {"apiBaseUrl": base_url}
```
hedhoud, line 54: **[P1] Normalize the grader base URL**

Could you please remove a trailing slash from this value? The bundled Helm configuration provides the grader URL ending in `/v1/`, and promptfoo appends `/chat/completions` directly. That makes grading requests use `/v1//chat/completions`, so both retrieval and answer evaluations fail with the default deployment settings. A regression test using a grader URL ending in `/` would cover this case.

hedhoud (changes requested):
I found one blocking issue in the default Helm deployment path. The rest of the configuration and focused tests look good.

PR #817, opened 2026-07-27
feat(evaluation) 6/15: eval_datasets and eval_runs persistence

Ahmath-Gadji:
Part **6 of 15** of the split of #811. Targets `eval/05-promptfoo-config` (#816).

## What

| | |
|---|---|
| `core/ports/evaluation_repo.py` | the `EvaluationRepository` contract |
| `services/persistence/evaluation_repo.py` | asyncpg implementation |
| `services/persistence/schema.py` | `eval_datasets`, `eval_runs` |
| `migrations/.../a7c9e1f2b3d4_add_evaluation_tables.py` | the migration |
| `postgres_store.py`, `catalog_store.py`, `di/container.py` | expose it on the store and the container |

## Notable

- **The store enforces one run at a time, not the orchestrator.** `ux_eval_runs_single_active` is a partial unique index over an always-true expression, so it admits exactly one row in `QUEUED`/`INDEXING`/`EVALUATING`; `create_run` maps the resulting `UniqueViolationError` to `ConflictError`. A read-then-insert in the service would be a race, and not a harmless one — starting a run regenerates a shared service user's token (part 8), so two racing starts that both passed a read check would revoke each other's credentials mid-indexing.
- **Metrics are JSONB, not columns.** The metric set is expected to grow, and the dataclasses in `core/models/evaluation.py` already define the shape. `_row_to_run` reconstructs them, including lifting `samples` back out of the indexing payload.
- **`_load` exists because asyncpg returns JSONB as `str`** unless a codec is registered on the pool; the repo does not own the pool, so it decodes defensively instead.
- **`created_by` is `ON DELETE SET NULL`**, matching how `files.created_by` behaves — deleting an admin should not cascade away the run history they produced. `eval_runs.dataset_id` *is* `ON DELETE CASCADE`: a run without its dataset is not meaningful.
- **The migration is idempotent** per this repo's alembic rules (`CLAUDE.md`): `Base.metadata.create_all()` runs at startup, so a freshly bootstrapped database already has these tables before alembic sees them. Every op is guarded by `table_exists` / `index_exists`, in both directions.

## Testing

`ruff`, format check, the layer-import guard, and the full unit suite (2258 passed) — including the container and catalog-store wiring tests, which the new abstract property could have broken. The one failure, `test_content_deduplication_can_be_disabled_by_env`, reproduces on `develop`.

### DIFF
--- a/openrag/core/ports/catalog_store.py
+++ b/openrag/core/ports/catalog_store.py
@@ -13,6 +13,7 @@ from .chunk_repo import ChunkRepository
 from .conversation_repo import ConversationRepository
 from .document_repo import DocumentRepository
 from .entity_repo import EntityRepository
+from .evaluation_repo import EvaluationRepository
 from .idempotency_repo import IdempotencyRepository
 from .job_repo import JobRepository
 from .model_endpoint_repo import ModelEndpointRepository
@@ -69,6 +70,10 @@ class CatalogStore(ABC):
     @property
     @abstractmethod
     def preset_repo(self) -> PresetRepository: ...
+
+    @property
+    @abstractmethod
+    def evaluation_repo(self) -> EvaluationRepository: ...
 
     @property
     @abstractmethod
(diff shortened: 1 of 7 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #818, opened 2026-07-27
feat(evaluation) 7/15: dataset upload, storage and deletion

Ahmath-Gadji:
Part **7 of 15** of the split of #811. Targets `eval/06-persistence` (#817).

## What

The dataset half of `EvaluationService`: `create_dataset`, `list_datasets`, `delete_dataset`. Files land under `<data_dir>/eval/<dataset_id>/` (`corpus/`, `testset.csv`) and a row records what is there. Run dispatch is part 8.

## Notable

- **Validation happens at upload, not at run time.** The CSV goes through `parse_testset` (part 2) here, so a malformed test set is rejected while the admin is still looking at the upload dialog — not thirty minutes into an indexing phase.
- **Corpus uploads stream to disk.** `create_dataset` takes open binary streams rather than `bytes`, and copies each in 1 MB blocks under a running byte budget. A 512 MB corpus is therefore a disk cost, not a RAM one. The blocking I/O runs on a worker thread so it cannot stall the event loop.
- **The size cap reads one byte past the limit** instead of trusting a client-supplied `Content-Length`, so an inflated header cannot get past it. `test_an_oversized_test_set_is_rejected_without_buffering_it_all` asserts the stream position, i.e. that the cap actually stopped the read.
- **Duplicate corpus basenames are rejected.** Silently overwriting would leave `corpus_file_count` overstating what a run will actually index. Path components a browser may have sent are flattened.
- **A partial write cleans up after itself** — any failure after the directory exists removes it before the error propagates.
- **Deleting a dataset in use is refused.** The runner reads the corpus off disk for the whole indexing phase; removing it mid-run would surface as a `FileNotFoundError` rather than a clear 409. Only an *active* run blocks — history keeps its results.

## Testing

3 unit tests against in-memory fakes: refused delete (and that the files survive it), allowed delete after an idle run, and the streaming size cap. `ruff`, format check and the layer-import guard pass.

### DIFF
--- a/openrag/services/orchestrators/evaluation_service.py
+++ b/openrag/services/orchestrators/evaluation_service.py
@@ -1,0 +1,174 @@
+"""EvaluationService — datasets on disk, runs dispatched to the worker layer.
+
+This slice covers dataset storage: an admin uploads a corpus plus a test set,
+both land under ``<data_dir>/eval/<dataset_id>/``, and a row records what is
+there. Run dispatch follows.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import shutil
+import uuid
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from core.evaluation import parse_testset
+from core.models.evaluation import EvalDataset
+from core.utils.exceptions import ConflictError, NotFoundError, ValidationError
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from typing import IO
+
+    from core.config.root import Settings
+    from core.ports.evaluation_repo import EvaluationRepository
+
+TESTSET_FILENAME = "testset.csv"
+CORPUS_DIRNAME = "corpus"
+
+#: Block size for streaming an upload to disk.
+_COPY_CHUNK_BYTES = 1024 * 1024
+
+
+class EvaluationService:
+    """Dataset storage plus run dispatch for the admin evaluation page."""
+
+    def __init__(
+        self,
+        *,
+        repo: EvaluationRepository,
+        config: Settings,
+    ) -> None:
+        self._repo = repo
+        self._config = config
+        self._settings = config.evaluation
+        self._root = Path(config.paths.data_dir) / "eval"
+
+    # ── datasets ─────────────────────────────────────────────────────
+
+    def _dataset_dir(self, dataset_id: str) -> Path:
+        return self._root / dataset_id
+
+    async def create_dataset(
+        self,
+        *,
+        name: str,
+        corpus: Sequence[tuple[str, IO[bytes]]],
+        testset: IO[bytes],
+        user_id: int | None,
+    ) -> EvalDataset:
+        """Validate and store a corpus + test set.
+
+        The CSV is parsed here so a malformed test set is rejected at upload
+        rather than after a run has already spent minutes indexing.
+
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
@coderabbitai full review

openrag/services/orchestrators/evaluation_service.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Refused while a run is using it: the runner reads the corpus from disk
+        for the whole indexing phase, so removing it mid-run would surface as a
+        confusing FileNotFoundError instead of a clear conflict.
+        """
+        active = await self._repo.active_run()
+        if active is not None and active.dataset_id == dataset_id:
+            raise ConflictError(f"Evaluation run '{active.id}' is still using this dataset.")
+
+        if not await self._repo.delete_dataset(dataset_id):
```
hedhoud, line 169: **[P1] Preserve run history when a dataset is deleted**

The persistence layer currently cascades dataset deletion into `eval_runs`. This call therefore removes every completed run and its saved metrics, even though the product flow promises that past run results remain available. That is unrecoverable admin data loss, and the in-memory fake does not expose it. Please retain the historical run rows and cover this behavior with a repository-backed test.

openrag/services/orchestrators/evaluation_service.py
```diff
@@ -0,0 +1,174 @@
...
+        return await self._repo.list_datasets()
+
+    async def delete_dataset(self, dataset_id: str) -> None:
+        """Delete a dataset and its stored files.
+
+        Refused while a run is using it: the runner reads the corpus from disk
+        for the whole indexing phase, so removing it mid-run would surface as a
+        confusing FileNotFoundError instead of a clear conflict.
+        """
+        active = await self._repo.active_run()
```
hedhoud, line 165: **[P1] Make the active-run protection atomic**

The active-run check and dataset deletion are separate database operations. A run can be created between them; the current cascade then removes its row and files while startup continues provisioning and dispatching it. The worker is left without its corpus or a run row to update. Please enforce the check and deletion atomically in the repository and add a test for this interleaving.

openrag/services/orchestrators/evaluation_service.py
```diff
@@ -0,0 +1,174 @@
...
+    ) -> int:
+        """Write the corpus and test set to disk. Blocking; call in a thread."""
+        corpus_dir = directory / CORPUS_DIRNAME
+        corpus_dir.mkdir(parents=True, exist_ok=True)
+
+        written = 0
+        budget = self._settings.max_corpus_bytes
+        for filename, stream in corpus:
+            # Flatten any path components a browser may have sent.
+            target = corpus_dir / Path(filename).name
```
hedhoud, line 128: **[P2] Check duplicates after file ID normalization**

The runner indexes files using their normalized `file_id`, but this check only compares the stored basenames. Distinct names such as `report 1.txt` and `report?1.txt` are both accepted here and later normalize to `report_1.txt`, so one document fails to index and source ground truth becomes ambiguous. Please reject collisions using the same normalization as the runner and add a regression test.

hedhoud (changes requested):
I found three correctness issues that need handling before this is safe to use. The upload streaming, size limits, and cleanup path otherwise look good.

PR #820, opened 2026-07-27
feat(evaluation) 9/15: run start, dispatch and cancellation

Ahmath-Gadji:
Part **9 of 15** of the split of #811. Targets `eval/08-runner-port` (#819).

## What

The run half of `EvaluationService` — `start_run`, `cancel_run`, and the provisioning around them — dispatched through the `EvaluationRunner` port so the orchestrator never imports Ray.

Setup and teardown of a run's *identity* live here rather than in the worker, because creating users and partitions is orchestration the API layer already owns. The worker receives a partition it may write to and a token it may use, and nothing else about the system.

## Notable

**The ordering inside `start_run` is the whole design.** Each step exists because of a specific failure, and each has a regression test:

1. **Ping the runner before claiming anything.** Dispatch is fire-and-forget, so an unreachable worker would otherwise strand the run in `QUEUED` with a partition and a token provisioned for nobody. Surfaces as `503`. → `test_start_run_refuses_when_the_runner_cannot_be_reached`
2. **Insert the run row before provisioning.** The partial unique index from part 6 is the mutual exclusion. A read-then-insert would let two racing requests both regenerate the shared eval user's token, the second revoking the credentials the first is still indexing with. The loser gets its `409` *before* touching anything. → `test_a_second_start_is_refused_before_the_token_is_regenerated` asserts `regenerated == 0`.
3. **A failed provision releases the row and drops the partition.** A run left in an active status holds the lock and would block every later run. → `test_a_failed_provision_releases_the_run_lock`

**Auth.** Runs authenticate as one long-lived non-admin service user, `__openrag_eval__`, whose token is regenerated at the start of every run — a single service account in the DB, and no usable plaintext token at rest. It gets `file_quota=-1` because a corpus is re-uploaded on every run and a quota would fail the second one for reasons unrelated to the eval.

**Cancellation asks the worker first.** A worker that owns the run writes its own terminal status, including the metrics — so the service must not overwrite it. `False` means *nobody* owns the run (an actor restart orphaned it); the row is reaped here instead, because nothing else would ever move it out of an active status. An unreachable runner is logged, not raised — an orphan still has to be reapable.

## Testing

10 unit tests, all against in-memory fakes — no Ray, no actor, no worker process, which is exactly what the port bought. Covers the three orderings above, both cancellation paths, the already-finished 409, and that `api_base_url` comes from `Settings` rather than the environment. `ruff`, format check and the layer-import guard pass.

### DIFF
--- a/openrag/services/orchestrators/evaluation_service.py
+++ b/openrag/services/orchestrators/evaluation_service.py
@@ -1,8 +1,19 @@
 """EvaluationService — datasets on disk, runs dispatched to the worker layer.
 
-This slice covers dataset storage: an admin uploads a corpus plus a test set,
-both land under ``<data_dir>/eval/<dataset_id>/``, and a row records what is
-there. Run dispatch follows.
+Setup and teardown of a run's *identity* live here rather than in the worker,
+because creating users and partitions is orchestration the API layer already
+owns. The worker receives a partition it may write to and a token it may use,
+and nothing else about the system.
+
+Dispatch goes through the :class:`~core.evaluation.runner.EvaluationRunner`
+port, so this orchestrator stays Ray-free; the Ray actor lives behind the
+adapter in ``services/workers/eval_dispatcher.py``.
+
+The bearer token handed to the worker belongs to a single long-lived service
+user (``__openrag_eval__``) whose token is **regenerated at the start of every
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #821, opened 2026-07-27
feat(evaluation) 10/15: the EvalRunner Ray actor and its dispatcher

Ahmath-Gadji:
Part **10 of 15** of the split of #811. Targets `eval/09-run-lifecycle` (#820). This is where a run actually happens.

## What

```
EvalRunner.run()
  1. update status → INDEXING
  2. upload + time each corpus file over the real HTTP API   → indexing metrics
  3. update status → EVALUATING
  4. render both promptfoo configs, exec them                → raw results.json
  5. summarize(), persist, drop the throwaway partition
```

Plus `RayEvaluationRunner`, the adapter binding the actor to the part-8 port, and the container wiring that hands it to `EvaluationService`.

## Notable

- **The runner drives OpenRAG through its own HTTP API**, not in-process calls. Two reasons: it is the path a real user's documents take, so the indexing timings mean something; and it is the same surface promptfoo talks to, so an eval can never pass against a code path the API does not expose.
- **The actor handle resolves on first use, not in `__init__`.** `EvalRunner` is a *detached* actor, so merely building the adapter must not be what spawns it — listing datasets should not start a worker process. The container therefore builds `EvaluationService` eagerly and the worker lazily.
- **`run()` never raises.** The caller dispatched it fire-and-forget and has nobody to catch for, so cancellation and failure are both written to the run row. The `finally` drops the partition whatever happened.
- **One bad corpus file is a failed sample, not a dead run** — it still contributes to `files_failed`. *Every* file failing is an error worth surfacing, since the retrieval phase would otherwise grade an empty index.
- **promptfoo exits non-zero when assertions fail**, which is a result, not an error. The presence of the output file is what decides. Both streams are captured for the failure message because promptfoo reports config errors on **stdout**, not stderr.
- **Each run gets its own `PROMPTFOO_CONFIG_DIR`.** promptfoo keeps a SQLite eval history under it, defaulting to `$HOME/.promptfoo`, which is not guaranteed writable in a container; a per-run temp dir is also never contended. WAL mode is disabled because some filesystems don't support it.
- **`max_concurrency=4`** so `cancel()` and `is_busy()` still land while `run()` holds a slot — otherwise cancellation would deadlock behind the run it is trying to cancel.

## Testing

`ruff`, format check, the layer-import guard, and the full unit suite (2270 passed) including the DI wiring test, which now covers `evaluation_service`. The actor itself is exercised end to end on a real deployment rather than in unit tests — it is a thin shell over HTTP and a subprocess, both mocked out of meaning. The pure logic it calls is tested in parts 3–5.

The one failure, `test_content_deduplication_can_be_disabled_by_env`, reproduces on `develop`.

> **Deployment note.** The images that install promptfoo land in part 11. On a *separate* Ray cluster (not compose, which runs Ray inside the API container), `<data_dir>` must be on shared storage — the runner reads dataset files by path.

### DIFF
--- a/openrag/di/container.py
+++ b/openrag/di/container.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
     from core.vector_stores import VectorStore
     from services.orchestrators.auth_service import AuthService
     from services.orchestrators.conversion_service import ConversionService
+    from services.orchestrators.evaluation_service import EvaluationService
     from services.orchestrators.indexing_service import IndexingService
     from services.orchestrators.job_service import JobService
     from services.orchestrators.mcp_service import MCPService
@@ -117,6 +118,7 @@ class ServiceContainer:
         self._partition_service: PartitionService | None = None
         self._model_endpoint_service: ModelEndpointService | None = None
         self._preset_service: PresetService | None = None
+        self._evaluation_service: EvaluationService | None = None
         self._workspace_service: WorkspaceService | None = None
         self._retrieval_service: RetrievalService | None = None
         self._query_service: QueryService | None = None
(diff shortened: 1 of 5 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #822, opened 2026-07-27
feat(evaluation) 11/15: admin API routes, and promptfoo in both images

Ahmath-Gadji:
Part **11 of 15** of the split of #811. Targets `eval/10-ray-worker` (#821). Completes the backend — the feature is fully usable over HTTP after this; parts 12–14 are the UI.

## What

| Method | Path | |
|---|---|---|
| GET | `/evaluation/datasets` | list |
| POST | `/evaluation/datasets` | upload corpus + test set |
| DELETE | `/evaluation/datasets/{id}` | delete |
| GET | `/evaluation/runs` | history |
| POST | `/evaluation/runs` | queue a run (`202`, or `409` / `503`) |
| GET | `/evaluation/runs/{id}` | one run with metrics and per-question detail |
| POST | `/evaluation/runs/{id}/cancel` | abandon an in-flight run |

Plus the response schemas, and Node 22 + a pinned promptfoo in both images.

## Notable

- **Every route is admin-only**, via a router-level `Depends(require_admin)`. A run indexes a corpus, spends grader tokens and occupies the single runner slot — not something a partition editor should be able to trigger.
- **The router passes streams, not bytes.** `create_dataset` hands `UploadFile.file` straight to the service. Starlette has already spooled a large upload to disk; reading it here would pull it back into memory unbounded and undo part 7's streaming.
- **promptfoo is installed in `api.Dockerfile` *and* `ray.Dockerfile`.** Compose runs Ray inside the API container, so the API image needs it; a deployment with a separate Ray cluster needs the ray image. Missing either produces a run that fails only at the promptfoo step, minutes in.
- **Pinned, not `npx promptfoo@latest`.** A run should not depend on npm reachability, nor on CLI behaviour changing under a deployment that was never rebuilt. Node comes from NodeSource because the distro package predates promptfoo's floor.

## Testing

`ruff`, format check, the layer-import guard and the full unit suite (2270 passed; the one failure reproduces on `develop`). The routes are thin adapters over the service tested in parts 7 and 9.

### DIFF
--- a/infra/docker/api.Dockerfile
+++ b/infra/docker/api.Dockerfile
@@ -18,7 +18,21 @@ RUN apt-get update && apt-get install -y \
 
 # install ffmpeg
 RUN apt update && \
-    apt install -y ffmpeg 
+    apt install -y ffmpeg
+
+# Node + promptfoo back the admin evaluation page, where EvalRunner shells out
+# to `promptfoo eval`. Also installed in ray.Dockerfile: Ray runs inside this
+# container unless the deployment uses a separate cluster. Pinned rather than
+# resolved at run time so a run never depends on npm reachability.
+ARG PROMPTFOO_VERSION=0.121.19
+# Node comes from NodeSource: the distro package predates promptfoo's floor.
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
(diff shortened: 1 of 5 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #823, opened 2026-07-27
feat(evaluation) 12/15: UI API client for the evaluation endpoints

Ahmath-Gadji:
Part **12 of 15** of the split of #811. Targets `eval/11-api-routes` (#822). First of three UI parts: data layer → components → page.

## What

`ui/src/lib/api/evaluation.ts` — typed wrappers over the seven `/evaluation` routes, with interfaces mirroring part 11's response schemas one-for-one.

## Notable

- **Two shared definitions live here rather than in the views.** `isActiveStatus` / `ACTIVE_RUN_STATUSES` gives "can this run still change?" one definition instead of a status list repeated in each view, and `EVAL_POLL_MS` gives the poll cadence one home — both the run list and the run detail poll while a run is in flight, and they should not be able to drift apart.
- **`createEvalDataset` appends `corpus` repeatedly**, which is how FastAPI reads `list[UploadFile]`.
- **`status-badge` learns `EVALUATING`** so a run in its promptfoo phase renders like every other in-flight status rather than falling through to the default styling.

## Testing

`tsc -b` clean; `eslint` reports only the one warning that is already on `develop` (`react-hooks/exhaustive-deps` in an unrelated file). No behaviour yet — nothing imports this until part 13.


Ahmath-Gadji:
@coderabbitai full review

PR #824, opened 2026-07-27
feat(evaluation) 13/15: dataset card and run detail components

Ahmath-Gadji:
Part **13 of 15** of the split of #811. Targets `eval/12-ui-api-client` (#823). Two self-contained components; the page composing them is part 14.

## What

- **`DatasetCard`** — lists stored datasets, owns the upload dialog, the start button and the delete confirmation.
- **`RunDetail`** — one run: header, error, indexing panel, quality panel, per-question table.

## Notable

- **Starting is gated on `runActive`**, passed down from the run list. Runs are serialised server-side, so the button should be disabled rather than reliably return `409` — the API still enforces it, this just stops the UI from inviting a click it knows will fail.
- **`RunDetail` polls only while the run can still change** (`isActiveStatus`, part 12). A finished run stops refetching rather than polling a row that will never change again.
- **Errors are rendered with `[overflow-wrap:anywhere]`** — a run error can be a 2000-char promptfoo stderr tail with no spaces in it, which would otherwise blow out the layout.
- **The metric panels degrade rather than disappear.** A run that failed during indexing has `retrieval`/`answer` as `null`; each panel renders "Not measured yet" so the reader can tell *how far* the run got.

## Testing

`tsc -b` clean; `eslint` reports only the warning already present on `develop`. Rendering tests land with the page in part 14, where the components are reachable.

### DIFF
--- a/ui/src/pages/admin/evaluation/dataset-card.tsx
+++ b/ui/src/pages/admin/evaluation/dataset-card.tsx
@@ -1,0 +1,222 @@
+import { useState } from "react";
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { toast } from "sonner";
+import { Play, Trash2, Upload } from "lucide-react";
+import {
+  createEvalDataset,
+  deleteEvalDataset,
+  listEvalDatasets,
+  startEvalRun,
+} from "@/lib/api/evaluation";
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #825, opened 2026-07-27
feat(evaluation) 14/15: Evaluation tab on the admin System page

Ahmath-Gadji:
Part **14 of 15** of the split of #811. Targets `eval/13-ui-components` (#824). **This is the point the feature becomes reachable** — everything before it is inert.

## What

`EvaluationTab` composes the dataset card and the run detail, owns the run history table, and is registered as a tab on the admin System page.

## Notable

- **The run list is the source of truth for "is a run active".** Runs are serialised server-side, so that is a property of the list, not of the dataset card — which is why `runActive` is computed here and passed down rather than derived twice.
- **Polling stops when nothing is active.** The same `EVAL_POLL_MS` / `isActiveStatus` pair from part 12, so the list and the detail view cannot drift apart on cadence.

## Testing

9 UI tests covering dataset listing, run start/cancel gating, the metric panels and error display. The full `ui` suite is green here: **23 files / 167 tests, 0 failures**, plus `tsc -b` clean and `eslint` reporting only the warning already on `develop`.

> Note for anyone comparing with #811: that description reported 31 pre-existing UI failures under jsdom. They do not reproduce on a clean `npm ci` in this environment — the suite passes in full.

## What's left

Part 15 is docs and the sample dataset only — no code.

### DIFF
--- a/ui/src/pages/admin/evaluation/index.test.tsx
+++ b/ui/src/pages/admin/evaluation/index.test.tsx
@@ -1,0 +1,212 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  cancelEvalRun,
+  getEvalRun,
+  listEvalDatasets,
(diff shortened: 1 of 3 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #826, opened 2026-07-27
docs(evaluation) 15/15: feature guide and a sample dataset

Ahmath-Gadji:
Part **15 of 15** of the split of #811. Targets `eval/14-ui-page` (#825). Docs and data only — no code.

## What

- **`CLAUDE.md`** — the Evaluation section: the flow, the layering rule (the service dispatches through a port and never touches Ray), why there are two promptfoo configs, and which values are deliberately *not* configuration.
- **`tests/evaluation/`** — a ready-made test set: 11 questions over 6 documents, every answer checked against the source rather than generated.

## Notable

- **The corpus is not committed** — third-party PDFs, ~36 MB. `corpus.txt` lists the filenames and the README gives the two commands to assemble and upload it.
- **24 documents for 11 questions, on purpose.** Only 6 are the subject of a question; the other 18 are distractors, each topically adjacent to a question's source (other AI-policy papers, other gut/neuro medical papers, other agricultural releases). A corpus where every document is on a different subject makes retrieval look better than it is — any half-working retriever scores a perfect hit rate when there is one candidate per topic.
- **The repo ignores `*.csv`**, so the committed test set is un-ignored explicitly. It is source, not scratch data.

## Carried over from #811 — known limitations, unchanged

Flagged for a decision rather than fixed:

- Indexing is **sequential**, so `files_per_minute` measures serialised throughput, not the system's real concurrency.
- `/search` defaults `similarity_threshold` to `0.75` and the eval does not override it, so retrieval can return fewer than `top_k` and depress recall for reasons unrelated to ranking.
- A **separate** Ray cluster needs `<data_dir>` on shared storage — the runner reads dataset files by path.
- Ordinary users can still create `__eval_`-prefixed partitions, which are merely hidden from listings.

## Testing

`ruff` and the layer-import guard pass. No code changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,8 @@ volumes/*
 !/services/.gitkeep # Keep the placeholder
 
 *.csv
+# The committed evaluation test set is source, not scratch data.
+!tests/evaluation/*.csv
 *.pkl
 
 #helm
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -242,6 +242,59 @@ Optional web search augmentation via the Staan API, allowing the LLM to combine 
 - `openrag/services/orchestrators/query_service.py` — `_prepare_for_web_only()`, web search logic in `_prepare_for_chat_completion()`
 - `openrag/api/routers/user/chat.py` — `__prepare_sources()` merges document and web sources
 
+### Evaluation (admin System page → Evaluation tab)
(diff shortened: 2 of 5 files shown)


Ahmath-Gadji:
@coderabbitai full review

PR #828, opened 2026-07-28
Preserve cross-page PDF table rows during indexing

hedhoud:
## Context

Large PDF tables can split one logical row over several pages. Continuation pages often omit the identifying cells, so the parser output no longer carries enough context for reliable retrieval.

## Problem

OpenRAG currently chunks those continuation fragments independently. This can separate supporting text from its row identity and allow unrelated rows to be mixed in retrieval results.

## Expected behavior

Indexation presets can opt into conservative automatic reconstruction. High-confidence continuations across adjacent pages are indexed as self-contained, row-aware chunks with source provenance. Uncertain cases preserve the original parser output, and the feature remains disabled by default for backward compatibility.

## Validation

- Backend unit suite: 2,233 passed.
- Admin UI suite: 171 passed.
- UI lint and production build passed.
- The regression fixture based on original pages 803–805 reconstructs the `CST salarié` / `L. 421-1` row across all three pages while keeping the next `L. 421-3` row separate.
- The complete 904-page source PDF was processed successfully with fail-open fallback for uncertain boundaries.

No database or vector-store migration is required.


hedhoud:
Superseded by #830, which uses the branch pushed directly to `linagora/openrag` as requested. Closing this fork-based draft to avoid duplicate review.

PR #835, opened 2026-07-29
feat(prompts): DB-backed prompt library with per-preset and per-partition selection

### LINKED ISSUE #772: Prompt management: DB-backed prompt library with global defaults + per-partition overrides + admin UI
## Problem
Prompts are currently **disk-only**: every consumer reads a fixed `.txt` template via `load_template_by_key(cfg.paths.prompts_dir, cfg.prompts, key)`. Operators can't view or edit prompts without redeploying, and there's no way to vary a prompt per partition. The `PgPromptRepository` is a stub (every method raises), `get_active`/`get_by_type` have no consumers, and there is no `prompts` table — so the "DB-first" resolution the module docstring describes doesn't exist yet.

## Goal
DB-backed prompt management, editable from the admin UI, with per-partition overrides — following the established `model_endpoints`/`pipeline_presets` pattern (table → asyncpg repo → service with seed/load → admin router → schemas → DI wiring).

## Scope
- **Schema:** `prompts` (library; one global default per type via a partial-unique index) + `partition_prompts` junction (one override per `(partition, type)`).
- **Resolution seam (single resolver):** `partition override → global default → disk-seed fallback`. Resolved **request-time** on the chat/query path; resolved **once per job** on the indexing path (avoids per-chunk DB hits at batch scale and keeps horizontal-scale workers coordination-free).
- **Backend:** real `PgPromptRepository`, `PromptService` (seed-from-disk, CRUD, set-default, assign/unassign per partition), admin router + per-partition assignment routes, schemas, DI/container wiring, idempotent Alembic migration.
- **Frontend:** admin prompt-management page (global defaults, library CRUD, per-partition config) + sidebar nav.
- **Tests:** revert-prove that editing a prompt changes generation output; repo/service/route unit coverage.

## Non-goals
- Prompt versioning/history and a user-facing "custom prompts" surface (future, net-new features).

andyne13:
Adds DB-backed prompt management: a global prompt library with a default per
type, selected per preset (indexation/retrieval) and per partition
(generation), and wired into every stage that sends a prompt to a model.

Closes #772.

## What it does

Seven prompt types (`sys_prompt`, `query_contextualizer`, `chunk_contextualizer`,
`image_captioning`, `hyde`, `multi_query`, `topic_tagger`) become editable rows
instead of read-only files on disk. Resolution is one seam:

```
resolve_prompt(prompt_type, names=[...]) -> str
```

first resolvable candidate name → the type's global default → the bundled disk
seed. Callers pass precedence-ordered names, so per-user personalisation can
later prepend a name without changing the signature.

Selection follows where the setting belongs:

| Prompt | Selected on |
|---|---|
| `sys_prompt` | partition `generation_prompt_names` |
| `query_contextualizer`, `hyde`, `multi_query` | retrieval preset |
| `chunk_contextualizer`, `image_captioning`, `topic_tagger` | indexation preset |

Preset fields serialise into the existing preset JSONB (no preset migration);
partitions get one JSONB column.

## Notable choices

- **Resolution timing.** Query/generation prompts resolve per request, so an
  edit takes effect with no restart. Ingest prompts resolve once per file, not
  per chunk — at 2K-document batches a per-chunk lookup would be a hot-path DB
  read, and per-actor caching would need cross-Ray cache coherence.
- **Templates validated at write time.** Types rendered with `str.format` accept
  only their own placeholders, so a bad template is a 422 on save instead of a
  500 on every later chat. The other three are sent verbatim and may contain
  braces.
- **One default per type** is a partial unique index, promoted by clear-then-set
  inside a locked transaction, not application-level bookkeeping.
- **References are soft.** Presets and partitions name prompts by string; a
  deleted or renamed prompt falls back to the default rather than breaking
  retrieval. Deleting a type's default is refused.
- **`spoken_style_answer` is removed** — it had no consumer; `sys_prompt` is the
  only answer prompt.

## Verification

All seven prompts were proven end-to-end on a live deployment: each was given a
uniquely-worded library prompt, selected through its own config surface, and the
logs confirmed both that it resolved (`prompt.resolve … <- named:…`) and that
its text reached the model (`llm.call … system[…]: …`). `sys_prompt` was also
confirmed behaviourally — the answer came back in the style the prompt asked
for, with no restart. Image captioning logs `<image_url>` in place of the
base64 payload.

Both log lines sit at DEBUG (`LOG_LEVEL=DEBUG`), so production stays quiet.

Unit suite 2258 green; repo-integration green against Postgres (two pre-existing
failures in `test_partition_repo.py` also fail on `develop`, unrelated).

## Follow-ups (deliberately not in scope)

- Preset `*_prompt_name` references are not validated at write time, while
  partition ones are — a preset typo silently falls back to the default.
- `used_by` counts *effective* resolution, so a type's default shows every
  partition that would fall back to it, including partitions whose retriever
  never invokes that type.
- Seeding races if several replicas boot into an empty database at once
  (a pre-existing pattern shared with the other seeded tables).

## Summary by CodeRabbit

* **New Features**
  * Added an admin prompt library for creating, listing, updating, deleting, and promoting prompts as defaults.
  * Added support for the `spoken_style_answer` prompt type.
  * Enabled partition-level generation prompt overrides (with prompt resolution and automatic prompt seeding).
  * Added configurable per-call system prompt overrides for contextualization and topic tagging.
  * Added bounded debug logging for LLM calls across supported inference backends.
* **Bug Fixes**
  * Improved validation for generation prompt overrides, including missing-prompt detection and conflict-safe default handling.
  * Made prompt resolution resilient by falling back to bundled prompt templates when repository lookups fail.


andyne13:
@coderabbitai review

openrag/services/inference/_call_log.py
```diff
@@ -0,0 +1,143 @@
...
+            return _clip(" || ".join(rendered), MAX_DETAIL_CHARS)
+        text = prompt or ""
+        return _clip(f"prompt[{len(text)}]: {_preview(text)}", MAX_DETAIL_CHARS)
+
+    # The lazy preview is passed positionally on purpose: loguru copies **kwargs
+    # into ``record["extra"]``, and the terminal formatter appends every extra —
+    # so a ``detail=`` kwarg would print the whole payload a second time on each
+    # line. Positional args are formatted into the message only.
+    logger.bind(caller=safe_caller, model=safe_model, endpoint=safe_endpoint, stream=stream).opt(lazy=True).debug(
+        f"llm.call {safe_caller} model={safe_model} stream={stream} | " + "{}",
```
Ahmath-Gadji: **Bug: a brace in the interpolated metadata raises `KeyError` out of the request path**

`log_llm_call` interpolates `safe_caller` / `safe_model` / `stream` into the *format string* and passes the detail as a positional arg, so loguru runs `message.format(detail)` over the whole string — the interpolated identifiers included. `_meta()` flattens newlines and bounds length, but it doesn't escape braces, so any surviving `{`/`}` becomes a format field.

`model` is client-controlled (`metadata.llm_override.model` — the comment on line 30 says so), and `log_llm_call` is called *outside* the `try/except` in `VLLMClient.generate` / `chat` / `stream_chat`, so the exception propagates straight out of the client into the request.

Confirmed against this branch, DEBUG sink attached:

```python
log_llm_call(caller="VLLMClient.chat", model="gpt{x}", endpoint="http://h",
             messages=[{"role": "user", "content": "hi"}])
# KeyError: 'x'
```

So `POST /v1/chat/completions` with `metadata: {"llm_override": {"model": "gpt{x}"}}` is a 500 on any deployment running at `LOG_LEVEL=DEBUG` — the level this PR's verification steps ask operators to use.

Same shape at `openrag/services/orchestrators/prompt_service.py:247` (separate comment): there the interpolated value is `name`, an admin-chosen prompt name, so a prompt called `sales{v2}` breaks *every* chat request at DEBUG until it's renamed.

Fix: keep caller-controlled text out of the format string — render the whole line inside the single lazy `{}` argument, which `str.format` never rescans. Note that `opt(lazy=True)` calls *every* positional arg, so the identifiers can't simply be moved out one at a time (`TypeError: 'str' object is not callable`).

`openrag/services/inference/_call_log.py:140-143`

openrag/services/orchestrators/prompt_service.py
```diff
@@ -0,0 +1,354 @@
...
+        deployment pays nothing for it.
+        """
+        logger.bind(
+            prompt_type=prompt_type,
+            candidates=candidates,
+            source=source,
+            resolved_name=name,
+            length=len(content),
+        ).opt(lazy=True).debug(
+            f"prompt.resolve {prompt_type} <- {source}{f':{name}' if name else ''} | " + "{}",
```
Ahmath-Gadji: **Bug: an admin-chosen prompt name containing a brace raises `KeyError` here**

Same defect as `openrag/services/inference/_call_log.py:140-143`: `name` is interpolated into the *format string*, and the trailing `"{}"` makes loguru run `message.format(...)` over the whole thing — so a brace in `name` is parsed as a format field.

`name` is free text (`CreatePromptRequest.name` only checks non-empty), so it can contain braces. Confirmed on this branch:

```python
PromptService._log_resolution("sys_prompt", ["my{tmpl}"], "named", "my{tmpl}", "content")
# KeyError: 'tmpl'
```

Because resolution happens per request, a partition pointed at a prompt named `my{tmpl}` fails **every** chat request while a DEBUG sink is attached — and it fails inside `resolve_prompt`, i.e. before the prompt is ever used.

`openrag/services/orchestrators/prompt_service.py:246-249`

openrag/services/workers/indexer_pool.py
```diff
@@ -231,6 +232,47 @@ def _reload_decision(self, required_model_names: dict[str, list[str]] | list[str
...
         )
 
+    # Enrichment stage → (enable flag, prompt_type, preset name-field, row key).
+    # The name-field is the indexation-preset config key naming a library prompt
+    # for that stage. Only enabled stages are resolved, so a file that neither
+    # contextualizes nor tags nor captions pays no prompt-resolution cost.
+    _INGEST_PROMPTS = (
+        ("enable_contextualization", "chunk_contextualizer", "contextualization_prompt_name", "contextualizer_prompt"),
+        ("enable_topic_tagging", "topic_tagger", "topic_tagging_prompt_name", "topic_tagger_prompt"),
+        ("enable_image_captioning", "image_captioning", "image_captioning_prompt_name", "caption_prompt"),
```
Ahmath-Gadji, line 242: **`indexation_config.get(flag)` drops the model default, so a sparse config silently skips caption-prompt selection**

`enable_image_captioning` defaults to **`True`** (`IndexationPipelineConfig`, and `pipeline_builder._captioning_enabled` returns `True` when there's no config at all). A bare `.get("enable_image_captioning")` returns `None` for an absent key, so a config that omits the flag captions during ingest but never resolves the `image_captioning` prompt — the preset's `image_captioning_prompt_name` and the type's library default are both ignored, and captioning quietly keeps using the disk seed. Nothing surfaces the divergence: `_resolve_ingest_prompts` just returns no key for that stage.

Today's dispatch path masks it (`_resolve_indexation_dispatch_config` sends a full `model_dump(mode="json")`, so every flag is present), which is exactly what makes it a latent trap for the next caller.

The convention elsewhere in this file is to restate the model default at the lookup — `indexation_config.get("enable_topic_tagging", False)` at `indexer_pool.py:571`, and the same at `indexer_actor.py:256`. Carrying the default in the `_INGEST_PROMPTS` tuple keeps the three flags honest in one place.

`openrag/services/workers/indexer_pool.py:239-243`

openrag/services/persistence/schema.py
```diff
@@ -92,6 +92,61 @@
 )
 
 
+# The 8 canonical prompt types — kept in sync with core.models.prompt.PromptType.
```
Ahmath-Gadji, line 95: **Comment nits left behind by the `spoken_style_answer` removal / reshuffle** (grouping three trivia into one thread)

1. Here: "The 8 canonical prompt types" — the tuple below lists **7**. `spoken_style_answer` went away with this PR, so the count is stale.
2. `openrag/core/ports/prompt_repo.py:53` and `:65` both read `# Global default (one per type)`, and the first one sits above `get_by_name` — a selection primitive, not the default. The second block then covers `reference_counts` *and* the default accessors.
3. `openrag/services/persistence/prompt_repo.py:32-33` — `# Indexation/retrieval preset config field -> the prompt_type it names…` is stranded above `_as_conflict`; the map it describes (`_PRESET_FIELD_TO_TYPE`) is 20 lines further down.

`openrag/services/persistence/schema.py:95`

openrag/services/orchestrators/query_service.py
```diff
@@ -296,14 +295,53 @@ def _default_llm_name(self) -> str:
...
         last_user = messages[-1]["content"]
         if RAGMODE(self._rag_mode) is RAGMODE.SIMPLERAG:
             return SearchQueries(query_list=[Query(query=last_user)])
 
         chat_history = "".join(f"{m['role']}: {m['content']}\n" for m in messages)
-        prompt = self._query_contextualizer_prompt.format(
+        contextualizer = await self._prompt_service.resolve_prompt(
+            "query_contextualizer",
+            names=[self._retrieval_prompt_name("query_contextualizer_prompt_name", partition)],
+        )
```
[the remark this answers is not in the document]
Ahmath-Gadji: Confirmed, and worth fixing at the choke point rather than at the three call sites.

The underlying change in behaviour: before this PR, `sys_prompt` / `query_contextualizer` / `hyde` / `multi_query` were read from disk **once, at construction** (`QueryService.__init__`, `RetrievalService._build_legacy_pipeline`), so serving a chat needed no prompt I/O at all. Now every request resolves through `PromptService`, which means chat and `/search` have taken on a new per-request Postgres dependency. A transient pool error inside `get_by_name` / `get_default` propagates out as an unhandled asyncpg exception (a 500), and it never reaches the disk-seed fallback — the fallback only runs when the DB *successfully* reports no default.

`resolve_prompt`'s own docstring already frames the disk seed as the "resolution must not fail" tier, so the resilient behaviour belongs inside it: catch repository failures, log them once at WARNING, and continue to the disk seed. Then the typed `ConfigError` stays reserved for the genuinely unrecoverable case (no library default *and* no readable bundled template), all three request-path callers keep working unchanged, and `indexer_pool._resolve_ingest_prompts` keeps its own belt-and-braces catch.

Retrieval-side note for `retrieval_service._resolve_query_template`: with the catch centralised, its `load_template_by_key` branch stays as-is (it's the "no PromptService wired" case), and a DB failure now lands on the same disk template via `resolve_prompt` instead of propagating.

Fixing this next.

Ahmath-Gadji:
## Review notes (three items that aren't inline-anchorable)

### 1. The `/prompts` API isn't documented — fixing in this branch

`docs/content/docs/documentation/API.mdx` is already touched here (to drop `spoken_style_answer`), and it documents every sibling admin resource — `/presets` at L473, `/model-endpoints` at L520 — so the seven new prompt-library routes plus the partition-level `generation_prompt_names` field are the one part of this feature an operator can't discover from the docs. I'll add a `Prompts` section in the same shape as the two above.

### 2. `ui/src/pages/admin/presets.tsx` still writes the old preset key — **not** fixing here

`IndexationPipelineConfig.vlm_caption_prompt_name` was renamed to `image_captioning_prompt_name`, but the preset editor still reads and writes the old name:

```ts
// ui/src/pages/admin/presets.tsx:390-391
promptValue={configGet(config, "vlm_caption_prompt_name", "")}
onPromptChange={(v) => set("vlm_caption_prompt_name", v || null)}
```

`IndexationPipelineConfig` is `extra="ignore"`, so this doesn't break anything — the key is just stored and never read, and `reference_counts` won't count it either. Not a regression (`vlm_caption_prompt_name` had no consumer on `develop` either), but the "Caption prompt" selector is now inert while a working backend field sits next to it. Left alone deliberately: this is the backend PR, and the prompt-library UI (which will need real `promptsByType` data anyway) is where that field should be rewired.

### 3. Seeding race — respecting your stated scope, with one correction

Follow-up #3 in the description calls the concurrent-boot seeding race "a pre-existing pattern shared with the other seeded tables". Half true: `preset_service.seed_defaults` goes through `PgPresetRepository.upsert`, which is `INSERT … ON CONFLICT DO UPDATE`, so it's race-safe. `model_endpoint_service.seed_defaults` does share the check-then-insert shape.

What's different here is the blast radius: `PgPromptRepository.create` maps a unique violation to a `ValidationError`, `_initialize_step` re-raises, and `ServiceContainer.initialize` aborts — so the replica that loses the race doesn't lose prompt seeding, it fails to boot. With N replicas starting against an empty database that's a crash-loop until one wins. A `except ValidationError: continue` around the `create` in `seed_defaults` (another replica already seeded this type — the idempotency check at the top of the loop just lost a race) would close it in three lines.

Not touching it since you've marked it out of scope; flagging it so the decision is explicit rather than inherited from the "pre-existing" framing.

openrag/services/inference/_call_log.py
Ahmath-Gadji: Suggested shape — one literal `{}` as the whole format string, everything else inside the lazy callable, so `str.format` never rescans user text:

```diff
-    logger.bind(caller=safe_caller, model=safe_model, endpoint=safe_endpoint, stream=stream).opt(lazy=True).debug(
-        f"llm.call {safe_caller} model={safe_model} stream={stream} | " + "{}",
-        _detail,
-    )
+    def _line() -> str:
+        return f"llm.call {safe_caller} model={safe_model} stream={stream} | {_detail()}"
+
+    # The message is a single literal placeholder: loguru runs
+    # ``message.format(*args)``, so any identifier interpolated into the format
+    # string itself would have its braces parsed as fields — and ``model`` is
+    # client-controlled. Positional args are substituted, never rescanned.
+    logger.bind(caller=safe_caller, model=safe_model, endpoint=safe_endpoint, stream=stream).opt(lazy=True).debug(
+        "{}", _line
+    )
```

Same for `_log_resolution`, folding the prefix into the existing lambda.

Two notes:

- `terminal_formatter` (`openrag/core/utils/logging.py:49`) already does `.replace("{", "{{").replace("}", "}}")` on the `extra` suffix, so the brace hazard is known in this codebase — it's the *message* path that isn't covered, because loguru formats it before any handler format runs.
- A regression test needs a DEBUG sink attached and a braced `model`, in the shape of the existing `test_a_newline_in_a_client_supplied_model_cannot_forge_log_lines`: the call must not raise, and the braces should survive verbatim in the emitted line.

openrag/services/orchestrators/query_service.py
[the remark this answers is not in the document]
Ahmath-Gadji: Correction to my previous reply: I'm leaving the fix to you (review-only pass, nothing pushed). Suggested shape, for whenever you pick it up:

```diff
         candidates = [n for n in (names or ()) if n]
-        for name in candidates:
-            prompt = await self._repo.get_by_name(prompt_type, name)
-            if prompt is not None:
-                self._log_resolution(prompt_type, candidates, "named", name, prompt.content)
-                return prompt.content
-        default = await self._repo.get_default(prompt_type)
-        if default is not None:
-            self._log_resolution(prompt_type, candidates, "default", default.name, default.content)
-            return default.content
+        try:
+            for name in candidates:
+                prompt = await self._repo.get_by_name(prompt_type, name)
+                if prompt is not None:
+                    self._log_resolution(prompt_type, candidates, "named", name, prompt.content)
+                    return prompt.content
+            default = await self._repo.get_default(prompt_type)
+            if default is not None:
+                self._log_resolution(prompt_type, candidates, "default", default.name, default.content)
+                return default.content
+        except Exception as exc:  # noqa: BLE001 - a store failure degrades to the bundled seed
+            # Serving a chat needed no prompt I/O before the library existed; a
+            # pool blip must not become a 500 when a usable template is on disk.
+            logger.warning(f"Prompt store unreachable for '{prompt_type}'; using the bundled template: {exc}")
         try:
             content = self._disk_seed(prompt_type)
```

Worth covering with a stub repo whose `get_by_name`/`get_default` raise, asserting `resolve_prompt` still returns the disk seed — that pins the contract the docstring claims.

One ordering caveat: a broad catch here would also swallow the `KeyError` from `_log_resolution` that I raised separately (a braced prompt name), turning a hard failure into a silent every-request fallback to the seed. So that one wants fixing on its own merits first, not left to be masked by this.

Ahmath-Gadji:
## Review summary

Review-only pass — nothing pushed to this branch. Correcting my two earlier "I'll fix this" lines: the docs section and the code fixes are all yours. CI is green on `8c207a9` (api-tests, milvus-integration, lint, layer-import-guard, tests) and the local unit suite runs 2289 passed.

The design holds up well: one resolution seam, selection filed where the setting belongs, soft name references with a default fallback, and the one invariant that needs SQL atomicity pushed into a partial unique index rather than application bookkeeping. The migration chain is a single head (`…d4e5f6a7b8c9 → e5f6a7b8c9d0 → e8f9a0b1c2d3`), every op is inspector-guarded, the JSONB round-trip is covered by the pool-level codec, and the removed `spoken_style_answer` leaves no dangling reference anywhere in the repo (`extra="ignore"` on `IndexationPipelineConfig` also keeps stored presets carrying the old `vlm_caption_prompt_name` loadable).

Findings, most serious first:

| # | Where | What |
|---|---|---|
| 1 | [`_call_log.py:141`](openrag/services/inference/_call_log.py) + [`prompt_service.py:247`](openrag/services/orchestrators/prompt_service.py) | **Braces in interpolated log metadata raise `KeyError` out of the request path.** Confirmed on this branch. Client-controlled `model`, and admin-chosen prompt `name`, both land in loguru's format string. At DEBUG this 500s the request. |
| 2 | `prompt_service.py:185-221` (CodeRabbit thread) | Chat and `/search` have taken on a new **per-request Postgres dependency**; a pool error can't reach the disk-seed fallback, so it surfaces as a 500 where the pre-PR code used an in-memory template. |
| 3 | `indexer_pool.py:239-243` | `indexation_config.get(flag)` drops the model default, and captioning defaults to **on** — a sparse config would caption while silently ignoring the selected caption prompt. Latent today (dispatch sends a full `model_dump`). |
| 4 | `API.mdx` | The seven `/prompts` routes and partition `generation_prompt_names` are undocumented, while `/presets` (L473) and `/model-endpoints` (L520) are. |
| 5 | `schema.py:95`, `ports/prompt_repo.py:53`, `prompt_repo.py:32` | Stale count and misfiled/duplicated section comments. |

Plus the two I flagged as deliberately-not-fixed above: the inert `vlm_caption_prompt_name` field in the preset UI, and the boot-crash blast radius of the seeding race (`preset_service` is race-safe via upsert, so only the `model_endpoint_service` half of the "pre-existing pattern" framing holds).

Nothing here touches the feature's core; #1 is the only one I'd consider a merge blocker, and it's a two-line change plus a test.

openrag/services/inference/_call_log.py
andyne13: Confirmed and fixed in `309f6140` — reproduced first:

```python
log_llm_call(caller="VLLMClient.chat", model="gpt{x}", endpoint="http://e", messages=[...])
# KeyError: 'x'
```

Your call-path reasoning was exactly right: it is emitted before the `try:` in `generate`/`chat`/`stream_chat`, and `.env.example` ships `LOG_LEVEL=DEBUG`, so the lazy path is live in a default deployment and the request dies.

Took your suggested shape — one literal `{}` as the whole message, everything built inside the lazy callable, so the substituted value is never rescanned. Braces in the caller, model, endpoint *and* message content are all inert now, covered by a test that asserts `gpt{x}` and `brace {in} content too` survive to the sink.

Good catch. `_meta()` was added to stop newline log-forging through that same client-controlled field, and I simply didn't think about braces.

openrag/services/orchestrators/prompt_service.py
andyne13: Confirmed and fixed in `309f6140`. Your repro reproduced verbatim (`KeyError: 'tmpl'`), and the blast radius is as you describe — resolution is per request, so a partition pointed at a prompt named `my{tmpl}` fails *every* chat, not an edge case.

Same single-literal-placeholder shape as the `_call_log` one, with a regression test covering a braced name and braced content.

openrag/services/workers/indexer_pool.py
andyne13: Confirmed and fixed in `309f6140`. `enable_image_captioning` defaults to `True`, so a sparse config captioned while never resolving its prompt — the preset's `image_captioning_prompt_name` and the library default both silently ignored.

Defaults now come from `IndexationPipelineConfig.model_fields` rather than being restated, so the three stages can't drift (captioning on, contextualization and topic tagging off).

Worth noting it caught a second thing on the way in: four `test_indexer_pool` tests started failing, because the fix works — a bare config now reaches prompt resolution where it previously skipped it. Those tests build the actor with `__new__`, bypassing `__init__`, so they'd never hit that branch. Production sets `_prompt_service` normally, so I stubbed the doubles rather than making the production path defensive about a state it can't be in.

openrag/services/persistence/schema.py
andyne13: Fixed in `309f6140`, with one already resolved:

1. The "8 canonical prompt types" count is correct again — `spoken_style_answer` is back. Removing it was my mistake: I justified it as unused, but it's a public API flag (`metadata.spoken_style_answer`) and a persistent Chainlit command, so deleting it was an undeclared breaking change. It's now an eighth library type, and I added the regression test that was missing — nothing asserted the flag swapped the prompt, which is why the removal kept the suite green.
2. `ports/prompt_repo.py` — the first block is now `# Selection by name` (it sits above `get_by_name`), the second `# Usage counts and the global default (one per type)`.
3. `prompt_repo.py` — the stranded `_PRESET_FIELD_TO_TYPE` comment is back above its map; `_as_conflict` had been inserted between them.

openrag/services/orchestrators/query_service.py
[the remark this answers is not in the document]
andyne13: Agreed, and fixed at the choke point in `309f6140`.

This was the most valuable point in the review: it names a property change I hadn't stated anywhere. Prompts used to be read from disk once at construction, so serving a chat needed no prompt I/O at all; per-request resolution quietly gave chat and `/search` a Postgres dependency, and a transient pool error propagated as a 500 without ever reaching the disk-seed fallback — the fallback only ran when the DB answered "no rows".

`resolve_prompt` now treats repository lookups as best-effort: a failure logs one warning and degrades to the bundled template. `ConfigError` still covers the genuinely unrecoverable case where the disk seed is also unreadable. Swallowed here rather than at the callers, so chat, query expansion, retrieval and indexing share the guarantee — a test drives it with a repository that raises on both `get_by_name` and `get_default` and asserts the bundled template comes back.

andyne13:
@Ahmath-Gadji all five addressed in `309f6140` — replies inline on each thread. Both brace bugs reproduced exactly as you described before I touched anything, and the `resolve_prompt` resilience point was the most valuable one in the review: per-request resolution had quietly given chat and `/search` a Postgres dependency they didn't have when prompts were read once at boot, and a pool error bypassed the disk-seed fallback entirely. That's now handled at the single choke point.

Also restored `spoken_style_answer` (your feature — I'd removed it claiming no consumer, which was wrong: it's a public API flag and a Chainlit command). It's back as an eighth library type, with the regression test that was missing.

2294 unit tests, ruff, layer guard and the prompt integration suite are green.

andyne13:
@Ahmath-Gadji thanks — I'd missed your two non-inline comments when I replied earlier. Both now addressed in `ea730de8`, and your correction on the seeding race was right and worth more than the framing I gave it.

**Seeding race** — you're right that "pre-existing pattern" was half true: `preset_service` goes through `INSERT … ON CONFLICT DO UPDATE` and is race-safe, so only the model-endpoint half shares the shape. And the blast radius is as you describe — worse still because my own 409 mapping is what turns the unique violation into a `ValidationError` that `_initialize_step` re-raises, so the losing replica fails to *boot*, not just to seed. Closed with the `except ValidationError: continue` you suggested, plus a test driving a repo whose `create` always raises.

**API docs** — added a `Prompt Library` section in the same shape as `/presets` and `/model-endpoints`: the seven routes, the resolution order, a table of where each type is selected, the plain-placeholder rule with its 422, and the partition `generation_prompt_names` PATCH.

**`vlm_caption_prompt_name`** — already fixed in the UI PR (#836): `presets.tsx` there reads and writes `image_captioning_prompt_name`, no occurrences of the old key remain. You were reviewing this branch, where `ui/` is still at develop's state, so the inert selector you saw is gone by the time the two land.

CI green on `ea730de8`; 2295 unit tests locally.

openrag/services/orchestrators/prompt_service.py
```diff
@@ -0,0 +1,389 @@
...
+                return default.content
+        except Exception as exc:  # noqa: BLE001 - a DB blip must not fail the request
+            logger.warning(f"Prompt lookup failed for '{prompt_type}'; falling back to the bundled template: {exc}")
+        try:
+            content = self._disk_seed(prompt_type)
+        except (FileNotFoundError, ValueError, KeyError) as exc:
+            raise ConfigError(
+                f"No prompt available for type '{prompt_type}': no library default and "
+                f"no readable bundled template ({exc}).",
+                code="PROMPT_UNAVAILABLE",
```
Ahmath-Gadji, line 246: **Bug: this `raise` itself raises `TypeError` — the typed `ConfigError` can never be constructed**

`ConfigError` hard-codes its code and forwards the rest, so a `code=` kwarg collides:

```python
class ConfigError(OpenRAGError):
    def __init__(self, message: str, **kwargs):
        super().__init__(message, code="CONFIG_ERROR", status_code=500, **kwargs)
```

`ConfigError(msg, code="PROMPT_UNAVAILABLE")` → `TypeError: OpenRAGError.__init__() got multiple values for keyword argument 'code'`. Confirmed on `ea730de8` — a valid type whose bundled template is missing and which has no library default:

```
PROMPTS_DIR=/nonexistent-templates  →  resolve_prompt("hyde")
FileNotFoundError: Prompt file not found: `/nonexistent-templates/hyde.txt`
  during handling of the above exception, another exception occurred:
TypeError: core.utils.exceptions.OpenRAGError.__init__() got multiple values for keyword argument 'code'
```

That state isn't hypothetical — it's exactly what `seed_defaults` leaves behind. A type whose disk template is missing is logged and **skipped**, so it ends up with no default, and every later resolution for it walks to the disk seed and fails. `PROMPTS_DIR` is a documented deployment knob, so a custom directory missing one file lands here.

The effect is the opposite of this branch's intent: the docstring promises "a typed `ConfigError` naming the type instead of letting a bare `FileNotFoundError` surface as an opaque 500", and what surfaces is an even more opaque `TypeError` about keyword arguments, with the prompt type nowhere in the message. On the ingest path the broad `except Exception` in `_resolve_ingest_prompts` masks it; on the chat path it's a 500.

It's the only `ConfigError(..., code=...)` call in the package — every other site takes the default code — and nothing covers it (no test references `ConfigError` or `PROMPT_UNAVAILABLE`), which is why a green suite doesn't catch it.

Fix is either dropping the kwarg, or giving `ConfigError` a `code` parameter with `"CONFIG_ERROR"` as its default the way `ValidationError` and `NotFoundError` already do. The second keeps the distinguishable code you clearly wanted here. Either way the path is worth one test, since by construction it only executes when something is already wrong.

`openrag/services/orchestrators/prompt_service.py:243-247`

openrag/services/inference/_call_log.py
Ahmath-Gadji: Verified fixed in https://github.com/linagora/openrag/commit/309f6140f386997d87e434ba93e400c440e46709.

Re-ran the repro on `ea730de8` with a DEBUG sink attached — braces in `model`, `endpoint` *and* message content all pass through now, and they survive verbatim rather than being eaten or doubled:

```
llm.call VLLMClient.chat model=gpt{x} stream=False | user[6]: hi {z}
```

Laziness is intact: an INFO-only sink still builds no preview (re-checked with a payload whose `split`/`__len__` raise). `test_a_brace_in_the_client_supplied_model_does_not_raise` pins it.

openrag/services/orchestrators/prompt_service.py
Ahmath-Gadji: Verified fixed in https://github.com/linagora/openrag/commit/309f6140f386997d87e434ba93e400c440e46709.

```
prompt.resolve sys_prompt <- named:my{tmpl} | 'content {c}'
```

No raise, name intact. Covered by `TestLoggingIsBraceSafe::test_a_brace_in_a_prompt_name_does_not_raise`.

openrag/services/workers/indexer_pool.py
Ahmath-Gadji: Verified fixed in https://github.com/linagora/openrag/commit/309f6140f386997d87e434ba93e400c440e46709.

Reading the defaults out of `IndexationPipelineConfig.model_fields` is better than what I suggested — the two can't drift at all now. Checked on `ea730de8`:

```
enable_image_captioning   -> True
enable_contextualization  -> False
enable_topic_tagging      -> False
bogus_flag                -> False        # unknown flag stays off
```

and a config that omits every flag now resolves `['image_captioning']` instead of `[]`.

openrag/services/persistence/schema.py
Ahmath-Gadji: Verified fixed in https://github.com/linagora/openrag/commit/309f6140f386997d87e434ba93e400c440e46709 — the port headers now read "Selection by name" / "Usage counts and the global default (one per type)", and the `_PRESET_FIELD_TO_TYPE` comment sits on the map it describes.

On the count: `spoken_style_answer` came back in https://github.com/linagora/openrag/commit/60aac502623d9e2551bedd1cb95cc08f15cf300c, so "8 canonical prompt types" is correct again — and it is genuinely 8 in every place that enumerates them (`PromptType`, `_PROMPT_TYPE_VALUES`, the migration CHECK, `PromptTypeName`, `_TYPE_TO_CONFIG_KEY`, `PROMPT_TYPE_KEYS`, `PromptsConfig`, plus the bundled template). Nothing left at 7.

openrag/services/orchestrators/query_service.py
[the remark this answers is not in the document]
Ahmath-Gadji: Verified fixed in https://github.com/linagora/openrag/commit/309f6140f386997d87e434ba93e400c440e46709 — caught at the single choke point, as discussed.

Checked on `ea730de8` with a repo whose `get_by_name`/`get_default` both raise `RuntimeError("pool exhausted")`:

```
WARNING  Prompt lookup failed for 'sys_prompt'; falling back to the bundled template: pool exhausted
DEBUG    prompt.resolve sys_prompt <- disk-seed | 'You are an AI conversational assistant…'
```

Returns the bundled template instead of propagating, one warning, and the `disk-seed` source is still logged so the degradation is visible. `TestResolveSurvivesRepositoryFailure` pins it.

`except Exception` is the right width here — `CancelledError` is a `BaseException` on 3.12, so request cancellation still propagates. One knock-on: the catch now also spans `_log_resolution`, so a logging failure would silently degrade to the seed. That was my worry about masking the brace bug, and it's moot now that the brace bug is fixed at the source rather than swallowed here.

Ahmath-Gadji:
## Verification of `ea730de8`

Re-checked every finding against the branch, not just the diff — each repro from the original review re-run, plus the local suite (**2295 passed**, up 6) and `ruff check` / `ruff format --check` clean. CI green on `ea730de8`.

| # | Finding | Status |
|---|---|---|
| 1 | Braces in log metadata → `KeyError` out of the request path | ✅ fixed `309f6140` — both sites, re-ran both repros; braces pass through verbatim, laziness intact |
| 2 | Per-request Postgres dependency with no fallback | ✅ fixed `309f6140` — repo raising `RuntimeError` now degrades to the bundled template with one warning |
| 3 | `.get(flag)` dropping the model default | ✅ fixed `309f6140` — defaults read from `model_fields`, so they can't drift; a sparse config resolves `['image_captioning']` |
| 4 | `/prompts` undocumented | ✅ fixed `ea730de8` — accurate against the code, including the `used_by` semantics and the write-time template rules |
| 5 | Stale count / misfiled comments | ✅ fixed `309f6140`, count correctly back to 8 via `60aac502` |
| — | Seeding race crash-loop (I'd flagged, not asked you to fix) | ✅ fixed `ea730de8` |

Each fix has a regression test that fails without it, which is the part that matters — three of these were invisible to a green suite before.

`60aac502` is a good catch of your own and I'd have missed it: `metadata.spoken_style_answer` is a documented public flag and a Chainlit command, so dropping it was a breaking change, and the reason nothing caught it is that nothing asserted the flag swapped the prompt. Restoring it as a library type is consistent everywhere I checked — `PromptType`, `_PROMPT_TYPE_VALUES`, the migration CHECK, `PromptTypeName`, `_TYPE_TO_CONFIG_KEY`, `_PROMPT_FORMAT_FIELDS` (same `{context}`/`{current_date}` pair as `sys_prompt`, correct since the same call site renders both), the partition key whitelist, `PromptsConfig`, the bundled template, docs and i18n. Migration chain is still a single head.

**One new finding**, which the fixes brought within reach rather than caused — [`prompt_service.py:243-247`](openrag/services/orchestrators/prompt_service.py#L243): the `raise ConfigError(..., code="PROMPT_UNAVAILABLE")` throws `TypeError` instead of the typed error, because `ConfigError` hard-codes `code` and forwards `**kwargs`. Reachable via a `PROMPTS_DIR` missing one template — `seed_defaults` skips that type, so it has no default and every resolution walks to the failing disk seed. Details and repro in the inline thread.

Two things left open by design, unchanged: the inert `vlm_caption_prompt_name` field in the preset UI (frontend PR), and preset `*_prompt_name` references still not validated at write time (your follow-up #1).

Ahmath-Gadji (dismissed):
Approving on `764a79e1`. Every finding from the review is addressed, each with a regression test that fails without its fix.

The `ConfigError` correction went the way I'd argued against, and it's the better call: making `code` **keyword-only** is safer than the variant I sketched (it can't be passed positionally, so it can't shadow anything), the four existing `ConfigError` sites are untouched, and the coverage added with it addresses the actual root cause rather than the symptom — a `raise` that was reasoned about but never executed. 91% → 96% branch coverage on `prompt_service`, with every raise this PR adds now driven by a test. That's worth more than the one-line deletion I proposed.

Verified on this head:

- The original repro now yields the intended error: `ConfigError | code=PROMPT_UNAVAILABLE | status=500`, message naming the type.
- Re-swept every exception construction site in `openrag/` (42 sites passing `code=`/`status_code=`, checked against what each class accepts): **0 raise `TypeError`**, down from 1.
- Local unit suite **2303 passed**, `ruff check` and `ruff format --check` clean.

Nothing outstanding on my side. The two items left open are deliberate and stated: the inert `vlm_caption_prompt_name` field in the preset UI (frontend PR), and preset `*_prompt_name` references not validated at write time (follow-up #1 in the description).

Ahmath-Gadji (approved):
Re-approving on `6cbc40dd` after the `develop` merge. Reviewed the conflict resolution rather than just the result — both sides survive in both files:

**`query_service.py`** — `develop`'s `prepend_system_prompt`, `force_retrieval`, the `requires_retrieval` conversational skip, the `web_source_numbers` filtering and the reshaped `_prepare_completions` are all intact, and each of the three answer-prompt sites now resolves through the library instead of the `__init__`-time snapshots this branch deletes.

**`retrieval_service.py`** — `develop`'s `_resolve_reranker` / `_default_reranker_name` land intact, with the reranker line correctly switched to the new helper, alongside this branch's `async _pipeline_for_partition` and hyde/multi_query template resolution.

`6cbc40dd` is the part worth calling out: the no-retrieval conversational reply from #807 read `self._sys_prompt_tmplt`, and git auto-merged that reference **without** reporting a conflict — so the resolution was live but unasserted. Adding the test that proves the conversational path resolves through `PromptService` (and honours the partition's selection) is exactly the right response to a silent auto-merge, and the commit message says so plainly.

Verified on this head:

- No leftover references to the three removed `__init__` prompt attributes anywhere in `openrag/` or `tests/`.
- Migration graph still a **single head** (`e8f9a0b1c2d3`, 15 revisions) — the merge brought no new revisions, so the multi-head hazard this branch already hit once didn't recur.
- All five original findings still hold: braces inert in both log lines, a repo failure degrading to the bundled seed, the captioning flag defaulting to `True`, and `ConfigError` carrying `PROMPT_UNAVAILABLE`.
- **2429 unit tests pass** (2303 → 2429 with `develop`'s), `ruff check` + `ruff format --check` clean, `check_layer_imports.py` OK.

Nothing outstanding. The two open items remain deliberate: the inert `vlm_caption_prompt_name` field in the preset UI (frontend PR), and preset `*_prompt_name` references not validated at write time (follow-up #1).

PR #836, opened 2026-07-29
feat(ui): admin prompt library and prompt selection in the preset/partition editors

andyne13:
Admin UI for the prompt library added in #835. **Stacked on that PR — review
and merge it first**; this branch targets `feat/pm-backend`, so the diff here
is UI-only.

Part of #772.

## What it adds

**Prompt Library page** (`/admin/prompts`, new sidebar entry) — prompts grouped
by concern (Answer / Indexation / Retrieval) as cards, with a drawer editor
offering an edit tab and a preview tab that renders the template against sample
values. Insert-variable buttons only offer the placeholders the type actually
accepts, so the editor can't produce a template the API would reject.

Each card carries a used-by badge and a default badge, and exposes the same
card-action affordances as Presets and Models rather than a new pattern.

**Selection lives in the editors that own the setting**, not in a separate
assignment screen:

- Preset editor — indexation presets pick contextualization / image-captioning /
  topic-tagging prompts; retrieval presets pick query-contextualizer, plus HyDE
  or multi-query, shown only when the retriever type actually uses them.
- Partition editor — an Answer section picks the final answer prompt.

Every picker offers "Use default", which clears the name and falls back to the
type's global default.

## Also in here

A few consistency fixes this page surfaced: the Jobs, Presets and Models tabs
now match the Library's tab sizing, the Jobs status tabs are title-cased, the
Default badge renders identically in Prompts and Models, and the partition
config rows no longer overflow their grid cells.

## Verification

`tsc -b` clean, 176 vitest tests green, production build clean. eslint reports
one pre-existing warning in `models.tsx`, untouched here.

Driven end-to-end against a live deployment: creating a prompt, selecting it in
a preset and in a partition, and confirming the selection reached the model.

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only Prompt Library (create/edit, preview with placeholder validation, set default, delete) and a new `/prompts` admin route.
  * Added “Prompts” to the sidebar, visible only with prompt-management permission.

* **Improvements**
  * Expanded prompt API and updated prompt/partition/preset wiring to support prompt-type–driven selection with “use default”.
  * Enhanced General Settings to edit `generation_prompt_names`; refined several admin UI details (tabs, badges, layout/spacing).

* **Tests**
  * Added pagination coverage for prompt listing and updated prompt/template utility and related admin UI tests.

### DIFF
--- a/ui/src/lib/prompt-meta.ts
+++ b/ui/src/lib/prompt-meta.ts
@@ -92,49 +92,77 @@ export interface TemplateScan {
   error?: string;
 }
 
-/** Scan a template the way Python's ``string.Formatter().parse`` does.
+type TemplateToken =
+  | { kind: "literal"; text: string }
+  | { kind: "field"; root: string; raw: string };
+
+interface TokenizeResult {
+  tokens: TemplateToken[];
+  malformed: boolean;
+  error?: string;
+}
+
+/** Tokenize a template the way Python's ``string.Formatter().parse`` does.
  *
- *  The backend validates with the real thing, so approximating it with a
- *  `/\{(\w+)\}/` regex let templates through that were rejected only on save:
- *  `{{` / `}}` are literal braces, a lone brace is an error, and a field may
- *  carry a conversion (`!r`), a format spec (`:>10`) or attribute/index access
- *  (`a.b`, `a[0]`) — all of which reduce to the root name.
+ *  Single source of grammar for both validation and the preview, so the two can
+ *  never disagree about what a template means: `{{`/`}}` are literal braces, a
+ *  lone brace is an error, and a field may carry a conversion (`!r`), a format
+ *  spec (`:>10`) or attribute/index access (`a.b`, `a[0]`), all of which reduce
+ *  to the root name the backend validates against.
  */
-export function scanTemplate(content: string): TemplateScan {
-  const fields: string[] = [];
+function tokenizeTemplate(content: string): TokenizeResult {
+  const tokens: TemplateToken[] = [];
+  let literal = "";
   let i = 0;
+  const flush = () => {
+    if (literal) {
+      tokens.push({ kind: "literal", text: literal });
+      literal = "";
+    }
+  };
   while (i < content.length) {
     const ch = content[i];
     if (ch === "{") {
       if (content[i + 1] === "{") {
+        literal += "{";
         i += 2;
         continue;
       }
       const end = content.indexOf("}", i + 1);
       if (end === -1) {
-        return { fields, malformed: true, error: "Single '{' encountered in format string" };
+        flush();
+        return { tokens, malformed: true, error: "Single '{' encountered in format string" };
       }
-      const root = content
-        .slice(i + 1, end)
-        .split("!")[0]
-        .split(":")[0]
-        .split(".")[0]
-        .split("[")[0]
-        .trim();
-      if (!fields.includes(root)) fields.push(root);
+      const raw = content.slice(i + 1, end);
+      const root = raw.split("!")[0].split(":")[0].split(".")[0].split("[")[0].trim();
+      flush();
+      tokens.push({ kind: "field", root, raw });
       i = end + 1;
       continue;
     }
     if (ch === "}") {
       if (content[i + 1] === "}") {
+        literal += "}";
         i += 2;
         continue;
       }
-      return { fields, malformed: true, error: "Single '}' encountered in format string" };
+      flush();
+      return { tokens, malformed: true, error: "Single '}' encountered in format string" };
     }
+    literal += ch;
     i += 1;
   }
-  return { fields, malformed: false };
+  flush();
+  return { tokens, malformed: false };
+}
+
+export function scanTemplate(content: string): TemplateScan {
+  const { tokens, malformed, error } = tokenizeTemplate(content);
+  const fields: string[] = [];
+  for (const token of tokens) {
+    if (token.kind === "field" && !fields.includes(token.root)) fields.push(token.root);
+  }
+  return { fields, malformed, error };
 }
 
 export function extractPlaceholders(content: string): string[] {
@@ -173,34 +201,30 @@ export interface PreviewSegment {
   varName?: string;
 }
 
-/** Substitute each known `{var}` with its sample value, tracking which spans
- *  were substituted so the preview can highlight them. */
+/** Render the template the way the pipeline will: substitute each known
+ *  `{var}` with its sample value, unescape `{{`/`}}`, and track which spans were
+ *  substituted so the preview can highlight them.
+ *
+ *  Shares `tokenizeTemplate` with validation so what an author sees here matches
+ *  what the model receives — a separate regex used to leave escaped braces
+ *  doubled and skip formatted fields entirely.
+ */
 export function renderPreview(content: string, promptType: string): PreviewSegment[] {
+  // Verbatim types are never `.format`-ed at runtime, so they preview as-is.
+  if (!FORMATTED_PROMPT_TYPES.has(promptType)) {
+    return content ? [{ value: content, isVariable: false }] : [];
+  }
   const vars = PROMPT_TYPE_VARIABLES[promptType] ?? [];
   const varMap = Object.fromEntries(vars.map((v) => [v.name, v.sample]));
+  const { tokens } = tokenizeTemplate(content);
 
-  const segments: PreviewSegment[] = [];
-  let lastIndex = 0;
-  let m: RegExpExecArray | null;
-  const regex = /\{(\w+)\}/g;
-
-  while ((m = regex.exec(content)) !== null) {
-    if (m.index > lastIndex) {
-      segments.push({ value: content.slice(lastIndex, m.index), isVariable: false });
-    }
-    const varName = m[1];
-    const sample = varMap[varName];
-    if (sample !== undefined) {
-      segments.push({ value: sample, isVariable: true, varName });
-    } else {
-      segments.push({ value: m[0], isVariable: false });
-    }
-    lastIndex = m.index + m[0].length;
-  }
-  if (lastIndex < content.length) {
-    segments.push({ value: content.slice(lastIndex), isVariable: false });
-  }
-  return segments;
+  return tokens.map((token) => {
+    if (token.kind === "literal") return { value: token.text, isVariable: false };
+    const sample = varMap[token.root];
+    return sample !== undefined
+      ? { value: sample, isVariable: true, varName: token.root }
+      : { value: `{${token.raw}}`, isVariable: false };
+  });
 }
 
 /* ---------- Prompt picker option values ---------- */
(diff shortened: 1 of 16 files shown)


ui/src/pages/admin/prompts.tsx
```diff
@@ -0,0 +1,575 @@
...
+    if (!name.trim()) {
+      toast.error("Name is required");
+      return;
+    }
+    if (!content.trim()) {
+      toast.error("Content is required");
+      return;
+    }
+    if (editing) {
+      updateMut.mutate({ id: editing.id, name, content });
```
hedhoud, line 398: Renaming a prompt that is already in use silently disconnects its name-based references. Existing presets and partitions keep the old name and then fall back to the global default, even though this drawer says edits apply to current users. Please preserve those assignments during rename, make the name immutable while it is used, or require an explicit warning that explains this behavior.

ui/src/pages/admin/partitions/detail.tsx
```diff
@@ -150,6 +182,7 @@ function GeneralTab({ partition }: { partition: PartitionConfig }) {
         indexation_preset: indexationPreset,
         retrieval_preset: retrievalPreset,
         chat_llm: chatLlm === "__default__" ? null : chatLlm,
+        generation_prompt_names: generationPrompts,
```
hedhoud: This resubmits the answer-prompt mapping on every partition save. If the selected prompt was renamed or deleted, the backend revalidates the stale name and rejects unrelated changes such as the description or chat depth. A non-admin owner cannot recover because this picker is disabled for them. Please avoid sending the mapping when it was not changed, or otherwise preserve the backend's fail-open behavior for stored stale references.

ui/src/lib/prompt-meta.ts
```diff
@@ -0,0 +1,141 @@
...
+  const matches: string[] = [];
+  let m: RegExpExecArray | null;
+  PLACEHOLDER_REGEX.lastIndex = 0;
+  while ((m = PLACEHOLDER_REGEX.exec(content)) !== null) {
+    if (!matches.includes(m[1])) matches.push(m[1]);
+  }
+  return matches;
+}
+
+export function validatePlaceholders(content: string, promptType: string) {
```
hedhoud, line 197: This validator does not match the backend's `str.format` rules. Escaped braces, format expressions, and unbalanced braces can therefore look valid here while producing an incorrect preview or a 422 only after Save. Please use the same template grammar as the backend and prevent submission when the template is malformed or contains an unsupported field.

ui/src/pages/admin/prompts.tsx
```diff
@@ -0,0 +1,575 @@
...
+// Filter chips: "all" plus one per concern group.
+const CONCERN_FILTERS = ["all", ...PROMPT_GROUPS.map((g) => g.name)] as const;
+
+export default function PromptsPage() {
+  const queryClient = useQueryClient();
+  const [concern, setConcern] = useState<(typeof CONCERN_FILTERS)[number]>("all");
+  const [editorOpen, setEditorOpen] = useState(false);
+  const [editing, setEditing] = useState<PromptResponse | null>(null);
+
+  const { data, isLoading } = useQuery({
```
hedhoud: A failed prompt-list request is currently treated as an empty successful result because only the loading state is handled. That tells an administrator there are no prompts and shows zero counts during an API or database failure. Please surface a clear error state and keep cached data visible during refetch failures.

ui/src/pages/admin/presets.tsx
```diff
@@ -598,6 +602,45 @@ function FeatureToggle({
...
+  onChange: (v: string) => void;
+}) {
+  return (
+    <div className="space-y-1.5">
+      <div className="flex items-center gap-1">
+        <Label className="text-xs">{label}</Label>
+        {prompts.length > 0 && <PromptViewButton prompts={prompts} selectedName={value} />}
+      </div>
+      <Select
+        value={value || "__default__"}
```
hedhoud: `__default__` is also a valid prompt name according to the API, but this control reserves it as the fallback sentinel. A prompt with that name can be created yet can never be selected. Please use a non-colliding internal value or reject the reserved name consistently across creation and every prompt picker.

hedhoud (changes requested):
The latest update is a clean rebase, so the functional patch is unchanged. The inline comments cover two behavior blockers and three user-facing correctness issues that should be addressed before approval. The existing unresolved cache-invalidation comment also remains valid.

andyne13:
@hedhoud thanks for the review — all five points are addressed, could you take another look when you have a moment?

- **Template validation** now mirrors the backend's `str.format` grammar instead of a `{(\w+)}` regex: `{{`/`}}` are literal, a lone brace is an error, and conversions / format specs / attribute access reduce to the root field. Malformed or unknown-variable templates are blocked before the request rather than coming back as a 422. Verbatim prompt types aren't format-validated at all, matching `_PROMPT_FORMAT_FIELDS`.
- **Rename of a prompt in use** now warns explicitly (inline on the field and again on submit), naming how many partitions lose their selection and that they fall back to the default.
- **`generation_prompt_names`** is only sent when this editor actually changed it, so an untouched-but-stale reference can no longer 422 an unrelated description or chat-depth edit — which a non-admin owner couldn't have recovered from.
- **Failed prompt-list request** shows an error with a retry instead of rendering as an empty library, and cached data stays visible across a failed refetch.
- **`__default__`** is no longer a reserved value: picker options are namespaced, so a prompt with that name stays selectable.

Also from CodeRabbit's follow-up: the preview shared the old regex, so it disagreed with the new validation on escaped braces and formatted fields — both now share one tokenizer.

Covered by new unit tests in `ui/src/lib/prompt-meta.test.ts` (204 green overall, `tsc` and build clean).

ui/src/lib/prompt-meta.ts
```diff
@@ -0,0 +1,260 @@
...
+        i += 2;
+        continue;
+      }
+      const end = content.indexOf("}", i + 1);
+      if (end === -1) {
+        flush();
+        return { tokens, malformed: true, error: "Single '{' encountered in format string" };
+      }
+      const raw = content.slice(i + 1, end);
+      const root = raw.split("!")[0].split(":")[0].split(".")[0].split("[")[0].trim();
```
hedhoud, line 140: Reducing an expression to its root makes unsupported formatting look safe. For example, `{context!x}` and `{context.missing}` both pass this check and preview as the sample value, but Python raises when the prompt is formatted at runtime. If such a prompt becomes the default, every affected request can fail. Please either restrict managed templates to plain supported placeholders or validate the complete expression on the backend and mirror that verdict here.

ui/src/pages/admin/presets.tsx
```diff
@@ -387,9 +393,9 @@ function IndexationPresetForm({
           onModelChange={(v) => set("vlm", v)}
           models={vlms}
           promptLabel="Caption prompt"
-          promptValue={configGet(config, "vlm_caption_prompt_name", "")}
-          onPromptChange={(v) => set("vlm_caption_prompt_name", v || null)}
-          prompts={promptsByType("vlm_caption")}
+          promptValue={configGet(config, "image_captioning_prompt_name", "")}
+          onPromptChange={(v) => set("image_captioning_prompt_name", v || null)}
+          prompts={promptsByType("image_captioning")}
```
hedhoud, line 398: Please preserve the empty value here as a real “Use default” selection. `FeatureToggle` automatically selects `prompts[0].name` whenever this list contains one entry; after seeding, that entry is normally the global default. Merely opening and saving an indexation preset then stores an explicit name, so promoting another global default later no longer affects that preset. Prompt options should not inherit the model auto-selection behavior.

ui/src/pages/admin/prompts.tsx
```diff
@@ -0,0 +1,629 @@
...
+
+export default function PromptsPage() {
+  const queryClient = useQueryClient();
+  const [concern, setConcern] = useState<(typeof CONCERN_FILTERS)[number]>("all");
+  const [editorOpen, setEditorOpen] = useState(false);
+  const [editing, setEditing] = useState<PromptResponse | null>(null);
+
+  const { data, isLoading, isError, error, refetch, isFetching } = useQuery({
+    queryKey: ["prompts-library"],
+    queryFn: () => listPrompts({ limit: 500 }),
```
hedhoud: This request exposes only the first 500 prompts as if it were the complete library. Because the API supports offset pagination, prompt 501 and later become invisible in the management page and selectors, while the displayed counts are partial. Please fetch all pages or expose real pagination so every stored prompt remains manageable.

hedhoud (changes requested):
The latest update addresses the earlier rename, stale-reference, error-state, sentinel, and cache findings. Three correctness issues still need attention before approval: runtime-unsafe template expressions, silent pinning of the current default in indexation presets, and the one-page prompt-library limit.

ui/src/lib/prompt-meta.ts
andyne13: Good catch — and it was worse than the editor: the backend had the same flaw, so this was reachable through the API too. `_validate_template` reduced a field to its root name, so `{context!x}` and `{context.missing}` were accepted and then raised `ValueError` / `AttributeError` at render time. As a type's global default that fails every request falling back to it.

Taking your first option: managed templates are now restricted to plain placeholders — no conversion, format spec, or attribute/index access. Fixed on the backend in #835 (`8c207a9e`, authoritative) and mirrored here so the editor reaches the same verdict before the request. A test asserts all seven bundled templates still pass, so the stricter rule can't make the seed path skip a type and leave it with no default.

ui/src/pages/admin/presets.tsx
andyne13: You're right, and the consequence is exactly as you describe: after seeding a type usually has precisely one prompt — its own global default — so merely opening and saving an indexation preset pinned that name, and promoting a different default later silently stopped affecting the preset.

Removed the prompt auto-selection (both the toggle handler and the effect that fired when the list resolved later). An empty value is now preserved as a real "Use default". Models keep the behaviour, since they have no equivalent fallback — noted in a comment so the asymmetry is deliberate rather than accidental.

ui/src/pages/admin/prompts.tsx
andyne13: Agreed — the cap made prompts past it invisible in the library *and* in every picker, while the header count read as the total.

Added `listAllPrompts()`, which walks the offset pages until a short page and is now used by the library page, the preset editor and the partition editor. Covered by tests for the multi-page walk, the single short page, and the empty library.

andyne13:
@hedhoud all three are addressed — replies inline on each thread.

The placeholder one turned out to be a backend bug as well: the same root-reduction was in `_validate_template`, so `{context!x}` and `{context.missing}` were reachable through the API and would have crashed at render time. Managed templates are now restricted to plain placeholders on the backend (#835, `8c207a9e`) with the editor mirroring that verdict here. Thanks for spotting it — that one would have been ugly as a global default.

Ready for another look when you have a moment.

hedhoud (dismissed):
Rechecked the latest update. The three blocking issues are addressed, and the focused tests and UI build pass locally. This is ready from my side.

hedhoud (approved):

PR #838, opened 2026-07-30
Release 2.1.0

andyne13:
Cuts v2.1.0 from `develop` (`187a359e`) per the adopted git flow. Feature release, not a patch bundle: it ships a DB-backed prompt library, user-management search and a partition-member picker.

## What's in it

| PR | |
|---|---|
| #835 / #836 | Prompt management — backend + admin UI (new prompt library, per-type defaults) |
| #807 | Hide empty Chainlit sources |
| #809 | User search in user management |
| #810 | Partition-member picker |
| #769 | Reject unsafe model-endpoint names; cascade renames safely (#768) |
| #633 | Helm chart refactor + admin-ui arbitrary-UID fix |
| #767 | Harden the GA publish guard (exact tag format, no shell interpolation) |
| #765 | Back-merge of v2.0.1 |

## Version bump (`4cb7e7ff`)

- `pyproject.toml` -> `2.1.0`, `uv.lock` regenerated (1 line, no dependency churn)
- `Chart.yaml` -> `appVersion: "2.1.0"`, chart `version: 0.6.1` (on top of #633's `0.6.0`)
- `values.yaml` -> 3 image tags to `v2.1.0` (openrag, openrag-ray, openrag-admin-ui)
- `docker-compose.yaml` -> 2 image pins to `v2.1.0`

No `v2.0.1` string remains outside docs and the lock file.

## Verification

- `develop` deployed and tested by a teammate at `376d7c33`; the delta to `187a359e` touches zero files under `openrag/`, `conf/`, `pyproject.toml` or `uv.lock` — chart, docs, `ui.Dockerfile` and infra tests only.
- `ui.Dockerfile` is the only image input changed since v2.0.1, and it had never been built by CI (the nightly builds api + ray only). Built locally from this branch: serves `GET /app/` -> 200 both as the image's own user (uid 10001, gid 0) and as an arbitrary UID with gid 0 (restricted-v2 simulation), nginx workers start, no permission errors.
- api + ray build green on every push to `develop`, most recently on `187a359e`.
- Locally: ruff check, ruff format --check, layer-import guard, `tests/unit/infra` all pass.

## Also adds `.github/RELEASING.md` (`184153ad`)

v2.0.1 produced a green `build.yml` run that built nothing — three jobs `skipped` behind an unusable `base_ref` guard. The guard is fixed, but nothing in our process would have caught it. This records the checks that prove a release actually shipped: per-job conclusions rather than the run's, manifest digests across all five registry coordinates, `latest` matching the release, ghcr and Docker Hub carrying the same build, and the version baked into the image.

## After merge

1. Tag `v2.1.0` on the merge commit **on main** -> `build.yml` publishes GA + `latest`
2. Work through `.github/RELEASING.md` before announcing anything
3. Publish the GitHub Release (manual — `build.yml` doesn't create one)
4. Back-merge `main` -> `develop`

Note: `main`'s hardened `verify-tag` guard has never run on a GA tag; this will be its first. Its logic was pre-flighted locally (tag regex, anonymous fetch of main) and it fails loud rather than skipping silently.


Ahmath-Gadji (approved):
LGTM

PR #840, opened 2026-07-31
fix(release): make the release checklist able to fail

andyne13:
Makes `.github/RELEASING.md` able to fail. Every check in it looked like a gate but could pass on a broken release; each one below was reproduced before being fixed, and the fixed version was executed rather than reasoned about.

### Round 1 — the four issues raised on #839

| | |
|---|---|
| `dg()` | Piped a failed registry lookup into `sha256sum`, which hashes empty input and returns `sha256:e3b0c442…` — a real-looking digest. Step 2 reported a **missing image as present**, and two missing tags compared equal so steps 3 and 4 printed `OK`. |
| Step 1 | The job-conclusion check only *printed* the count of non-success jobs and exited 0 regardless. |
| Step 5 | Compared `RepoDigests` (`repo@sha256:…`) against a bare `sha256:…`. Those can never match literally. |
| Step 8 | Matched any version-shaped tag instead of comparing to `$VER`, so a pin left at the previous release satisfied it. |

### Round 2 — four more false-pass paths

| | |
|---|---|
| Step 1 | Validated only the jobs the API returned, so a job that never ran contributed no failing conclusion and the gate passed on the strength of the others. Now requires each of the four jobs present exactly once and `success`. |
| Step 5 | Ignored the `docker pull` exit status, letting a failed pull fall through to an older cached image, and read `{{index .RepoDigests 0}}` from an unordered list. Now gates on the pull and selects the entry by repository. |
| Step 8 (chart) | `grep -A4` matched a repository named inside a comment, then read a neighbouring block's tag. |
| Step 8 (compose) | Counted `$VER` interpolated into a regex, so `v2x1y0` passed; a count of two was also reached by a duplicate pin, or by two commented-out lines. Now matches active `image:` values as fixed whole strings, one per repository, rejecting strays. |

### Round 3 — the chart lookup, parsed properly

The round-2 replacement for `grep -A4` was itself wrong, in two ways both reviewers found:

- **Unbounded block scope** — with a repository declaring no `tag:`, the pending lookup outlived its `image:` block and consumed a later block's tag. A completely unpinned image reported a valid version.
- **Duplicate sibling `tag:`** — the scanner stopped at the first value; YAML resolves duplicates last-wins, which is what Helm deploys. A correct pin followed by a stale one passed.

That was the third text-matching bug in one check, so the scanner is gone. `values.yaml` is now parsed with PyYAML: comments disappear by construction, block boundaries are real object boundaries, and a strict loader **raises on duplicate keys** rather than guessing which value Helm would use.

The chart half now needs `python3` + PyYAML. If it is missing the check exits non-zero and step 8 fails — it does not skip.

### Verification

- Step 1 run against the **v2.0.1 run that built nothing** (`30034799802`): fails, and reports `verify-tag` absent from that run entirely. Against the v2.1.0 run (`30620163814`): passes.
- Step 5 run end-to-end against the live registry and a real pull: pulled digest matches.
- Step 8 run verbatim against the real `v2.1.0` tag: `step 8 PASS`, exit 0 — the stricter checks do not regress the release that already shipped.
- Every adversarial fixture fails: unpinned image, duplicate keys, commented-out repository, duplicated image block, regex-wildcard compose version, duplicate compose pin, stray stale pin.

No change to the v2.1.0 release, which was verified with real, distinct digests and independently confirmed by pulling the image and reading `2.1.0` out of it.


hedhoud (commented):
There are four cases where this checklist can still report success without proving the release is valid:

1. The workflow check validates only the jobs GitHub returned. If one of the four required publishing jobs is missing entirely, the remaining successful jobs still make the gate pass.

2. The image check continues after a failed pull. When the same tag already exists locally, Docker can inspect that cached image and compare an old digest, so the step may print OK even though the release image was not fetched.

3. The chart lookup is text-based and can match a repository name inside a comment. It may then read a nearby active tag and validate the wrong image block.

4. The Compose check relies on a combined regex count. A comment, duplicate entry, or version containing characters matched by the unescaped dots can produce a count of two without having one correct API pin and one correct Admin UI pin.

These are release gates, so a false pass is risky: each check should fail unless the exact expected job or image pin is present and was successfully verified.

.github/RELEASING.md
```diff
@@ -166,14 +233,96 @@ The chart and compose pins are part of the release surface; shipping them
...
+pin_for() {
+  awk -v want="$1" '
+    { sub(/#.*/, "") }
+    match($0, /^[[:space:]]*repository:[[:space:]]*"[^"]*"[[:space:]]*$/) {
+      ind = index($0, "repository") - 1
+      v = $0; sub(/^[^"]*"/, "", v); sub(/".*$/, "", v)
+      if (v == want) { n++; pend = 1; pind = ind } else { pend = 0 }
+      next
+    }
+    pend && match($0, /^[[:space:]]*tag:[[:space:]]*"[^"]*"[[:space:]]*$/) {
```
hedhoud: The lookup is not bounded to the current `image:` block. If the target repository has no `tag:`, `pend` remains active and can consume a same-indented tag from a later image block. When that unrelated tag equals `$VER`, the release check reports success even though the target image is unpinned. Please reset the pending lookup at the image-block boundary or parse the YAML structurally.

.github/RELEASING.md
```diff
@@ -166,14 +233,96 @@ The chart and compose pins are part of the release surface; shipping them
...
+      if (index($0, "tag") - 1 == pind) {
+        v = $0; sub(/^[^"]*"/, "", v); sub(/".*$/, "", v)
```
hedhoud: This also accepts duplicate sibling `tag:` keys. The first matching tag is stored and clears `pend`, so a correct first value followed by a stale duplicate can pass this check while Helm uses the later effective value. Please require exactly one sibling tag in the target image block before accepting the pin.

hedhoud (commented):
The Helm pin gate still has two false-pass cases that can let a release proceed without validating the value Helm will actually deploy.

.github/RELEASING.md
andyne13: Confirmed, and reproduced before changing anything: with `linagoraai/openrag` declaring `repository` but no `tag`, the pending lookup stayed armed across the block boundary and returned the *next* block's `v2.1.0`. The gate passed on a completely unpinned image.

Taking the second option you offered — parsing structurally. `pin_for` is gone; the chart pins are now read with PyYAML by walking the document for mappings that declare a `repository`, so block boundaries are real object boundaries rather than indentation guesses, and comments disappear by construction. That also retires the earlier `grep -A4` failure mode for good.

The same fixture now reports `FAIL  linagoraai/openrag -> no sibling tag: (image is unpinned)`.

.github/RELEASING.md
andyne13: Also confirmed. I checked which value actually wins: PyYAML resolves a duplicate `tag` last-wins (so `v2.0.0` in my fixture), while the scanner stopped at the first and reported `v2.1.0` — a correct pin followed by a stale one passed, exactly as you described.

Rather than require exactly one sibling tag, the loader now rejects the file: a strict constructor raises on any duplicate key in `values.yaml`, so the check refuses to guess which value Helm would deploy instead of silently picking one.

Output on that fixture: `FAIL  values.yaml did not parse: duplicate key 'tag' at line 5 — refusing to guess which value Helm would use`.

Both fixtures from the previous round still fail, and step 8 still passes against the real `v2.1.0` tag. Note the chart half now needs python3 + PyYAML; if it is missing the check exits non-zero and step 8 fails rather than skipping.

hedhoud (approved):
The revised YAML validation resolves both reported false-pass cases: image pins are now scoped structurally, and duplicate keys are rejected. I also verified the valid release case and the malformed fixtures, and all CI checks are green.

PR #842, opened 2026-07-31
fix(retrieval): return no matches instead of erroring when the vector collection doesn't exist yet

Ahmath-Gadji:
## What

`search()`/`multi_query_search()` on `VectorStoreSearcher` raise instead of
returning an empty result when the shared Milvus collection doesn't exist yet.

## Root cause

The shared collection (`config.vectordb.collection_name`) is created lazily
on the first indexed file *system-wide* (`services/workers/stages/store.py`).
On a fresh deployment — or any retrieval-needing chat/search query before
that first file is indexed, regardless of partition — Milvus raises
`CollectionNotExists`, wrapped as `VDBSearchError`, and propagates as a real
error instead of "no matches."

This is the same class of bug #508 already fixed for `delete_partition`
(`PartitionService`), using the existing `VectorStore.collection_exists()`
port method as the guard. That fix only covered delete; search was still
exposed.

## Fix

Mirror #508's pattern in `VectorStoreSearcher.search()` and
`multi_query_search()`: check `collection_exists()` before embedding/
searching, and return `[]` immediately if the collection is absent — no
collection means no chunks to find, and skipping straight to an empty result
also avoids an unnecessary embedding-API call.

## Test

Adds `test_search_returns_empty_when_collection_does_not_exist` and
`test_multi_query_search_returns_empty_when_collection_does_not_exist`,
asserting the embedder and store are never called when the collection is
absent. Full unit suite: 2433 passed (1 pre-existing unrelated failure on
`develop`, unaffected by this change), ruff clean, layer-import guard clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Search now safely returns no results when the configured vector collection is unavailable.
  * Prevented unnecessary query processing and access attempts for missing collections.
* **Tests**
  * Added coverage for missing-collection search scenarios.
  * Updated existing search tests to reflect collection availability checks.

### DIFF
--- a/openrag/services/storage/vector_store_searcher.py
+++ b/openrag/services/storage/vector_store_searcher.py
@@ -69,6 +69,9 @@ class VectorStoreSearcher(RetrievalSearcher):
         similarity_threshold: float = 0.0,
         with_surrounding_chunks: bool = True,
     ) -> list[Chunk]:
+        # Collection is created lazily on first insert (#508); absent means no chunks.
+        if not await self._store.collection_exists(self._collection):
+            return []
         (embedding,) = await self._embedder.embed([query])
         filters: dict[str, Any] = {"partition": partition}
         if filter:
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
Reopening under the correct account/commit identity.

Issue #778, opened 2026-07-27
Chainlit should hide Sources when no relevant evidence supports the answer


hedhoud:
Context

A source list is useful only when retrieved evidence supports the answer. In the `openrag_all` partition, a conversational message such as "Hello" can receive a normal greeting while still displaying unrelated sources.

Problem

Chainlit can append a Sources section when the source collection is empty or when the request does not need document evidence. This makes unrelated documents appear to support the response and can reduce trust in valid citations.

Expected behavior

- No Sources element is rendered when the source collection is empty, filtered, or unavailable.
- Greetings and other conversational requests that do not require document evidence do not display sources, even if retrieval produced candidates.
- Sources are shown only when they are valid and relevant to the answer.
- Knowledge questions with supporting evidence keep their current citation behavior.

PR #847, opened 2026-08-02
fix(rag): keep sources when the LLM omits the citation tag

Ahmath-Gadji:
## Why

Sources are not cited in a lot of messages: PR #807 changed the fallback so a missing `[Sources: ...]` tag hides all sources instead of keeping them. In practice the model often skips the tag on answers that are genuinely grounded in retrieved documents, so those sources were being dropped from the response.

## Fix

- `filter_sources_by_citations` now keeps all presented sources when no tag is found, instead of returning none.
- `extra.citations_reported` (bool) tells a client whether the model actually emitted a `[Sources: ...]` tag (even an empty one) — `false` is the only case where `sources` falls back to keeping everything, so a client can distinguish "cited every source" from "didn't report citations at all".
- `extra.presented_sources`: everything actually shown to the LLM (after context-budget truncation), regardless of citation.
- `extra.cited_sources`: strictly what the model cited via the tag — unlike `sources`, never falls back to "everything". `sources` itself is left untouched for backward compatibility with existing clients (e.g. Twake); new consumers (e.g. Chainlit) should move to `cited_sources`, falling back to `presented_sources` in the UI when it's empty.
- `extra.all_retrieved_sources`: the complete retrieval set, captured before context-budget truncation (and, on the map-reduce path, before map-reduce replaces retrieved docs with LLM-generated summaries) — so it now genuinely reflects what retrieval returned in `chat`, `chat_stream`, and `complete`. Gated behind `metadata.include_all_retrieved_sources: true` (default off) since it's debug/eval telemetry — retrieval is uncapped up to `retriever.top_k` while the context budget only fits a handful of documents, so this can be a large payload.

## Summary by CodeRabbit

* **New Features**
  * Source metadata now distinguishes presented, cited, and fully retrieved sources.
  * Added citation status reporting for responses with, without, or invalid citation markers.
  * Chat and completion requests can optionally include all unfiltered retrieved sources for evaluation and troubleshooting.
  * Source metadata is now consistent across chat, streaming, conversational, completion, web search, and map-reduce workflows.

* **Bug Fixes**
  * Sources are retained when responses do not include citation markers.
  * Improved source selection in streaming responses when citations are unavailable.


openrag/services/orchestrators/query_service.py
```diff
@@ -670,13 +670,7 @@ async def chat(
...
         chunk["choices"][0]["message"]["content"] = clean
         chunk["extra"] = json.dumps(
-            {
-                "sources": filter_sources_by_citations(
-                    sources,
-                    citations,
-                    allow_uncited=structured_output,
-                )
-            }
+            {"sources": filter_sources_by_citations(sources, citations), "all_retrieved_sources": sources}
```
hedhoud: `all_retrieved_sources` is populated after the context-budget selection, so it excludes documents and web results that retrieval returned but could not fit in the prompt. That makes the field incomplete for debugging and evaluation, which is its stated purpose. Please preserve the complete retrieval set before context truncation and use it here, as well as in the streaming and completion paths.

openrag/core/utils/source_filtering.py
```diff
@@ -263,8 +262,8 @@ async def stream_with_source_filtering(
     else:
         final_clean, citations = pending, None
 
-    filtered = filter_sources_by_citations(sources, citations, allow_uncited=allow_uncited_sources)
-    extra_payload = {"sources": filtered}
+    filtered = filter_sources_by_citations(sources, citations)
+    extra_payload = {"sources": filtered, "all_retrieved_sources": sources}
```
andyne13: Suggest adding a third key here so clients can tell "the model cited exactly these" from "the model reported nothing, so we kept everything". Right now those two states are indistinguishable: with no tag `sources == all_retrieved_sources`, but that is also true when the model genuinely cites every source.

```suggestion
    extra_payload = {"sources": filtered, "all_retrieved_sources": sources, "citations_reported": citations is not None}
```

Same addition would be needed in the two `json.dumps` sites in `query_service.py` (`chat` and `complete`) to keep the three paths consistent.

andyne13 (commented):
Verified the branch locally at `f4bd28a6`: full unit suite passes (2431), `ruff check`, `ruff format --check` and `check_layer_imports.py` are all clean.

Two notes before the findings, since this PR partially reverts #807:

- **It does not reopen #778.** #807's conversational fix works by *not retrieving at all* — the `requires_retrieval=False` branch in `_prepare_chat` returns `payload, [], [], True`, so `sources` is already `[]` and the citation filter never runs. Greetings and identity questions still show no sources.
- The `allow_uncited` removal is correct dead-weight cleanup: structured output previously reached "keep everything" via `allow_uncited=True` and now reaches it via `citations is None`. Same result, and `_allows_uncited_sources` → `_is_structured_output` is the honest name.

### Confirming @hedhoud's point, with a wider scope

The post-truncation issue is real, and the context budget isn't the only thing narrowing the set before it reaches `all_retrieved_sources`:

- `docs = [docs[i] for i in included]` — `query_service.py:510` (context budget), and again at `:574` in `_prepare_completions`, so the `complete()` path has the same gap.
- `web_results = [web_results[number - web_start_index] for number in web_source_numbers]` — `:524`, same for web results.
- `docs = await self._map_reduce(...)` — `:494` **replaces** the retrieved docs with generated summary docs when `use_map_reduce` is set, so on that path `all_retrieved_sources` contains no retrieved document at all.

If the field is meant for debugging and evaluation, it needs to be captured right after retrieval, before all three of these.

### Missing: nothing distinguishes "cited" from "unreported"

This is the one I'd most like to see addressed. After this change `extra.sources` conflates two very different states, and `all_retrieved_sources` doesn't disambiguate them either — when the tag is missing, `sources == all_retrieved_sources`, but that is *also* true when the model legitimately cited every source. Neither the Chainlit front nor the admin UI can tell the two apart, so the #778 failure mode becomes unobservable rather than fixed.

One extra key solves it:

```python
extra_payload = {
    "sources": filtered,
    "all_retrieved_sources": sources,
    "citations_reported": citations is not None,
}
```

That lets a client render "Sources cited" vs "Sources consulted", and makes the tag-omission rate measurable instead of invisible.

### The public API contract wasn't updated

`CLAUDE.md` documents the new field, but the OpenAPI descriptions that clients actually read still say `extra` contains only `sources`:

- `openrag/api/routers/user/chat.py:448` (chat completions)
- `openrag/api/routers/user/chat.py:554` (completions)

Nothing under `docs/content/` describes `extra`, so those two docstrings are the whole public contract for this field.

### PR description is stale

The description says `extra.retrieved_sources`; commit `f4bd28a6` renamed it to `all_retrieved_sources`. The auto-generated walkthrough inherited the old name too. Worth fixing before merge so the changelog is right. The summary bullet "Structured-output responses no longer process citation markers incorrectly" also isn't part of this diff — that gate already exists on `develop`.

### Nits, non-blocking

- **On-wire duplication.** `filtered_json` is serialized once and attached to both the tail chunk and the finish chunk, so a streamed answer now carries the source array four times instead of twice — and in exactly the case this PR targets (no tag), the two arrays are byte-identical. It's bounded, since `_dict_to_chunk` strips `text` from metadata, so I would keep it if unconditional presence is what makes the field useful for eval. Just flagging the cost.
- **Truncated streams.** Upstream dying before the tag now yields `citations is None` → every source marked cited, alongside `truncated: true`. Defensible, but the warning at `source_filtering.py:275` logs `sources=len(filtered)`, which will read as "cited N sources" for answers that cited none.
- **Test coverage asymmetry.** `chat` got `test_chat_without_citation_keeps_retrieved_sources`; `complete()` has no equivalent, and no test asserts that `complete()` emits `all_retrieved_sources` at all. Also worth noting the `prepare_sources=lambda d, w: [...] if d or w else []` stub change is load-bearing — it is more faithful than the old stub (real `__prepare_sources` at `chat.py:214` does return `[]` for empty docs), but it makes the surrounding `assert sources == []` assertions trivially true.

andyne13:
Design note before this gets implemented — shape only, no implementation detail.

Three sets nest here: **retrieved ⊇ presented ⊇ cited**. Retrieved measures recall, presented is what the model could possibly have used, cited is what it claims it used. They answer different questions, so each wants its own stage-accurate name, and no field should mean two different stages depending on circumstance. That's also the cleanest argument against the current fallback: `sources` now flips between "cited" and "everything" with nothing on the wire saying which.

One caution on the fix @hedhoud asked for, worth knowing up front: citation numbers are **positions into the presented list**. The citation-bearing field therefore can't simply be swapped for the fuller set. Web sources are numbered starting from the document count *after* truncation, so widening that list silently misaligns them. The retrieved set needs to arrive as its own field that nothing indexes into.

Also worth flagging that the gap is large rather than marginal: the context budget is sized for ten documents while the retrieval path is uncapped and can return up to `retriever.top_k`. This field would routinely omit most of what was retrieved. Arguably that cap/budget mismatch is the real defect and the new field mostly makes it visible.

Last thought: a full retrieval dump on every response is debug telemetry on the hot path. Behind a request flag, or in traces, would serve evaluation better without growing every answer.

openrag/services/orchestrators/query_service.py
Ahmath-Gadji: Fixed — retrieved_docs/retrieved_web_results are now snapshotted right after retrieval, before the context-budget truncation drops anything that didn't fit the prompt, and used for all_retrieved_sources in chat, chat_stream, and complete (f3c5b2f8). Also caught (thanks to andyne13's follow-up review) that map-reduce was reassigning docs to LLM-generated summaries *before* that snapshot on the chat path, so all_retrieved_sources held summaries instead of anything retrieval actually returned on that path — fixed separately in c564472c, with a test exercising chat() end-to-end with use_map_reduce=True asserting sources (summaries) differs from all_retrieved_sources (originals).

One behavior change worth flagging: all_retrieved_sources is now opt-in via `metadata.include_all_retrieved_sources` (default off) rather than always present — see the design-note thread for why.

openrag/core/utils/source_filtering.py
Ahmath-Gadji: Added exactly as suggested (f782b071) — citations_reported: citations is not None, in the streaming terminal payload plus the two json.dumps sites in query_service.py (chat and complete), same as you flagged in the full review. Test coverage: explicit [Sources: none] vs no-tag-at-all vs citing every source, all asserting citations_reported distinguishes them (the last one is the case you called out as otherwise indistinguishable from the no-tag fallback).

Ahmath-Gadji:
Thanks for both passes — replying to the review and the design note together since they overlap.

**From the review:**

- Confirmed both notes on the #807 revert scope — no action needed there, and agreed `allow_uncited` → `_is_structured_output` is honest cleanup.
- **Post-truncation gap, wider scope**: fixed the context-budget/web-truncation part in f3c5b2f8 (also closes the `complete()` gap you called out at `:574`). The map-reduce replacement at `:494` was a separate miss on my part — `retrieved_docs` was being snapshotted *after* map-reduce reassigned `docs` to LLM summaries, so `all_retrieved_sources` held summaries instead of anything retrieved on that path. Fixed in c564472c, with a `chat()`-level test asserting `sources` (summaries) differs from `all_retrieved_sources` (originals) when `use_map_reduce=True`.
- **citations_reported**: added exactly as you specified, across all three paths (f782b071).
- **API docs**: `chat.py:448`/`:554` docstrings now document all four extra fields and the new opt-in flag (9429a7b9).
- **PR description**: updated — no more `retrieved_sources`, reflects the current field set.
- **Nits**: kept the on-wire duplication as you suggested. Added the missing `complete()` no-citation test. Left the truncated-stream log wording (`sources=len(filtered)` reading as "cited N" when nothing was cited) as-is for now — happy to adjust if it's still bothering anyone, but it's just a log line.

**From the design note — went with option 1's non-breaking version:**

Kept `sources` untouched for backward compatibility (existing clients, e.g. Twake). Added `presented_sources` (everything shown to the LLM, pre-filter) and `cited_sources` (strictly what was cited, empty — not falling back — whenever no tag was found). Chainlit's `app_front.py` now reads `cited_sources`, falling back to `presented_sources` in the UI when it's empty (4c71a7e5) — so the naming ambiguity you flagged is resolved for new consumers without breaking old ones.

On the correctness caution: confirmed already safe — `filter_sources_by_citations` only ever indexes into the presented list; `all_retrieved_sources`/`all_sources` is a fully separate field nothing indexes into, in both the pre-existing and current code.

On the budget/cap mismatch: agreed it's the more fundamental issue, but leaving that out of scope here — happy to open a follow-up issue if you'd like to track it separately.

On gating the full dump: done — `all_retrieved_sources` is now opt-in via `metadata.include_all_retrieved_sources` (default off, and the underlying `prepare_sources` call on the full retrieval set is skipped entirely when not requested, not just excluded from the payload). Turned up a side issue while wiring this for `/completions`: `OpenAICompletionRequest` had no `metadata` field at all, so no metadata flag — including the pre-existing `spoken_style_answer` — could ever reach that endpoint. Added it (87c07b07).

Ahmath-Gadji:
Filed the context-budget/reranker.top_k mismatch as a separate issue: #851 — keeping it out of this PR's scope.

hedhoud (approved):
The follow-up changes address the earlier source-truncation and citation-state concerns across chat, streaming, completion, and map-reduce. The focused tests and CI are green, and the backward-compatible fallback remains intact.

PR #856, opened 2026-08-25
docs: update architecture diagram and correct README format list

andyne13:
## Why

The architecture diagram is replaced with the new animated version. While updating the reference, the supported-format list was checked against `conf/config.yaml` and had drifted.

## Diagram

- **README** carries the animated GIF. It plays **once** instead of looping: a 7.4s animation looping indefinitely with no pause control conflicts with WCAG 2.2.2, which is worth avoiding given public-sector deployments. The loop block was removed at byte level, so all 148 frames, timings and transparency are unchanged.
- **Documentation site** keeps a static image. Its `<Image>` component processes assets through Sharp, which strips GIF animation — it would have rendered a frozen first frame with no error. The static file is the composited final frame, so both surfaces show the same diagram.
- Declared height moves from `350` to `420` to match the new 1012×708 ratio (the old value was already mismatched against the previous 1920×903 image).
- The orphaned root `RAG_architecture.png` is removed. The only remaining mention is in `tests/resources/test_file.txt`, a frozen snapshot of an older README used as text-extraction test input — not a live reference, and deliberately left untouched.

## Format list corrections

Checked against `conf/config.yaml`:

- **`eml` added** — supported via `EmlLoader`, previously absent from the list while the diagram advertises e-mail.
- **`flac` added** — supported, previously absent.
- **`mp4` and `flv` moved** out of *Audio Files* into their own *Video Files* row. Both are classified `DocumentType.VIDEO` and routed to the transcription backend, which extracts the audio track.
- **Captions wording corrected** — images are replaced by generated descriptions only when image captioning is enabled, not unconditionally.

## Also

The documentation homepage meta description was still the Starlight scaffold placeholder (*"This is a page in my Starlight-powered site"*), which is what search engines and link previews were showing.

## Out of scope

Positioning is deliberately untouched. The "Experimentation Playground" heading and the "lightweight… focused on experimentation" paragraph remain in both files, and are proposed separately so they can be reviewed alongside the website messaging.

## Summary by CodeRabbit

* **Documentation**
  * Updated the architecture visual with an animated image and improved its display size.
  * Expanded supported file formats to include email, audio, and video.
  * Corrected file-extension details and refined email and image descriptions.
  * Clarified Markdown conversion behavior, including AI-generated descriptions for parser-extractable images.
  * Identified OpenRag as a sovereign, open source Retrieval-Augmented Generation framework built by Linagora.


andyne13:
Thanks @coderabbitai — all four addressed in 44fe893f.

**E-mail description (README:34)** — you were right, and this was the one that mattered. I had written "thread relationships preserved", which the parser does not do. `EmlParser` extracts subject, from, to, date and message-id, but there is no `In-Reply-To` or `References` handling and it never populates `relationship_id`. Threading is caller-supplied, as the document-relationships documentation states.

Replaced with what the parser actually does — and it turns out to be a better line, since attachment dispatch through each format's own pipeline is the more interesting capability:

> **E-mail**: `eml` - Message body and headers, with attachments parsed through their own format's pipeline

**Meta description (index.mdx:3)** — agreed, the value described the concept rather than the product. Adopted with one change: "framework" rather than "tool", to match how the README and website describe OpenRag.

**VLM hyphenation (README:37)** — applied.

**Captioning scope (README:39)** — narrowed to "images the parser can extract", since replacement depends on what each format's parser surfaces.

Ahmath-Gadji:
Reviewed — verified the format-list corrections against `conf/config.yaml` (eml/flac/mp4/flv all check out) and the GIF loop-removal claim at byte level (no `NETSCAPE2.0` extension block present). Two things worth flagging before merge:

1. **Format-list drift wasn't fully fixed**: [`docs/content/docs/documentation/features_in_details.md`](https://github.com/linagora/openrag/blob/develop/docs/content/docs/documentation/features_in_details.md#L8-L13) has the same stale content this PR corrects elsewhere — `mp4`/`flv` still lumped into "Audio Files" instead of a Video row, no `eml` entry, and the same "images replaced by AI-generated captions... consistent processing" wording implying unconditional captioning (already reworded in README.md, not here). After this merges, that page will disagree with README.md/index.mdx on supported formats.

2. **Minor/inert**: [`tests/resources/test_file.txt:3`](https://github.com/linagora/openrag/blob/develop/tests/resources/test_file.txt#L3) still references `./RAG_architecture.png` (a static fixture — old README snapshot used for text-extraction tests, not a live link, doesn't break CI). Just means "nothing referenced it after the swap" isn't literally true.

Nothing blocking otherwise — CI is green and CodeRabbit's earlier comments were already addressed in the second commit.

andyne13:
Thanks @Ahmath-Gadji — both findings were correct, and both were mine.

**1. Format-list drift — fixed in 11f1c6fc.** You're right that merging as-is would have left the documentation site contradicting the README. `features_in_details.md` carried a byte-identical copy of the stale block, so it now gets the same corrections: `eml` added, `mp4`/`flv` moved to a Video row, `flac` added, and the captioning sentence reworded.

Worth noting the page already disagreed with *itself* before this: the "Audio & Video File Processing" section immediately below listed `flac` correctly, while the format list above omitted it. Both are now consistent.

Docs site rebuilt after the change — 28 pages, clean.

**2. Test fixture — leaving it, correcting my claim instead.** You're right that "nothing referenced it after the swap" wasn't literally true. `tests/resources/test_file.txt` is a snapshot of the old README used as text-extraction input, so its content is deliberately frozen test data rather than a live reference — editing it would change the fixture for no benefit. I've corrected the PR description rather than the file, since the inaccuracy was in my wording, not in the code.

One knock-on for whoever picks up the positioning work: that fixture also contains the old `Experimentation Playground` heading, so it will keep diverging from the real README as the copy changes. Harmless for the tests either way, just worth knowing it isn't a mirror.

Ahmath-Gadji:
Confirmed — both addressed. \`11f1c6f\` brings \`features_in_details.md\` in line with README.md/index.mdx and \`conf/config.yaml\`, and fixes the pre-existing self-contradiction (flac was already correct in the "Audio & Video" section below). Agreed the test fixture is frozen test data, not a live doc — fine to leave as-is with the PR description corrected instead. CI is green. LGTM from my side.

Ahmath-Gadji (approved):

PR #857, opened 2026-08-25
chore(deps): refresh npm lockfiles

andyne13:
## What

Refreshes both npm lockfiles to the newest versions permitted by the semver ranges already declared in `package.json`.

**No manifest changes.** Nothing moves outside a range the project had already accepted, so this is a lockfile-only diff.

## Verification

Run after the refresh, in this branch:

| Check | Result |
|---|---|
| Documentation site build | 28 pages, complete |
| Admin UI build (includes `tsc -b`) | passes |
| Admin UI test suite | 212 tests across 27 files, all passing |

## Deliberately out of scope

The Astro toolchain is left alone. Moving `astro` and `@astrojs/starlight` forward — along with the `esbuild` and `sharp` versions pinned beneath them — means major version bumps with genuine breaking-change risk for the documentation site. That deserves its own PR with its own testing, rather than riding along with a routine lockfile refresh.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,30 +33,30 @@
       }
     },
     "node_modules/@astrojs/compiler": {
-      "version": "2.13.0",
-      "resolved": "https://registry.npmjs.org/@astrojs/compiler/-/compiler-2.13.0.tgz",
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji (approved):
Reviewed: pure lockfile refresh (no package.json changes), all `resolved` URLs point to registry.npmjs.org, no lockfileVersion change, no suspicious install scripts beyond esbuild's expected one, CI green.

PR #858, opened 2026-08-25
docs(ci): correct the astro deploy trigger comment

andyne13:
## Problem

`.github/workflows/astro.yml` describes its own trigger incorrectly:

```yaml
on:
  # Runs on pushes targeting the default branch
  push:
    branches: ["main"]
```

The default branch is `develop`, not `main` — most likely a leftover from before the branch switch. Anyone reading the comment would reasonably expect the documentation site to publish from `develop`, when in fact it only updates when a release reaches `main`.

This came up while checking when a documentation change would actually become visible, and the comment sent me the wrong way.

## Change

**Comment only — the trigger is untouched.** Publishing the documentation site from the released branch looks deliberate, so this corrects the description rather than the behaviour.

The `workflow_dispatch` comment also now mentions that it can publish from a selected branch, which is the practical way to refresh the site ahead of a release.

## Verification

YAML still parses, and the resolved trigger is unchanged (`push.branches == ["main"]`).

## Summary by CodeRabbit

* **Documentation**
  * Clarified deployment workflow comments, including branch-based publishing behavior.
  * Documented how to manually publish documentation from a selected branch ahead of a release.


hedhoud (approved):
Looks good. The updated comment now matches the actual deployment triggers and makes the manual branch behavior clear.

PR #859, opened 2026-08-25
fix(ui): regenerate lockfile with npm 10 so the Docker build's npm ci works

Ahmath-Gadji:
## What

Regenerates `ui/package-lock.json` with **npm 10** so the admin-ui Docker build's `npm ci` works again.

`docker compose up --build` has been failing since #857 merged:

```
=> ERROR [admin-ui build 4/7] RUN npm ci
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json or npm-shrinkwrap.json are in sync.
npm error Invalid: lock file's ajv@6.15.0 does not satisfy ajv@8.20.0
npm error Invalid: lock file's ajv-formats@3.0.1 does not satisfy ajv-formats@2.1.1
npm error Invalid: lock file's json-schema-traverse@0.4.1 does not satisfy json-schema-traverse@1.0.0
```

## Root cause

Not a bad dependency version — an **npm-version tree-layout mismatch**.

`infra/docker/ui.Dockerfile` builds on `node:22-alpine`, which ships **npm 10.9.8**. The lockfile refreshed in #857 was generated with **npm 11/12**, and the two resolve the `ajv` cluster differently:

| hoisted at root | npm 11/12 (committed in #857) | npm 10.9.8 (what Docker runs) |
|---|---|---|
| `ajv` | 6.15.0 | 8.20.0 |
| `ajv-formats` | 3.0.1 | 2.1.1 |
| `json-schema-traverse` | 0.4.1 | 1.0.0 |

The trigger is that `ajv` is required at two incompatible majors at once — `eslint` wants `^6.14.0`, `@modelcontextprotocol/sdk` (via `shadcn`) wants `^8.17.1` — *while also* being an **optional peer dep** of `@hookform/resolvers` (`ajv ^8.12.0`, `ajv-formats ^2.1.1`). npm 10 hoists the v8 pair to the root to satisfy those optional peers and nests `ajv@6.15.0` under `eslint`; npm 11/12 do the opposite. `npm ci` builds its own ideal tree, compares it to the lockfile, and refuses the mismatch.

## Fix

Regenerated under npm 10.9.8, in an isolated copy so the existing `ui/node_modules` couldn't influence resolution:

```bash
npm install --package-lock-only --ignore-scripts   # npm 10.9.8
```

**This is a tree-layout change only.** I diffed every name→version pair between the old and new lockfiles:

- **Zero dependency version changes.** The `"version"` lines moving in the diff are the same packages relocating paths — e.g. `ajv@6.15.0` moving from the root down to `node_modules/eslint/node_modules/ajv`.
- The only additions are 5 optional wasm sub-deps of `@tailwindcss/oxide-wasm32-wasi` (`@emnapi/*`, `@napi-rs/wasm-runtime`, `@tybys/wasm-util`) that npm 10 records nested and npm 11/12 dedupe away.

No `package.json` change, no Dockerfile change, no upgrades or downgrades.

## Verification

| Check | Result |
|---|---|
| `npm ci` under npm **10.9.8** (node:22-alpine) | exit 0 |
| `npm ci` under npm **11** | exit 0 |
| `npm ci` under npm **12** | exit 0 |
| `npm@12 install --package-lock-only` re-run | lockfile **byte-identical** — newer-npm contributors won't churn it back |
| `docker build -f infra/docker/ui.Dockerfile .` | passes end to end (`npm ci` → `tsc -b` → `vite build` → nginx stage) |

## Why CI didn't catch this

Nothing runs `npm ci` on `ui/` at PR time. The admin-ui image is only built by `build.yml` and `build_rc.yml`, both triggered on **tags**; `astro.yml` is the only workflow that touches npm and it builds the docs site from the root lockfile on pushes to `main`. So #857 merged green and this would next have surfaced during a release build.

Worth adding a PR-time job that runs `npm ci` (and the 212-test `ui/` suite) so a desynced lockfile fails fast — happy to do that here or in a follow-up, whichever you prefer.

## Other lockfiles — checked, both fine

- **root `package-lock.json`** (docs site): `npm ci` passes on npm 10.8.2 / 10.9.8 / 11 / 12, and the full `astro.yml` job reproduces clean (`npm ci` + `astro build` → 28 pages). It has no `ajv` at all, and a `semver` sweep of every peer range against the resolved tree found no real violations.
- **`extern/chainlit-datalayer/package-lock.json`**: passes; #857 never touched `extern/` anyway.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1507,6 +1507,23 @@
         "url": "https://opencollective.com/eslint"
       }
     },
+    "node_modules/@eslint/eslintrc/node_modules/ajv": {
+      "version": "6.15.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.15.0.tgz",
+      "integrity": "sha512-fgFx7Hfoq60ytK2c7DhnF8jIvzYgOMxfugjLOSMHjLIPgenqa7S7oaagATUq99mV6IYvN2tRmC0wnTYX6iPbMw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "fast-deep-equal": "^3.1.1",
+        "fast-json-stable-stringify": "^2.0.0",
+        "json-schema-traverse": "^0.4.1",
+        "uri-js": "^4.2.2"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/epoberezkin"
+      }
+    },
     "node_modules/@eslint/eslintrc/node_modules/globals": {
       "version": "14.0.0",
       "resolved": "https://registry.npmjs.org/globals/-/globals-14.0.0.tgz",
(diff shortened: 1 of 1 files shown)


hedhoud (approved):
LGTM!

Issue #579, opened 2026-06-25
Chunker cuts mid-sentence on PDF line-wraps (recursive_splitter)


andyne13:
### Summary
The `recursive_splitter` chunker can end a chunk **in the middle of a sentence**, splitting on a visual line-wrap newline rather than on a paragraph or sentence boundary.

### Symptom
Indexing a text-heavy PDF (most visible with the `pymupdf` backend) produces adjacent chunks that break mid-sentence, e.g. (synthetic illustration):

> chunk N  (…) "… the committee reviewed the report and was assisted by the"
> chunk N+1 "… assisted by the **external auditors** before approving the budget."

chunk N ends at `…by the` — not a sentence end. The overlap region then repeats it. The break point is a single `\n`, not a `.`/`?`/`!`.

### Root cause
1. Backends that preserve the PDF's visual layout (notably `pymupdf4llm`) keep **line-wrap newlines inside paragraphs** — e.g. `…by the\nexternal auditors…`. (marker/docling reflow paragraphs, so they mostly avoid it.)
2. The splitter is configured `separators=["\n", r"(?<=[\.\?\!])"]` with a **single `\n` tried first**, so it breaks on every line-wrap before it ever considers sentence boundaries.
3. `sanitize_text` collapses repeated spaces and caps consecutive newlines at 2, but leaves single mid-paragraph `\n` intact.

The class docstring states *"Splits on paragraph boundaries first, then sentence terminators, then smaller separators"* — the implemented separators don't match that intent (no `\n\n`, and `\n` is first).

### Fix (first iteration)
- Reorder separators to paragraph → sentence → line → word: `["\n\n", r"(?<=[\.\?\!])", "\n", " "]`, and join text elements as paragraphs (`\n\n`).
- **Reflow** intra-paragraph line-wrap newlines into spaces before splitting, while preserving markdown structure (blank-line paragraph breaks, headings, list items, tables, code fences) so stored chunk text reads as clean prose.
- Unit test: a paragraph with mid-sentence `\n` must not produce a chunk that ends mid-sentence.

### Notes
Chunk quality is an iterative effort; this is the first item. Work will land on a `feat/` branch off `refactor/hexagonal`. Area: backend.

PR #861, opened 2026-08-27
feat(chunking): structure-aware structured_section strategy

andyne13:
Adds `structured_section`, an opt-in chunking strategy that cuts on document structure instead of character separators, and wires a per-partition safety bound derived from the embedder's context window.

## Why

The default `recursive_splitter` is page-aware but not structure-aware. On a 904-page French legal code it produced three defects: `chunk_overlap_rate: 0.2` re-embedded whole short Articles 2–3× as pure retrieval noise; articles were torn mid-unit with headings orphaned from their bodies; and `[PAGE_N]` markers were embedded inline as content.

Marker emits legal structure (`Titre`, `Article L110-1`) as plain-text lines rather than `##` headings, so boundary detection is regex-on-line-content, with markdown `#` recognised too. Patterns are config-driven so this generalises beyond French legal documents.

## Results

Measured on 91 documents parsed three ways (marker / docling / pymupdf), against `recursive_splitter`:

| metric | recursive | structured_section |
|---|---|---|
| cross-chunk duplication | 5.5% / 6.1% / 8.3% | **1.9% / 2.5% / 2.4%** |
| orphan headings ([see note](#a-correction-to-the-orphan-figure)) | 1129 / 1037 / 527 | **81 / — / —** |
| inline `[PAGE_N]` | 3751 / 3731 / 3662 | **5 / 2 / 0** |
| chunks under the size floor | 1.9% / 3.6% / 0.9% | **0.2% / 0.2% / 1.0%** |
| chunks opening mid-sentence | 4.38% | **1.68%** |

On the legal code, 65% of cuts land exactly on `Article`/`Titre` markers.

Cost: ~1.3–1.45× the wall time of `recursive_splitter` (median document 32ms → 39ms; the 904-page code 1.42s → 1.74s). It makes *fewer* token-counter calls; the extra time is regex work on line content. Negligible beside GPU parsing and per-chunk contextualization.

## Design notes

- **Atomic units are exempt from the packing ceiling.** `max_tokens` budgets prose packing; a figure caption or table is not split by it, because a fragment of one is not a smaller version of it. Enforcing the ceiling fragmented 55 captions for no embedding benefit — the largest unsplit caption was 1,211 tokens.
- **`hard_max_tokens` is a separate safety net**, derived from the context window of the embedder the partition actually indexes with (half of it, leaving headroom for the self-containment header and contextualization's `[CONTEXT]` block). Content past the window is silently truncated by vLLM, which is real data loss; `chunk_size` says nothing about that limit. Embedder endpoints declare their window via `extra["max_model_len"]`, the key `_build_embedder_factory` already reads — no migration needed.
- **Small neighbours merge across atomic types when they share a page.** A slide is page-atomic: its label, figure and table belong together. Restricting the permissive route to a shared page prevents chaining across slides.
- **Break-less paragraphs descend** sentence → `;` → `:` → em-dash → `,` before word-wrapping, so the only splitter that can land mid-sentence fires solely on punctuation-free text.
- **A warning is logged** when a chunk exceeds `truncate_prompt_tokens`, making the one case the chunker deliberately does not prevent — an indivisible table row larger than the window — visible instead of silent.

The strategy stays pure-domain: no Ray, no LLM, no config object; the token counter and the embedder window are injected.

## Notes for review

- Opt-in per preset; `recursive_splitter` remains the default. No behaviour change for existing partitions.
- Beyond the chunker, this touches `core/chunking/factory.py`, `core/config/{chunking,model_endpoints}.py` and `services/workers/{indexer_pool,pipeline_builder}.py` for the window plumbing. Happy to split that into a follow-up PR if preferred.
- Remaining known gaps, deliberately not addressed here: degenerate/looping VLM captions are chunked faithfully rather than detected (belongs in the captioning step), and `<span id="page-N">` parser anchors survive in ~7% of chunks (same class as `[PAGE_N]`, worth the same treatment).

## Summary by CodeRabbit

- **New Features**
  - Added structured-section chunking for Markdown and configurable heading- or article-style documents.
  - Preserves hierarchy, tables, captions, page boundaries, and optional heading paths.
  - Adapts chunk token limits to the selected embedder’s context window.

- **Improvements**
  - Splits oversized content safely while enforcing token limits.
  - Warnings identify chunks exceeding effective embedding limits.
  - Configuration validates positive token bounds and regular-expression patterns.

- **Documentation**
  - Clarified how endpoint-specific embedder context limits affect chunking.

### A correction to the orphan figure

An earlier revision of this description reported `5 / 4 / 1` orphan headings. That was **understated**, and the cause is worth stating plainly.

`chunk_eval.py` detects an orphan by matching a chunk's last line against `^#{1,6}\s+\S` or a French structural keyword. Until the latest commit, `structured_section` wrote heading text into the body *flattened* — `_clean_heading` had stripped the `#` markers — so a heading sitting at a chunk's end was **invisible to the metric**. It counted 4 where the real number was far higher.

Now that the body copy keeps its markdown (the reviewer's finding), the metric can see them and reports **81** on the marker corpus. That is the metric gaining sight, not a regression.

Counted independently of the regex — a chunk whose last line matches one of that document's own headings — the marker corpus has:

| | flattened body | markdown kept (current) |
|---|---|---|
| chunks ending on a heading | 87 | **66** |

So the behaviour predates this change and improves slightly. Of the 81 the metric now reports: 6 are heading-only chunks, 75 have a body and trail off on a heading — `_merge_small` folding a short heading unit backward into the preceding chunk. That is a real remaining defect, tracked rather than fixed here.

The other columns in the table above are unaffected: duplication, `[PAGE_N]` noise and under-min are measured on chunk text and never depended on the markdown being present.

### DIFF
--- a/openrag/core/chunking/structured_section.py
+++ b/openrag/core/chunking/structured_section.py
@@ -733,7 +733,15 @@ class StructuredSectionChunker(BaseChunker):
         ``_enforce_ceiling``."""
         emax = self._effective_max(filename, unit.heading_path, unit.pages)
         if unit.chunk_type is ChunkType.TABLE and unit.tokens > emax:
-            element = MDElement(type="table", content=unit.text, page_number=unit.start_page)
+            # chunk_table reads the *leading* lines as the column header and
+            # replays them on every piece. Since a heading now travels into the
+            # atomic unit, those leading lines can be the heading — which was
+            # then replayed instead of the column header, leaving every
+            # continuation piece a wall of unlabelled numbers. Split the table
+            # alone and re-attach the heading to the first piece, which is the
+            # one it introduces.
+            lead, table_text = _split_leading_headings(unit.text)
+            element = MDElement(type="table", content=table_text, page_number=unit.start_page)
             subs = chunk_table(element, chunk_size=emax, length_function=self.length_function)
             pieces = [
                 _Unit(
@@ -746,6 +754,9 @@ class StructuredSectionChunker(BaseChunker):
                 )
                 for s in subs
             ]
+            if lead and pieces:
+                pieces[0].text = f"{lead}\n\n{pieces[0].text}"
+                pieces[0].tokens = self.length_function(pieces[0].text)
         else:
             pieces = [unit]
         return self._enforce_ceiling(pieces, filename)
@@ -1044,6 +1055,18 @@ def _absorb(cur: _Unit, unit: _Unit, *, path: list[str] | None = None) -> None:
     cur.atomic = cur.atomic or unit.atomic
     if path is not None:
         cur.heading_path = list(path)
+
+
+def _split_leading_headings(text: str) -> tuple[str, str]:
+    """Separate a block's leading heading lines from its content."""
+    lines = text.splitlines()
+    cut = 0
+    lead: list[str] = []
+    while cut < len(lines) and (not lines[cut].strip() or _HEADING_LINE_RE.match(lines[cut])):
+        if lines[cut].strip():
+            lead.append(lines[cut].strip())
+        cut += 1
+    return "\n\n".join(lead), "\n".join(lines[cut:]).strip()
 
 
 def _is_headings_only(unit: _Unit) -> bool:
--- a/openrag/core/chunking/structured_section.py
+++ b/openrag/core/chunking/structured_section.py
@@ -226,7 +226,7 @@ class StructuredSectionChunker(BaseChunker):
         heading_keywords: tuple[str, ...] | list[str] | None = None,
         leaf_patterns: tuple[str, ...] | list[str] | None = None,
         prepend_heading_path: bool = True,
-        layout: str = "auto",
+        layout: str = "flowing",
         **kwargs: Any,
     ) -> None:
         # Overlap is forced to 0: leaves are atomic, so replaying a tail would
--- a/openrag/core/chunking/structured_section.py
+++ b/openrag/core/chunking/structured_section.py
@@ -1,0 +1,569 @@
+"""Structure-aware chunking strategy (``structured_section``).
+
+Where ``recursive_splitter`` cuts on character separators (paragraph → sentence
+→ line → word) with a fixed overlap, ``structured_section`` cuts on the
+document's own structure: it detects headings and leaf units (e.g. legal
+``Article`` markers) by **regex on line content**, keeps each leaf atomic,
+greedily packs consecutive leaves up to a token target, and prepends each chunk
+with its heading path so the chunk is self-describing for retrieval.
+
+Why regex-on-lines rather than markdown ``#`` parsing: the default PDF parser
+(Marker) emits legal/structural headings (``Titre``, ``Chapitre``,
+``Article L110-1`` …) as **plain-text lines, not** ``##`` headings, so a
+markdown-``#``-only detector would miss every boundary. Patterns are
+config-driven (``heading_patterns`` / ``leaf_patterns``) so the strategy
+generalizes beyond French legal codes; markdown ``#`` headings are recognized
+too.
+
+Design (in priority order), matching the redesign brief:
+1. Boundary detection on line content: markdown ``#``; keyword headings
+   (Livre/Titre/Chapitre/Section/…); leaf markers (``Article L###`` …).
+2. Atomicity: never split a leaf/section unless it alone exceeds ``max_tokens``;
+   then fall back to paragraph → sentence.
+3. Packing: greedily pack consecutive short leaves under the same heading up to
+   ``target_tokens`` (a chunk = one or a few whole articles under one heading).
+4. Self-containment: the heading path is prepended as a header.
+5. Overlap = 0 (leaves are atomic; overlap would just duplicate whole articles).
+6. ``[PAGE_N]`` markers become ``page_number`` / ``page_range`` metadata, never
+   inline text.
+
+Pure domain logic — no LLM, no Ray. The token counter is injected. Tables and
+image-description blocks are treated as atomic units via the shared
+``markdown_utils`` primitives, exactly like ``recursive.py``.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from typing import Any
+
+from core.chunking.markdown_utils import (
+    PAGE_RE,
+    MDElement,
+    chunk_table,
+    split_md_elements,
+)
+from core.chunking.recursive import (
+    _IMAGE_PLACEHOLDER_MARKER,
+    _INLINE_ELEMENT_TOKEN_THRESHOLD,
+    BaseChunker,
+    dewrap_paragraphs,
+)
+from core.chunking.registry import chunking_registry
+from core.models.chunk import Chunk, ChunkType
+from core.models.document import ProcessedDocument
+from core.utils.text import sanitize_text
+
+# Markdown ATX heading: ``## Title``. Level = number of ``#``.
+_MD_HEADING_RE = re.compile(r"^(#{1,6})\s+(\S.*?)\s*#*$")
+
+# A leading markdown heading prefix (``#### ``) and inline emphasis markers
+# (``**bold**``, ``_em_``, `` `code` ``) — stripped when testing a line for a
+# leaf pattern and when cleaning heading text for the breadcrumb.
+_HEADING_PREFIX_RE = re.compile(r"^\s*#{1,6}\s*")
+_EMPHASIS_RE = re.compile(r"[*_`]+")
+
+
+def _clean_heading(text: str) -> str:
+    """Strip markdown emphasis/heading noise from a heading's display text."""
+    return _EMPHASIS_RE.sub("", _HEADING_PREFIX_RE.sub("", text)).strip()
+
+
+# Keyword headings emitted as plain text by structure-unaware parsers. Ordered
+# outermost → innermost; the index gives the nesting level so a heading stack
+# can pop correctly. Config can replace this list.
+_DEFAULT_HEADING_KEYWORDS: tuple[str, ...] = (
+    "livre",
+    "titre",
+    "chapitre",
+    "section",
+    "sous-section",
+    "paragraphe",
+)
+
+# Leaf markers — a unit that is atomic and does NOT open a nesting level (an
+# article sits *inside* the current heading, it doesn't contain sub-articles).
+_DEFAULT_LEAF_PATTERNS: tuple[str, ...] = (r"^\s*Art(?:icle|\.)\s*[LRDOlrdo]?\.?\s*\d",)
+
+# A markdown heading counts as its own nesting level (by ``#`` depth); keyword
+# headings get a level *below* every markdown heading so a plain-text ``Titre``
+# nested under a real ``# Book`` heading still stacks sensibly.
+_MD_LEVEL_BASE = 0
+_KEYWORD_LEVEL_BASE = 100
+
+
+@dataclass
+class _Unit:
+    """A structural unit: a leaf (article/heading-intro) or an atomic block."""
+
+    heading_path: list[str]
+    text: str
+    tokens: int
+    pages: set[int] = field(default_factory=set)
+    chunk_type: ChunkType = ChunkType.TEXT
+    atomic: bool = False  # table / image — never merged into prose, never split by us
+
+    @property
+    def start_page(self) -> int | None:
+        return min(self.pages) if self.pages else None
+
+
+@chunking_registry.register("structured_section")
+class StructuredSectionChunker(BaseChunker):
+    """Structure-aware chunker keyed on headings + leaf markers."""
+
+    def __init__(
+        self,
+        chunk_size: int = 512,
+        chunk_overlap_rate: float = 0.0,
+        length_function: Callable[[str], int] | None = None,
+        *,
+        min_tokens: int | None = None,
+        max_tokens: int | None = None,
+        inline_threshold: int | None = None,
+        heading_keywords: tuple[str, ...] | list[str] | None = None,
+        leaf_patterns: tuple[str, ...] | list[str] | None = None,
+        prepend_heading_path: bool = True,
+        **kwargs: Any,
+    ) -> None:
+        # Overlap is forced to 0: leaves are atomic, so replaying a tail would
+        # duplicate whole articles across chunks (the defect this replaces).
+        super().__init__(
+            chunk_size=chunk_size,
+            chunk_overlap_rate=0.0,
+            length_function=length_function,
+            **kwargs,
+        )
+        # Sizing follows the current evidence sweet spot: ~512 target, with a
+        # min/max band (≈128 / ≈768 at target 512) so tiny sibling sections pack
+        # up and only genuinely oversize leaves get split.
+        self.target_tokens = chunk_size
+        self.max_tokens = max_tokens or int(chunk_size * 1.5)
+        self.min_tokens = min_tokens if min_tokens is not None else max(1, chunk_size // 4)
+        # Tables / image captions at or below this many tokens are inlined with
+        # surrounding prose instead of becoming standalone chunks (matches the
+        # recursive chunker's inline threshold).
+        self._inline_threshold = inline_threshold if inline_threshold is not None else _INLINE_ELEMENT_TOKEN_THRESHOLD
+        self.prepend_heading_path = prepend_heading_path
+
+        keywords = tuple(k.lower() for k in (heading_keywords or _DEFAULT_HEADING_KEYWORDS))
+        self._keyword_level = {kw: _KEYWORD_LEVEL_BASE + i for i, kw in enumerate(keywords)}
+        self._keyword_re = re.compile(
+            r"^\s*(" + "|".join(re.escape(k) for k in keywords) + r")\b",
+            re.IGNORECASE,
+        )
+        self._leaf_res = [re.compile(p, re.IGNORECASE) for p in (leaf_patterns or _DEFAULT_LEAF_PATTERNS)]
+
+    # ------------------------------------------------------------------
+    # ChunkingStrategy contract
+    # ------------------------------------------------------------------
+    def chunk(self, document: ProcessedDocument, partition: str = "default") -> list[Chunk]:
+        content = self._content_from(document)
+        if not content.strip():
+            return []
+
+        metadata = self._chunk_metadata_base(document, partition)
+        filename = str(metadata.get("filename") or metadata.get("source") or "")
+
+        units = self._build_units(content)
+        candidates = self._pack(units, filename)
+        candidates = self._merge_small(candidates, filename)
+
+        chunks: list[Chunk] = []
+        for idx, unit in enumerate(candidates):
+            header = self._build_header(filename, unit.heading_path, unit.pages)
+            body = unit.text.strip()
+            text = f"{header}\n\n{body}" if (header and self.prepend_heading_path) else body
+            chunks.append(
+                Chunk(
+                    document_id=metadata.get("file_id", ""),
+                    text=text,
+                    chunk_index=idx,
+                    chunk_type=unit.chunk_type,
+                    partition=partition,
+                    page_number=unit.start_page,
+                    token_count=self.length_function(text),
+                    header=header or None,
+                    content=body,
+                    metadata={
+                        **{k: v for k, v in metadata.items() if k not in ("file_id", "partition")},
+                        "hierarchy_path": list(unit.heading_path),
+                        "section_title": unit.heading_path[-1] if unit.heading_path else "",
+                        "page_range": _page_range(unit.pages),
+                    },
+                )
+            )
+        return chunks
+
+    # ------------------------------------------------------------------
+    # Step 1 — parse the document into ordered structural units
+    # ------------------------------------------------------------------
+    def _build_units(self, content: str) -> list[_Unit]:
+        """Walk the document in order, emitting one ``_Unit`` per leaf/atomic block.
+
+        A heading line updates the heading stack but is emitted only through the
+        chunk header — never as its own body — so a heading can never be
+        stranded as an orphan chunk. Any prose that follows a heading before the
+        next marker becomes that section's intro unit.
+        """
+        units: list[_Unit] = []
+        stack: list[tuple[int, str]] = []  # (level, text)
+        page = 1
+
+        # Current open leaf being accumulated (marker line + its body lines).
+        buf_path: list[str] = []
+        buf_lines: list[str] = []
+        buf_pages: set[int] = set()
+
+        def flush() -> None:
+            nonlocal buf_lines, buf_pages
+            text = "\n".join(buf_lines).strip()
+            if text:
+                units.append(
+                    _Unit(
+                        heading_path=list(buf_path),
+                        text=text,
+                        tokens=self.length_function(text),
+                        pages=set(buf_pages) or {page},
+                        chunk_type=ChunkType.TEXT,
+                    )
+                )
+            buf_lines = []
+            buf_pages = set()
+
+        for element in split_md_elements(content):
+            if element.type in ("table", "image"):
+                # Non-informative images (the captioner's "[Image Placeholder]")
+                # are dropped, matching the recursive chunker — otherwise every
+                # blank logo becomes a chunk.
+                if element.type == "image" and _IMAGE_PLACEHOLDER_MARKER in element.content.lower():
+                    continue
+                # Small tables / image captions inline with the surrounding prose
+                # (like recursive's _prepare_md_elements) so a slide's handful of
+                # logos don't each become their own tiny chunk; only large ones
+                # stay atomic (a big table splits via chunk_table downstream).
+                if self.length_function(element.content) <= self._inline_threshold:
+                    buf_lines.append(element.content.strip())
+                    buf_pages.add(element.page_number or page)
+                else:
+                    flush()
+                    units.append(self._atomic_unit(element, list(stack_path(stack)), page))
+                continue
+
+            for raw_line in dewrap_paragraphs(element.content).split("\n"):
+                page_match = PAGE_RE.fullmatch(raw_line.strip())
+                if page_match:
+                    # Content AFTER ``[PAGE_k]`` is page k+1 (see markdown_utils).
+                    page = int(page_match.group(1)) + 1
+                    continue
+
+                level_text = self._heading(raw_line)
+                if level_text is not None:
+                    level, text = level_text
+                    flush()
+                    _push_heading(stack, level, text)
+                    buf_path = list(stack_path(stack))
+                    continue
+
+                if self._is_leaf(raw_line):
+                    flush()
+                    buf_path = list(stack_path(stack))
+
+                if raw_line.strip():
+                    buf_lines.append(raw_line)
+                    buf_pages.add(page)
+            flush()
+
+        return units
+
+    def _atomic_unit(self, element: MDElement, heading_path: list[str], page: int) -> _Unit:
+        text = element.content.strip()
+        ctype = ChunkType.TABLE if element.type == "table" else ChunkType.IMAGE_CAPTION
+        return _Unit(
+            heading_path=heading_path,
+            text=text,
+            tokens=self.length_function(text),
+            pages={element.page_number or page},
+            chunk_type=ctype,
+            atomic=True,
+        )
+
+    def _heading(self, line: str) -> tuple[int, str] | None:
+        # A leaf marker (e.g. an Article) is *content*, even when the parser
+        # emitted it as a markdown heading ("#### **Article L110-5**"). It must
+        # NOT become a container heading, or its identifier would live only in
+        # the breadcrumb and be dropped when short sibling articles merge —
+        # losing the single most important retrieval key of a legal article.
+        if self._is_leaf(line):
+            return None
+        md = _MD_HEADING_RE.match(line)
+        if md:
+            return _MD_LEVEL_BASE + len(md.group(1)), _clean_heading(md.group(2))
+        kw = self._keyword_re.match(line)
+        if kw:
+            return self._keyword_level[kw.group(1).lower()], _clean_heading(line)
+        return None
+
+    def _is_leaf(self, line: str) -> bool:
+        # Strip a markdown heading prefix + emphasis so "#### **Article L110-5**"
+        # matches the same leaf patterns as a plain "Article L110-5" line.
+        core = _EMPHASIS_RE.sub("", _HEADING_PREFIX_RE.sub("", line)).strip()
+        return any(r.match(core) for r in self._leaf_res)
+
+    # ------------------------------------------------------------------
+    # Step 2 — pack consecutive units up to the token target
+    # ------------------------------------------------------------------
+    def _pack(self, units: list[_Unit], filename: str) -> list[_Unit]:
+        out: list[_Unit] = []
+        cur: _Unit | None = None
+
+        for unit in units:
+            if unit.atomic:
+                cur = self._flush_into(out, cur)
+                out.extend(self._emit_atomic(unit, filename))
+                continue
+            if unit.tokens > self._effective_max(filename, unit.heading_path, unit.pages):
+                cur = self._flush_into(out, cur)
+                out.extend(self._split_oversize(unit, filename))
+                continue
+            if cur is None:
+                cur = _copy_unit(unit)
+            elif cur.heading_path == unit.heading_path and cur.tokens + unit.tokens <= self._packing_budget(
+                filename, cur.heading_path, cur.pages | unit.pages
+            ):
+                _absorb(cur, unit)
+            else:
+                out.append(cur)
+                cur = _copy_unit(unit)
+        if cur is not None:
+            out.append(cur)
+        return out
+
+    def _header_reserve(self, filename: str, heading_path: list[str], pages: set[int]) -> int:
+        """Exact token cost of the prepended ``[Source | Page | Section]`` header
+        (plus its blank-line separator), reserved from the body budget so the
+        embedded ``chunk.text`` (header + body) stays within target/max.
+
+        Measured from the real header rather than estimated: a flat allowance
+        for the ``Source:``/``Page:`` fields silently under-reserves whenever
+        the filename is long, and the body then overshoots ``max`` by exactly
+        the amount the estimate missed.
+        """
+        if not self.prepend_heading_path:
+            return 0
+        header = self._build_header(filename, heading_path, pages)
+        return self.length_function(f"{header}\n\n") if header else 0
+
+    def _packing_budget(self, filename: str, heading_path: list[str], pages: set[int]) -> int:
+        """Body budget so a packed chunk lands near ``target`` *including* header."""
+        return max(self.min_tokens, self.target_tokens - self._header_reserve(filename, heading_path, pages))
+
+    def _effective_max(self, filename: str, heading_path: list[str], pages: set[int]) -> int:
+        """Body ceiling so ``chunk.text`` (body + header) never exceeds ``max``."""
+        return max(self.min_tokens, self.max_tokens - self._header_reserve(filename, heading_path, pages))
+
+    def _emit_atomic(self, unit: _Unit, filename: str) -> list[_Unit]:
+        """A table over the (header-adjusted) max is split by ``chunk_table``."""
+        emax = self._effective_max(filename, unit.heading_path, unit.pages)
+        if unit.chunk_type is ChunkType.TABLE and unit.tokens > emax:
+            element = MDElement(type="table", content=unit.text, page_number=unit.start_page)
+            subs = chunk_table(element, chunk_size=emax, length_function=self.length_function)
+            return [
+                _Unit(
+                    heading_path=list(unit.heading_path),
+                    text=s.content.strip(),
+                    tokens=self.length_function(s.content),
+                    pages={s.page_number} if s.page_number else set(unit.pages),
+                    chunk_type=ChunkType.TABLE,
+                    atomic=True,
+                )
+                for s in subs
+            ]
+        return [unit]
+
+    def _split_oversize(self, unit: _Unit, filename: str) -> list[_Unit]:
+        """Split a single over-``max`` leaf at paragraph, then sentence boundaries,
+        to header-adjusted target/max so each piece + header stays within max."""
+        path = unit.heading_path
+        pieces = _greedy_split(
+            unit.text,
+            self._packing_budget(filename, path, unit.pages),
+            self._effective_max(filename, path, unit.pages),
+            self.length_function,
+        )
+        return [
+            _Unit(
+                heading_path=list(unit.heading_path),
+                text=piece.strip(),
+                tokens=self.length_function(piece),
+                pages=set(unit.pages),
+                chunk_type=ChunkType.TEXT,
+            )
+            for piece in pieces
+        ]
+
+    @staticmethod
+    def _flush_into(out: list[_Unit], cur: _Unit | None) -> None:
+        if cur is not None:
+            out.append(cur)
+        return None
+
+    # ------------------------------------------------------------------
+    # Step 3 — merge under-min neighbours (combined <= max)
+    # ------------------------------------------------------------------
+    def _merge_small(self, units: list[_Unit], filename: str) -> list[_Unit]:
+        """Fold an under-``min`` neighbour into the previous chunk when they are
+        structurally compatible (same section, or one path a prefix of the
+        other) and the result stays under ``max``.
+
+        Merging *backward* (into the already-emitted previous chunk) keeps a
+        stray small leaf with its own section rather than pulling it forward
+        across a heading boundary into the next one — which would strand it
+        under a wrong/shallower heading path.
+
+        The ceiling is the *header-adjusted* max: ``_Unit.tokens`` counts the
+        body only, while the emitted ``chunk.text`` also carries the
+        ``[Source | Page | Section]`` header. Comparing bodies against the raw
+        ``max_tokens`` here let a merge fill the body to the ceiling and then
+        push past it once the header was prepended — worst on deep heading
+        paths, where the header is largest.
+        """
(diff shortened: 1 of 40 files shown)


openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,898 @@
...
+
+        Otherwise the structural route applies, which is stricter: tables stay
+        out (a table glued to unrelated prose loses the column headers that make
+        it readable) and the heading paths must be compatible.
+        """
+        if prev.tokens >= self.min_tokens and unit.tokens >= self.min_tokens:
+            return False
+        if prev.tokens + unit.tokens > self._effective_max(filename, merged_path, prev.pages | unit.pages):
+            return False
+        if prev.pages & unit.pages:
```
Ahmath-Gadji: **Major: the same-page route merges unrelated sections and erases the heading path.**

`_may_merge` returns `True` on a shared page *before* consulting `_may_absorb` / `_compatible`, and `_merge_small:635` then rewrites the merged unit's path to `_common_prefix(prev, unit)` regardless of which route allowed the merge. When two adjacent short units sit on the same page under different top-level headings, the common prefix is `[]` — so the section hierarchy is deleted from both the embedded header and `hierarchy_path`.

The docstring argues the page restriction "cannot chain content across slides or across distant sections". That holds for slide decks, where a page *is* a section. It does not hold for dense text, where several short units per page under different headings is the norm — i.e. exactly the 904-page legal code that motivates this PR.

Reproduced at production defaults (`chunk_size=512` → `min_tokens=128`, `cl100k_base` counter), one page, four articles under four different `Titre`s:

```
Titre Ier : Dispositions fiscales / Chapitre Ier : Taux
Article L110-1 — Le taux normal de la TVA est fixé à 20 %.
Titre II : Dispositions sanitaires / Chapitre Ier : Vaccination
Article L210-1 — La vaccination antidiphtérique est obligatoire ...
Titre III : Dispositions relatives au travail
Article L310-1 — La durée légale du travail effectif ...
Titre IV : Dispositions pénales
Article L410-1 — Est puni de deux ans d'emprisonnement ...
```

Output — **1 chunk**, `hierarchy_path: []`, `section_title: ''`, and no `Section:` field in the header at all:

```
[Source: code-general.pdf | Page: 1]

Article L110-1 Le taux normal de la taxe sur la valeur ajoutée est fixé à 20 %.

Article L210-1 La vaccination antidiphtérique est obligatoire pour les mineurs.

Article L310-1 La durée légale du travail effectif est fixée à trente-cinq heures.

Article L410-1 Est puni de deux ans d'emprisonnement le fait de contrevenir au présent code.
```

Two problems compound here:

1. A tax article, a public-health article, a labour article and a criminal article share one embedding vector.
2. Every `Titre`/`Chapitre` is erased — from the embedded text *and* from metadata. Only `text` feeds the BM25 function in the Milvus schema, and `hierarchy_path` lands in the dynamic field, so that heading text becomes unretrievable by either dense or sparse search. That is the orphan-heading / self-containment defect the PR sets out to fix, reintroduced one step later in the pipeline.

It chains, too: the gate is "at least one side under `min_tokens`", so a single chunk keeps absorbing successive small units from further sections while it stays under `_effective_max`.

Two orthogonal fixes, both small:

- Gate the permissive route on the merge not destroying context — e.g. require `_common_prefix(prev.heading_path, unit.heading_path)` to be non-empty whenever both sides have a path, so the same-page route still handles the slide case (identical paths, or one side path-less within a slide) without spanning `Titre` boundaries.
- Independently, never let `_absorb` shorten a path to `[]` when both inputs had one — losing the breadcrumb entirely is strictly worse than keeping the previous unit's.

`openrag/core/chunking/structured_section.py:663`, `openrag/core/chunking/structured_section.py:635`

openrag/services/workers/pipeline_builder.py
```diff
@@ -151,6 +182,10 @@ async def _timed(name: str, coro: Any) -> None:
                         per_chunk_timeout=self.timeouts.contextualize_per_chunk,
                     ),
                 )
+            # After contextualization, not before: the [CONTEXT] envelope is
+            # prepended here and is part of what gets embedded, so measuring
+            # earlier would miss exactly the chunks the envelope pushes over.
+            self._warn_on_embedder_overflow(row, embedder_window, getattr(chunker, "length_function", None))
```
Ahmath-Gadji, line 194: **Major: this re-tokenizes every chunk of every file, synchronously on the actor's event loop.**

`_warn_on_embedder_overflow` builds `sized = [(chunk, tokens(chunk)) for chunk in chunks]` over the whole list before filtering, so the full pass runs even in the overwhelmingly common case of zero offenders. It is a plain synchronous call inside `execute()`, so it blocks the loop for its whole duration.

That is the one thing the adjacent code goes out of its way to avoid — `stages/chunk.py:_chunk_with_timeout` wraps the chunker in `asyncio.to_thread` under the comment *"Run the synchronous chunker without blocking the event loop."* This pass re-does a subset of the same tokenization work, in the loop.

Measured on this branch, ~1000-token chunks of French prose:

| chunks | tiktoken `cl100k_base` | `ChatOpenAI.get_num_tokens` (the production default from `get_num_tokens()`) |
|---|---|---|
| 500 | 103 ms | 217 ms |
| 3000 | 609 ms | **1244 ms** |

A 904-page document like the one in the PR description lands in the thousands of chunks, so that is ~1.2 s of stalled loop per file — during which every other in-flight file's parse/caption/contextualize/embed/store continuation is frozen, not just this one's.

Blast radius is wider than the feature: the call is unconditional, so partitions on the default `recursive_splitter` — which use nothing else in this PR — pay it on every single file.

Three options, cheapest first:

1. Short-circuit on the stored count. `chunk.token_count` is only stale for chunks contextualization rewrote, and it is stale *low* — the `[CONTEXT]` envelope only adds. So `token_count > limit` is already conclusive, and the recount is only needed for chunks whose stored count is near the limit. Skipping any chunk with `token_count <= limit - envelope_headroom` cuts the pass to almost nothing on normal documents.
2. Have `contextualize_stage` refresh `token_count` when it rewrites `text` via `model_copy`, which removes the reason for the recount entirely and fixes the stale count for every other consumer of that field too.
3. Failing both, `await asyncio.to_thread(...)` it, so at least it stops blocking the loop.

`openrag/services/workers/pipeline_builder.py:188`, `openrag/services/workers/pipeline_builder.py:363-369`

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,898 @@
...
+        candidates = self._pack(units, filename)
+        candidates = self._merge_small(candidates, filename)
+
+        chunks: list[Chunk] = []
+        for idx, unit in enumerate(candidates):
+            # Only claim a header the chunk actually carries: ``header`` is
+            # persisted alongside ``text``, so populating it while
+            # ``prepend_heading_path`` is off would let a consumer rebuild
+            # ``header + content`` into a string that was never embedded.
+            header = self._build_header(filename, unit.heading_path, unit.pages) if self.prepend_heading_path else ""
```
Ahmath-Gadji: **`prepend_heading_path=False` silently drops every heading from the indexed corpus.**

`_build_units` never emits a heading line as body — it updates the stack and `continue`s (`structured_section.py:356-362`), by design, so a heading can't be stranded as an orphan chunk. The heading text then reaches the output through exactly one path: `_build_header`, here on line 270. Turn `prepend_heading_path` off and that path is gone.

Same document, only the flag changed:

```
prepend_heading_path=True
  "[Source: code.pdf | Page: 1 | Section: Livre Ier : Dispositions générales > Titre Ier : Principes > Chapitre unique]

   Article L110-1 Le présent code régit l'entrée et le séjour des étrangers."

prepend_heading_path=False
  "Article L110-1 Le présent code régit l'entrée et le séjour des étrangers.

   Article L110-2 L'étranger qui séjourne régulièrement peut solliciter une carte."
```

`Livre Ier`, `Titre Ier`, `Chapitre unique`, `Titre II` appear nowhere in any chunk's `text`. They survive only in `metadata["hierarchy_path"]`, which lands in Milvus's dynamic field — and the BM25 function is declared over `text` alone (`milvus_store.py:311-318`), so that heading text is unreachable by dense *and* sparse retrieval. Compare `recursive_splitter`, which keeps heading lines inline in the body and so loses nothing.

It's a plain `bool` in `ChunkerConfig` with no warning, so an operator flipping it to drop the `[Source | Page | Section]` preamble (a reasonable thing to want — it costs tokens and repeats across chunks) silently deletes their document structure from the index instead.

Either keep the heading lines in the body when the flag is off, or document on the field that `False` means "headings are not indexed at all" — the current comment doesn't hint at it.

`openrag/core/chunking/structured_section.py:270`, `openrag/core/config/chunking.py:41`

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,898 @@
...
+            max(1, target - reserve) if target is not None else self._packing_budget(filename, path, unit.pages),
+            max(1, ceiling - reserve) if ceiling is not None else self._effective_max(filename, path, unit.pages),
+            self.length_function,
+        )
+        return [
+            _Unit(
+                heading_path=list(unit.heading_path),
+                text=piece.strip(),
+                tokens=self.length_function(piece),
+                pages=set(unit.pages),
```
Ahmath-Gadji: **Major: every piece of a split unit inherits the parent's whole page set, so `page_number` / `page_range` are wrong.**

`_split_oversize` copies `pages=set(unit.pages)` verbatim into each piece. When the source unit spans pages — which is the normal case for any document whose structure isn't dense enough to cut per page — every piece claims the full span, and `Chunk.page_number` (`min(pages)`) becomes the unit's *first* page for all of them, including pieces physically several pages later. `_copy_unit:724` does the same.

Verified against a live 4-page PDF (`cispe_cloud_cispe_digital_sovereignty_principles.pdf`, PyMuPDF parser), two partitions differing only in chunker:

| chunk | 0 | 1 | 2 | 3 | 4 | 5 |
|---|---|---|---|---|---|---|
| `recursive_splitter` (`page`) | 1 | 1 | 2 | 3 | 3 | 4 |
| `structured_section` (`page`) | 1 | 1 | 1 | 1 | 1 | 1 |
| `structured_section` (`page_range`) | 1-4 | 1-4 | 1-4 | 1-4 | 1-4 | 1-4 |

Isolated to this line — 4 pages of prose, no headings, no leaf markers, so `_build_units` emits one unit:

```
A) units built:   tokens=816   pages=[1, 2, 3, 4]
B) chunks after splitting:
     page=1  page_range='1-4'   'Page 1 sentence one. Filler sentence about cloud'
     page=1  page_range='1-4'   ...  (×15, including pieces whose text is on page 4)
```

The strategy docstring lists as design goal 6 that `[PAGE_N]` markers "become `page_number` / `page_range` metadata, never inline text". The first half works — no marker leaked into any chunk body. The second half doesn't: the metadata they become is wrong for every split piece.

This is a user-visible regression, not just bookkeeping. `page` is what the citation UI shows and what `/extract` anchors on, so a source card points at page 1 of a 4-page document; the error grows with document length, and on the 904-page code in the PR description every chunk of a multi-page unit would carry the same misleading range.

`markdown_utils.get_chunk_page_number()` already solves exactly this for the recursive path — it resolves start/end pages from the `[PAGE_N]` markers present in a given piece of text, with a carry-in for the previous chunk's ending page. Threading the same resolution through `_greedy_split` (or re-deriving each piece's pages from the marker positions before they're stripped) would fix both call sites.

`openrag/core/chunking/structured_section.py:598`, `openrag/core/chunking/structured_section.py:724`

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,898 @@
...
+                    flush()
+                    _push_heading(stack, level, text)
+                    buf_path = list(stack_path(stack))
+                    continue
+
+                if self._is_leaf(raw_line):
+                    flush()
+                    buf_path = list(stack_path(stack))
+
+                if raw_line.strip():
```
Ahmath-Gadji, line 598: **Major: dropping blank lines here erases every paragraph break, which then cascades into the splitter.**

`if raw_line.strip():` skips blank lines, and `buf_lines` is joined with a single `\n` in `flush()`. Block-level separation is gone before packing ever runs.

Measured on a live 4-page PDF (PyMuPDF parser, 126 blank-line-separated blocks in the parsed markdown), two partitions differing only in chunker:

| | paragraph breaks preserved across all 6 chunks |
|---|---|
| `recursive_splitter` | **152** |
| `structured_section` | **7** |

Three consequences, all reproduced in the live index.

**1. Headings get glued onto the end of the preceding sentence.** With no `\n\n` left, `_greedy_split`'s paragraph splitter (`re.split(r"\n{2,}")`) never matches, so every oversize unit falls straight through to the sentence ladder — which rejoins with `" "`. Actual indexed text:

```
...portability of some providers' solutions. **CISPE and Digital Sovereignty**
...without being hindered by unfair license terms. **Recommended measures** :
...providers trade associations. **2/ Control of data protection**
```

Same source text through `recursive_splitter`:

```
choice for cloud customers.

**1/ Freedom of technological choice**

**Principle** : Customers should be able to choose...
```

This works against rule 4 of the strategy ("a heading always travels with its body") from the opposite side — `_rehome_trailing_headings` guards against a heading stranded at a piece's *end*, but a heading welded into the middle of a prose line is no longer detectable as a heading at all.

**2. Lettered list items are severed from their content across chunk boundaries.** Live chunk 2 ends on a bare `b.`; chunk 3 opens with `CSPs shall have services certified against...`. Chunk 3 ends on `d.`; chunk 4 opens with its content. Two causes compound:

```
_BLOCK_LINE_RE:  '1. item' → True     'a. item' → False
_SENTENCE_RE.split("b. CSP shall provide transparency.")
  → ['b.', 'CSP shall provide transparency.']
```

`_BLOCK_LINE_RE` (`recursive.py:43`) has `\d+[.)]\s` for ordered lists but no lettered equivalent, so `a.`/`b.`/`c.` are prose; `_SENTENCE_RE` (`:777`) then reads the marker's period as a sentence terminator and makes the marker its own unit, free to land at a chunk boundary. Adding `[a-z][.)]\s` to `_BLOCK_LINE_RE` and a negative lookbehind for a single-letter/digit token in `_SENTENCE_RE` would cover both.

**What does survive:** `- ` bullets come through intact (`_BLOCK_LINE_RE` covers them) — the four-attribute list and the nested `(1)`/`(2)` items were clean. And no `[PAGE_N]` leaked into any body.

**Related observation, not a bug:** PyMuPDF emitted exactly one `#` heading for this document; its ~10 real section headers (`Background at EU level`, `1/ Freedom of technological choice`, `About CISPE`, …) are all `**bold**` lines. Since detection covers `#` headings and configured keywords only, all six chunks sat under the single document title and the section structure never reached `hierarchy_path`. The PR's measurements are marker-based, where headings arrive differently — worth a line in the docs that on a PyMuPDF-parsed PDF the strategy's headline feature is largely inert unless `heading_keywords` is tuned for the corpus.

`openrag/core/chunking/structured_section.py:368`

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,1056 @@
...
+                    # The heading opens the body of the unit it introduces, the
+                    # way recursive_splitter keeps heading lines inline. Holding
+                    # it only in the breadcrumb made the breadcrumb its sole
+                    # home, so every path-shortening step deleted it outright:
+                    # _merge_small taking a _common_prefix dropped the differing
+                    # leaf, a sibling heading popped it before any body flushed,
+                    # and prepend_heading_path=False removed the last copy.
+                    # Coverage against the source ran 91-99% where
+                    # recursive_splitter ran ~100%.
+                    add_line(text, page, heading=True)
```
Ahmath-Gadji: **Headings reach the body stripped of their markdown, so the chunk loses the structure it just detected.**

`_heading()` returns `_clean_heading(...)` output — `#` markers and emphasis removed — and that flattened string is what `add_line(..., heading=True)` puts into the body. A source `### List of Figures` is indexed as bare `List of Figures`, indistinguishable from a prose line.

Measured on a live index of a 44-page report (`ecdpm_org_global_approaches_to_digital_sovereignty…pdf`, PyMuPDF), same file in two partitions differing only in chunker:

| | ATX `#` markers surviving in chunk bodies |
|---|---|
| parser output (`pymupdf4llm`) | 30 |
| `recursive_splitter` | 35 (overlap duplicates a few) |
| `structured_section` | **0** |

The actual chunk, which is what prompted this:

```
[Source: ecdpm.pdf | Page: 3-4 | Section: List of Figures]
List of Figures
Figure 1.1: Different approaches to governance of personal and non-personal data ....... 10
Figure 1.2: Approaches to data localisation ............................................ 15
Figure 2.1 **:** Leading tech firms are largely non-EU ................................. 24
Figure 2.2: The bloc has also fallen behind in investments ............................. 24
ii
Acknowledgements
The authors would like to thank the many policymakers and experts who shared their expertise…
```

Source for those two lines is `### List of Figures` and `### Acknowledgements`.

Three costs:

1. **The LLM can't see the hierarchy.** `## X` vs `#### X` is the cheapest possible signal of nesting depth, and it's dropped. Downstream the chunk is markdown-rendered (the admin UI above renders it as markdown), so a heading renders as a paragraph.
2. **It's lossy relative to the default chunker.** `recursive_splitter` keeps heading lines verbatim, so switching a partition to `structured_section` silently degrades the markdown of every chunk.
3. **The stated reason no longer applies.** The comment at `:437` says the markup is dropped so the body "isn't re-detected as a heading downstream" — but that reasoning is for *rejected* heading candidates on the `_clean_heading(raw_line) if _MD_HEADING_RE.match(raw_line)` path. For an *accepted* heading, re-detection is exactly what a downstream consumer should be able to do.

Keeping `"#" * level + " " + text` (or the original `raw_line`) would preserve the structure. `_clean_heading` output is still the right thing for the breadcrumb and `hierarchy_path` — only the body copy needs the markers.

Related, visible in the same chunk: `hierarchy_path` is `['List of Figures']` while the body also carries the whole `Acknowledgements` section, because `_merge_small` keeps the previous unit's path when it folds a sibling in. So `section_title` names only the first of the two sections the chunk actually contains.

`openrag/core/chunking/structured_section.py:424`

Ahmath-Gadji (commented):
Re-tested the five earlier findings end-to-end on a live instance running this head — all fixed.

What follows are open questions from that round rather than findings, with an example where one helps.

---

### 1. Should sibling sections in the mid-size band be allowed to pack together?

`_pack` merges two units only when `cur.heading_path == unit.heading_path` (`:525`), and `_merge_small` only fires when at least one side is under `min_tokens`. So two adjacent sections that are each *between* `min_tokens` and `target_tokens` can never be combined — too big for the small-merge pass, structurally barred from the packing pass. At defaults that band is 128-512 tokens.

```
6 adjacent ## sections @ ~216 tok  ->  6 chunks [216 ×6]  =  42% of the 512 target
```

Genuinely small sections are fine — a deck of ~25-token slides, and a run of law stubs that only cross-refer, both collapse to a single chunk because they fall under `min_tokens`. And a real 44-page report packs well (140 chunks, mean 487, 95% of target).

So: is the mid-band worth addressing, or rare enough in your corpora to leave alone? Would packing across sibling paths while the result stays under `max_tokens` — recording the shared ancestor as the path — be acceptable, or does that dilute the section guarantee too much?

### 2. Is 1.5x the intended tolerance — and should this chunker's knobs be exposed in the UI?

Answering the open question rather than raising one: an oversize section does split (`:519`), at `max_tokens` minus the header reserve, and `max_tokens` defaults to `int(chunk_size * 1.5)` (`:222`), overridable per preset.

```
~586 tok section  ->  1 chunk  [606]        (18% over target, kept whole)
~781 tok section  ->  2 chunks [515, 303]
```

Is 1.5x deliberate?

That leads into a broader one. `structured_section` adds six knobs to `ChunkerConfig` — `min_tokens`, `max_tokens`, `hard_max_tokens`, `prepend_heading_path`, `heading_keywords`, `leaf_patterns` — and they all round-trip through the preset API today. But the admin UI's chunking form (`ui/src/pages/admin/presets.tsx:300-346`) renders only **Strategy**, **Chunk size** and **Overlap rate**, so none of the six are reachable from the UI. `heading_keywords` and `leaf_patterns` are the ones that make boundary detection generalise past French legal codes, which is the whole premise of the strategy being config-driven — tuning them currently means editing a preset by API.

Two smaller things in that same form, both visible with `structured_section` selected:

- **Overlap rate is shown but ignored.** The constructor forces `chunk_overlap_rate=0.0` (`:214`), so the field silently does nothing for this strategy.
- **The show/hide logic branches on strategy names that are not registered here.** It tests against `"markdown_section"` and `"markdown_layout"`, while `/presets/options` returns `["recursive_splitter", "structured_section"]` — so the conditions never match and the form was never adapted for the new strategy.

Would it be worth surfacing the sizing band and the pattern lists in the preset editor, and gating Overlap rate on strategies that actually honour it? Happy for that to be a follow-up PR if it is out of scope here.

### 3. Should a split continuation replay its lead-in and its parent bullet?

`chunk_overlap_rate` is forced to `0.0` (`:214`). That reads right for whole atomic leaves. It reads less right once a nested list has to be split, because a sub-bullet can land at the top of a chunk with nothing above it:

```
chunk 0:  Recommended measures
          The following measures apply to every provider ... and are cumulative:
          - Measure 1: ...
              - A. The record shall state the purpose, the legal basis ...
              - B. The record shall state the purpose, the legal basis ...

chunk 1:      - C. The record shall state the purpose, the legal basis ...   <- no lead-in, no parent bullet
```

Reproduced on this head — a 12-measure list with four sub-items each splits into three chunks, and chunks 1 and 2 both open on a sub-item whose parent `- Measure N:` sits in the previous chunk:

```
chunk 0: opens on 'Recommended measures'                     lead-in: yes
chunk 1: opens on '- B. The record shall state the purpose'  lead-in: no,  parent bullet: no
chunk 2: opens on '- D. The record shall state the purpose'  lead-in: no,  parent bullet: no
```

The sub-item states a rule without saying what it is a rule about. The header still carries `Section: Recommended measures` on all three, so section context survives — what is gone is the enclosing bullet and the lead-in, i.e. the two levels between the section title and the text.

One aggravating factor, possibly worth its own fix: the above is the **loose** list form. Written **compactly** — no blank line between items, the usual markdown form — the list is also flattened, because `_greedy_split` splits paragraphs on `\n{2,}`, a compact list is a single paragraph, so it falls through to the sentence rung which rejoins with `" "`:

```
compact list:  60 source bullets ->  3/60 still on their own line
loose list  :  60 source bullets -> 60/60 still on their own line
```

So for a compact list the continuation chunk is not merely missing its parent — it is space-joined prose with no list structure left at all.

Would replaying the unit's opening line, and for a nested list the innermost enclosing bullet, at the top of each continuation piece be worth it? That stays bounded, unlike a token-based overlap, and fires only where a split actually happened.

### 4. With the section filled, is contextualization still worth its LLM call here?

`chunk_contextualizer_tmpl.txt` builds its summary from *Filename, First Chunks, Previous Chunks, Current Chunk*. This strategy already supplies filename, page and section path deterministically in `_build_header`, for free — so the remaining delta looks like mostly the **Previous Chunks** continuity, which is what question 3 is about.

If a lead-in replay lands, does `enable_contextualization` still pay for a call per chunk on this strategy? Might be worth an A/B on the eval corpus before pairing the two by default.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in 1664d397. Verified end-to-end on a live instance rebuilt at `3b12b96f`.

The flat `.txt` repro now yields 3 chunks with real breadcrumbs instead of 1 with `hierarchy_path: []`, and every `Titre`/`Chapitre` is present in the indexed text. The retrieval check that was the proof of the defect:

| query | before | after | recursive control |
|---|---|---|---|
| `Dispositions pénales` | 0 hits | 5 | 2 |
| `Dispositions sanitaires` | 0 hits | 6 | 2 |
| `Chapitre Ier : Taux applicables` | 0 hits | 5 | 2 |

The paged variant also keeps its paths, and the permissive route now requires a shared ancestor. Resolving.

openrag/services/workers/pipeline_builder.py
Ahmath-Gadji: Fixed in 1664d397. `token_count` is a lower bound (the envelope only adds), so it settles every chunk except the band near the limit and the full re-tokenisation no longer runs on the event loop for every file — including on `recursive_splitter` partitions, which paid it for nothing. Reported 361 ms → 0.30 ms at 3000 chunks. Resolving.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in cde6a982 / 1664d397 — the breadcrumb is now kept inline at the top of the body, so heading text reaches the index regardless of the flag, and the flag still does what turning it off asks for (drops the verbose `[Source | Page | Section]` preamble).

Verified with `prepend_heading_path=False` on the live build: `Titre Ier`, `Chapitre Ier`, `Titre II`, `Titre III`, `Titre IV` all present in the chunk bodies. Resolving.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in 99a8ecd1 via `_Unit.page_marks`. Verified on the same live 4-page PDF:

| | chunk pages |
|---|---|
| before | `1, 1, 1, 1, 1, 1` — every chunk `page_range: 1-4` |
| after | `1, 1-2, 2-3, 3-4, 4` |
| recursive control | `1, 1, 2, 3, 3, 4` |

Resolving.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in 99a8ecd1. Verified on the same live PDF:

| | before | after | recursive control |
|---|---|---|---|
| paragraph breaks preserved | 7 | 125 | 152 |
| orphaned `b.` / `d.` at boundaries | 2 | 0 | — |
| headings welded mid-line | 3+ | 0 | — |
| word coverage vs source | — | 100.00% | 100.00% |

One residual with a different cause, raised in the review rather than here: a **compact** list (no blank line between items) is still flattened by the sentence rung, since `_greedy_split` splits paragraphs on `\n{2,}` and a compact list is a single paragraph — 3/60 bullets keep their own line, against 60/60 for the loose form. Not this finding, so resolving this one.

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,1234 @@
...
+        """Body ceiling so ``chunk.text`` (body + header) never exceeds ``max``."""
+        return max(1, self.max_tokens - self._header_reserve(filename, heading_path, pages))
+
+    def _emit_atomic(self, unit: _Unit, filename: str) -> list[_Unit]:
+        """A table over the (header-adjusted) max is split by ``chunk_table``;
+        whatever still exceeds the ceiling then goes through
+        ``_enforce_ceiling``."""
+        emax = self._effective_max(filename, unit.heading_path, unit.pages)
+        if unit.chunk_type is ChunkType.TABLE and unit.tokens > emax:
+            element = MDElement(type="table", content=unit.text, page_number=unit.start_page)
```
Ahmath-Gadji: **Regression in `fb711ceb`: a split table now loses its column header on every continuation piece.**

`chunk_table` assumes `element.content` opens with the table — `parse_markdown_table` reads the leading lines as the header block and replays them on each piece. Since `fb711ceb` the heading travels into the atomic unit (`_build_units`: `atomic.text = f"{pending}\n\n{atomic.text}"`), so what arrives here starts with `## Regional breakdown`, and that heading is replayed instead of the column header.

Bisected on a table-only document (60 rows, `chunk_size=512`):

```
67670334  chunk 0: '## Regional breakdown' / '| Region | Q1 | Q2 | Q3 |'
          chunk 1: '| Region | Q1 | Q2 | Q3 |' / '| --- | --- | --- | --- |'   <- header replayed

fb711ceb  chunk 0: '## Regional breakdown' / '| Region | Q1 | Q2 | Q3 |'
          chunk 1: '## Regional breakdown' / '| Region 51 | 561 | 612 | 663 |' <- header gone
```

Piece 2 onward is a wall of unlabelled numbers. `_enforce_ceiling`'s docstring is explicit that this must not happen — *"`chunk_table` already divides them losslessly on row boundaries, replaying the column header"* — and that is exactly the property the "tables are never force-split" carve-out is resting on.

In isolation `chunk_table` is fine (7/7 pieces carry the header); it is only the heading-prefixed `unit.text` that breaks it.

Simplest fix is to keep the heading out of what `chunk_table` parses and re-attach it after — split on `unit.text` minus its leading heading lines, then prepend the heading to the first piece (or to each, matching whatever `fb711ceb` intended). Worth a test asserting the column header appears in every piece of a split table, since the full suite (2608 tests) passes with this broken.

`openrag/core/chunking/structured_section.py:736`

openrag/core/chunking/structured_section.py
```diff
@@ -0,0 +1,1234 @@
...
+        pages = [b for b in document.text_blocks if (b.text or "").strip()]
+        if len(pages) < _MIN_PAGES_FOR_DETECTION:
+            return False
+        counts = [self.length_function(b.text) for b in pages]
+        if statistics.median(counts) > _MAX_MEDIAN_TOKENS_PER_PAGE:
+            return False
+        short = sum(1 for c in counts if c < _SHORT_PAGE_TOKENS) / len(counts)
+        return short >= _MIN_SHORT_PAGE_FRACTION
+
+    def _page_units(self, content: str) -> list[_Unit]:
```
Ahmath-Gadji, line 380: **Paginated mode bypasses the markdown-element pass, so tables and line-wraps are handled differently from every other document.**

`_page_units` iterates `content.split("\n")` directly. `_build_units` instead goes through `split_md_elements`, `dewrap_paragraphs`, `is_placeholder_image` and `_atomic_unit`. None of those run here, so a document that trips `_looks_paginated` gets a different contract — and a slide deck is exactly where large tables and figures live.

Same 13-page deck with one 60-row table on page 8, only `layout` changed:

```
layout=flowing    chunk 1: type=table  '| Region | Q1 | Q2 | Q3 |'   (rows on their own lines)
                  chunk 2: type=table
layout=paginated  chunk 7: type=text   '| Region | Q1 | Q2 | Q3 | |---|---|---| | Region 1 | 11 …'
                  chunk 8: type=text   '| Region 30 | 330 | 360 | 390 | | Region 31 | …'
```

Three consequences:

1. **`chunk_type` is `text`, never `table`** — 0 table-typed chunks in paginated mode against 2 in flowing. Any consumer filtering on `chunk_type` stops seeing tables in these documents.
2. **`chunk_table` never runs**, so the split is the prose sentence-splitter and the column header is not replayed on later pieces.
3. **The table is flattened onto one line.** With no `\n\n` between rows the unit is a single paragraph, so `_greedy_split` falls through to the sentence rung, which rejoins with `" "`. This is the same compact-block cascade I raised for lists in the review — tables hit it here because nothing upstream made them atomic.

Separately, `dewrap_paragraphs` is skipped, so PDF visual line-wraps survive verbatim:

```
layout=paginated  | Le present rapport analyse les consequences economiques
                  | de la transformation numerique sur les entreprises de taille
layout=flowing    | Le present rapport analyse les consequences economiques de la transformation …
```

That is the defect #579 fixed for `recursive_splitter`, reappearing for any document `_looks_paginated` accepts. Since detection is `auto` by default and measures the parsed pages rather than the file type, an operator does not choose this — a report with enough sparse pages lands in it silently.

Would reusing the `split_md_elements` walk and cutting on `[PAGE_N]` boundaries within it — rather than a separate raw-line loop — be workable? That would keep one contract for atomicity, placeholders and reflow, with only the boundary rule differing.

`openrag/core/chunking/structured_section.py:380`

pyproject.toml
```diff
@@ -76,6 +76,18 @@ dev = [
...
+# group, not a runtime dependency: it must never land in the shipped image.
+#
+# Pinned to the same fork+rev as the eval-promptfoo repo. The metrics live in
+# ``ragas.metrics.collections`` (the "v2" API), which upstream PyPI ragas does
+# not expose — and matching the rev is what makes the scores here directly
+# comparable with that harness rather than merely similar.
+#
+#   uv sync --group eval
+eval = [
+    "ragas",
```
Ahmath-Gadji: **Scope check: this pulls a git dependency on a personal fork, for a script that is not in the PR.**

`scripts/rag_eval.py` — the stated and only consumer — does not exist anywhere on this branch. Nothing in the tree imports `ragas`; the sole references are these two comment blocks. So as it stands the group buys nothing here and the pin cannot be validated against the code that is supposed to use it.

`uv.lock` carries the fork and its transitive closure regardless (+288 lines), for everyone who resolves the lockfile:

```
uv.lock:2986  eval = [{ name = "ragas", git = "https://github.com/idostik/ragas?rev=561262f…" }]
```

`eval` is not a default group, so a plain `uv sync` will not install it — that part is fine. The concern is a dependency on an individual's fork rather than an org-controlled mirror: the rev pin protects against content drift, but not against the account or repo going away, and the lockfile now depends on that URL resolving.

Would it be cleaner to land this with the script that uses it, in its own PR, and to mirror the fork under `linagora/` if the `ragas.metrics.collections` API is genuinely needed?

Same question for `rag.expose_context` / `include_context` (`core/config/retrieval.py`, `services/orchestrators/query_service.py`, ~45 lines): the reasoning is sound and it is double-gated off by default, but it is a retrieval-API feature in a chunking PR. You already offered to split the embedder-window plumbing into a follow-up — these two look like better candidates for that than the window plumbing, which the chunker genuinely needs.

`pyproject.toml:79-89`

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in e214b225. Verified at `fb711ceb` — `### List of Figures` and `### Acknowledgements` now reach the body with their markers intact, where before 0 of the source document's 30 ATX headings survived. Resolving.

Ahmath-Gadji (commented):
Second pass, on `fb711ceb`. The six earlier threads are all resolved — the heading-markdown one by `e214b225`, verified. The PR has grown since though (structured_section.py is now 1234 lines, and `67670334` adds a layout mode), so this pass is on what arrived after `3b12b96f`.

**One regression, in the top commit.** `fb711ceb` moves the heading into the atomic unit's text, which is right for the orphan-heading problem it targets — but `_emit_atomic` then hands that text to `chunk_table`, which reads the leading lines as the table's header block. So every continuation piece of a split table now replays the *heading* instead of the column header. Bisected `67670334` → `fb711ceb`; details on the thread. The full suite (2608 tests) passes with it broken, so it needs a test as well as a fix.

**One design question on the new layout mode.** `_page_units` walks raw lines rather than going through `split_md_elements` / `dewrap_paragraphs` / `is_placeholder_image`, so paginated documents get a different contract from every other document: tables never get `chunk_type=table` or `chunk_table`, and PDF line-wraps are not reflowed (#579's defect, for any document `_looks_paginated` accepts). Since detection is `auto` and measures parsed pages rather than file type, an operator does not opt into that. Thread has the side-by-side.

**Two scope items**, both raised as questions rather than objections: the `eval` group adds a git dependency on a personal `ragas` fork whose only stated consumer (`scripts/rag_eval.py`) is not in this PR, and `rag.expose_context` is a retrieval-API feature living in a chunking PR.

Nothing else critical surfaced. The `67670334` trade-off you flagged yourself — `pct_under_min` rising to 45% on decks because `_merge_small` is skipped in paginated mode — I did not treat as a finding, since the commit message is explicit that whether a page-sized chunk retrieves better is unmeasured. That seems like the right thing to settle on the eval corpus rather than in review.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Fixed in 4064d582. Verified at `70c04eda` on the same table-only document that bisected the regression:

```
chunk 0: type=table tok=769  line0='## Regional breakdown'      line1='| Region | Q1 | Q2 | Q3 |'
chunk 1: type=table tok=149  line0='| Region | Q1 | Q2 | Q3 |'  line1='| --- | --- | --- | --- |'
>>> every piece carries the column header: True
```

Matches the pre-regression behaviour at `67670334`, and the heading stays attached to the first piece — the one it introduces. Also holds when the split produces more pieces: at `chunk_size=200` the column header is present in 4/4.

Resolving. Note the same guarantee still does not hold in paginated mode, since `_page_units` never makes the table atomic so `chunk_table` is not reached at all — that is the separate thread on `_page_units`, not this one.

openrag/core/chunking/structured_section.py
Ahmath-Gadji: Still reproduces at `70c04eda` — flagging because 4064d582 fixed the column-header guarantee for the flowing path only. Paginated mode never makes the table atomic, so `chunk_table` is not reached and the fix does not apply:

```
layout=flowing    chunk_type=table: 2   column header in 2/2 table chunks   line-wraps reflowed: True
layout=paginated  chunk_type=table: 0   column header in 1/2 table chunks   line-wraps reflowed: False
```

and the table still arrives as a single flattened line:

```
| Region | Q1 | Q2 | Q3 | |---|---|---|---| | Region 1 | 11 | 12 | 13 | | Region 2 | 22 | 24 | 26 | …
```

No action needed if you would rather take this as a follow-up — the mode is opt-in-ish (`auto` fires on 3 of your 30 corpus documents), and the rest of the PR is in good shape. Leaving the thread open so it does not get lost.

andyne13: Confirmed, and taking it as a follow-up.

One update since `70c04eda`: `d6f3dd68` defaults `layout` to `"flowing"`, so the paginated path is no longer reachable unless a preset sets `layout="paginated"` or `"auto"` explicitly — `auto` no longer fires on its own. That removes the part you flagged as not being an operator's choice (a report with enough sparse pages landing there silently), and it aligns the chunker's own default with `ChunkerConfig`'s, which had drifted to `"auto"` and would have made the eval harness disagree with production.

So the defect is unchanged but now dormant by default rather than opt-in-ish. The follow-up is the rewrite you suggested — build page units from the `split_md_elements` walk and cut on `[PAGE_N]` within it, so atomicity, placeholders and reflow keep one contract and only the boundary rule differs. Leaving the thread open for that.

Ahmath-Gadji (dismissed):
Approving. Every finding from the earlier passes is fixed and verified end-to-end, and the one thread left open (`_page_units` bypassing the markdown-element pass) is agreed as a follow-up now that `d6f3dd68` makes the paginated path opt-in rather than something `auto` could select on its own.

Two things for a future iteration, neither blocking this PR:

* **Expose this chunker's hyperparameters in the UI so an admin can tune them.** `structured_section` adds `min_tokens`, `max_tokens`, `hard_max_tokens`, `prepend_heading_path`, `heading_keywords` and `leaf_patterns` to `ChunkerConfig`, and they all round-trip through the preset API — but the admin form (`ui/src/pages/admin/presets.tsx`) renders only Strategy, Chunk size and Overlap rate, so tuning any of them means editing a preset by API. `heading_keywords` / `leaf_patterns` are the ones that let boundary detection generalise past French legal codes, which is the whole premise of the strategy being config-driven.

* **Table and image chunks lose their caption.** A caption — `Table 3: …`, `Source: …`, `Figure 2 — …` — sits on the line before or after the block, so it lands outside the atomic unit and the table chunk arrives with no statement of what it depicts. Retrieval then has only the cell contents to match against, which is exactly what a question about the table will not contain. This is not specific to this PR (`recursive_splitter` has the same gap), so it belongs in the strategy rather than in a fix here.

  The interesting part is that the detection already exists: `_CAPTION_RE` (`structured_section.py:89`) recognises `figure|table|tableau|source|photo|graphique|schéma|crédits`, and `:127` uses it to *reject* those lines as headings. So the signal is identified and then discarded. Since we already carry the section path, the natural place for it is the same breadcrumb:

  ```
  [source: XX, section: Section > Sub-section 1 > Table 3: quarterly revenue by region]
  ```

  That would make a table or figure retrievable by what it is about, not only by the tokens inside it.

Ahmath-Gadji (dismissed):
Approving. Every finding from the earlier passes is fixed and verified end-to-end, and the one thread left open (`_page_units` bypassing the markdown-element pass) is agreed as a follow-up now that `d6f3dd68` makes the paginated path opt-in rather than something `auto` could select on its own.

Two things for a future iteration, neither blocking this PR:

* **Expose this chunker's hyperparameters in the UI so an admin can tune them.** `structured_section` adds `min_tokens`, `max_tokens`, `hard_max_tokens`, `prepend_heading_path`, `heading_keywords` and `leaf_patterns` to `ChunkerConfig`, and they all round-trip through the preset API — but the admin form (`ui/src/pages/admin/presets.tsx`) renders only Strategy, Chunk size and Overlap rate, so tuning any of them means editing a preset by API. `heading_keywords` / `leaf_patterns` are the ones that let boundary detection generalise past French legal codes, which is the whole premise of the strategy being config-driven.

* **Table and image chunks lose their caption.** A caption — `Table 3: …`, `Source: …`, `Figure 2 — …` — sits on the line before or after the block, so it lands outside the atomic unit and the table chunk arrives with no statement of what it depicts. Retrieval then has only the cell contents to match against, which is exactly what a question about the table will not contain. This is not specific to this PR (`recursive_splitter` has the same gap), so it belongs in the strategy rather than in a fix here.

  The interesting part is that the detection already exists: `_CAPTION_RE` (`structured_section.py:89`) recognises `figure|table|tableau|source|photo|graphique|schéma|crédits`, and `:127` uses it to *reject* those lines as headings. So the signal is identified and then discarded. Since we already carry the section path, the natural place for it is the same breadcrumb:

  ```
  [source: XX, section: Section > Sub-section 1 > Table 3: quarterly revenue by region]
  ```

  That would make a table or figure retrievable by what it is about, not only by the tokens inside it.

Ahmath-Gadji (approved):
LGTM. All review threads resolved and CI green (api-tests, milvus-integration, tests, lint, layer-import-guard, CodeRabbit).

`structured_section` is opt-in per preset and `recursive_splitter` stays the default, so no behaviour change for existing partitions. The `hard_max_tokens` derivation from the embedder's `max_model_len` is the right safety net against silent vLLM truncation.

The remaining `_merge_small` orphan-heading case is documented and tracked rather than hidden — fine to land as-is.

PR #863, opened 2026-08-27
fix(vdb): stamp the schema version when the collection is created

### LINKED ISSUE #862: Concurrent indexing into a collection that does not exist yet silently loses a file
## Summary

The first indexing task to reach the store stage against a newly created collection fails, because the store stage issues its idempotency `delete` before the collection has been loaded. Milvus rejects a delete on an unloaded collection, and that rejection aborts the whole indexing task — even though, on a brand-new collection, the delete has nothing to remove by definition.

With more than one file indexing concurrently, one file is silently lost per fresh partition: `LoadCollection` happens while the first task is already failing, so whichever task arrives after it succeeds.

## Evidence

From a single stack boot (Milvus and API logs, same trace):

```
12:29:22.518  proxy/impl.go:522    CreateCollection received
12:29:22.872  utils/util.go:121    failed to GetShardLeaders
                                   error="collection not loaded[collection=468669888254444638]"
12:29:22.872  proxy/task_delete.go:568  fail to execute complex delete
12:29:22.872  proxy/impl.go:2708   Failed to run delete task: collection not loaded[...]
12:29:25.113  proxy/impl.go:792    LoadCollection received          <-- 2.24s after the delete
```

The rejected delete is the per-file idempotency delete:

```
expr="(partition == "orb-docling-...") and
      (file_id  == "bench-00002-facts2-885db8331e2d") and
      (_openrag_indexing_task_id == "720259a7f25c461ab1b8d904e91ac824")"
count=0
```

`count=0` — there was nothing to delete, as expected for a file being indexed for the first time.

Both files in that partition reported a complete stage-timing line, and the difference is entirely in `ms_store`:

```
task_id=720259a7... | filename=facts2.txt | n_chunks=1 | ms_embed=86 | ms_store=3    | ms_total=2698   <-- task FAILED
task_id=982db360... | filename=facts1.txt | n_chunks=1 | ms_embed=74 | ms_store=3663 | ms_total=6062   <-- task ok
```

The failing task still logs `indexing stage timings` at INFO as though it completed; only `ms_store` (3 ms — the time to be rejected) distinguishes it.

## Reproduction

1. Boot a fresh stack (empty Milvus volume).
2. Create a partition.
3. Upload 2+ files to it concurrently, immediately.

Observed over 12 consecutive fresh-stack boots (each with 2 files, 2 upload workers, across the pymupdf/marker/docling parsers):

| | |
|---|---|
| boots where one file failed | 11 / 12 |
| `collection not loaded` errors in each | 6 |
| boots where both files indexed | 1 (and it logged 0 such errors) |

The one clean boot is the case where both tasks happened to reach the store stage after `LoadCollection` had completed — both show `ms_store` ≈ 3880 ms.

## Impact

Silent partial data loss on first use of any partition: the file is accepted (`201`), the task ends `FAILED`, and the document is simply absent from the index. It is most visible in short-lived or per-run stacks (CI, benchmarks, ephemeral deployments), where every partition is a cold one. On a long-lived deployment it costs the first file or two after each new collection, which is easy to miss.

## Suggested fix

Either would close it:

- Ensure the collection is loaded before the store stage issues its delete, or
- Treat `collection not loaded` on the idempotency delete as "nothing to delete" and continue. The delete exists to remove prior rows for the same `file_id`/task; on a collection that has never been loaded there are none, so failing the task on it is never the right outcome.

The second is the more robust of the two, since it does not depend on load timing at all.

## Environment

- OpenRAG `v2.1.1` (`2198252`)
- `milvusdb/milvus:v3.0.0`
- Compose deployment, GPU (NVIDIA L4), external embedder endpoint
- 2 files, 1 chunk each; `uploadWorkers=2`

## Possibly related

#845 (readiness green while unusable) is a different mechanism, but the same shape: the stack reports itself ready before it can actually serve the request that follows.

EnjoyBacon7:
Fixes #862.

## The bug

Indexing several files into a collection that does not exist yet loses one of them. The upload is accepted, the task ends `FAILED`, and the document is simply absent from the index. On a stack booted fresh per run it reproduced on **11 of 12 consecutive boots** (2 files, 2 upload workers).

The failing task's traceback:

```
File "services/workers/stages/store.py", line 35, in store_stage
  await vector_store.ensure_collection("default", len(embedding))
File "services/storage/milvus_store.py", line 1079, in ensure_collection
  await self.initialize(dimension)
File "services/storage/milvus_store.py", line 213, in _ensure_loaded
  self._check_schema_version()
VDBSchemaMigrationRequiredError: Collection `vdb_test` is at schema version 0
but the application requires version 1. Please perform the migration script.
```

The collection had been created seconds earlier, by the neighbouring worker.

## Why

`_ensure_loaded` created the collection and stamped its version in a *second* call:

```python
create_collection(...)      # collection exists; version property absent
_store_schema_version()     # ...stamped only here
```

Indexer workers are separate Ray processes (`IndexerWorkerActor` pids 840 and 848 in the log below), so `_load_lock` — an asyncio lock, per instance — does not serialise them against one another. A worker reaching `has_collection()` inside that window finds the collection, reads no version property, defaults to `0`, and raises "migration required".

The one boot in twelve that indexed both files is the one where both workers arrived after the stamp.

## The fix

Stamp the version in the request that creates the collection, so it never exists without one. `properties` reaches the CreateCollection RPC via `Prepare.create_collection_request`; verified against Milvus v3.0.0 that `describe_collection` returns it immediately after creation:

```
properties -> {'openrag.schema_version': '1', 'timezone': 'UTC'}
```

The same race has a second window: both workers can see `has_collection()` as false and both attempt the create. The loser's `MilvusException` was fatal to its file — it now re-checks, and if the collection is there validates it like any other existing collection. A create that fails *without* leaving a collection still raises `VDBCreateOrLoadCollectionError` unchanged.

`_store_schema_version` had no other caller and is removed; the Milvus migration scripts write the property themselves.

## Tests

`TestEnsureLoadedConcurrentCreation`, four cases: the version is in the create request and no `alter_collection_properties` follows; losing the create race validates the winner's collection instead of failing; a create failure that leaves no collection still raises; and a genuinely stale collection is still refused a migration bypass.

The two race tests fail without the source change and pass with it:

```
FAILED  test_a_new_collection_is_stamped_in_the_create_request
FAILED  test_losing_the_create_race_validates_the_winners_collection
2 failed, 2 passed
```

`tests/unit` is green (2543 passed), `ruff check` and `ruff format` clean.

## Note

The failure is invisible from the API's task list — the state is `FAILED` and the reason only appears via `/indexer/task/{id}/error`. Diagnosing this took reading the Milvus container's own logs, where the visible symptom is a `collection not loaded` delete that is actually the post-failure cleanup in `_cleanup_indexed_vectors`, not the cause.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple processes create the same vector collection simultaneously.
  * Concurrent creation races now correctly recognize an existing collection and continue loading it after validating its schema version.
  * Non-duplicate collection creation failures are surfaced immediately with their original error context.
  * Duplicate-creation errors now remain failures when the collection is still unavailable.
  * Existing collections with incompatible schema versions continue to require migration.

### DIFF
--- a/openrag/services/storage/milvus_store.py
+++ b/openrag/services/storage/milvus_store.py
@@ -244,11 +244,22 @@ class MilvusVectorStore(VectorStore):
                         properties={SCHEMA_VERSION_PROPERTY_KEY: str(self._config.schema_version)},
                     )
                 except MilvusException as e:
-                    raise VDBCreateOrLoadCollectionError(
-                        f"Failed to create collection `{self._collection_name}`: {e!s}",
-                        collection_name=self._collection_name,
-                        operation="create_collection",
-                    ) from e
+                    # A duplicate-create error means another worker may have
+                    # won the race after the initial existence check. Other
+                    # Milvus errors must keep their original failure state.
+                    if "already exist" not in str(e).lower():
+                        raise VDBCreateOrLoadCollectionError(
+                            f"Failed to create collection `{self._collection_name}`: {e!s}",
+                            collection_name=self._collection_name,
+                            operation="create_collection",
+                        ) from e
+                    if not self._client.has_collection(self._collection_name):
+                        raise VDBCreateOrLoadCollectionError(
+                            f"Failed to create collection `{self._collection_name}`: {e!s}",
+                            collection_name=self._collection_name,
+                            operation="create_collection",
+                        ) from e
+                    self._check_schema_version()
 
             self._wait_for_vector_indexes()
             try:
--- a/tests/unit/services/storage/test_milvus_store.py
+++ b/tests/unit/services/storage/test_milvus_store.py
@@ -1055,6 +1055,20 @@ class TestEnsureLoadedConcurrentCreation:
         assert properties == {SCHEMA_VERSION_PROPERTY_KEY: "1"}
         store._client.alter_collection_properties.assert_not_called()  # type: ignore[attr-defined]
 
+    def test_losing_the_create_race_validates_the_winners_collection(self, store: MilvusVectorStore) -> None:
+        """A worker that loses collection creation should continue indexing."""
+        store._embedding_dimension = 8
+        store._client.has_collection.side_effect = [False, True]
+        store._client.create_collection.side_effect = MilvusException(message="collection already exists")
+        store._client.describe_collection.return_value = {
+            "properties": {SCHEMA_VERSION_PROPERTY_KEY: "1"},
+            "fields": [{"name": "vector"}, {"name": "sparse"}],
+        }
+
+        store._ensure_loaded()
+
+        store._client.load_collection.assert_called_once_with(store._collection_name)
+
     def test_duplicate_collection_with_different_parameters_preserves_error(self, store: MilvusVectorStore) -> None:
         store._embedding_dimension = 8
         store._client.has_collection.return_value = False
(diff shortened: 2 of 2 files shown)


hedhoud:
Issue #862 is addressed by #907, which covers the concurrent collection initialization race. That PR is still awaiting merge, so the issue is not yet resolved on develop. We can close this PR in favor of #907 once it lands.

hedhoud (approved):

PR #866, opened 2026-08-27
fix(chunking): stop dropping real captions beside an image placeholder

andyne13:
Silent content loss in `recursive_splitter`, the current default chunker. Split out of #861, where it was found — it stands alone and needs no part of that PR.

## The bug

`BaseChunker` skipped an image block whenever `[Image Placeholder]` appeared **anywhere** in it:

```python
if element.type == "image" and _IMAGE_PLACEHOLDER_MARKER in element.content.lower():
    continue
```

The marker means the captioner found nothing useful in *one* image — but a single block can describe several. A marker sitting next to real caption text took that text with it, before the chunk was ever built. No error, no log; the content simply never reaches the index and is unretrievable.

## How it was found

By reassembling every chunk and diffing against the parsed source across a 30-document corpus — a check none of the existing cleanliness metrics can make, since a chunker that silently drops text scores well on all of them.

Two examples of what never reached the index:

- a 132-token chart caption — `92 000 COLLABORATEURS / 55% DU GROUP / 27 000 AVTOVAZ / 65 000 RENAULT GROUP / 16 PAYS …`
- 225 words of a tourism description — `tout Étretat est une destination riche en attractions naturelles et culturelles …`

## The fix

`is_placeholder_image()` skips a block only when nothing but the wrapper and the marker(s) remains. Anything with characters left is real caption text and is kept. Placeholder-only blocks are still dropped, so the noise this guard exists to remove is unaffected.

## Effect

Coverage against the source, marker corpus, `recursive_splitter`:

| | before | after |
|---|---|---|
| mean | 99.55% | **99.81%** |
| worst document | 95.16% | **99.10%** |
| documents below 99% | 3 | **0** |

## Notes

- Affects indexing only; no API or config change, nothing to migrate. Documents already indexed keep their current chunks until re-indexed.
- `structured_section` (#861) shares this code path and inherits the fix; that PR currently carries the same change and it will drop out of its diff once this merges.
- Three tests: placeholder-only blocks still skipped, a mixed block keeps its caption, and an end-to-end check that the caption survives chunking.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image filtering to skip only image blocks without meaningful caption text.
  * Preserved captions that include placeholder markers alongside descriptive content.
  * Removed placeholder markers while retaining meaningful image captions during content chunking.
  * Recognized punctuation-only captions as meaningful content while continuing to ignore whitespace-only captions.

* **Tests**
  * Added coverage for placeholder-only images and mixed image blocks with real captions.

### DIFF
--- a/openrag/core/chunking/recursive.py
+++ b/openrag/core/chunking/recursive.py
@@ -33,6 +33,31 @@ from core.utils.text import sanitize_text
 # they don't pollute the index.
 _IMAGE_PLACEHOLDER_MARKER = "[image placeholder]"
 
+# Strips the block's own wrapper so only the caption text is weighed.
+_IMAGE_BLOCK_TAG_RE = re.compile(r"</?image_description>", re.IGNORECASE)
+_IMAGE_PLACEHOLDER_RE = re.compile(re.escape(_IMAGE_PLACEHOLDER_MARKER), re.IGNORECASE)
+
+
+def is_placeholder_image(content: str) -> bool:
+    """True when an image block carries no caption beyond the placeholder.
+
+    The marker means the captioner found nothing useful in *one* image, but a
+    single block can describe several. Dropping the whole block on any
+    occurrence therefore deleted real caption text along with it: on the corpus
+    that cost a 132-token chart caption ("92 000 COLLABORATEURS…") and 225 words
+    of a tourism description, silently, from the index — and from *both*
+    chunkers, since they share this test.
+
+    So a block is only skipped when nothing but the wrapper and the marker(s)
+    remains. Anything with actual characters left is real caption text and is
+    kept.
+    """
+    if _IMAGE_PLACEHOLDER_MARKER not in content.lower():
+        return False
+    remainder = _IMAGE_PLACEHOLDER_RE.sub(" ", _IMAGE_BLOCK_TAG_RE.sub(" ", content))
+    return not any(char.isalnum() for char in remainder)
+
+
 # Tables/images smaller than this token count are inlined with surrounding
 # text rather than emitted as standalone chunks.
 _INLINE_ELEMENT_TOKEN_THRESHOLD = 100
@@ -228,7 +253,7 @@ class BaseChunker(ChunkingStrategy):
 
         for element in md_elements:
             if element.type in ("table", "image"):
-                if element.type == "image" and _IMAGE_PLACEHOLDER_MARKER in element.content.lower():
+                if element.type == "image" and is_placeholder_image(element.content):
                     continue
                 if self.length_function(element.content) <= _INLINE_ELEMENT_TOKEN_THRESHOLD:
                     texts.append(element)
--- a/openrag/core/chunking/recursive.py
+++ b/openrag/core/chunking/recursive.py
@@ -36,26 +36,47 @@ _IMAGE_PLACEHOLDER_MARKER = "[image placeholder]"
 # Strips the block's own wrapper so only the caption text is weighed.
 _IMAGE_BLOCK_TAG_RE = re.compile(r"</?image_description>", re.IGNORECASE)
 _IMAGE_PLACEHOLDER_RE = re.compile(re.escape(_IMAGE_PLACEHOLDER_MARKER), re.IGNORECASE)
+# Removing a marker leaves the blank lines that surrounded it.
+_COLLAPSE_BLANK_RE = re.compile(r"\n{3,}")
 
 
 def is_placeholder_image(content: str) -> bool:
     """True when an image block carries no caption beyond the placeholder.
 
-    The marker means the captioner found nothing useful in *one* image, but a
-    single block can describe several. Dropping the whole block on any
-    occurrence therefore deleted real caption text along with it: on the corpus
-    that cost a 132-token chart caption ("92 000 COLLABORATEURS…") and 225 words
-    of a tourism description, silently, from the index — and from *both*
-    chunkers, since they share this test.
+    ``wrap_caption`` emits one wrapper per image, so the shape that matters is
+    the marker and real caption text inside the *same* wrapper: a composite
+    figure where the VLM could describe some parts and not others. Dropping the
+    whole block on any occurrence of the marker deleted that caption text along
+    with it — on the corpus, a 132-token chart caption
+    ("92 000 COLLABORATEURS…") and 225 words of a tourism description,
+    silently, from the index, and from *both* chunkers since they share this
+    test. (Two *adjacent* wrappers were never affected: ``IMAGE_RE`` is
+    non-greedy, so they parse as separate elements and only the placeholder-only
+    one was dropped.)
 
     So a block is only skipped when nothing but the wrapper and the marker(s)
     remains. Anything with actual characters left is real caption text and is
-    kept.
+    kept — see ``strip_placeholder_markers`` for removing the marker from it.
     """
     if _IMAGE_PLACEHOLDER_MARKER not in content.lower():
         return False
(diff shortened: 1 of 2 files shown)


tests/unit/core/chunking/test_recursive.py
```diff
@@ -306,3 +306,47 @@ def test_dewrap_preserves_code_fence_verbatim():
...
+                    "<image_description>\n\n92 000 COLLABORATEURS 27 000 AVTOVAZ\n\n</image_description>\n\n"
+                    "Closing paragraph."
+                ),
+                page_number=1,
+            )
+        ],
+        metadata={"source": "test.md"},
+    )
+    joined = "\n".join(c.text for c in splitter.chunk(doc, partition="p"))
+    assert "92 000 COLLABORATEURS" in joined, "real caption text dropped alongside the placeholder"
```
Ahmath-Gadji, line 355: **The end-to-end test passes on the unfixed code — it can't catch this bug coming back**

Revert only the call site (`recursive.py:256`) to the old `_IMAGE_PLACEHOLDER_MARKER in element.content.lower()`, keep this whole test file, and all 22 tests still pass. I ran it on `e44b3a0`.

The reason is `IMAGE_RE` in `markdown_utils.py:39` — it is non-greedy, so the two adjacent `<image_description>` blocks in this fixture never form one element:

```
--- ADJACENT (this fixture), pre-fix
    text  'Intro paragraph.'
    image '<image_description>\n\n[Image Placeholder]\n\n</image_description>'
    image '<image_description>\n\n92 000 COLLABORATEURS 27 000 AVTOVAZ\n\n</image_...'
    text  'Closing paragraph.'
    caption kept: True      <-- already true before the fix
```

The old code skipped the placeholder-only element and kept the caption element on its own. So neither `test_mixed_image_block_survives_chunking` nor `test_image_block_keeps_real_caption_beside_a_placeholder` (same shape, two wrappers) describes a case that was ever broken.

The shape that actually regresses is the marker and the caption inside **one** wrapper — which, given `wrap_caption()` emits one wrapper per image, is also the only shape the bug can take in practice (a VLM caption for a composite image that placeholders one part and describes the rest):

```
--- SINGLE_WRAPPER, pre-fix
    image '<image_description>\n\n[Image Placeholder]\n\n92 000 COLLABORATEURS 27 0...'
    caption kept: False     <-- the bug
--- SINGLE_WRAPPER, post-fix
    caption kept: True
```

Suggested fixture for `test_mixed_image_block_survives_chunking` (fails pre-fix, passes post-fix):

```python
"<image_description>\n\n[Image Placeholder]\n\n92 000 COLLABORATEURS 27 000 AVTOVAZ\n\n</image_description>\n\n"
```

Worth applying the same change to the two-wrapper string in `test_image_block_keeps_real_caption_beside_a_placeholder`, and rewording the "a single block can describe several" line in the `is_placeholder_image` docstring — as written it points the next reader at adjacent blocks, which were never affected.

openrag/core/chunking/recursive.py
```diff
@@ -33,6 +33,31 @@
...
+    of a tourism description, silently, from the index — and from *both*
+    chunkers, since they share this test.
+
+    So a block is only skipped when nothing but the wrapper and the marker(s)
+    remains. Anything with actual characters left is real caption text and is
+    kept.
+    """
+    if _IMAGE_PLACEHOLDER_MARKER not in content.lower():
+        return False
+    remainder = _IMAGE_PLACEHOLDER_RE.sub(" ", _IMAGE_BLOCK_TAG_RE.sub(" ", content))
```
Ahmath-Gadji, line 66: **The kept block now carries `[Image Placeholder]` into the chunk text**

Right call to keep the caption, but the block is passed through verbatim, so the marker the guard exists to suppress now reaches the index. Same probe on `e44b3a0`, single-wrapper mixed block:

```
pre-fix :  caption kept: False   marker in chunks: False
post-fix:  caption kept: True    marker in chunks: True
```

`milvus_store.py:127-129` has the marker in the BM25 stop-word list, so lexical scoring is unaffected — but the dense embedding is computed over the chunk text, and the chunk is what `format_context()` puts in front of the LLM. So a mixed block now embeds and displays `[Image Placeholder]` as if it were caption content.

The function already computes exactly the string you'd want to keep. Something like:

```python
def strip_placeholder_markers(content: str) -> str:
    """Drop the marker(s) from a block that also carries real caption text."""
    return _IMAGE_PLACEHOLDER_RE.sub("", content)
```

applied at the call site when `is_placeholder_image()` returns `False` (or fold both into one helper returning `str | None` — `None` to skip, otherwise the cleaned content). Keeps the caption, keeps the noise out, and makes the guard's original intent hold for the mixed case too.

Note this also changes the inline-vs-standalone decision at `recursive.py:257`: the marker's tokens currently count toward `_INLINE_ELEMENT_TOKEN_THRESHOLD`, so stripping first is the more correct measurement as well.

tests/unit/core/chunking/test_recursive.py
Ahmath-Gadji: Fixed in 97238d61789ae6876d173244c7ac133e3ea1bac1 — verified.

Both fixtures now put the marker and the caption in one wrapper, and the end-to-end test earns its keep: reverting `_prepare_md_elements` to the old `_IMAGE_PLACEHOLDER_MARKER in element.content.lower()` rule on the new head gives

```
FAILED test_mixed_image_block_survives_chunking
E   AssertionError: real caption text dropped alongside the placeholder
E   assert '92 000 COLLABORATEURS' in 'Intro paragraph.\n\nClosing paragraph.'
1 failed, 21 passed
```

where it was 22/22 green before. The docstring reword covers the adjacent-wrapper case explicitly too. 55/55 in `tests/unit/core/chunking/` pass on `97238d6`.

openrag/core/chunking/recursive.py
Ahmath-Gadji: Fixed in 97238d61789ae6876d173244c7ac133e3ea1bac1 — verified.

`strip_placeholder_markers` runs on kept blocks before the size test, so the marker no longer reaches the chunk text and the inline-vs-standalone decision measures the caption. The new `assert "[Image Placeholder]" not in joined` in `test_mixed_image_block_survives_chunking` locks it in, and `_COLLAPSE_BLANK_RE` handles the blank lines the removal leaves behind.

One note on CodeRabbit's follow-up about `any(char.isalnum())` on line 64: I'd leave it. Its case is a block whose only content beside the marker is something like `©` or `—`, and `not remainder.strip()` would keep that — but stripping the marker then leaves a chunk of bare punctuation in the index, which is the noise this guard exists to remove. The alnum test reads as deliberate; worth a short reply on that thread rather than a change.

Ahmath-Gadji (approved):
Approved.

Both review findings are addressed in 97238d6 and I verified each on that head:

- The end-to-end test now fails on the pre-PR rule (`22/22 green` → `1 failed, 21 passed`, `test_mixed_image_block_survives_chunking`), so it actually defends the fix. Fixtures use the one-wrapper shape, which is the only shape the bug can take given `wrap_caption()`.
- `strip_placeholder_markers()` keeps `[Image Placeholder]` out of the chunk text, and running it before the size test makes the inline-vs-standalone decision measure the caption rather than the marker.

`tests/unit/core/chunking/` is 55/55 locally and all checks are green.

On CodeRabbit's open point about `any(char.isalnum())`: I'd leave it as is — `not remainder.strip()` would keep a block whose only content beside the marker is something like `©`, and after stripping the marker that lands a chunk of bare punctuation in the index, which is the noise the guard exists to remove. Not blocking either way.

PR #871, opened 2026-08-28
fix(vectordb): make BM25 lexical search case-insensitive

### LINKED ISSUE #870: BM25 lexical search is case-sensitive: the custom analyzer drops the lowercase filter
## Summary

The BM25 (sparse) leg of hybrid search is **case-sensitive**: `Rapport` and `rapport` are two different terms. A query typed in a different case than the source document scores 0 on the lexical leg, so hybrid retrieval silently degrades to dense-only for that term.

## Root cause

`analyzer_params` in [`openrag/services/storage/milvus_store.py`](../blob/develop/openrag/services/storage/milvus_store.py#L121-L138) declares a **custom** analyzer:

```python
analyzer_params = {
    "tokenizer": "standard",          # <- custom analyzer, not the built-in one
    "filter": [{"type": "stop", "stop_words": [...]}],
}
```

Milvus's built-in `standard` **analyzer** is defined as `{"tokenizer": "standard", "filter": ["lowercase"]}` ([Standard Analyzer](https://milvus.io/docs/standard-analyzer.md)). Writing `"tokenizer": "standard"` instead of `"type": "standard"` selects only the *tokenizer* and opts out of the whole built-in filter chain — a custom analyzer applies only the filters you list ([Analyzer Overview](https://milvus.io/docs/analyzer-overview.md)). The `lowercase` filter went with it.

Introduced in cef0e878 (Oct 2025) when the stop words were added, and carried over verbatim into the Milvus 3 store.

## Evidence

Verified with `run_analyzer` against Milvus 3.0.0:

```
analyzer as shipped  -> ['Le', 'Président', 'PARIS', 'paris', 'The', 'AND', 'rapport', 'Rapport']
builtin {"type": "standard"} -> ['le', 'président', 'paris', 'paris', 'the', 'and', 'rapport', 'rapport']
shipped + "lowercase"        -> ['président', 'paris', 'paris', 'rapport', 'rapport']
```

The same analyzer runs on the query at search time as on the text at insert time ([Full Text Search](https://milvus.io/docs/full-text-search.md)), so index and query must agree — there is no query-side normalization compensating for this.

## Secondary finding: the stop words never fire

The standard tokenizer splits on `<`, `>`, `[`, `]` and `_`, so no token can ever equal `"<image_description>"`:

```
'<image_description>a cat</image_description>' -> ['image', 'description', 'cat', 'image', 'description']
'[CHUNK_START] hello [CHUNK_END] [CONTEXT] x'  -> ['CHUNK', 'START', 'hello', 'CHUNK', 'END', 'CONTEXT', 'x']
```

Only `_english_` / `_french_` do anything, and today only on tokens that are already lowercase (`'le the and de rapport'` -> `['rapport']`, while `The` / `Le` / `AND` above survive). The marker entries tokenize into ordinary words (`image`, `context`, `start`), so extending the stop list is *not* the fix — it would strip real content. They are left inert; the real fix, if wanted, is to keep the markers out of the indexed text.

## Impact

- Lexical recall loss on any case mismatch — capitalized entities (`PARIS`, `Président`), sentence-initial words, acronyms, headings.
- Stop-word removal is inconsistent: `the` is stripped, `The` is not, which also skews BM25 IDF.
- Dense retrieval is unaffected, which is why this has gone unnoticed.

## Fix

Put `lowercase` ahead of the stop filter. Order matters — the built-in stop-word lists are lowercase:

```
lowercase then stop: ['rapport', 'rapport']
stop then lowercase: ['the', 'le', 'and', 'rapport', 'rapport']
```

## Migration

`analyzer_params` is **not** alterable after collection creation — `alter_collection_field` only exposes `max_length`, `max_capacity` and `mmap.enabled` ([Alter Collection Field](https://milvus.io/docs/alter-collection-field.md)). Applying the fix to an existing deployment therefore means rebuilding the collection, gated behind a `vectordb.schema_version` bump (1 -> 2) and a migration script.

Re-embedding is *not* required: dense vectors can be copied verbatim and the sparse vectors are regenerated server-side by the BM25 `Function` on insert.

Ahmath-Gadji:
Closes #870.

`analyzer_params` declared a **custom** analyzer (`"tokenizer": "standard"` plus an explicit filter chain) instead of the built-in `{"type": "standard"}`. A custom analyzer inherits nothing, so this silently dropped the `lowercase` filter the built-in one bundles: `Rapport` and `rapport` were distinct BM25 terms, and since Milvus applies the same analyzer to the query, a case mismatch scored zero on the lexical leg of hybrid search. Dense retrieval was unaffected, which is why it went unnoticed. Present since cef0e878.

Fix: `lowercase` first in the chain — order matters, the `_english_` / `_french_` stop-word lists are lowercase, so today `the` is stripped but `The` is not. The marker stop words (`<image_description>`, `[CHUNK_START]`, ...) are left alone and documented as inert: the tokenizer splits them, and their split forms are ordinary words that would strip real content.

## Migration

`analyzer_params` can't be altered while `enable_match` is set (and `enable_match` is immutable), and `add_function_field` — which would re-attach BM25 and backfill the sparse column in place — needs Storage V3, off by default. Both probed against Milvus 3.0.0; transcript in the comment below.

So `schema_version` goes 1 → 2 and `2.lowercase_bm25_analyzer.py` copies the rows into `<collection>_v2_rebuild`, verifies the counts and swaps the names, keeping the old collection as `<collection>_v1_backup`. Nothing is re-embedded — dense vectors are copied verbatim, sparse ones regenerated server-side on insert. Dense-only deployments are only re-stamped. Chunk `_id`s change (`auto_id`); Postgres keys on `file_id`, so only MCP chunk IDs handed out earlier go stale.

## Testing

2542 unit tests, ruff, layer-import guard. New unit tests cover the filter, its ordering and the wiring onto `text`. The migration was run end to end against a real Milvus 3.0.0 on a throwaway 25-row collection with dynamic fields, nulls and a `TIMESTAMPTZ`:

| | rows | `bm25("rapport")` | `bm25("RAPPORT")` | `bm25("Le")` |
|---|---|---|---|---|
| v1 | 25 | 12 | 0 | 13 |
| upgraded | 25 | 25 | 25 | 0 |
| downgraded | 25 | 12 | 0 | 13 |

Dry-run changed nothing, re-running the upgrade was a no-op, rollback restored the collection.

**Where the mismatch bites:** the check lives in `ensure_collection`, which only the indexing store stage calls — so an un-migrated deployment still boots and still answers searches (with the old analyzer), and fails the first upload with `Collection ... is at schema version 1 but the application requires version 2`. Verified against a live v1 collection. The rebuild, its ~2× disk need and the backup cleanup are documented in `milvus_migration.mdx`.

### DIFF
--- a/conf/config.yaml
+++ b/conf/config.yaml
@@ -66,7 +66,7 @@ vectordb:
   hybrid_search: true
   enable: true
   timeout: 120.0 # per-request timeout (s) for the Milvus sync/async clients
-  schema_version: 1 # Increment when the collection schema changes and a migration is required
+  schema_version: 2 # Increment when the collection schema changes and a migration is required
 
 # --- Relational Database (PostgreSQL) ---
 # Env: POSTGRES_HOST, POSTGRES_PORT, POSTGRES_USER, POSTGRES_PASSWORD,
--- a/docs/content/docs/documentation/milvus_migration.mdx
+++ b/docs/content/docs/documentation/milvus_migration.mdx
@@ -238,6 +238,40 @@ Migrations are versioned. The runner reads the current schema version stored in 
 Stop the OpenRAG application before running any migration.
 :::
 
+### Version 2 — case-insensitive BM25 analyzer
+
+Schema version 2 adds the `lowercase` filter to the `text` analyzer, so BM25 stops treating `Rapport` and `rapport` as different terms. Milvus applies the analyzer to the query as well as to the indexed text, so before this fix a case mismatch scored zero on the lexical leg.
+
+It **rebuilds** the collection rather than altering it: Milvus 3.0 refuses to change `analyzer_params` while `enable_match` is set (and `enable_match` is itself immutable), and [attaching a BM25 function](https://milvus.io/docs/add-fields-to-an-existing-collection.md) to an existing collection — which would backfill the sparse column — requires [Storage V3](https://milvus.io/docs/storage-v3.md), off by default. The migration copies the rows into `<collection>_v2_rebuild`, checks the counts, then swaps the names, keeping the old collection as `<collection>_v1_backup`.
+
+:::caution[Plan for the rebuild]
+- **Disk** — both copies exist at once; allow roughly twice the collection size until the backup is dropped.
+- **No re-embedding** — dense vectors are copied verbatim and the sparse ones are regenerated server-side on insert, which is what applies the new analyzer.
+- **Chunk IDs change** — `_id` is `auto_id`. Postgres keys on `file_id`, but a chunk ID handed out earlier by the MCP `get_chunk_by_id` tool no longer resolves.
+- **`hybrid_search: false`** deployments have no BM25 leg and are only re-stamped — no rebuild.
+- A failed copy leaves the original untouched; drop the partial `<collection>_v2_rebuild` and retry.
+- The version check runs on the indexing path only. An un-migrated deployment still boots and still answers searches — with the old analyzer — and fails the first upload with `Collection ... is at schema version 1 but the application requires version 2`.
+:::
+
+Once validated, drop the backup:
+
+```python
+from pymilvus import MilvusClient
+
+MilvusClient(uri="http://localhost:19530").drop_collection("<collection>_v1_backup")
+```
+
+:::note[Where to run these]
+The commands below run from `infra/compose/`. From the repo root, point Compose at the file (and at your project name, if you use one):
+
+```bash
+docker compose -p <project> -f infra/compose/docker-compose.yaml run --no-deps --rm --build --entrypoint "" openrag \
+    uv run python services/persistence/migrations/milvus/migrate.py --dry-run
+```
+
+The migration runner is a file, not a package — pass the full path to `migrate.py`, and keep Milvus running.
+:::
+
 ### Step 1 — Start only the Milvus container
 
 ```bash
@@ -267,7 +301,7 @@ docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-
   </TabItem>
 </Tabs>
 
-Review the output to confirm which migrations are pending and what changes they would apply.
+Review the output to confirm which migrations are pending and what changes they would apply. The migration scripts ship **inside the image**, which is why `--build` is there: if the `Discovered N migration(s)` line does not list the migration you expect, the image is stale — and on a shared host another checkout may have rebuilt the same `linagoraai/openrag` tag, so add `--build` to the apply step too.
 
 ### Step 3 — Apply all pending migrations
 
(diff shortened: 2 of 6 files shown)


Ahmath-Gadji:
### Why the rebuild, given Milvus 3.0's schema evolution

Fair challenge — Milvus 3.0 is much less rigid here than 2.x, and my first description ("the analyzer is immutable") was too strong. pymilvus 3.0.1 exposes `alter_collection_field`, `drop_function_field` and `add_function_field`, and the docs confirm you can [attach a BM25 function to an existing collection](https://milvus.io/docs/add-fields-to-an-existing-collection.md) with the output field backfilled over existing rows: *"Milvus generates the Function output for new entities as they are written. For existing entities, Milvus populates the generated vector field asynchronously through background compaction."*

So I probed both in-place routes against Milvus 3.0.0 before keeping the rebuild.

**1. Alter the analyzer in place** — blocked by `enable_match`:

```
alter_collection_field(c, "text", {"analyzer_params": ...})
  -> code=1100 "can not alter analyzer params for field text after text match is
     enabled or a BM25/MinHash function depends on it"

drop_function_field(c, "text_bm25_emb")          -> OK (drops function + sparse + index)
alter_collection_field(c, "text", {"analyzer_params": ...})
  -> same error: enable_match is still true

alter_collection_field(c, "text", {"enable_match": False})
  -> code=1100 "enable_match does not allow update in collection field param"
```

Dropping the function clears half the condition — on a collection created *without* `enable_match` the alter then succeeds (passing `analyzer_params` as a JSON string; a dict fails validation). But `enable_match` is immutable, and every collection OpenRAG has created sets it. The live `vdb_test` confirms it:

```
text params: {'max_length': 65535, 'enable_match': 'true', 'enable_analyzer': 'true',
              'analyzer_params': '{"tokenizer":"standard","filter":[{"type":"stop", ...}]}'}
```

**2. Re-attach the BM25 function and let it backfill** — blocked by Storage V3:

```
add_function_field(c, field_schema=sparse, func=BM25(text -> sparse), index_params=...)
  -> code=1100 "adding a function field requires StorageV3; enable common.storage.useLoonFFI"
```

[Storage V3](https://milvus.io/docs/storage-v3.md) is off by default in Milvus 3.0, and enabling it changes the serialized data format irreversibly (no rollback to 2.6). Our compose deployment does not enable it.

The in-place path therefore needs *both* a collection that never enabled text match *and* a Storage-V3 deployment. Neither holds for collections already in the field, so the row copy stays. The reasoning is now recorded in the migration docstring and the migration guide rather than living in a PR thread.

### Follow-up worth deciding separately

`enable_match=True` is set on `text` but nothing in the codebase issues a `TEXT_MATCH` query. Dropping it from the schema would make future analyzer changes alterable in place (no rebuild), and would shrink the text-match inverted index. The catch: it is reachable today through the raw `expr` escape hatch in `_build_filter_expr`, so removing it is a silent behaviour change for anyone using `TEXT_MATCH` that way. Happy to do it here (the collection is being rebuilt anyway, so it costs nothing extra) or leave it — your call.

PR #877, opened 2026-08-31
feat(ui): mark new features with a badge that expires on its own

Ahmath-Gadji:
## Why

A "NEW" marker is only useful if someone removes it. Nobody does — it becomes permanent decoration, and the cleanup lands in a release that has no other reason to touch the file.

So the marker is data. `ui/src/lib/whats-new.ts` maps a feature key to the version it shipped in, and the marker stops rendering `NEW_FOR_MINORS` (2) minors later. Forgetting to delete an entry costs nothing: it expires on its own. Entries never interact — each answers the same independent question, *is the running version inside the window of the version I shipped in?*

## What ships here

The mechanism, its guide and its tests — **deliberately without a call site**. `NEW_SINCE` is empty and no page renders a marker yet.

The first candidate, `chunking.structured_section`, belongs to #861, which isn't on `develop`; registering it here would badge nothing (the form takes `chunking_strategies` from the API, so an option the backend doesn't offer renders none) and would tie this PR's merge order to that one. An empty registry is also the state the design keeps returning to, every time entries expire and get swept up — so it is worth supporting and testing, not a gap. Adoption is one registry line plus one hook call.

| file | role |
|---|---|
| `ui/src/lib/whats-new.ts` | the registry, the window rule, and nothing about the UI |
| `ui/src/components/shared/new-badge.tsx` | `NewBadge`, `NewDot`, `useNewOptions`, `useIsNew` |
| `FEATURE_TAG.md` | the guide: concept, the two markers, worked examples per surface |
| `CONTRIBUTING.md` | the release-time command that resolves pending entries |
| `.github/workflows/build.yml` | the gate that blocks GA images with unresolved entries |

## How it behaves

For an entry pinned to `2.2.0`:

| app reports | marker | why |
|---|---|---|
| 2.1.1 | ✅ | predates the feature — the option isn't there anyway |
| 2.2.0 → 2.3.0 | ✅ | inside the window |
| 2.4.0 | ❌ | window passed |
| 3.0.0 | ❌ | a major bump always expires it |
| `UNRELEASED` entry | ✅ | merged but untagged — new to everyone running that build |
| unreadable / not a version | ❌ | fails closed |

Failing closed is deliberate: one unbadged feature beats every feature badged forever on a deployment whose `/version` can't be read. "Not a version" means the whole string — `2.2rubbish`, `2.2.0.wat` and out-of-order PEP 440 (`2.2.0.post1a1`) are rejected, not read as 2.2. What *is* accepted is wider than SemVer, because `/version` comes from `importlib.metadata` and reports PEP 440: `2.2.0rc1`, `2.1.1.post1`, `2.2.0.dev3` all compare, so badges don't blank out on an RC deployment.

Keys are `"<group>.<value>"` for an option inside a control — two dropdowns can offer the same string, and a bare key would badge both — and any stable dotted string elsewhere (`nav.workspaces`).

## Two markers

The **word** (`NEW` pill) goes on the thing itself: an option in an open list, a tab, a nav item. The **dot** (small amber circle) goes on a *collapsed* control's label, where the new thing is out of sight. A dropdown needs both — the dot gets the reader to open the list, the word tells them what they came for. The dot carries an accessible name, and is amber because a crimson dot beside a form label reads as *required*.

Radix mirrors a selected option's children into the closed trigger, which would show a second marker there; the marker hides that copy itself, which is what keeps it a plain child on any surface.

## Release step and gate

A feature PR can't know its release number — the release branch chooses it — so entries are authored as `UNRELEASED` and resolved by that branch, in the commit that bumps `pyproject.toml`, reading the version back from it (**CONTRIBUTING.md → Releases**).

An unresolved entry badges forever and silently, so `verify-tag` refuses to publish GA images while any entry is unresolved, and fails loudly when the registry is missing or unreadable rather than reporting success. It covers GA tags only; `develop`, RC and self-built images badge `UNRELEASED` entries freely, which is what `UNRELEASED` means — FEATURE_TAG.md says so.

## Testing

31 tests across `whats-new.test.ts` (window rule, `UNRELEASED`, malformed and PEP 440 versions, the registry lookup against fixtures) and `new-badge.test.tsx` (attachment on any surface, the dot's accessible name, the group hook, and the documented dropdown example rendered as written, including the Radix trigger case).

The lookup is tested through `isNewInRegistry` / `hasNewInRegistry`, which take the registry as an argument, rather than by reaching into `NEW_SINCE` for a key — the registry is legitimately empty, and that is exactly when a regression there would go unnoticed. No test skips itself.

Full UI suite: **243 passed, 0 skipped** (29 files). `tsc --noEmit`, eslint and `npm run build` clean. The release command and the gate were verified by extracting them verbatim from the files and running them against fixtures — entries with and without a trailing comma, a missing `version` line, and a missing or unreadable registry.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,54 @@ jobs:
             exit 1
           fi
 
+      # A NEW badge is data, not markup: each entry in ui/src/lib/whats-new.ts
+      # records the version its feature shipped in, and the badge expires on its
+      # own once the running version is far enough past it. An entry still
+      # reading UNRELEASED has no version to expire against, so it badges every
+      # user forever — and silently, because nobody inspects a dropdown
+      # expecting an option *not* to say NEW. Resolving them belongs to the
+      # release branch, in the same commit that bumps pyproject.toml (see
+      # "Releases" in CONTRIBUTING.md). This is the gate that turns forgetting
+      # that step into a failed release rather than a permanent badge.
+      - name: Verify every NEW badge has been resolved to a version
+        env:
+          TAG_NAME: ${{ github.ref_name }}
+        run: |
+          registry=ui/src/lib/whats-new.ts
+
+          # A registry we cannot read is not a clean registry. Asked as an `if`
+          # condition, grep's "file missing" (exit 2) is indistinguishable from
+          # "no matches" (exit 1), so a moved or renamed file would report OK
+          # here and the backstop would be gone without anyone noticing.
+          if [ ! -r "$registry" ]; then
+            echo "::error::$registry is missing or unreadable, so the NEW badge gate cannot run. If the registry moved, update this step."
+            exit 1
+          fi
+
+          # Anchored to a whole registry line — indent, quoted key, colon, and
+          # an optional trailing comma, which the last entry in the object may
+          # legally omit — so the sentinel's own definition and the prose that
+          # explains it cannot trip the gate, and an entry written without the
+          # comma cannot slip past it.
+          set +e
+          unresolved=$(grep -nE '^ *"[^"]+": *"UNRELEASED",? *$' "$registry")
+          status=$?
+          set -e
+
+          if [ "$status" -eq 0 ]; then
+            echo "::error::$registry has unresolved NEW badge entries; they would badge forever. Set them to ${TAG_NAME#v} on the release branch (CONTRIBUTING.md -> Releases) before tagging."
+            echo "$unresolved"
+            exit 1
+          fi
+
+          # Only exit 1 means a genuine no-match; anything else is grep failing.
+          if [ "$status" -ne 1 ]; then
+            echo "::error::could not search $registry for unresolved NEW badge entries (grep exited $status)."
+            exit 1
+          fi
+
+          echo "OK: no unresolved NEW badge entries"
+
   build-and-push-image:
     needs: verify-tag
     runs-on: ubuntu-latest
--- a/ui/src/lib/whats-new.ts
+++ b/ui/src/lib/whats-new.ts
@@ -1,0 +1,179 @@
+/**
+ * "NEW" badge registry.
+ *
+ * A badge is data, not markup, so that removing it is never a manual step that
+ * someone has to remember in a release which otherwise doesn't touch the file.
+ * Each entry records the version a feature shipped in; `isNew` compares that
+ * against the running app version and stops returning true once the feature is
+ * `NEW_FOR_MINORS` minor releases old. Forgetting to delete an entry therefore
+ * costs nothing — the badge disappears on its own, and the line can be swept up
+ * whenever anyone happens to notice.
+ *
+ * A key is any stable dotted string — `nav.workspaces`, `presets.tab.retrieval`,
+ * `chunking.structured_section` — and nothing here knows what surface it names.
+ * For an option inside a control the convention is `"<group>.<value>"` rather
+ * than the bare option value, so that `useNewOptions(group, values)` can derive
+ * every key from the group alone, and so that two dropdowns offering the same
+ * string do not badge each other the day that happens.
+ *
+ * Attaching one is components/shared/new-badge.tsx. FEATURE_TAG.md, at the repo
+ * root, is the worked guide: registering a key, the two ways to attach it, and
+ * what the release branch does with it.
+ */
+
+/** How many minor releases a feature keeps its badge. */
+export const NEW_FOR_MINORS = 2;
+
+/**
+ * Placeholder for a feature that has merged but is not in a tagged release yet.
+ *
+ * The version a feature ships in is not knowable when the feature PR is
+ * written: a `release/X.Y.Z` branch is cut once the features intended for that
+ * version are complete, so the number is chosen after the feature merges —
+ * sometimes weeks later. Any pin written during development is a prediction,
+ * and wrong the moment the release plan moves. Author entries with this
+ * sentinel instead, and let the release branch resolve them in the same commit
+ * that bumps `pyproject.toml`, reading the version back from there rather than
+ * retyping it. See "Releases" in CONTRIBUTING.md for the command; it is kept
+ * there alone so the two cannot drift.
+ *
+ * Entries are written as the string literal, not as a reference to the constant
+ * below, so that a one-liner can find them. That one-liner anchors on a whole
+ * registry line — indent, quoted key, colon, and an optional trailing comma,
+ * since the last entry in the object may legally omit it — so it matches
+ * nothing else. A looser pattern silently corrupts two things at once: this
+ * constant's own definition, turning the sentinel into a version and breaking
+ * every future entry, and any prose that mentions the sentinel while explaining
+ * the step.
+ *
+ * An unresolved entry badges unconditionally, which is correct — everyone
+ * running a build that contains the feature is seeing it for the first time —
+ * but it also never expires. Forgetting the step is therefore silent and
+ * permanent, so `verify-tag` in .github/workflows/build.yml refuses to publish
+ * GA images while any entry is still unresolved.
+ */
+export const UNRELEASED = "UNRELEASED";
+
+/**
+ * Feature key -> the app version it first shipped in (`major.minor[.patch]`),
+ * or `UNRELEASED` until a release pins it.
+ */
+export const NEW_SINCE: Readonly<Record<string, string>> = {
+  // Nothing is marked right now. Entries look like this — see FEATURE_TAG.md:
+  //   "chunking.structured_section": "UNRELEASED",
+  //   "models.reranker": "2.4.0",
+  // A commented example is invisible to both the release one-liner and the CI
+  // gate: each anchors on a whole registry line, which starts with the quoted
+  // key and nothing else.
+};
+
+type Version = { major: number; minor: number };
+
+/**
+ * Parse `major.minor` out of a version string. Returns null for anything
+ * unparseable so callers fail closed (no badge) rather than badging everything.
+ *
+ * The whole string has to be a version, not merely start like one: a prefix
+ * match reads `2.2rubbish` as `2.2` and badges on it, which is the opposite of
+ * failing closed.
+ *
+ * What counts as a version is wider than SemVer, because `/version` serves
+ * `importlib.metadata`, which reports **PEP 440**: a release candidate arrives
+ * as `2.2.0rc1`, not `2.2.0-rc.1`, and `.post1` / `.dev0` builds are legal too.
+ * Rejecting those would blank every badge on an RC deployment.
+ *
+ * Each accepted suffix is spelled out rather than admitted as "some dotted
+ * thing after the digits" — that shortcut lets `2.2.0.wat` through and reads it
+ * as 2.2, which is the prefix hole again, one dot along. In order:
+ *
+ *   1. numeric segments — `2.2`, `2.2.0`, or more;
+ *   2. PEP 440 markers in their defined order — at most one pre-release
+ *      (`a` `b` `rc`), then at most one `post`, then at most one `dev`, each
+ *      attached or separated by `.` `-` `_` (`2.2.0rc1`, `2.1.1.post1`,
+ *      `2.2.0a1.post2.dev3`). Ordering them is what makes the grammar reject
+ *      `2.2.0.dev1.post1.dev5` and `2.2.0.post1a1`, which are version-shaped but
+ *      not versions;
+ *   3. an optional SemVer pre-release — `-` then dot-separated alphanumerics,
+ *      which is how a hand-written registry pin spells `2.2.0-rc.1`;
+ *   4. an optional `+` build or local segment (`+build.7`, `+local.7`).
+ */
+const VERSION_PATTERN =
+  /^\s*v?(\d+)\.(\d+)(?:\.\d+)*(?:[._-]?(?:a|b|rc)\d*)?(?:[._-]?post\d*)?(?:[._-]?dev\d*)?(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?\s*$/i;
+
+function parseVersion(raw: string | undefined | null): Version | null {
+  if (!raw) return null;
+  const match = VERSION_PATTERN.exec(raw);
+  if (!match) return null;
+  return { major: Number(match[1]), minor: Number(match[2]) };
+}
+
+/**
+ * Whether `featureKey` should still be badged when the app reports
+ * `appVersion`.
+ *
+ * False when either version is unknown or unparseable: an unbadged new feature
+ * is a much smaller problem than every feature being badged forever on a
+ * deployment whose `/version` we can't read.
+ *
+ * A major bump always expires the badge — 3.0 is not "two minors after 2.2" in
+ * any sense a reader would accept.
+ */
+export function isNew(featureKey: string, appVersion: string | undefined | null): boolean {
+  return isNewInRegistry(NEW_SINCE, featureKey, appVersion);
+}
+
+/**
+ * The lookup itself, over any registry — `isNew` is this against `NEW_SINCE`.
+ *
+ * Separate for the same reason `isNewSince` is separate from the registry: the
+ * registry is *data*, and legitimately empty between the day the last entry
+ * expires and the day the next feature registers one. A test that reaches into
+ * `NEW_SINCE` for a key to try can only skip itself on those days — which is
+ * precisely when a regression in the lookup would go unnoticed. Given a
+ * registry, this can be tested against a fixture on any day.
+ */
+export function isNewInRegistry(
+  registry: Readonly<Record<string, string>>,
+  featureKey: string,
+  appVersion: string | undefined | null
+): boolean {
+  return isNewSince(registry[featureKey], appVersion);
+}
+
+/**
+ * The window rule itself, independent of the registry — `isNew` is this plus a
+ * lookup. Separate so the rule can be tested against arbitrary versions without
+ * the tests having to be edited every time a registry entry is added, pinned by
+ * a release, or retired.
+ */
+export function isNewSince(since: string | undefined, appVersion: string | undefined | null): boolean {
+  if (since === undefined) return false;
+  // Merged but not yet tagged: new to everyone running a build that has it, and
+  // there is no version to compare against — not even the reader's own, since
+  // they are ahead of the last release rather than behind it.
+  if (since === UNRELEASED) return true;
+  const from = parseVersion(since);
+  const current = parseVersion(appVersion);
+  if (!from || !current) return false;
+  if (current.major !== from.major) return current.major < from.major;
+  return current.minor - from.minor < NEW_FOR_MINORS;
+}
+
+/**
+ * Whether any of `values` is still new within `group` — so a collapsed control
+ * can carry the badge for options the reader would otherwise have to open a
+ * dropdown to discover.
+ */
+export function hasNewIn(group: string, values: string[], appVersion: string | undefined | null): boolean {
+  return hasNewInRegistry(NEW_SINCE, group, values, appVersion);
+}
+
+/** `hasNewIn` over any registry — see `isNewInRegistry` for why it is separate. */
+export function hasNewInRegistry(
+  registry: Readonly<Record<string, string>>,
+  group: string,
+  values: string[],
+  appVersion: string | undefined | null
+): boolean {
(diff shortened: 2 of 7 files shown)


ui/src/lib/whats-new.ts
```diff
@@ -0,0 +1,125 @@
...
+ * or `UNRELEASED` until a release pins it.
+ */
+export const NEW_SINCE: Readonly<Record<string, string>> = {
+  // Nothing is marked right now. Entries look like this — see FEATURE_TAG.md:
+  //   "chunking.structured_section": "UNRELEASED",
+  //   "models.reranker": "2.4.0",
+  // A commented example is invisible to both the release one-liner and the CI
+  // gate: each anchors on a whole registry line, which starts with the quoted
+  // key and nothing else.
+};
```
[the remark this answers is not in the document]
Ahmath-Gadji: Deliberate — this PR ships the mechanism, not an adoption of it.

`chunking.structured_section` is #861's strategy, which isn't on `develop`. Registering it here would badge nothing anyway (the form takes `chunking_strategies` from the API, so an option the backend doesn't offer renders none) and would tie this PR's merge order to that one.

An empty `NEW_SINCE` is a supported state, not a broken one — it's where the registry lands every time entries expire and get swept up — and it's covered by tests. Adoption is one registry line plus one hook call, documented in FEATURE_TAG.md.

andyne13 (commented):
CodeRabbit's items look addressed to me — the trailing-comma anchor and the version regex in e5b7c88, the `$VER` guard and `sed -i` portability in f0fd6d7d. These are separate.

## The wiring was dropped in the force-push, and the description still describes it

The badge has **no call sites**. Nothing under `ui/src` imports `NewBadge`, `useNewOptions`, `useIsNew` or `NewDot` outside the two test files, and `NEW_SINCE` is `{}` — its entries are commented-out examples. No user sees a badge from this branch.

That wasn't always true. The pre-force-push head, `03ef5c4b`, had:

```
modified  13+  2-  ui/src/pages/admin/presets.tsx
```

which is the Strategy-select wiring the description still describes ("Where it shows", and "`NEW_SINCE` currently holds a single key, `chunking.structured_section`"). The current head touches `presets.tsx` in none of its four commits, so both paragraphs describe a version of the PR that no longer exists.

Worth flagging that CodeRabbit reviewed the earlier revision — `presets.tsx` is in its first run's file list — so its later "add the registry entry" note is only half of it, and its Low merge-risk verdict was partly formed against a PR that had a call site.

Restoring the `presets.tsx` change and the registry entry is the fix; it also resolves the next point.

## Three tests silently skip themselves, and two more assert nothing

`whats-new.test.ts`:

```ts
const KEY = Object.keys(NEW_SINCE)[0] ?? "";
describe.skipIf(!KEY)("for a key the registry does carry", () => { … })
```

With an empty registry that disables exactly the three cases covering `isNew` and `hasNewIn` — the two functions the components actually call. The "registry hygiene" block degrades the same way: both loops iterate `Object.entries(NEW_SINCE)` zero times, so the typo'd-pin check the description credits catches nothing.

What's left is `isNewSince` (the pure window rule) plus the component tests against a mocked registry. The registry path itself is untested. The run also doesn't match the description's "228 passed (28 files)" — I get **235 passed, 3 skipped** across 29 files, and the skips aren't mentioned.

## The CI gate fails open if the registry file moves

```bash
registry=ui/src/lib/whats-new.ts
if grep -qE '^ *"[^"]+": *"UNRELEASED",? *$' "$registry"; then
```

`grep` on a missing path exits **2**, so the `if` takes the else branch and the step prints `OK: no unresolved NEW badge entries` and passes. Renaming or moving `whats-new.ts` silently disables the guard — which matters more here than usual, since the failure it exists to prevent is permanent and invisible. One line:

```bash
[ -f "$registry" ] || { echo "::error::$registry not found"; exit 1; }
```

## The gate only covers GA tags

`verify-tag` is gated on `startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-rc.')`. Anyone running `develop`, an RC image, or a self-built image sees an `UNRELEASED` badge that never expires and that no gate catches. That's inherent to hanging the check off the GA release and is probably the right trade, but FEATURE_TAG.md presents the gate as the backstop without naming the deployments it doesn't reach — and RC deployments are real here.

---

Two things I checked that hold up, for the record: `VERSION_PATTERN` handles PEP 440 correctly, which is not incidental — `/version` is served from `importlib.metadata`, so `2.2.0rc1` is what an RC deployment actually reports and a SemVer-only pattern would blank every badge there. And the Radix trigger-mirror fix works: Tailwind emits `[data-slot=select-value] .\[\[data-slot\=select-value\]_\&\]\:hidden{display:none}` into the built CSS, and it matches the real `data-slot` on `SelectValue`. The test asserts the class string rather than computed style (jsdom has no Tailwind), so that one is worth knowing is verified at the CSS level rather than by the test.

ui/src/lib/whats-new.ts
```diff
@@ -0,0 +1,139 @@
...
+ * permanent, so `verify-tag` in .github/workflows/build.yml refuses to publish
+ * GA images while any entry is still unresolved.
+ */
+export const UNRELEASED = "UNRELEASED";
+
+/**
+ * Feature key -> the app version it first shipped in (`major.minor[.patch]`),
+ * or `UNRELEASED` until a release pins it.
+ */
+export const NEW_SINCE: Readonly<Record<string, string>> = {
```
hedhoud, line 61: The PR still promises a NEW marker on the chunking Strategy control, but the registry is empty and none of the badge helpers has a production call site. As a result, no user can see a badge; only the mocked examples exercise it. Please restore the registry entry and presets integration so the stated behavior is actually reachable.

.github/workflows/build.yml
```diff
@@ -52,6 +52,32 @@ jobs:
...
+        env:
+          TAG_NAME: ${{ github.ref_name }}
+        run: |
+          registry=ui/src/lib/whats-new.ts
+          # Anchored to a whole registry line — indent, quoted key, colon, and
+          # an optional trailing comma, which the last entry in the object may
+          # legally omit — so the sentinel's own definition and the prose that
+          # explains it cannot trip the gate, and an entry written without the
+          # comma cannot slip past it.
+          if grep -qE '^ *"[^"]+": *"UNRELEASED",? *$' "$registry"; then
```
hedhoud: This treats an unreadable or missing registry the same as a clean registry. grep exits with an error, but because it is the if condition the step continues and reports success, silently disabling the release backstop. Please make an unreadable registry fail the gate and reserve success for a genuine no-match result.

ui/src/lib/whats-new.ts
```diff
@@ -0,0 +1,139 @@
...
+ * What counts as a version is wider than SemVer, because `/version` serves
+ * `importlib.metadata`, which reports **PEP 440**: a release candidate arrives
+ * as `2.2.0rc1`, not `2.2.0-rc.1`, and `.post1` / `.dev0` builds are legal too.
+ * Rejecting those would blank every badge on an RC deployment. So: an optional
+ * `v`, two or more numeric segments, an optional PEP 440 marker attached
+ * directly (`a` `b` `rc` `post` `dev` — the normalized set), and an optional
+ * `-`/`+`/`.` suffix for SemVer pre-releases and build metadata.
+ */
+const VERSION_PATTERN =
+  /^\s*v?(\d+)\.(\d+)(?:\.\d+)*(?:(?:a|b|rc|post|dev)\d*)?(?:[-+.][0-9A-Za-z][0-9A-Za-z.+-]*)?\s*$/i;
```
hedhoud: The generic dotted suffix still accepts malformed versions such as 2.2.0.wat. That contradicts the fail-closed contract and can show badges when /version or a registry pin is corrupt. Please limit suffixes to the explicitly supported PEP 440 and SemVer forms and cover malformed dotted suffixes with a regression test.

hedhoud (commented):
I found three cases where the current implementation does not meet the user-visible and fail-closed guarantees described by the PR.

.github/workflows/build.yml
Ahmath-Gadji: Good catch — fixed in 9a11e2ca.

`if grep` can't distinguish exit 1 (no match) from exit 2 (file missing): I ran the old step against an absent registry and it printed grep's error, then `OK`, and exited 0.

The step now refuses to run without a readable registry and treats only exit 1 as clean — any other status fails loud with grep's exit code. `chmod 000` is covered too, which a `[ -f ]` check alone would miss.

ui/src/lib/whats-new.ts
Ahmath-Gadji: Right — `2.2.0.wat` was read as 2.2. Fixed in 9a11e2ca.

The previous commit closed the prefix hole but left a generic dotted tail, which is the same hole one dot along. Each suffix is now spelled out: numeric segments, PEP 440 markers (`a b rc post dev`), an optional SemVer pre-release, an optional `+build`/local segment.

Tested both ways — `2.2.0.wat`, `2.2.0.x9`, `2.2.0-`, `2.2.0+` rejected; `2.2.0rc1`, `2.1.1.post1`, `2.2.0-rc.1`, `2.2.0.dev3+local.7` still compare. A corrupt *pin* is covered as well, since the same parser reads both sides. The PEP 440 forms are deliberate: `/version` comes from `importlib.metadata`, so an RC deployment reports `2.2.0rc1` and a SemVer-only pattern would blank every badge there.

ui/src/lib/whats-new.ts
Ahmath-Gadji: The mismatch is real, but it was in the description rather than the code — rewritten, thanks for catching it.

The empty registry is deliberate: `chunking.structured_section` is #861's strategy, which isn't on `develop`, so the entry would badge nothing here (the form takes `chunking_strategies` from the API) and would couple the two PRs. Adoption is one registry line plus one hook call, documented in FEATURE_TAG.md with this dropdown as the worked example.

Happy to land the wiring here instead if you'd rather the PR demonstrate itself end to end — say the word and I'll add it on top of #861.

Ahmath-Gadji:
@andyne13 all four addressed — thanks, the test one was the most valuable.

- **Self-skipping tests** (70cd571c). Right, and the worst of them: an empty registry disabled exactly the `isNew` / `hasNewIn` cases, on the days a regression there would go unnoticed. `isNewInRegistry` / `hasNewInRegistry` now take the registry as an argument, so the lookup is tested against fixtures — pin, `UNRELEASED`, typo'd pin, empty. **243 passed, 0 skipped.**
- **Gate failing open** (9a11e2ca). Confirmed against the old step: it exits 0 on a missing registry. Now it requires a readable one and reserves success for grep's exit 1; `chmod 000` is covered, which `[ -f ]` alone would miss.
- **GA-only gate.** Agreed on both halves. FEATURE_TAG.md now names what it doesn't reach — `develop`, RC and self-built images — and what it does guarantee: an unresolved entry can't reach GA.
- **Wiring and description.** Description rewritten with the real numbers. The empty registry is a scope decision rather than force-push fallout: #861 isn't on `develop`, so the entry would badge nothing here. Happy to wire it up if you'd prefer the PR demonstrate itself — that thread is open.

Noted on CodeRabbit's verdict predating the force-push; its merge-risk line still cites `30af4`. And you're right that the Radix mirror is verified at the CSS level rather than by the test — the test says so in a comment.

ui/src/lib/whats-new.ts
```diff
@@ -0,0 +1,148 @@
...
+ *   1. numeric segments — `2.2`, `2.2.0`, or more;
+ *   2. any run of PEP 440 pre/post/dev markers, attached or separated by one of
+ *      `.` `-` `_` (`2.2.0rc1`, `2.1.1.post1`, `2.2.0a1.post2.dev3`), limited to
+ *      the normalized set since that is all the endpoint can emit;
+ *   3. an optional SemVer pre-release — `-` then dot-separated alphanumerics,
+ *      which is how a hand-written registry pin spells `2.2.0-rc.1`;
+ *   4. an optional `+` build or local segment (`+build.7`, `+local.7`).
+ */
+const VERSION_PATTERN =
+  /^\s*v?(\d+)\.(\d+)(?:\.\d+)*(?:[._-]?(?:a|b|rc|post|dev)\d*)*(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?\s*$/i;
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed in 4b2ea444.

The marker run was "any number of `a`/`b`/`rc`/`post`/`dev` in any order", which accepts both of your examples and reduces them to 2.2. PEP 440 fixes the order, so the grammar now does too: at most one pre-release, then at most one `post`, then at most one `dev` — simpler than the free-form run it replaces.

Both examples are covered by regression tests; the accepted forms are unchanged (`2.2.0rc1`, `2.1.1.post1`, `2.2.0a1.post2.dev3`, `2.2.0.dev3+local.7`).

andyne13 (commented):
Decision on the open point: **877 merges as the mechanism, and the wiring goes into #861.**

@Ahmath-Gadji your reply offers two different things in one sentence — "land it here" and "on top of #861 once it's in". The second means 877 waits for 861, which is the merge-order coupling you argued against. To be explicit: neither. This PR merges as-is, and I add the registry entry and the presets wiring to #861, the PR that ships the strategy the badge announces.

@hedhoud that should cover your point. The badge does get a production call site and a user-visible marker — in the PR that introduces the option it marks, rather than this one. Holding 877 for it would block a ready PR on an unready one and put the marker in a different commit from the feature it announces.

Order matters and only works one way round: 877 merges → 861 picks up develop → 861 adds `"chunking.structured_section": "UNRELEASED"` plus the `useNewOptions("chunking", chunkingStrategies)` call on the Strategy label and items. `@/components/shared/new-badge` does not exist on 861 until this lands.

Two notes for whoever writes that wiring, since it is not a revert of the dropped diff:

- `NewInGroupBadge` is gone — the current API is the `useNewOptions(group, values)` hook returning `{ dot, badge, badgeFor }`.
- The dropped version had no `textValue` on the `SelectItem`, which `new-badge.tsx`'s own docblock calls for so the marker stays out of Radix's typeahead.

One consequence worth stating here, since it is this PR's gate: from the moment that `UNRELEASED` entry merges, `verify-tag` blocks GA image publishing until a release branch resolves it. That is the design working, and it fails loud rather than silently — but it makes `chore(release): bump version to X.Y.Z` a two-file commit from then on. I'll add it to the release runbook.

Re-checked at 70cd571c: 243 tests pass with no skips (was 235 + 3 skipped), `tsc --noEmit` and `npm run build` clean. Two things I verified independently rather than from the commit messages:

- The gate, under `bash -eo pipefail`, in seven states — clean registry, unresolved with a trailing comma, unresolved without one, resolved pin, missing file, unreadable file, and registry-is-a-directory. All seven correct; the directory case exits 2 and fails loud naming the status, which is a case the fix did not claim.
- The release one-liner, extracted verbatim from CONTRIBUTING.md and run against a fixture: it resolves both comma forms and preserves which form was there, and leaves the `UNRELEASED` constant, the surrounding prose, the commented examples and already-pinned entries untouched. The gate then passes on the output.

Ahmath-Gadji:
@andyne13 agreed on the order, and thanks for taking the 861 wiring — both notes for it are right (`useNewOptions` is the API, and `textValue` is needed or the option answers to `"structured_sectionNEW"`).

Two things, since you re-checked at `70cd571c`:

- One commit landed after it. `4b2ea444` tightens the version grammar to PEP 440's marker order — at most one pre-release, then `post`, then `dev` — so `2.2.0.dev1.post1.dev5` and `2.2.0.post1a1` are rejected rather than read as 2.2. The gate, the release one-liner and the registry lookups are untouched, so your seven-state and verbatim-fixture checks still hold. Still 243 passing, 0 skipped.
- For the runbook edit: `.github/RELEASING.md:109` tells the releaser that a loud `verify-tag` failure means the tag is not an ancestor of `origin/main`. From the 861 entry onwards there is a second cause with a different fix — resolve the registry on the release branch and re-cut the tag — so that line wants a clause. Worth doing in the same pass as the two-file bump note.
- One more since, `bbcb7ce7`: `NewDot` was a bare span, so `display: inline` — width and height do not apply there, and its `size-1.5` was ignored anywhere the parent was not a flex or grid container. It only ever showed because the surfaces tried (the Strategy label, the sidebar item) are flex rows. `inline-block` fixes it and costs nothing in a flex parent. Relevant to the 861 wiring if the dot ends up on a label that is not a flex row, and its docblock claimed the opposite, so the guide is corrected too.

Ahmath-Gadji:
Squashed the eight commits into one, `6a8b1647`.

Since the last force-push here dropped content: this one changes no content at all. `git diff bbcb7ce7 6a8b1647` is empty — the tree is identical to what you both reviewed. The SHAs the reviews cite (`e5b7c88e`, `f0fd6d7d`, `9a11e2ca`, `70cd571c`, `4b2ea444`, `bbcb7ce7`) stay resolvable from the review timeline, but are no longer on the branch.

ui/src/lib/whats-new.ts
```diff
@@ -0,0 +1,179 @@
...
+ *      attached or separated by `.` `-` `_` (`2.2.0rc1`, `2.1.1.post1`,
+ *      `2.2.0a1.post2.dev3`). Ordering them is what makes the grammar reject
+ *      `2.2.0.dev1.post1.dev5` and `2.2.0.post1a1`, which are version-shaped but
+ *      not versions;
+ *   3. an optional SemVer pre-release — `-` then dot-separated alphanumerics,
+ *      which is how a hand-written registry pin spells `2.2.0-rc.1`;
+ *   4. an optional `+` build or local segment (`+build.7`, `+local.7`).
+ */
+const VERSION_PATTERN =
+  /^\s*v?(\d+)\.(\d+)(?:\.\d+)*(?:[._-]?(?:a|b|rc)\d*)?(?:[._-]?post\d*)?(?:[._-]?dev\d*)?(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?\s*$/i;
```
hedhoud: The parser still accepts a hybrid of the two supported suffix syntaxes because the PEP 440 markers and the SemVer prerelease are optional in sequence. For example, `2.2.0rc1-beta` and `2.2.0.post1-rc.1` both match and are reduced to `2.2`, so a malformed `/version` value or registry pin can still show a NEW badge instead of failing closed. Could we make the suffix grammars alternatives—ordered PEP 440 markers or a SemVer `-...` prerelease, followed by optional build metadata—and add these two examples as regression cases? This keeps both valid formats without allowing them to be combined.

Ahmath-Gadji: Right on both — `2.2.0rc1-beta` and `2.2.0.post1-rc.1` matched and reduced to 2.2, which is the fail-closed contract broken the same way as the last round.

Fixed in `1c672976`: the ordered PEP 440 run and the SemVer pre-release are now alternatives, with the `+build` segment left outside the alternation so it stays legal after either. Verified across 18 accepted and 16 rejected forms — I added `2.2.0-rc.1+build.7` and `2.2.0-1` to the accept set, since those are what an alternation could plausibly have broken; only your two examples change verdict, and both are regression tests now.

andyne13 (approved):
Approving at 1c672976. Everything raised across the four review rounds is addressed, and I re-verified the whole change rather than just the last fix, since the branch was squashed and re-merged since my first pass.

- All six CI checks pass. Locally: 243 UI tests with no skips, `tsc --noEmit`, `npm run build` and `eslint` clean.
- The version grammar: 19 cases covering every example from all four rounds behave correctly — both PEP 440/SemVer hybrids reject, the marker-ordering pair rejects, `2.2.0.wat` / `2.2rubbish` / `2.2.0-` / `2.2.0+` reject, and every legitimate form still accepts (plain, `v`-prefixed, PEP 440 pre/post/dev chains, SemVer pins, both `+build` variants). Making the two grammars alternatives closed the last class without narrowing anything valid.
- The gate: re-simulated under `bash -eo pipefail` in seven states — clean, unresolved with and without a trailing comma, resolved pin, missing file, unreadable file, and registry-is-a-directory. All correct.
- The release chain end to end: the `CONTRIBUTING.md` command resolves both comma forms against a fixture, leaves the `UNRELEASED` constant, the prose and the commented example untouched, and the gate then passes on its output.
- No skipped tests: moving the lookup behind `isNewInRegistry` / `hasNewInRegistry` means the registry path is covered against fixtures even while `NEW_SINCE` is empty, which was the point.

Merging as the mechanism, with an empty registry and no call site, is deliberate per the plan agreed above: the registry entry and the presets wiring land in #861, which ships the strategy the badge announces. That still needs 877 to merge first, since `@/components/shared/new-badge` does not exist on 861 until it does.

PR #880, opened 2026-09-01
fix(vectordb): make BM25 lexical search case-insensitive

### LINKED ISSUE #870: BM25 lexical search is case-sensitive: the custom analyzer drops the lowercase filter

Ahmath-Gadji:
BM25 lexical search is case-sensitive: a query for `rapport` scores zero against text that says `Rapport`. Closes #870.

## Why it happened

The `text` field declares a **custom** analyzer — `"tokenizer": "standard"` plus an explicit filter chain — not the built-in `standard` analyzer, which is `{"type": "standard"}`. A custom analyzer inherits nothing, so this silently opted out of the `lowercase` filter the built-in one bundles. Milvus runs the same analyzer over the query text at search time, so a case mismatch on either side scores zero on the lexical leg of hybrid search. The dense leg still returns something, which is why this looked like mediocre recall rather than a bug.

Order matters, so `lowercase` goes first: the `_english_` / `_french_` stop lists are lowercase and previously only matched tokens that happened to already be folded — `the` was stripped, `The` was not, which also skewed IDF.

## Why `enable_match` goes too

Nothing in the codebase issues a `TEXT_MATCH` query; it is reachable only through the raw `expr` escape hatch. It is not free, though: Milvus refuses to alter `analyzer_params` while text match is enabled, **and** refuses to turn `enable_match` off (`enable_match does not allow update in collection field param`). While it is set, the analyzer is frozen forever.

So both changes ride one rebuild — and it is the last one. With `enable_match` gone, a future analyzer change is `drop_function_field` → `alter_collection_field` → `add_function_field` on the live collection, with the sparse column backfilled by compaction.

## The migration

`2.rebuild_text_analyzer.py`, schema version 1 → 2. Copies the rows into `<collection>_v2_rebuild`, verifies the counts, swaps the names, keeps the old collection as `<collection>_v1_backup` (never dropped automatically).

Nothing is re-embedded — dense vectors are copied verbatim and the sparse ones are regenerated server-side on insert, which is what applies the new analyzer. Per-deployment values (vector dim, VARCHAR lengths, hybrid or not) are read off the live collection rather than assumed, and a declared field the migration does not know about aborts the run rather than being silently dropped.

Chunk IDs survive. `_id` is `auto_id`, which normally refuses caller-supplied keys, so the rebuild collection carries `allow_insert_auto_id` for the duration of the copy and drops it again before the swap — the collection ends up with the same semantics as a freshly created one, and an ID handed out earlier by MCP still resolves.

The copy reads a snapshot, so the migration re-counts the source afterwards and aborts, original untouched, if the count moved — i.e. if something was still writing when the runbook said to stop OpenRAG.

It reuses the copy/swap logic from #871 (closed) with both of that PR's open review findings fixed: the swap now renames the backup back into place if the second rename fails — the one moment the collection name resolves to nothing — and the copy pages on the collection's *declared* `text` length instead of a guessed fixed allowance, so a page of maximally long chunks cannot exceed the RPC cap.

## Making the mismatch visible

`_check_schema_version` raises, but only on the indexing path, so an un-migrated deployment boots, answers searches against the old schema, and only learns about it at the first upload. The store now probes the collection when it is **constructed** — which is where the API process first reaches Milvus, since it never calls `initialize` — and logs a warning naming the runner command:

```
WARNING | Collection `openrag` is at schema version 1, but this build expects 2.
          Indexing fails until the pending migration(s) are applied. Run, with
          OpenRAG stopped: uv run python .../migrate.py --dry-run (then without).
```

It never raises: a collection that does not exist yet is not a mismatch, and an unreadable one is logged and swallowed. A collection *ahead* of the build gets the opposite advice. Construction already connected (`MilvusClient` does an eager `_wait_for_channel_ready`, and `__init__` already converted that failure into `VDBConnectionError`), so this adds one RPC to a connection that was being made anyway.

**This does not need Storage V3** (#876). It was verified on a Storage V2 server precisely to keep the two independent; V3 is what makes the *next* such change cheap, not this one.

## Verification

End to end against a throwaway Milvus 3.0.0 on **Storage V2**, seeded with a v1-shaped collection (old analyzer, `enable_match`, 20 rows across 3 partitions):

| | before | after |
|---|---|---|
| `bm25("rapport")` | 10 / 20 rows | **20 / 20** |
| `bm25("RAPPORT")` | 0 | **20** |
| `enable_match` on `text` | `"true"` | absent |

Also: all 20 `_id` values came back identical; `allow_insert_auto_id` was not left on the collection and a subsequent normal insert still auto-generated; `--dry-run` created nothing and left the stamp at 1; a second `upgrade` is a no-op; the partition key survived the copy; `downgrade` put the v1 collection back with all 20 rows and kept the v2 one aside.

In-repo:

- `uv run pytest tests/unit/` — 2571 passed. 17 new tests drive the migration against a fake client (skip paths, the interrupted-previous-run guard, the row-count mismatch abort, the failed-swap restore, copy paging, the `auto_id` property lifecycle, the concurrent-write abort), 4 pin the analyzer shape and assert `enable_match` is no longer wired onto the field, and 7 cover the startup warning (stale, ahead, current, absent, unreachable, and log-before-raise).
- `uv run pytest tests/integration/repos/test_milvus_store_integration.py -m integration` — 19 passed against a live Milvus, including two new ones: `run_analyzer` folds case (`le` / `de` are now stripped, which only happens because the fold precedes the `_french_` list), and a lowercase query matches capitalised text with the dense leg range-filtered out, so only BM25 can be answering.
- ruff clean; `scripts/check_layer_imports.py` OK.

## Operator impact

- **A migration is required.** `vectordb.schema_version` goes to 2, so an un-migrated collection is stale. The version check runs on the indexing path only: the deployment still boots and still answers searches — with the old analyzer — and fails the first upload with `Collection ... is at schema version 1 but the application requires version 2`. Run `migrate.py` before restarting, with OpenRAG stopped — the copy reads a snapshot, and the migration aborts if the source changes underneath it. Allow roughly twice the collection size while both copies exist. Chunk IDs are preserved. `hybrid_search: false` deployments have no BM25 leg and are only re-stamped.

Documented under `## Schema Migrations` → "Version 2" in `milvus_migration.mdx`, along with what the v2 → v1 rollback can and cannot recover.

## Summary by CodeRabbit

* **New Features**
  * BM25 text search is now case-insensitive, improving matches regardless of capitalization.
  * Added a Milvus schema migration that rebuilds collections while preserving data and supporting rollback.
  * Migration previews include consistency checks, backups, and safeguards before collections are replaced.
  * Startup checks warn when a collection requires migration.

* **Documentation**
  * Added upgrade, dry-run, rollback, naming, and operational guidance.

* **Bug Fixes**
  * Improved consistency between indexed text and search queries through case normalization.

### DIFF
--- a/docs/content/docs/documentation/milvus_migration.mdx
+++ b/docs/content/docs/documentation/milvus_migration.mdx
@@ -238,6 +238,47 @@ Migrations are versioned. The runner reads the current schema version stored in 
 Stop the OpenRAG application before running any migration.
 :::
 
+### Version 2 — case-insensitive BM25 analyzer
+
+Schema version 2 makes two changes to the `text` field:
+
+- adds the `lowercase` filter to the analyzer, so BM25 stops treating `Rapport` and `rapport` as different terms. Milvus applies the analyzer to the query as well as to the indexed text, so before this fix a case mismatch scored zero on the lexical leg;
+- drops `enable_match`. Nothing in OpenRAG issues a `TEXT_MATCH` query, and while it is set Milvus refuses to alter the analyzer.
+
+It **rebuilds** the collection rather than altering it, because both changes are refused in place: `analyzer_params` cannot change while text match is enabled, and `enable_match` itself cannot be turned off. The migration copies the rows into `<collection>_v2_rebuild`, checks the counts, then swaps the names, keeping the old collection as `<collection>_v1_backup`.
+
+This is the last rebuild. Once `enable_match` is gone, a later analyzer change is [three calls on the live collection](https://milvus.io/docs/add-fields-to-an-existing-collection.md) with the sparse column backfilled by compaction — provided [Storage V3](https://milvus.io/docs/storage-v3.md) is enabled.
+
+:::caution[Plan for the rebuild]
+- **Disk** — both copies exist at once; allow roughly twice the collection size until the backup is dropped.
+- **Memory** — checking the row counts loads both collections, and a rename does not unload one, so the migration releases the backup from query-node memory after the swap. The doubling above is disk, not RAM.
+- **No re-embedding** — dense vectors are copied verbatim and the sparse ones are regenerated server-side on insert, which is what applies the new analyzer.
+- **Chunk IDs are preserved** — `_id` is `auto_id`, which normally refuses caller-supplied keys, so the rebuild collection carries `allow_insert_auto_id` for the duration of the copy and drops it again before the swap. A chunk ID handed out earlier by the MCP `get_chunk_by_id` tool still resolves.
+- **Stop OpenRAG first**, as the banner above says — the copy reads a snapshot, so a row written while it runs would be missing from the rebuild. The migration re-counts the source afterwards and aborts, leaving the original untouched, if the count moved.
+- **`hybrid_search: false`** deployments have no BM25 leg and are only re-stamped — no rebuild, and no backup collection, in either direction.
+- A failed copy leaves the original untouched; drop the partial `<collection>_v2_rebuild` and retry.
+- The version check runs on the indexing path only. An un-migrated deployment still boots and still answers searches — with the old analyzer — and fails the first upload with `Collection ... is at schema version 1 but the application requires version 2`.
+:::
+
+Once validated, drop the backup:
+
+```python
+from pymilvus import MilvusClient
+
+MilvusClient(uri="http://localhost:19530").drop_collection("<collection>_v1_backup")
+```
+
+:::note[Where to run these]
+The commands below run from `infra/compose/`. From the repo root, point Compose at the file (and at your project name, if you use one):
+
+```bash
+docker compose -p <project> -f infra/compose/docker-compose.yaml run --no-deps --rm --build --entrypoint "" openrag \
+    uv run python services/persistence/migrations/milvus/migrate.py --dry-run
+```
+
+The migration runner is a file, not a package — pass the full path to `migrate.py`, and keep Milvus running.
+:::
+
 ### Step 1 — Start only the Milvus container
 
 ```bash
@@ -267,20 +308,20 @@ docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-
   </TabItem>
 </Tabs>
 
-Review the output to confirm which migrations are pending and what changes they would apply.
+Review the output to confirm which migrations are pending and what changes they would apply. The migration scripts ship **inside the image**, which is why `--build` is there: if the `Discovered N migration(s)` line does not list the migration you expect, the image is stale — and on a shared host another checkout may have rebuilt the same `linagoraai/openrag` tag, so add `--build` to the apply step too.
 
 ### Step 3 — Apply all pending migrations
 
 <Tabs>
   <TabItem label="GPU">
 ```bash
-docker compose run --no-deps --rm --entrypoint "" openrag \
+docker compose run --no-deps --rm --build --entrypoint "" openrag \
     uv run python services/persistence/migrations/milvus/migrate.py
 ```
   </TabItem>
   <TabItem label="CPU">
 ```bash
-docker compose --profile cpu run --no-deps --rm --entrypoint "" openrag-cpu \
+docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-cpu \
     uv run python services/persistence/migrations/milvus/migrate.py
 ```
   </TabItem>
@@ -314,22 +355,22 @@ To upgrade or downgrade to a specific schema version rather than the latest:
   <TabItem label="GPU">
 ```bash
 # Upgrade to version 2 only
-docker compose run --no-deps --rm --entrypoint "" openrag \
+docker compose run --no-deps --rm --build --entrypoint "" openrag \
     uv run python services/persistence/migrations/milvus/migrate.py --target 2
 
 # Downgrade to version 0 (resets version stamp and drops indexes)
-docker compose run --no-deps --rm --entrypoint "" openrag \
+docker compose run --no-deps --rm --build --entrypoint "" openrag \
     uv run python services/persistence/migrations/milvus/migrate.py --downgrade --target 0
 ```
   </TabItem>
   <TabItem label="CPU">
 ```bash
 # Upgrade to version 2 only
-docker compose --profile cpu run --no-deps --rm --entrypoint "" openrag-cpu \
+docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-cpu \
     uv run python services/persistence/migrations/milvus/migrate.py --target 2
 
 # Downgrade to version 0 (resets version stamp and drops indexes)
-docker compose --profile cpu run --no-deps --rm --entrypoint "" openrag-cpu \
+docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-cpu \
     uv run python services/persistence/migrations/milvus/migrate.py --downgrade --target 0
 ```
   </TabItem>
@@ -337,18 +378,29 @@ docker compose --profile cpu run --no-deps --rm --entrypoint "" openrag-cpu \
 
 ### Rollback
 
-Milvus does not support dropping fields. A downgrade only removes indexes and resets the version stamp — fields remain in the schema but are unused by the application:
+What a downgrade can undo depends on the migration:
+
+- **Version 2 → 1** swaps the `<collection>_v1_backup` collection back into place and sets the version stamp to 1. The version-2 collection is kept aside as `<collection>_v2_rolled_back` rather than dropped, so nothing is lost — but this is a point-in-time rollback: **rows indexed after the upgrade exist only in the collection set aside**.
+
+  The rollback needs that backup collection. If it is gone, the downgrade fails rather than reporting a rollback it did not perform, and there are three ways forward:
+
+  - Restore `<collection>_v1_backup` from a Milvus backup and re-run the downgrade.
+  - Re-create the collection with the v1 schema and re-index it.
+  - Move the version stamp back by hand — but **only** if you know this collection was re-stamped rather than rebuilt. A collection that already had the v2 analyzer and no text match before the upgrade was moved to version 2 by the stamp alone, so it correctly has no backup. Afterwards it is indistinguishable from one that was rebuilt and lost its backup, which is why the migration refuses to guess between them.
+
+  A `hybrid_search: false` collection never had a backup — it was only re-stamped on the way up, so the rollback moves its stamp back and nothing else.
+- **Version 1 → 0** only removes indexes and resets the version stamp, since Milvus does not support dropping fields — they remain in the schema but are unused by the application.
 
 <Tabs>
   <TabItem label="GPU">
 ```bash
-docker compose run --no-deps --rm --entrypoint "" openrag \
+docker compose run --no-deps --rm --build --entrypoint "" openrag \
     uv run python services/persistence/migrations/milvus/migrate.py --downgrade
 ```
   </TabItem>
   <TabItem label="CPU">
 ```bash
-docker compose --profile cpu run --no-deps --rm --entrypoint "" openrag-cpu \
+docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-cpu \
     uv run python services/persistence/migrations/milvus/migrate.py --downgrade
 ```
   </TabItem>
@@ -368,9 +420,10 @@ Files must follow the pattern `N.short_description.py`, where `N` is the **targe
 
 ```
 openrag/services/persistence/migrations/milvus/
-  1.add_temporal_fields.py    ← brings the schema to version 1
-  2.your_new_migration.py     ← brings the schema to version 2
-  migrate.py                  ← generic runner (do not rename)
+  1.add_created_at_temporal_fields.py  ← brings the schema to version 1
+  2.rebuild_text_analyzer.py           ← brings the schema to version 2
+  3.your_new_migration.py              ← brings the schema to version 3
+  migrate.py                           ← generic runner (do not rename)
 ```
 
 The numeric prefix determines execution order. Never reuse or change an existing version number.
@@ -383,7 +436,7 @@ Each migration script must expose the following at module level:
 |------|------|-------------|
 | `TARGET_VERSION` | `int` | The schema version this script brings the collection to |
 | `upgrade(client, collection_name, dry_run)` | `function` | Applies the migration |
-| `downgrade(client, collection_name, dry_run)` | `function` | Reverts the migration (indexes only — fields cannot be dropped) |
+| `downgrade(client, collection_name, dry_run)` | `function` | Reverts the migration, as far as Milvus allows — a field added in `upgrade` cannot be dropped, so such a migration only removes its indexes and resets the stamp |
 
 ### Minimal template
 
--- a/openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
+++ b/openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
@@ -1,0 +1,613 @@
+"""
+Milvus migration: rebuild the ``text`` field  (schema version 1 → 2)
+====================================================================
+Two changes to the same field, in one pass:
+
+* adds the ``lowercase`` filter to the analyzer, so BM25 stops treating
+  ``Rapport`` and ``rapport`` as distinct terms;
+* drops ``enable_match``, which nothing in OpenRAG queries — no code path
+  issues ``TEXT_MATCH`` — and which is what makes the analyzer immutable.
+
+Both need a rebuild: Milvus 3.0 refuses to alter ``analyzer_params`` while text
+match is enabled, and refuses to turn ``enable_match`` off ("enable_match does
+not allow update in collection field param"). Probed against 3.0.0, on
+collections created both before and after Storage V3.
+
+This is the last rebuild. Once ``enable_match`` is gone, a later analyzer change
+is ``drop_function_field`` → ``alter_collection_field`` → ``add_function_field``
+on the live collection, with the sparse column backfilled by compaction.
+
+Copies rows into ``<collection>_v2_rebuild``, verifies the counts, then swaps
+names, keeping the old collection as ``<collection>_v1_backup`` — never dropped
+automatically, but released from query-node memory so it does not sit there at
+full size next to the live one. Dense vectors are copied verbatim and the
+sparse vectors are regenerated server-side on insert, so nothing is
+re-embedded. Collections with no ``sparse`` field have no BM25 leg and are only
+re-stamped, in either direction.
+
+``_id`` is ``auto_id``, which normally rejects caller-supplied keys; the rebuild
+collection carries ``allow_insert_auto_id`` for the duration of the copy so the
+original chunk IDs survive, and the property is dropped again before the swap.
+
+OpenRAG must be stopped: rows written to the source while the copy runs are not
+picked up, so the migration re-counts the source afterwards and aborts if it
+moved.
+
+Usage — prefer the generic runner (from repo root, inside the container):
+    docker compose run --no-deps --rm --entrypoint "" openrag \\
+        uv run python services/persistence/migrations/milvus/migrate.py [--dry-run]
+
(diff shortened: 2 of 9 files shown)


openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,509 @@
...
+_COPY_PAGE_BUDGET_BYTES = 32 * 1024 * 1024
+
+#: Per-row allowance for everything but ``text`` and ``vector`` — the scalar
+#: columns plus any dynamic fields travelling in ``$meta``.
+_COPY_ROW_OVERHEAD_BYTES = 2_048
+
+#: Keys that must never be written back on insert: ``_id`` is an ``auto_id``
+#: primary key and ``sparse`` is a BM25 ``Function`` output — Milvus rejects
+#: caller-supplied values for both.
+_UNSETTABLE_KEYS = frozenset({"_id", "sparse"})
```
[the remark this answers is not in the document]
Ahmath-Gadji: Good catch, and it turns out Milvus has a first-class answer for it. Probed against 3.0.0 (Storage V2): without the property an explicit `_id` on an `auto_id` field is rejected outright (`more fieldData has pass in: invalid parameter[expected=4][actual=5]`); with `allow_insert_auto_id` set the value round-trips; and once the property is dropped again a normal insert auto-generates as before.

So the rebuild collection now carries `allow_insert_auto_id` for the duration of the copy, `_id` stays in the copied rows, and the property is dropped before the swap so the collection ends up with the same semantics as a freshly created one. Failing to drop it aborts inside the existing try/except, which leaves the source untouched.

Covered by two unit tests (the copied row keeps `_id` and drops `sparse`; the property is set → insert → dropped → renamed, in that order) and by an end-to-end run against a live Milvus asserting the 20 source `_id` values come back identical, that `allow_insert_auto_id` is not left on the collection, and that a subsequent normal insert still works.

The "chunk IDs change" caveat is gone from the docs as a result. ede05c12

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,509 @@
...
+    for name in (rebuild_name, backup_name):
+        if client.has_collection(name):
+            raise RuntimeError(
+                f"Collection '{name}' already exists — a previous run of this migration was "
+                "interrupted. Inspect it and drop it before retrying."
+            )
+
+    dim = _vector_dim(desc)
+    text_max_length = _max_length(desc, "text")
+    source_count = _row_count(client, collection_name)
```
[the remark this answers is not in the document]
Ahmath-Gadji: Partly taken. Milvus has no write-barrier primitive to acquire, and the barrier would have to live in the application rather than the migration — the runbook contract is that OpenRAG is stopped for the whole procedure (`:::danger[OpenRAG must be stopped]`, and step 1 starts *only* the Milvus container).

What was missing was the migration checking that the contract was honoured. It now re-counts the source after `_copy_rows()` and aborts if the count moved, before anything is renamed:

```
The source collection changed during the copy: 20 rows before, 21 after.
Something is still writing to 'openrag' — stop OpenRAG, drop 'openrag_v2_rebuild', and retry.
```

That catches inserts and deletes during the copy; the source is untouched either way, so the fix is to stop the writer and re-run. A net-zero in-place update would still slip through, which is why the docs keep the "stop OpenRAG first" bullet rather than presenting the check as a substitute for it. Capture-and-replay would be the alternative, but it is a lot of machinery for a procedure that is already required to run against a quiet collection.

Unit test: `test_a_source_that_changes_during_the_copy_aborts_the_swap`. ede05c12

docs/content/docs/documentation/milvus_migration.mdx
```diff
@@ -267,7 +306,7 @@ docker compose --profile cpu run --no-deps --rm --build --entrypoint "" openrag-
   </TabItem>
 </Tabs>
 
-Review the output to confirm which migrations are pending and what changes they would apply.
+Review the output to confirm which migrations are pending and what changes they would apply. The migration scripts ship **inside the image**, which is why `--build` is there: if the `Discovered N migration(s)` line does not list the migration you expect, the image is stale — and on a shared host another checkout may have rebuilt the same `linagoraai/openrag` tag, so add `--build` to the apply step too.
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed, and applied more widely than flagged: every `migrate.py` invocation in the page now carries `--build`, not just the two apply examples — the "targeting a specific version" and rollback snippets had the same staleness risk. 11 commands, all consistent now. ede05c12

openrag/services/storage/milvus_store.py
```diff
@@ -178,6 +184,12 @@ def __init__(self, config: VectorDBConfig) -> None:
         # production drops surface a real issue, revisit with evidence
         # rather than racing pymilvus's internal channel state.
 
+        # One describe_collection at construction so a pending migration shows
+        # up in the startup logs. The API process builds this store while
+        # wiring the container and never calls `initialize` afterwards, so
+        # without this the mismatch stays invisible until the first upload.
+        self.warn_if_migration_pending()
```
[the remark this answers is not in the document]
Ahmath-Gadji: Half right, so half taken.

The unreachable case is already handled and predates this change: `MilvusClient(uri=..., timeout=...)` connects eagerly (`_wait_for_channel_ready`), and `__init__` has always converted that failure into `VDBConnectionError` at line 163 — before the probe runs. Verified against a closed port: construction raises from the client constructor, never reaching `warn_if_migration_pending`. So the probe does not introduce a 120s wait on an unreachable server; that cost was already there and already fatal.

What it *could* add is latency against a server that completes the gRPC handshake but answers metadata slowly — two extra RPCs bounded by the client default. Fixed by giving the probe its own timeout rather than moving it:

```python
_SCHEMA_PROBE_TIMEOUT = 5.0
...
self._client.has_collection(self._collection_name, timeout=_SCHEMA_PROBE_TIMEOUT)
self._read_schema_version(timeout=_SCHEMA_PROBE_TIMEOUT)
```

`_read_schema_version` takes the timeout as a keyword so `_check_schema_version` — the indexing path, which is not a probe — keeps the configured client timeout. Two tests pin both halves: the probe passes `_SCHEMA_PROBE_TIMEOUT` on both calls, and `_check_schema_version` passes `None`.

Not moving it to the startup lifecycle: construction is where the API process first reaches Milvus (it never calls `initialize`), which is exactly why the check lives there — that placement is the feature, and bounding the RPCs removes the cost. df30439b

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,541 @@
...
+def downgrade(client: MilvusClient, collection_name: str, dry_run: bool = False) -> None:
+    """Swap the v1 backup collection back into place.
+
+    Rows indexed after the upgrade live only in the v2 collection, so this is a
+    point-in-time rollback: the v2 collection is kept aside rather than dropped.
+    """
+    backup_name = f"{collection_name}{BACKUP_SUFFIX}"
+    aside_name = f"{collection_name}{ROLLBACK_ASIDE_SUFFIX}"
+
+    if not client.has_collection(backup_name):
```
hedhoud, line 536: **P1 — Handle downgrades that did not require a rebuild.** Dense-only collections are upgraded by changing only the schema-version property, so they correctly have no backup collection. Returning here reports a successful downgrade while leaving version 2 stamped, and the version 1 application will reject indexing. Please restamp collections that legitimately used the no-rebuild path, and fail loudly when a backup should exist but is missing.

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,541 @@
...
+        client.drop_collection_properties(rebuild_name, property_keys=[ALLOW_INSERT_AUTO_ID])
+        _stamp_version(client, rebuild_name, TARGET_VERSION)
+    except Exception:
+        logger.error(
+            f"Rebuild failed — '{collection_name}' is untouched. "
+            f"Drop the partial collection '{rebuild_name}' before retrying."
+        )
+        raise
+
+    _swap_in_rebuild(client, collection_name, rebuild_name, backup_name)
```
hedhoud: **P1 — Release the collection moved aside after the swap.** Count verification loads both the source and rebuilt collections. Renaming the source does not unload it, so retaining the backup as instructed leaves two complete collections resident and can exhaust QueryNode memory on a large deployment. Please release the backup after upgrade, and release the version 2 collection moved aside during rollback as well.

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,541 @@
...
+        f"{'[DRY-RUN] ' if dry_run else ''}Rolling back:\n"
+        f"  '{collection_name}' → '{aside_name}' (v2, kept aside)\n"
+        f"  '{backup_name}' → '{collection_name}' (v1, restored)"
+    )
+    if dry_run:
+        logger.info("[DRY-RUN] Dry-run complete. No changes were made.")
+        return
+
+    client.rename_collection(collection_name, aside_name)
+    client.rename_collection(backup_name, collection_name)
```
hedhoud: **P2 — Restore the live name if the second rollback rename fails.** If this rename fails, the version 2 collection has already moved aside while the version 1 backup still has its backup name. The configured live collection name then does not exist, so OpenRAG cannot restart without manual repair. Please add the same restore-on-failure protection used by the upgrade swap.

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
```diff
@@ -0,0 +1,541 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _copy_batch_size(dim: int, text_max_length: int = MAX_LENGTH) -> int:
+    """Rows per page, kept under Milvus's result-size cap.
+
+    Sized on the *declared* ``text`` length rather than the average one: a page
+    of maximally long chunks has to fit too.
+    """
+    per_row = dim * 4 + text_max_length + _COPY_ROW_OVERHEAD_BYTES
```
hedhoud: **P2 — Include every variable-size field in the page budget.** The fixed 2 KiB allowance does not cover long partition or file identifiers, and dynamic metadata can be much larger. A page of valid rows can therefore exceed Milvus's result-size limit and make this required migration impossible to complete. Please size pages from all variable fields or reduce the batch adaptively when Milvus rejects it.

hedhoud (commented):
I found four migration paths that can leave a supported deployment unusable or prevent a valid collection from being migrated safely.

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
Ahmath-Gadji: Taken — and it was worse than "reports success": the downgrade returned 0 with version 2 still stamped, so the next upload failed the version check with no indication why.

The two paths are now told apart by the collection itself. A collection with no `sparse` field never had an analyzer to fix, so the upgrade only moved its stamp and the rollback does the same — no backup involved in either direction. Anything else reaching that branch has a BM25 leg, which means the upgrade rebuilt it and left a backup that is now gone; that raises, naming the collection and the version it is stuck at, and pointing at the two ways out (restore the backup, or recreate and re-index). The error also mentions the one ambiguous case — a hybrid collection that was already in its v2 shape at v1 and so was only re-stamped — and says to move the stamp by hand there, since nothing on the collection distinguishes it from a lost backup after the fact.

The predicate the upgrade uses to decide "rebuild or re-stamp" is now a named `_needs_rebuild()` shared by both directions, so the two cannot drift apart.

Verified live against Milvus 3.0.0: a dense-only collection goes 1 → 2 → 1 by stamp alone, and a rebuilt collection whose backup has been dropped raises instead of returning. `67e90ddc`

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
Ahmath-Gadji: Confirmed and fixed. Both counts call `load_collection`, and `rename_collection` does not unload, so the backup was left fully resident next to the live collection — on a deployment sized for one copy, that is the whole margin.

Whichever collection is moved aside is now released after each swap: the backup after the upgrade, the version-2 collection after the rollback. Best effort — a release that fails logs a warning rather than failing a migration that has already completed, since the consequence is memory that frees on the next Milvus restart, not data loss.

Verified live: after the upgrade `get_load_state` reports `NotLoad` for the backup and `Loaded` for the live collection, and the same after the rollback for the collection set aside. The runbook now says the doubling it warns about is disk, not RAM. `67e90ddc`

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
Ahmath-Gadji: Agreed, and rather than copy the protection I generalised the one that already had it. `_swap_in_rebuild` became `_swap_collections(client, live_name, incoming_name, aside_name)`; the upgrade calls it with the rebuild, the rollback with the backup. Same restore-on-failure, same last-resort message naming where both collections ended up if even the restore fails — and no second copy to keep in sync.

Covered by a unit test that fails the rollback's second rename and asserts the version-2 collection is renamed back to the live name. `67e90ddc`

openrag/services/persistence/migrations/milvus/2.rebuild_text_analyzer.py
Ahmath-Gadji: Right, and understated: `partition` and `file_id` are declared at the same 64 KiB as `text`, so the 2 KiB allowance was covering two fields that could each be as large as the one it was sized on.

The budget now comes from the schema rather than from two arguments. Every declared VARCHAR contributes its own `max_length`, the dense vector its `dim * 4`, and anything with no size in its params a fixed 16 bytes. `sparse` gets charged too — `output_fields=["*"]` brings it back over the wire even though the copy strips it before inserting — at the width of the `text` it derives from, which is generous for one entry per unique term. Dynamic metadata is the one part with no declared width, so it keeps a flat allowance, raised 2 KiB → 8 KiB.

I took the deterministic half of your suggestion rather than the adaptive one: a `query_iterator` surfaces the rejection on `.next()`, mid-stream, and restarting it loses the position, so shrinking on rejection would mean restarting the copy rather than retrying the page.

On a default collection this pages at 124 rows instead of ~490, still inside the 32 MiB budget with the 64 MiB cap left as headroom. Verified live end to end at that size. `67e90ddc`

docs/content/docs/documentation/milvus_migration.mdx
```diff
@@ -314,41 +355,44 @@ To upgrade or downgrade to a specific schema version rather than the latest:
...
 ```
   </TabItem>
 </Tabs>
 
 ### Rollback
 
-Milvus does not support dropping fields. A downgrade only removes indexes and resets the version stamp — fields remain in the schema but are unused by the application:
+What a downgrade can undo depends on the migration:
+
+- **Version 2 → 1** swaps the `<collection>_v1_backup` collection back into place and sets the version stamp to 1. The version-2 collection is kept aside as `<collection>_v2_rolled_back` rather than dropped, so nothing is lost — but this is a point-in-time rollback: **rows indexed after the upgrade exist only in the collection set aside**. The rollback needs that backup collection; if it has already been dropped, the downgrade fails rather than reporting a rollback it did not perform, and the only way back is to recreate the collection with the old analyzer and re-index. A `hybrid_search: false` collection never had a backup — it was only re-stamped on the way up, so the rollback moves its stamp back and nothing else.
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fair — the migration's error message already names three ways out of a missing backup, and the doc only named one. Fixed in `7ecd69b9`.

The bullet now splits into the three routes explicitly: restore `<collection>_v1_backup` from a Milvus backup and re-run the downgrade, re-create with the v1 schema and re-index, or move the stamp back by hand.

On that third one I kept a caveat your suggestion did not have. A hybrid collection that was already in the v2 shape is re-stamped without a backup, yes — but after the fact it is *indistinguishable* from one that was rebuilt and lost its backup: both end up hybrid, v2-shaped, stamped 2, no backup. That is exactly why `downgrade` raises instead of picking one. So the doc says to move the stamp by hand **only if you know** the collection was re-stamped rather than rebuilt, and says why the migration will not guess. Presenting it as a plain third option would invite an operator to re-stamp a rebuilt collection back to v1 while it still holds v2 data.

hedhoud (approved):
Reviewed the latest changes. The actionable review findings are resolved, and the required checks pass.

PR #883, opened 2026-09-01
feat(ui): highlight new STT and Jobs capabilities

hedhoud:
## Context

The v2.2 release introduces several STT configuration surfaces and makes ongoing Jobs activity easier to discover. Without a temporary visual cue, these capabilities are easy to miss after an upgrade.

## Expected behavior

Use the existing self-expiring NEW marker on the STT model tab, transcription prompts, preset STT controls, the MOSS speaker-aware setting, and Jobs. The marker remains independent from the active-job count and does not change stable navigation labels.

## Dependencies

This is intentionally a draft stacked on #875 and also includes #881 for integration verification. After both dependencies merge, it will be retargeted to develop and the resulting focused diff will be verified again.

## Validation

All 282 UI tests pass. Lint completes with zero errors, and the production build succeeds.


hedhoud:
Superseded by a clean follow-up from develop. This draft includes unrelated dependency history, so the replacement will contain only the temporary NEW badge mechanism and its intended UI placements.

PR #892, opened 2026-09-02
fix: remove leftover refactor scaffolding

### LINKED ISSUE #888: Leftover refactor scaffolding: openrag/consts.py shim with 4 consumers, unused SingletonMeta, empty services/events/
Three artefacts of the hexagonal migration are still in the tree:

- `openrag/consts.py` is a re-export shim over `core.utils.conts` with four live consumers: `api/dependencies/files.py:8`, `api/dependencies/llm.py:1`, `api/routers/user/chat.py:18` and `app_front.py:16`. Because the shim sits outside the four layer roots, an `api/` file importing it registers as importing nothing at all as far as `check_layer_imports.py` is concerned.
- `core/utils/singleton.py` defines `SingletonMeta`, described in its own docstring as being "for legacy adapter objects". It has zero importers in `openrag/`, `tests/` or `scripts/`.
- `openrag/services/events/` contains only a 0-byte `__init__.py`.

Suggested fix: repoint the four consumers at `core.utils.conts`, delete the shim, and delete `singleton.py` and the empty package. The target module name is also misspelled (`conts.py` rather than `consts.py`), which is worth correcting while its 8 importers are being touched anyway.

Ahmath-Gadji:
## Summary
- Deletes `openrag/consts.py` (a re-export shim outside the layer roots, which hid its four importers from `check_layer_imports.py`), the unused `SingletonMeta` metaclass, and the empty `services/events/` package
- Repoints the four shim consumers (`api/dependencies/llm.py`, `api/dependencies/files.py`, `api/routers/user/chat.py`, `app_front.py`) directly at `core.utils.consts`
- Renames `core/utils/conts.py` -> `consts.py` to fix the module name typo while its importers are already being touched

Closes #888

## Test plan
- [x] `uv run ruff check openrag/ tests/` / `ruff format --check`
- [x] `python scripts/check_layer_imports.py` passes
- [x] `uv run pytest tests/unit/` — 2764 passed (1 pre-existing, unrelated failure in `test_compose_storage.py` from a local docker-compose bind-mount edit, not part of this branch)
- [x] Verified at runtime that `import consts` now raises `ModuleNotFoundError` and all repointed modules import cleanly

## Summary by CodeRabbit

* **New Features**
  * Added centralized handling for shared constants and metadata sanitization.
  * Added protection against exposing or spoofing server-managed metadata fields.

* **Bug Fixes**
  * Corrected module references for constants and metadata utilities, improving reliability across uploads, indexing, chat, and search workflows.

* **Refactor**
  * Consolidated shared constants under a common utility module.
  * Removed legacy constant re-exports and singleton enforcement.


EnjoyBacon7 (approved):
Reviewed the full diff. Mechanically complete and behavior-preserving.

Verified:
- No surviving references to `conts`, `SingletonMeta`, `utils.singleton`, or `services.events` anywhere in the repo, including docs, .github/, infra/ and conf/.
- AST-based import resolution across `openrag/` and `tests/`: zero unresolvable internal imports.
- No dynamic references that a symbol grep would miss (importlib, sys.modules stubbing, monkeypatch/patch string targets, spec_from_file_location). `tests/unit/test_app_front_secret.py` loads app_front.py by path but only syspath_prepends `openrag/`, so the new import resolves identically.
- All attributes used via the widened module import exist (`FILE_READ_CHUNK_SIZE`, `PARTITION_PREFIX`, `LEGACY_PARTITION_PREFIX`). The shim only re-exported 4 names, so this is strictly additive, with no shadowing and no cycle risk.
- `scripts/check_layer_imports.py` passes. The repoint turns an invisible top-level import into a properly classified api -> core edge.
- `ruff check` and `ruff format --check` both clean.
- Packaging unaffected: pyproject uses `packages.find include = ["openrag*"]`, so removing `services/events` needs no manifest edit.

Fully closes #888.

Minor note, not blocking and pre-existing: `consts.py` also holds behavior (`is_internal_metadata_key`, `strip_internal_metadata`, `strip_protected_metadata`), which is an odd fit for a constants module.

PR #896, opened 2026-09-03
docs(releasing): add step 9 — verify the GitHub Release object

andyne13:
v2.2.0 shipped with an empty release `name`, so GitHub rendered the tagged commit's subject as the title: "Merge pull request #895 from linagora/release/v2.2.0". Steps 1-8 verify what was built; nothing checked the Release object itself.

Step 9 gates on title, tag, draft, prerelease and non-empty notes. Verified against v2.2.0 and v2.1.1 (PASS) and five synthetic broken states (all FAIL).

The v2.2.0 title is already corrected — metadata only, no retag.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the release-title failure example by noting that it was manually corrected after version 2.2.0.
  * Added context explaining why release notes must exceed 200 characters.
  * Documented historical release-note lengths for additional reference.
  * Expanded release guidance to make validation requirements and prior examples easier to understand during future release preparation.

### DIFF
--- a/.github/RELEASING.md
+++ b/.github/RELEASING.md
@@ -397,7 +397,10 @@ git show "$VER:infra/charts/openrag-stack/Chart.yaml" | grep -E '^version:'
 Steps 1-8 verify what was *built*. Nothing so far looks at the Release object
 users actually land on. A release whose `name` is empty renders with the tagged
 commit's subject as its title — so a perfectly good release displays as
-`Merge pull request #NNN from linagora/release/vX.Y.Z`. That happened on v2.2.0.
+`Merge pull request #NNN from linagora/release/vX.Y.Z`. v2.2.0 published with
+the commit subject as its title and was corrected by hand minutes later — which
+is the point: nothing catches it, so the title is only right when someone
+happens to notice.
 
 `gh release create` leaves `name` empty unless `--title` is passed, so this is a
 recurring trap rather than a one-off slip: **always pass `--title "$VER"`.**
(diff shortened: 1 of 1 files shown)


andyne13:
Context on where this came from, since it touches @hedhoud's v2.2.0 release.

**The release itself was fine.** I checked the substance before touching anything: the tag matches the GA regex and is an ancestor of `origin/main`, `pyproject.toml` = `2.2.0`, chart `appVersion` = `2.2.0` with chart `version` bumped `0.6.2` → `0.6.3`, notes present, not a draft, not a prerelease. Only the display title was wrong, and I have already corrected it with `gh release edit v2.2.0 --title v2.2.0` — metadata only, no retag and no rebuild.

**This was a hole in the process, not a mistake.** `gh release create` leaves `name` empty unless `--title` is passed, and nothing in `RELEASING.md` said so. When `name` is empty GitHub renders the tagged commit's subject instead, which is why the release displayed as *"Merge pull request #895 from linagora/release/v2.2.0"*. Anyone following the documented process would have produced exactly the same result.

That is the actual gap this PR closes: steps 1-8 verify what was **built** — workflow, registries, digests, image contents, tag placement, chart pins — and none of them look at the Release object users land on. A release can be correct in every respect and still present itself as a merge-commit subject.

Step 9 was executed in both directions rather than reasoned about: PASS against v2.2.0 and v2.1.1, FAIL against five synthetic broken states (empty name, commit-subject name, draft, GA flagged prerelease, empty notes) — including the exact v2.2.0 case.

hedhoud (commented):
Minor: the release-notes check counts whitespace characters. A body containing more than 200 spaces or newlines would pass even though it has no meaningful release notes. Please ignore surrounding or all-whitespace content before applying the 200-character threshold.

PR #901, opened 2026-09-04
feat(api)!: extra as JSON object, nested chunk metadata, reranker score

Ahmath-Gadji:
Three changes to the chat response's `extra` payload. Two are breaking; landing them together means clients rewrite their reader once, not twice.

## 1. `extra` is a JSON object, not a JSON-encoded string

It was double-encoded — a string of JSON inside a JSON body — so every client had to `json.loads` an already-parsed field, and it rendered as an unreadable blob in Swagger.

**Watch out:** the mid-stream empty sentinel goes from `"{}"` (truthy) to `{}` (falsy), so a client guarding `if chunk.extra:` takes the other branch. Same outcome, but it's the only part that fails silently — everything else raises `TypeError`.

## 2. Source entries nest the chunk's metadata under `chunk`

```json
{
  "source_type": "document",
  "chunk": { "...the chunk's metadata, unchanged..." },
  "rerank_score": 0.646,
  "chunk_url": "...",
  "file_url": "..."
}
```

Nothing dropped — every field the flat entry had is still there. Also makes the spoofing guard structural: `source_type`/`chunk_url`/`file_url` can no longer collide with metadata, and are scrubbed from inside `chunk` too. Web entries stay flat (no chunk, no file).

## 3. The reranker score is no longer discarded

`_rerank_chunks` used the index from `[(index, score), ...]` and dropped the score. It now rides on `ScoredChunk` — declared and exported since the refactor, never instantiated. It was a flat sibling of `Chunk` that couldn't travel a `list[Chunk]` pipeline, so it's now a `Chunk` subclass and the score is a typed field.

Scores default to `None`, not `0.0`, so "no reranker ran" stays distinguishable from "scored zero".

## Compatibility

Chainlit accepts both shapes for `extra` and for source entries, so a rolling deploy works in either direction.

## Decided, not changed here

- **The three source arrays stay as they are.** `sources` and `cited_sources` are the same list object and `presented_sources` is a superset, so entries are serialised 2–3× (10 sources / 5 cited ≈ 62 KB, ~31 KB of it duplication). `sources` is kept for backward compatibility; the intent is to drop it in a future release, which takes most of the duplication with it.
- **`content` stays** — it's still in use.

## Open question

- **Internals ship to clients**: `source` (absolute on-disk path), `content_sha256`, raw Milvus `section_id`/`prev`/`next`.

## Testing

`2918 passed`, ruff + format clean, layer-import guard OK. 13 new tests. Reverting the `extra` producers fails 43 tests; all 61 existing app_front tests pass unchanged against the nested shape, which is what proves the fallback.

Integration tests in `test_openai_compat.py` are updated but unrun — they need a live server.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that API `extra` fields are structured JSON objects, a breaking change introduced in v2.2.0.
  * Documented source entry formats, retrieval scores, URLs, and scoring caveats.

* **Enhancements**
  * Grouped source metadata under `chunk`, with scores and authoritative links presented separately.
  * Source rendering supports both current and previous metadata formats.

* **Bug Fixes**
  * Prevented metadata from overriding server-generated source links or supplying stale retrieval scores.
  * Distinguished scoring stages when no score is available.


openrag/api/routers/user/source_links.py
```diff
@@ -30,14 +49,17 @@ def build_document_source_link(
...
+
+    link = {
+        "source_type": "document",
+        "chunk": {
+            key: value
+            for key, value in doc_metadata.items()
+            if key not in _AUTHORITATIVE_KEYS and key not in _SCORE_KEYS
+        },
+    }
+    link.update({key: doc_metadata[key] for key in _SCORE_KEYS if key in doc_metadata})
```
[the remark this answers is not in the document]
Ahmath-Gadji: Valid, and worse than stale — spoofable. Milvus collections have a dynamic field, so `metadata: {"rerank_score": 0.99}` sent at upload is persisted verbatim and read back on every query, then promoted here as if this retrieval had computed it.

Closed at both boundaries rather than by threading `ScoredChunk` through the response builder (the typed provenance is already gone by `chat.py::__prepare_sources`, which works on `Document.metadata`):

- `vector_store_searcher._dict_to_chunk` drops the three score keys along with `score`, so persisted ones never enter `Chunk.metadata`.
- `ScoredChunk.to_langchain()` clears them from inherited metadata before stamping its typed fields, so the emitted keys are exactly the non-null ones this retrieval set.

Both ends share `RETRIEVAL_SCORE_KEYS` (`core/utils/consts.py`) so the guard and the promotion can't drift apart. Tests added in `test_vector_store_searcher.py` and `test_pipeline.py`. Amended into the same commit.

tests/load/automatic-evaluation-pipeline/benchmark.py
```diff
@@ -39,7 +39,7 @@ async def retrieve_response_and_docs_openrag(query: str, partition: str, _base_u
         try:
             res = await client.chat.completions.create(**settings)
             response_llm = res.choices[0].message.content
-            list_source_chunk_ids = [item["_id"] for item in json.loads(res.extra)["sources"]]
+            list_source_chunk_ids = [item["_id"] for item in res.extra["sources"]]
```
hedhoud: The benchmark still reads document IDs from the old top-level source shape. With this PR, document metadata is nested under `chunk`, so a valid response can raise here and the broad exception handler turns the evaluation into an empty ID list, producing misleading retrieval metrics. Please update this consumer to read the nested document ID while keeping non-document sources safe.

hedhoud:
I checked the other compatibility concern about preserving top-level document metadata aliases. I do not think aliases are needed here: this is an intentional breaking API change, the new contract keeps chunk metadata under `chunk`, and the existing tests explicitly protect that separation. The current in-repository UI already handles both the legacy flat shape and the new nested shape.

tests/load/automatic-evaluation-pipeline/benchmark.py
Ahmath-Gadji: Valid — and the failure mode is worse than a raised error, as you say: the `except` returns `(None, [])`, so a question that retrieved perfectly scores 0 on hit-rate, MRR, recall and nDCG. A regression in the benchmark would read as a regression in retrieval.

Fixed in f4d51bc0. Extracted to `_source_chunk_ids`, which reads `chunk["_id"]` and falls back to the flat `_id` — the same tolerance `app_front._chunk_field` grew in this branch, so a benchmark run works against either side of a rolling deploy. `extra` is normalised the same way, since a server from before this branch sends it as a JSON string.

Entries with no chunk id are skipped rather than left to raise. That part was already latent before the nesting change: web results are flat and keyed by `url`, so any run with `websearch` on would have zeroed the same metrics.

Checked the other consumers: `eval_frames.py` only reads top-level `chunk_url`, which stays authoritative in both shapes, and it already normalises a string `extra` — so it needs nothing. `benchmark.py` was the only one.

The helper is verified against the seven entry shapes (nested, flat, mixed, web-only, `None`, non-dict entry, malformed `chunk`); 2920 unit tests pass, ruff clean. Agreed on your second point — no aliases; the nested contract stands on its own.

hedhoud (approved):
Reviewed the latest commit. All checks pass, the prior review comments are addressed, and no actionable issues remain.

Issue #846, opened 2026-07-31
Default PDF parser holds the whole document in memory (2 copies on the #640 recovery path); Indexer actors have no max_restarts


andyne13:
## Summary

The default PDF parser holds the entire document in memory, and on one recovery path holds two full copies at once. The Indexer actors that run it have no restart policy, so a worker that OOMs is not recovered the way Marker/Docling workers are.

This is the memory-residency half of a third-party audit finding. The *admission control* half is already tracked in #663 — this issue is deliberately scoped to per-file residency and worker recovery.

## 1. The default PDF parser is fully resident

`openrag/core/indexing/parsers/pdf/pymupdf.py` opens from **bytes**, not a path:

```python
def _extract_markdown(raw: bytes, filename: str):
    with pymupdf.open(stream=raw, filetype="pdf") as doc:      # whole file resident
        try:
            chunks = _to_markdown(doc)
        except RuntimeError:
            cleaned = doc.tobytes(garbage=4, clean=True)       # second full copy
            with pymupdf.open(stream=cleaned, filetype="pdf") as clean_doc:
                chunks = _to_markdown(clean_doc)
```

`raw` and `cleaned` are both live inside the `except` block, so peak usage is roughly double — on the malformed-document recovery path added in #640, which is exactly where unusual (often large) documents end up.

`pymupdf4llm.to_markdown(doc, page_chunks=True, ...)` chunks the *output* per page; the document is already fully loaded by then.

**This is the shipped default in all three places:** `conf/config.yaml:202` (`pdf: PyMuPDFLoader`), `infra/compose/.env.example:46`, and `infra/charts/openrag-stack/values.yaml:666`. So a default chart or compose deployment is on this path.

**Marker is not affected** — it splits into page ranges (`marker_workers.py:293-298`) and passes a *path* plus `page_range` to each worker (`marker_workers.py:176-191`), so its memory is bounded per batch. Anyone who has switched `PDFLOADER=MarkerLoader` is not exposed here.

## 2. Indexer actors have no restart policy

`indexer_pool.py:41` and `:355` declare bare `@ray.remote`, with no `max_restarts`. Marker and Docling deliberately set it:

- `marker_workers.py:82` and `docling_workers.py:85,97` → `@ray.remote(max_restarts=5)`, with comments explaining that a worker which OOMs or CUDA-faults must come back.

So the tier most likely to OOM on a large PDF is the one with no recovery configured, while the GPU tiers have it.

[...]

PR #903, opened 2026-09-07
fix(jobs): bound in-memory task state retention

EnjoyBacon7:
Fixes the memory leak in #660.

Finished indexing tasks were never removed from the in-memory task state actor.
It is detached, so it survives restarts and grows until it runs out of memory:
51 MB for 20,000 files, about 2.5 KB each.

Finished tasks are now dropped after 1h or past a 2,000 record cap. Running
tasks and unsettled cancellations are kept, so cancellation still fences its
worker. Tracebacks are capped at 8,000 characters. Status polling now 404s for
tasks finished over 1h ago, except a cancellation: its record is the fence that
stops a late worker, so it is held for the tombstone TTL instead.

The retention window is set independently of the cancellation tombstone TTL,
which stays at 24h: the tombstone fences late workers rather than answering
reads. Postgres holds the durable job history (#904), so the in-memory copy
only has to cover the reads that follow a job settling.

## Summary by CodeRabbit

* **Improvements**
  * Terminal task records are retained for a limited time and capped in number, reducing stale task data while preserving active tasks.
  * Recovered completed tasks retain their original expiration schedule.
  * Retention cleanup runs when checking task status or listing tasks.
  * Terminal records are evicted promptly after settlement, including during rapid bursts of completed tasks.
  * Recovery stops monitoring tasks that are no longer available instead of polling indefinitely.
* **Tests**
  * Added coverage for retention, eviction, recovery, active-task preservation, and cancellation-fence handling.

### DIFF
--- a/openrag/services/workers/task_state.py
+++ b/openrag/services/workers/task_state.py
@@ -302,8 +302,8 @@ class TaskStateManager:
         self.lock = threading.Lock()
 
     def _ensure_task(self, task_id: str) -> TaskInfo:
-        # Every path that can add a task goes through here, and none of them
-        # iterate ``self.tasks``, which makes this the one safe place to evict.
+        # Every path that can add a task goes through here, so admission always
+        # sheds history first.
         self._evict_terminal_tasks_locked()
         if task_id not in self.tasks:
             self.tasks[task_id] = TaskInfo()
@@ -331,7 +331,11 @@ class TaskStateManager:
         _delete_recoverable_task(task_id)
 
     def _evict_terminal_tasks_locked(self, *, now: float | None = None) -> None:
-        """Drop settled tasks once they age out or the retention cap is exceeded."""
+        """Drop settled tasks once they age out or the retention cap is exceeded.
+
+        This removes entries from ``self.tasks`` and ``self.user_index``, so call
+        it before reading either, never while iterating one.
+        """
         deadline = (time.time() if now is None else now) - _TERMINAL_TASK_RETENTION_SECONDS
         examined = 0
         while self.terminal_tasks and examined < len(self.terminal_tasks):
@@ -641,8 +645,6 @@ class TaskStateManager:
     @ray.method(concurrency_group="get")
     async def get_state(self, task_id: str) -> str | None:
         with self.lock:
-            # Status polling is the one read that cannot iterate the task map,
-            # so it can evict. Without this, retention would need new work.
             self._evict_terminal_tasks_locked()
             info = self.tasks.get(task_id)
             if info is not None:
@@ -747,12 +749,16 @@ class TaskStateManager:
     @ray.method(concurrency_group="queue_info")
     async def get_all_states(self) -> dict[str, str | None]:
         with self.lock:
+            # A queue that only gets polled admits no task, so the listing reads
+            # have to shed history or retention never runs.
+            self._evict_terminal_tasks_locked()
             self._expire_refless_tasks_if_stale_locked()
             return {tid: info.state for tid, info in self.tasks.items()}
 
     @ray.method(concurrency_group="queue_info")
     async def get_all_info(self) -> dict[str, dict]:
         with self.lock:
+            self._evict_terminal_tasks_locked()
             self._expire_refless_tasks_if_stale_locked()
             return {
                 task_id: {
@@ -768,6 +774,8 @@ class TaskStateManager:
     @ray.method(concurrency_group="queue_info")
     async def get_all_user_info(self, user_id: int) -> dict[str, dict]:
         with self.lock:
+            # Before the index lookup: eviction can drop the user's whole entry.
+            self._evict_terminal_tasks_locked()
             task_ids = self.user_index.get(user_id, set())
             self._expire_refless_tasks_if_stale_locked(task_ids)
             return {
--- a/openrag/services/workers/task_state.py
+++ b/openrag/services/workers/task_state.py
@@ -4,6 +4,7 @@ import base64
 import json
 import threading
 import time
+from collections import OrderedDict
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from datetime import UTC, datetime, timedelta
@@ -37,6 +38,13 @@ _RECOVERABLE_TASK_KV_PREFIX = b"recoverable-task-v1:"
 _CANCELLATION_TOMBSTONE_TTL_SECONDS = 24 * 60 * 60
 _FILE_DELETE_FENCE_TTL_SECONDS = 2 * 60
 _CONTENT_CLAIM_REGISTRATION_GRACE_SECONDS = 60
+# Terminal task records are progress receipts, not the system of record: the
+# durable per-file state lives in the Postgres catalog. Retention is aligned
+# with the cancellation tombstone TTL so the in-memory row and its durable
+# tombstone disappear together, with a hard cap so a burst cannot outrun it.
+_TERMINAL_TASK_RETENTION_SECONDS = _CANCELLATION_TOMBSTONE_TTL_SECONDS
+_MAX_TERMINAL_TASKS = 10_000
+_MAX_TASK_ERROR_CHARS = 8_000
 STALE_REFLESS_TASK_ERROR = (
     "Indexing task never exposed a worker reference within the registration grace period; marking it failed as stale."
 )
@@ -196,6 +204,23 @@ def _save_recoverable_task(task_id: str, info: TaskInfo) -> None:
         _internal_kv_del(key, namespace=_task_state_kv_namespace())
 
 
+def _delete_recoverable_task(task_id: str) -> None:
+    from ray.experimental.internal_kv import _internal_kv_del
+
+    if not _task_state_storage_available():
+        return
+    _internal_kv_del(_recoverable_task_key(task_id), namespace=_task_state_kv_namespace())
+
+
+def _truncate_error(tb_str: str | None) -> str | None:
+    """Keep the tail of a traceback: the raising frame and message live there."""
+    if tb_str is None or len(tb_str) <= _MAX_TASK_ERROR_CHARS:
+        return tb_str
+    marker = "...[truncated]...\n"
+    keep = max(_MAX_TASK_ERROR_CHARS - len(marker), 0)
+    return marker + tb_str[len(tb_str) - keep :]
+
+
 try:
     from core.config import load_config as _load_config
 
@@ -254,17 +279,64 @@ class TaskStateManager:
     def __init__(self) -> None:
         self.tasks = _load_recoverable_tasks()
         self.user_index: dict[int | None, set[str]] = {}
+        # Terminal task ids in eviction order, mapped to the time they settled.
+        self.terminal_tasks: OrderedDict[str, float] = OrderedDict()
+        now = time.time()
         for task_id, info in self.tasks.items():
             self.user_index.setdefault(info.details.get("user_id"), set()).add(task_id)
+            if info.state in TERMINAL_TASK_STATES:
+                self.terminal_tasks[task_id] = now
         self.file_delete_fences = _load_file_delete_fences()
         # Ray runs each concurrency group on a separate event loop. A single
         # asyncio lock cannot safely coordinate methods across those loops.
         self.lock = threading.Lock()
 
     def _ensure_task(self, task_id: str) -> TaskInfo:
+        # Every path that can add a task goes through here, and none of them
+        # iterate ``self.tasks``, which makes this the one safe place to evict.
+        self._evict_terminal_tasks_locked()
         if task_id not in self.tasks:
             self.tasks[task_id] = TaskInfo()
         return self.tasks[task_id]
+
+    def _persist_task_locked(self, task_id: str, info: TaskInfo) -> None:
+        """Persist a task mutation and keep the terminal-retention ledger in sync."""
+        _save_recoverable_task(task_id, info)
+        if info.state in TERMINAL_TASK_STATES:
+            self.terminal_tasks.pop(task_id, None)
+            self.terminal_tasks[task_id] = time.time()
+        else:
+            self.terminal_tasks.pop(task_id, None)
+
+    def _forget_task_locked(self, task_id: str) -> None:
+        info = self.tasks.pop(task_id, None)
+        if info is None:
+            return
+        user_id = (info.details or {}).get("user_id")
+        owned = self.user_index.get(user_id)
+        if owned is not None:
+            owned.discard(task_id)
+            if not owned:
+                self.user_index.pop(user_id, None)
+        _delete_recoverable_task(task_id)
+
+    def _evict_terminal_tasks_locked(self, *, now: float | None = None) -> None:
+        """Drop settled tasks once they age out or the retention cap is exceeded."""
+        deadline = (time.time() if now is None else now) - _TERMINAL_TASK_RETENTION_SECONDS
+        examined = 0
+        while self.terminal_tasks and examined < len(self.terminal_tasks):
+            task_id, settled_at = next(iter(self.terminal_tasks.items()))
+            if len(self.terminal_tasks) <= _MAX_TERMINAL_TASKS and settled_at > deadline:
+                break
+            info = self.tasks.get(task_id)
+            if info is not None and _cancelled_task_has_worker_fence(info):
+                # An unsettled cancellation still fences a live worker. Losing it
+                # would let that worker write after the file was cancelled.
+                self.terminal_tasks.move_to_end(task_id)
+                examined += 1
+                continue
+            self.terminal_tasks.pop(task_id)
+            self._forget_task_locked(task_id)
 
     def _record_details(
         self,
@@ -298,7 +370,7 @@ class TaskStateManager:
 
     def _set_cancelled_locked(self, task_id: str, info: TaskInfo) -> None:
         info.state = "CANCELLED"
-        _save_recoverable_task(task_id, info)
+        self._persist_task_locked(task_id, info)
 
     def _expire_refless_task_if_stale_locked(self, task_id: str, info: TaskInfo) -> bool:
         ref = info.object_ref.get("ref") if isinstance(info.object_ref, dict) else info.object_ref
@@ -316,7 +388,7 @@ class TaskStateManager:
             return False
         info.state = "FAILED"
         info.error = STALE_REFLESS_TASK_ERROR
-        _save_recoverable_task(task_id, info)
+        self._persist_task_locked(task_id, info)
         return True
 
     def _expire_refless_tasks_if_stale_locked(self, task_ids: Iterable[str] | None = None) -> None:
@@ -396,15 +468,15 @@ class TaskStateManager:
             if state == "SERIALIZING":
                 info.worker_submitted = True
                 info.submission_started_at = None
-            _save_recoverable_task(task_id, info)
+            self._persist_task_locked(task_id, info)
             return True
 
     @ray.method(concurrency_group="set")
     async def set_error(self, task_id: str, tb_str: str) -> None:
         with self.lock:
             info = self._ensure_task(task_id)
-            info.error = tb_str
-            _save_recoverable_task(task_id, info)
+            info.error = _truncate_error(tb_str)
+            self._persist_task_locked(task_id, info)
 
     @ray.method(concurrency_group="set")
     async def set_failed_if_not_cancelled(self, task_id: str, tb_str: str) -> bool:
@@ -420,8 +492,8 @@ class TaskStateManager:
                 return False
             if info is not None:
                 info.state = "FAILED"
-                info.error = tb_str
-                _save_recoverable_task(task_id, info)
+                info.error = _truncate_error(tb_str)
+                self._persist_task_locked(task_id, info)
             return True
 
     @ray.method(concurrency_group="set")
@@ -446,7 +518,7 @@ class TaskStateManager:
             info.object_ref = None
             info.worker_submitted = False
             info.submission_started_at = None
-            _save_recoverable_task(task_id, info)
+            self._persist_task_locked(task_id, info)
             return True
 
     @ray.method(concurrency_group="set")
@@ -462,7 +534,7 @@ class TaskStateManager:
             if info.state in CANCELLABLE_INDEXING_STATES:
                 info.state = "FAILED"
                 info.error = "Indexer worker submission was rejected after the worker settled."
-            _save_recoverable_task(task_id, info)
+            self._persist_task_locked(task_id, info)
             return True
 
     @ray.method(concurrency_group="set")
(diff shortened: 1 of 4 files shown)


openrag/services/workers/task_state.py
```diff
@@ -254,18 +279,65 @@ class TaskStateManager:
     def __init__(self) -> None:
         self.tasks = _load_recoverable_tasks()
         self.user_index: dict[int | None, set[str]] = {}
+        # Terminal task ids in eviction order, mapped to the time they settled.
+        self.terminal_tasks: OrderedDict[str, float] = OrderedDict()
+        now = time.time()
         for task_id, info in self.tasks.items():
             self.user_index.setdefault(info.details.get("user_id"), set()).add(task_id)
+            if info.state in TERMINAL_TASK_STATES:
```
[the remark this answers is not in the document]
EnjoyBacon7: Not applying this one. The constructor uses `TERMINAL_TASK_STATES` from `core.models.catalog`, which includes `CANCELLED`. `TERMINAL_INDEXING_STATES` is a different constant that this code does not use. The assertion at line 1028 passes on this branch.

Tracking fenced cancellations in the ledger is deliberate. Eviction already skips them while the fence is held, so the record becomes evictable once cancellation settles. Excluding them from the ledger would leave a settled cancellation that never gets touched again untracked forever.

openrag/services/workers/task_state.py
```diff
@@ -254,18 +279,65 @@ class TaskStateManager:
...
+                self.terminal_tasks[task_id] = now
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in 88fc6233. `_load_recoverable_tasks` now returns the persisted deadlines alongside the tasks, and the constructor seeds the settle time from the deadline rather than the recovery time. A record recovered close to its deadline now expires on its original schedule instead of getting a second full window.

Covered by `test_recovery_preserves_the_original_retention_deadline`.

openrag/services/workers/task_state.py
```diff
@@ -254,18 +279,65 @@ class TaskStateManager:
...
         self.file_delete_fences = _load_file_delete_fences()
         # Ray runs each concurrency group on a separate event loop. A single
         # asyncio lock cannot safely coordinate methods across those loops.
         self.lock = threading.Lock()
 
     def _ensure_task(self, task_id: str) -> TaskInfo:
+        # Every path that can add a task goes through here, and none of them
+        # iterate ``self.tasks``, which makes this the one safe place to evict.
+        self._evict_terminal_tasks_locked()
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in 88fc6233. `get_state` now evicts before the lookup, so a queue that goes quiet still sheds its history and the status route returns 404 once a record ages out.

Scoped to that one method on purpose. The queue listing reads iterate the task map, so evicting inside them would mutate a dict mid-iteration.

Covered by `test_expired_task_is_evicted_when_its_status_is_polled`.

hedhoud:
Please handle the remaining retention gap from review: terminal task records should be evicted when clients poll queue/list data, not only when new tasks arrive or an individual task is queried. Please cover `get_all_states`, `get_all_info`, and `get_all_user_info` so the 24-hour retention window and 10,000-task cap are enforced even when the workload is read-only.

EnjoyBacon7:
Fixed in e8c86ad7. `get_all_states`, `get_all_info` and `get_all_user_info` now evict before they read the task map, so a read-only workload still enforces the 24h window and the 10,000 cap.
My earlier reason for scoping this to `get_state` was wrong: the mid-iteration hazard only rules out evicting during the comprehension, not before it. In `get_all_user_info` the call has to come before the `user_index` lookup, since eviction can drop the user's entry.
Covered by a parametrized regression test over the three reads.

hedhoud:
Please address these task-state retention and rolling-deployment issues before merge:

- Settled terminal records are only evicted on a later admission or queue read, so a burst of completed tasks can exceed the cap and TTL.
- Receipt eviction currently removes cancellation tombstones after the one-hour receipt window, although the cancellation fence is intended to last 24 hours. These lifetimes need to be independent.
- Recovered cancelled tasks derive receipt retention from the tombstone deadline, causing receipts to live for roughly 24 hours instead of one hour.
- The detached `TaskStateManager` can be reused during a rolling deployment because the compatibility check does not identify the retention-aware generation, leaving the old actor active until manual restart.

Please include regression coverage for the retention boundaries and actor replacement/reuse behavior.

openrag/services/workers/task_state.py
```diff
@@ -252,20 +283,77 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
         if task_id not in self.tasks:
             self.tasks[task_id] = TaskInfo()
         return self.tasks[task_id]
 
+    def _persist_task_locked(self, task_id: str, info: TaskInfo) -> None:
+        """Persist a task mutation and keep the terminal-retention ledger in sync."""
+        _save_recoverable_task(task_id, info)
+        if info.state in TERMINAL_TASK_STATES:
+            self.terminal_tasks.pop(task_id, None)
+            self.terminal_tasks[task_id] = time.time()
```
hedhoud: This records the terminal timestamp but does not trigger `_evict_terminal_tasks_locked()`. If tasks settle in a burst without a later admission or queue read, the in-memory ledger can exceed both the TTL and cap. Please evict here or arrange an independent cleanup path.

openrag/services/workers/task_state.py
```diff
@@ -252,20 +283,77 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
+        now = time.time()
         for task_id, info in self.tasks.items():
             self.user_index.setdefault(info.details.get("user_id"), set()).add(task_id)
+            if info.state not in TERMINAL_TASK_STATES:
+                continue
+            # A restart must not restart the clock, so recover the settle time
+            # from the persisted deadline where the record carries one.
+            expires_at = expiries.get(task_id)
+            self.terminal_tasks[task_id] = (
+                expires_at - _TERMINAL_TASK_RETENTION_SECONDS if expires_at is not None else now
```
hedhoud: For a recovered terminal task, this reconstructs `settled_at` from the persisted expiry. A cancellation tombstone can carry the 24-hour deadline, which makes the receipt live until that deadline rather than for the intended one-hour retention. Persist a separate receipt deadline.

openrag/services/workers/task_state.py
```diff
@@ -252,20 +283,77 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
+        info = self.tasks.pop(task_id, None)
+        if info is None:
+            return
+        user_id = (info.details or {}).get("user_id")
+        owned = self.user_index.get(user_id)
+        if owned is not None:
+            owned.discard(task_id)
+            if not owned:
+                self.user_index.pop(user_id, None)
+        _delete_recoverable_task(task_id)
```
hedhoud, line 354: Receipt eviction reaches this method and deletes the recoverable record, including a cancellation tombstone. That couples the one-hour terminal receipt lifetime to the 24-hour cancellation fence, so a restart after one hour can lose sticky cancellation state. Keep tombstone deletion on its own expiry.

openrag/services/workers/task_state.py
```diff
@@ -37,6 +38,15 @@
 _CANCELLATION_TOMBSTONE_TTL_SECONDS = 24 * 60 * 60
 _FILE_DELETE_FENCE_TTL_SECONDS = 2 * 60
 _CONTENT_CLAIM_REGISTRATION_GRACE_SECONDS = 60
+# Terminal task records are progress receipts, not the system of record: the
+# durable per-file state lives in the Postgres catalog. They are kept only long
+# enough to answer the reads that follow a job settling, then dropped, with a
+# hard cap so a burst cannot outrun the time bound. The cancellation tombstone
+# keeps its own, longer TTL: it fences late workers rather than answering reads,
+# so the two lifetimes are deliberately not tied together.
+_TERMINAL_TASK_RETENTION_SECONDS = 60 * 60
```
hedhoud, line 48: The retention behavior changes the detached actor's lifecycle, but there is no generation/feature marker for bootstrap to distinguish this actor from an older retained `TaskStateManager`. During a rolling deployment, `get_if_exists` can therefore keep the old implementation running. Please version the actor or add a capability check that forces replacement.

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in b65ebd64. The settle paths (`set_state`, `set_failed_if_not_cancelled`, `set_cancelled_if_active`) sweep after persisting, so a burst that ends the queue sheds its own history instead of waiting for the next caller. The listing reads now sweep after the stale pass rather than before it, so anything that pass just failed is covered in the same call.

Not inside `_persist_task_locked` itself, though: `_expire_refless_tasks_if_stale_locked` iterates `self.tasks` and reaches it, and eviction pops from that dict. Sweeping there would raise mid-iteration on the queue listings.

Covered by `test_a_settling_burst_sheds_history_with_nothing_else_running`.

openrag/services/workers/task_state.py
EnjoyBacon7: Half fixed in b65ebd64, and I would like your read on the other half.

The arithmetic is gone: the ledger stores the deadline itself, so recovery reuses the persisted value instead of reconstructing a settle time from it. Eviction timing there is unchanged, since `expires_at - retention <= now - retention` and `expires_at <= now` are the same test.

What I have not done is split the receipt from the fence. The in-memory record is the fence: `set_state`, `set_failed_if_not_cancelled` and `set_queued_details` all read `info.state` to refuse a late writer, so dropping the record at the receipt window drops that guard whatever the KV still holds. A 1h receipt with a 24h fence needs a second structure consulted by each of those guards, plus `_ensure_task` no longer being allowed to recreate a blank record for a fenced id. That is a fair amount of new surface in the cancellation path for memory the 2,000 record cap already bounds.

So the rule is now explicit in both directions: a record lives to the later of the two deadlines, a fence is never cut short by the receipt window, and a receipt never inherits a fence deadline. The cost is that a cancelled task stays readable for as long as its fence is held. Say the word and I will do the split instead.

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in b65ebd64. A record with a live tombstone is no longer evicted on the receipt window, so `_forget_task_locked` and the KV delete only run once the fence itself has expired.

The ledger holds absolute deadlines now, and a fence sitting at the head no longer stalls the sweep for the receipts queued behind it.

Covered by `test_cancellation_fence_outlives_the_receipt_window`, which asserts `_delete_recoverable_task` is not called at the receipt window and is at the fence deadline, and by `test_an_expired_receipt_behind_a_fence_is_still_evicted`.

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in b65ebd64. `TaskStateManager.supports_bounded_task_retention` is the marker and `_supports_task_state_recovery` requires it, so an actor left detached by an earlier deployment is replaced during bootstrap the way one without restart support already is.

Covered by `test_task_state_manager_without_bounded_retention_is_replaced`.

Ahmath-Gadji: The marker does its job, but replacing the actor has a knock-on effect on the detached indexer actors.

`ray.kill(actor, no_restart=True)` plus re-creation gives `TaskStateManager` a new actor id. `_INDEXER_ACTOR_PROTOCOL_VERSION` is still `v7`, so on an external Ray cluster (Helm `ray.enabled`) the upgraded API reattaches to the existing `IndexerPoolDispatcher-v7` / `IndexerWorker-v7-*` actors. Those still hold handles to the killed actor:

- `IndexerPool._task_state_manager`, cached in `__init__` (`indexer_pool.py:139`) and in `_task_state_actor()` (`indexer_pool.py:737`)
- `IndexerWorkerActor._tsm`, passed in at construction

From then on `set_object_ref`, `set_state(SERIALIZING/COMPLETED)` and `set_failed_if_not_cancelled` all target a dead actor. `retry_idempotent_ray_actor_method` retries for 30s and then raises, so every upload fails until someone retires the v7 generation by hand. Compose isn't affected, because Ray restarts with the API container there.

Could this PR bump the version to `v8`? That also means updating the actor names in `test_indexer_pool.py` and the rolling-deploys note in `CLAUDE.md`.

openrag/services/workers/task_state.py
```diff
@@ -252,20 +293,96 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
         # Ray runs each concurrency group on a separate event loop. A single
         # asyncio lock cannot safely coordinate methods across those loops.
         self.lock = threading.Lock()
 
     def _ensure_task(self, task_id: str) -> TaskInfo:
+        # Every path that can add a task goes through here, so admission always
+        # sheds history first.
+        self._evict_terminal_tasks_locked()
         if task_id not in self.tasks:
             self.tasks[task_id] = TaskInfo()
```
Ahmath-Gadji, line 319: **Evicted tasks come back as `state=None` records that are never evicted**

Now that records get dropped, a write that arrives after eviction comes through here and creates a blank `TaskInfo()`. Its state is `None`, so `_persist_task_locked` never adds it to `terminal_tasks`, and it stays in `tasks` / `user_index` forever.

Two ways to hit it:
- `TaskCompletionTracker._record_finished_at` (`task_completion.py:162`) calls `get_details`, which doesn't sweep, and then `set_details`. If the record expires between those two calls, `set_details` evicts it inside `_ensure_task` and then recreates it with details but no state. `dispatcher._record_finished_at` follows the same pattern.
- A late `set_state(task_id, "SERIALIZING")` on an evicted id. The write is refused because there's no ref, but the blank record has already been inserted, and it is never persisted or tracked.

The effect downstream: `JobService.list_tasks(task_status="completed")` calls `i["state"].lower()` (`job_service.py:109`), which raises `AttributeError` and returns a 500 on that user's task list (or the admin's).

Reproduced on b65ebd64:

```python
monkeypatch.setattr(task_state_module, "_TERMINAL_TASK_RETENTION_SECONDS", 60.0)
# t=1000
await m.set_queued_details("done", file_id="f", partition="p", metadata={}, user_id=7)
await m.set_state("done", "COMPLETED")
details = await m.get_details("done")
# t=1061: expired between the tracker's two calls
await m.set_details("done", file_id="f", partition="p", metadata=details["metadata"], user_id=7)
assert m.tasks["done"].state is None
assert "done" not in m.terminal_tasks
# t=1_000_000
assert (await m.get_all_user_info(7))["done"]["state"] is None
```

Two possible fixes:
1. Only admission paths (`set_queued_details`, `set_state(..., "QUEUED")`) create records.
2. Put stateless records in the ledger too, and don't leave a record behind when a write is refused.

The second is less invasive, since a few existing tests rely on writes creating records (e.g. `test_delete_cleanup_still_fences_legacy_indexing_states`).

openrag/services/workers/task_state.py
```diff
@@ -252,20 +293,96 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
+        Admission-time eviction always runs one settle behind, so a burst that
+        ends the queue would leave its last records in memory until something
+        else touched the actor. Only call this from a public entry point:
+        eviction mutates ``self.tasks`` and would break a caller iterating it.
+        """
+        self._persist_task_locked(task_id, info)
+        self._evict_terminal_tasks_locked()
+
+    def _forget_task_locked(self, task_id: str) -> None:
+        info = self.tasks.pop(task_id, None)
```
Ahmath-Gadji, line 356: **`recover_refless` polls forever once its task is evicted**

`TaskCompletionTracker.recover_refless` (`task_completion.py:106-152`, outside this diff) only exits on one of these: `finished_at`, a stale expiry, an object ref, `CANCELLED`, or a terminal state. Once this method removes the record, none of them can happen. `get_details`, `get_state` and `get_object_ref` all return `None`, and `expire_refless_task_if_stale` returns `False`. So the loop sleeps 5s and repeats indefinitely, making 4 TSM calls per iteration, and the id stays in `_tracked_task_ids`.

This is reachable when a recovered task settles and gets evicted between two polls, for example by the cap during a burst of settles. I reproduced it on b65ebd64 by running the real tracker against a `TaskStateManager` with `_MAX_TERMINAL_TASKS = 1`: the coroutine never returns.

An early return when `get_details` returns `None` (the actor no longer knows the task) would cover it. #904 keeps the same exit conditions, so it needs the same guard.

Ahmath-Gadji:
**Merge order with #904**

With only this PR merged, a finished task's status disappears after 1h, or once 2,000 newer tasks have settled. After that, `GET /indexer/task/{id}` returns 404 and `/task/{id}/error` has nothing, and nothing else can answer either request. The description relies on Postgres holding the durable history, but that is #904, which is still open and targets `develop` independently of this branch.

A client that polls its `task_status_url` late, such as cozy-stack/Twake after a bulk upload of more than 2,000 files, would read that 404 as a lost job. Could this PR wait until #904 is merged, or go in together with it (for example with #904 stacked on this branch)?

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in 6465c53b. `_INDEXER_ACTOR_PROTOCOL_VERSION` is now `v8`, so a bootstrap replacement rolls the dispatcher and workers to a fresh generation alongside the TaskStateManager instead of leaving them attached to the killed actor. Updated the actor names in `test_indexer_pool.py` and the rolling-deploys note in CLAUDE.md.

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in b1fdc023. `_ensure_task` now gives a freshly created record a receipt deadline as soon as it exists, and `_persist_task_locked` treats `state=None` the same as a terminal state for retention (receipt window, no fence). That covers both reproductions: `set_details` recreating a stateless record after an eviction race, and `set_state(..., "SERIALIZING")` refusing a write on an unknown id without ever persisting the blank record it created. Both now age out with the rest of the ledger instead of leaking.

Covered by `test_details_written_on_an_evicted_id_does_not_leak_a_stateless_record` (your first repro) and `test_refused_write_on_an_unknown_id_does_not_leak_a_blank_record` (your second).

openrag/services/workers/task_state.py
EnjoyBacon7: Fixed in 0ab69188. `recover_refless` now returns as soon as `get_details` comes back `None`, before touching `expire_refless_task_if_stale`/`get_state`/`get_object_ref`, since none of them can tell it anything once the actor has forgotten the task.

You're right that #904 needs the same guard; applying it there separately since it carries its own copy of `task_completion.py`.

Covered by `test_recover_refless_stops_once_the_actor_forgets_the_task`, which fails an assertion if the loop ever reaches `asyncio.sleep` instead of returning on the first read.

openrag/services/workers/task_state.py
```diff
@@ -252,20 +293,96 @@ def _content_claim_registration_expired(details: dict[str, Any]) -> bool:
...
+        self.terminal_tasks: OrderedDict[str, float] = OrderedDict()
+        now = time.time()
         for task_id, info in self.tasks.items():
             self.user_index.setdefault(info.details.get("user_id"), set()).add(task_id)
+            if info.state not in TERMINAL_TASK_STATES:
+                continue
+            # A restart must not restart the clock: a record that was persisted
+            # with a fence deadline is held to that deadline, not to a fresh
+            # window. Only a fence has one, so the rest get the receipt window.
+            self.terminal_tasks[task_id] = expiries.get(task_id, now + _TERMINAL_TASK_RETENTION_SECONDS)
```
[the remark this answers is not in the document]
EnjoyBacon7: Fixed in 21c3c849. `_evict_terminal_tasks_locked`'s cap branch only checked `_cancelled_task_has_worker_fence`, which clears once the worker goes quiet, so a settled cancellation's tombstone could be forgotten under cap pressure well before its own 24h deadline, and a later `set_state` would recreate the task through `_ensure_task` with no fence. It now also checks the record's own already-computed deadline (the later of receipt and fence) before forgetting a settled cancellation or stale-refless tombstone, and steps over it instead, same as it already does outside the cap branch.

On the second point: I traced through whether the FIFO break heuristic can strand an already-expired receipt behind a not-yet-expired fence long-term, and I don't think it can. `move_to_end` runs every time a fence is stepped over, which re-appends it after whatever is currently at the tail; anything genuinely behind it afterward was inserted (or itself last moved) no earlier than that. So a receipt that's actually expired can't stay stuck behind a fence indefinitely, just possibly for one sweep. I didn't add a test for that since I couldn't construct a case where it doesn't self-correct within a call or two; happy to be shown one.

Covered by `test_capacity_pressure_does_not_evict_an_unexpired_cancellation_tombstone`, which reproduces the scenario (a burst pushing the ledger over a cap of 1 while a settled cancellation still has most of its 24h left) and confirms a late `set_state` is still refused afterward.

EnjoyBacon7:
Pushed four fixes for the outstanding review round:

- 6465c53b: bumped the indexer actor protocol so a TaskStateManager replacement rolls the dispatcher/workers with it (your first finding).
- b1fdc023: stopped state=None records leaking past eviction, both from `set_details` racing an eviction and from a refused `set_state` on an unknown id (your second finding).
- 0ab69188: stopped `recover_refless` polling forever once its task is evicted (your third finding); applied the same guard to #904 separately since it carries its own copy of the file.
- 21c3c849: fixed a related gap CodeRabbit had flagged and nobody had picked up yet, capacity eviction could forget an unexpired cancellation tombstone and let a late write resurrect the task.

On merge order: still open. Worth noting the branch is now CONFLICTING against `develop` for an unrelated reason, develop landed its own `_INDEXER_ACTOR_PROTOCOL_VERSION` bump to v8 (#846, actor `max_restarts`) after this branch was cut, so my v8 collides with theirs the same way the v6 lineages did before v7. That'll need a rebase/merge to v9 (folding both rationales) before this can land regardless of what we decide about #904. I didn't do that merge myself since it pulls in the full develop diff (~180 files) onto this branch, wanted to flag it rather than push something that size unprompted.

EnjoyBacon7:
Merged develop in and bumped to v9 (1e46b1f3), folding in develop's own v8 (#846 max_restarts) alongside this branch's v8 (TaskStateManager retention), same pattern as the v6→v7 merge above it. Updated the actor names in the test file and the CLAUDE.md note accordingly. Branch is MERGEABLE now and CI is running.

CodeRabbit's summary review has come back clean since the last round.

EnjoyBacon7:
Deciding the merge order: #904 first, then #903.

#903's own description already says why: "Postgres holds the durable job history (#904), so the in-memory copy only has to cover the reads that follow a job settling." That only holds if #904 is live first. Merged alone, #903 makes finished-task status 404 after 1h or a 2,000-task cap with nothing to fall back on, exactly the regression you flagged for a client like cozy-stack/Twake polling late after a bulk upload. #904 merged alone has no such risk, it just adds durable history on top of today's unbounded (leaky) retention, no worse than the current behavior.

No version collision blocks this order either: #904 doesn't touch `_INDEXER_ACTOR_PROTOCOL_VERSION`, so #903's v9 bump won't need to change again once #904 lands.

Plan: land #904 into `develop`, then update #903 from `develop` (should be clean) and land that.

Ahmath-Gadji (dismissed):

Ahmath-Gadji (approved):

PR #904, opened 2026-09-07
feat(jobs): persist indexing job state to Postgres

EnjoyBacon7:
Second half of #660. The memory leak half is #903.

Indexing job state lived only in memory, so a restart erased all job history
and left in-flight work invisible with nothing to clean it up.

Each indexing task now gets a row in a new jobs table. It is written when the
task is admitted, updated when it settles, and read back when the actor no
longer knows the task, so status and the task list survive a restart. On
startup, records left in flight by a restart are marked failed, and history
older than 30 days is deleted. A failed history write never blocks indexing.

The schema is the one from #677, which cannot merge on its own. user_id has a
foreign key to users.id with ON DELETE SET NULL, and a CHECK pins status to the
DocumentStatus values (five, not #677's seven: #721 removed CHUNKING and
INSERTING after that PR was written). finished_at is replaced by started_at and
completed_at, so queue wait and service time are separable instead of collapsed
into one settle timestamp; started_at is stamped by the indexer worker when the
task leaves the queue. Indexes are (status, created_at) and (user_id, status),
plus an expression index on COALESCE(completed_at, created_at) that the
retention sweep now filters on.

A settled row freezes status, error and completed_at together. Freezing the
status alone let a FAILED write that lost the cancel race staple its traceback
onto a row reading CANCELLED.

## Summary by CodeRabbit

* **New Features**
  * Added durable indexing job history with statuses, errors, timestamps, and optional user/file details.
  * Job lists and task details remain available after live task state is no longer retained.
  * Added user filtering, automatic recovery, retry handling, and cleanup of old completed jobs.

* **Bug Fixes**
  * Indexing continues when history persistence is temporarily unavailable.
  * Workspace ownership is transferred only after all attachments succeed.

* **Tests**
  * Added coverage for persistence, recovery, cleanup, migrations, and worker integration.

### DIFF
--- a/openrag/core/models/catalog.py
+++ b/openrag/core/models/catalog.py
@@ -58,9 +58,14 @@ class DocumentRecord(BaseModel):
 class IndexationJob(BaseModel):
     """The durable record of one indexing task.
 
-    Mirrors a ``TaskStateManager`` entry into Postgres so job state survives a
-    restart and stays visible to operators. ``status`` reuses
-    :class:`DocumentStatus`, the state machine the indexing path already writes.
+    The Postgres row is what remains of a task after the ``TaskStateManager``
+    actor evicts it, so job state survives a restart and stays visible to
+    operators. ``status`` reuses :class:`DocumentStatus`, the state machine the
+    indexing path already writes.
+
+    ``started_at`` is stamped when the task leaves the queue and ``completed_at``
+    when it settles, so queue wait and service time are separable rather than
+    collapsed into one settle timestamp.
     """
 
     id: str = Field(default_factory=lambda: str(uuid.uuid4()))
@@ -71,4 +76,5 @@ class IndexationJob(BaseModel):
     error: str | None = None
     created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
     updated_at: datetime | None = None
-    finished_at: datetime | None = None
+    started_at: datetime | None = None
+    completed_at: datetime | None = None
--- a/openrag/core/models/catalog.py
+++ b/openrag/core/models/catalog.py
@@ -37,14 +37,6 @@ INDEXING_CONTENT_CLAIM_TOKEN_PREFIX = "task:"
 COPY_CONTENT_CLAIM_TOKEN_PREFIX = "copy:"
 
 
-class JobStatus(str, Enum):
-    QUEUED = "QUEUED"
-    RUNNING = "RUNNING"
-    SUCCESS = "SUCCESS"
-    FAILED = "FAILED"
-    PARTIAL = "PARTIAL"
-
-
 class DocumentRecord(BaseModel):
     """A document entry in the catalog (PostgreSQL)."""
 
@@ -64,12 +56,19 @@ class DocumentRecord(BaseModel):
 
 
 class IndexationJob(BaseModel):
-    """An indexation job tracking batch document processing."""
+    """The durable record of one indexing task.
+
+    Mirrors a ``TaskStateManager`` entry into Postgres so job state survives a
+    restart and stays visible to operators. ``status`` reuses
+    :class:`DocumentStatus`, the state machine the indexing path already writes.
+    """
 
     id: str = Field(default_factory=lambda: str(uuid.uuid4()))
-    status: JobStatus = JobStatus.QUEUED
-    total_documents: int = 0
+    status: DocumentStatus = DocumentStatus.QUEUED
     partition: str = "default"
+    file_id: str | None = None
+    user_id: int | None = None
+    error: str | None = None
     created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
-    started_at: datetime | None = None
-    completed_at: datetime | None = None
+    updated_at: datetime | None = None
+    finished_at: datetime | None = None
--- a/openrag/services/orchestrators/job_service.py
+++ b/openrag/services/orchestrators/job_service.py
@@ -197,7 +197,7 @@ class JobService:
 def _job_to_info(job: Any) -> dict[str, Any]:
     """Render a durable row in the same shape the actor returns."""
     created_at = job.created_at.isoformat() if job.created_at else None
-    finished_at = job.finished_at.isoformat() if job.finished_at else None
+    completed_at = job.completed_at.isoformat() if job.completed_at else None
     state = job.status.value
     return {
         "state": state,
@@ -209,7 +209,7 @@ def _job_to_info(job: Any) -> dict[str, Any]:
             "user_id": job.user_id,
         },
         "created_at": created_at,
-        "duration_ms": _duration_ms(created_at, finished_at, state=state, now=datetime.now(UTC)),
+        "duration_ms": _duration_ms(created_at, completed_at, state=state, now=datetime.now(UTC)),
     }
 
 
--- a/openrag/services/orchestrators/job_service.py
+++ b/openrag/services/orchestrators/job_service.py
@@ -107,7 +107,10 @@ class JobService:
 
         # The actor only remembers live and recent tasks. Durable rows fill in
         # history it has evicted, and everything dispatched before a restart.
-        all_info = {**await self._durable_task_info(is_admin=is_admin, user_id=user_id), **all_info}
+        all_info = {
+            **await self._durable_task_info(is_admin=is_admin, user_id=user_id, task_status=task_status),
+            **all_info,
+        }
 
         if task_status is None:
             filtered = list(all_info.items())
@@ -180,11 +183,18 @@ class JobService:
             logger.warning("Failed to read durable job", task_id=task_id, error=str(exc))
             return None
 
-    async def _durable_task_info(self, *, is_admin: bool, user_id: int | None) -> dict[str, dict]:
+    async def _durable_task_info(
+        self,
+        *,
+        is_admin: bool,
+        user_id: int | None,
+        task_status: str | None,
+    ) -> dict[str, dict]:
         if self._job_repo is None:
             return {}
         try:
             jobs = await self._job_repo.list_jobs(
+                statuses=_durable_statuses(task_status),
                 user_id=None if is_admin else user_id,
                 limit=_DURABLE_TASK_LIMIT,
             )
@@ -192,6 +202,19 @@ class JobService:
             logger.warning("Failed to list durable jobs", error=str(exc))
             return {}
         return {job.id: _job_to_info(job) for job in jobs}
+
+
+def _durable_statuses(task_status: str | None) -> list[str] | None:
+    """The statuses the durable query should return, or ``None`` for all.
+
+    The row limit applies to what the query returns, so an unfiltered read
+    would hide older matches behind newer rows of every other status.
+    """
+    if task_status is None:
+        return None
+    if task_status.lower() == "active":
+        return list(_ACTIVE_STATES)
+    return [task_status.upper()]
 
 
 def _job_to_info(job: Any) -> dict[str, Any]:
--- a/openrag/services/orchestrators/job_service.py
+++ b/openrag/services/orchestrators/job_service.py
@@ -23,17 +23,22 @@ from core.models.catalog import (
     TASK_FINISHED_AT_METADATA_KEY,
     TERMINAL_TASK_STATES,
 )
+from core.utils.logging import get_logger
+
+logger = get_logger()
 
 _ACTIVE_STATES = ("QUEUED", "SERIALIZING")
+_DURABLE_TASK_LIMIT = 500
 _TERMINAL_STATES = frozenset(state.value for state in TERMINAL_TASK_STATES)
 
 
 class JobService:
     """Queue/worker introspection over the TaskStateManager actor."""
 
-    def __init__(self, task_state_manager: Any, timeout: float = 60.0) -> None:
+    def __init__(self, task_state_manager: Any, timeout: float = 60.0, *, job_repo: Any = None) -> None:
         self._tsm = task_state_manager
         self._timeout = timeout
+        self._job_repo = job_repo
 
     async def _call(self, submit: Any, task_description: str) -> Any:
         """Route TaskStateManager calls through the centralized Ray helper.
@@ -100,6 +105,10 @@ class JobService:
                 f"get_all_user_info({user_id})",
             )
 
+        # The actor only remembers live and recent tasks. Durable rows fill in
+        # history it has evicted, and everything dispatched before a restart.
+        all_info = {**await self._durable_task_info(is_admin=is_admin, user_id=user_id), **all_info}
+
         if task_status is None:
             filtered = list(all_info.items())
         elif task_status.lower() == "active":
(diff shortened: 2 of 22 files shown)


openrag/services/workers/dispatcher.py
```diff
@@ -307,6 +311,14 @@ async def dispatch_indexing(
                 f"Task {task_id} was rejected because file {file_id!r} in partition {partition!r} is being deleted"
             )
 
+        await self._record_job(
+            task_id,
+            status=DocumentStatus.QUEUED,
+            partition=partition,
+            file_id=file_id,
+            user_id=task_details["user_id"],
+        )
```
hedhoud: Please persist the dispatch failure to the durable job record when submission fails after the queued row is created and before completion tracking is established.

Right now the in-memory task state can record the failure while the database row remains QUEUED. If the actor is restarted or evicts the task before reconciliation runs, clients can see a job stuck in QUEUED even though dispatch has already failed.

The durable record should retain the terminal failure state and completion timestamp before the exception is returned. Please keep the existing cleanup and uncertainty handling for cases where the worker may have started successfully.

hedhoud (dismissed):

openrag/services/workers/dispatcher.py
hedhoud, line 320: Please persist the dispatch failure to the durable job record when submission fails after the queued row is created and before completion tracking is established.

Right now the in-memory task state can record the failure while the database row remains QUEUED. If the actor is restarted or evicts the task before reconciliation runs, clients can see a job stuck in QUEUED even though dispatch has already failed.

The durable record should retain the terminal failure state and completion timestamp before the exception is returned. Please keep the existing cleanup and uncertainty handling for cases where the worker may have started successfully.

EnjoyBacon7: Fixed in 662cc3fd. The dispatch-failure path now writes FAILED, the traceback and `finished_at` to the job row right after it marks the actor, so the row no longer sits at QUEUED once the actor forgets the task.
The uncertainty handling is unchanged: the write is behind the same `not submission_outcome_unknown and mark_submit_failed` guard, so a submission that may have started a worker still leaves the row alone.
Two tests cover it, one per branch.

EnjoyBacon7:
@hedhoud ready for another look. The dispatch-failure path is fixed in 662cc3fd, details in the thread on `dispatcher.py`.

api-tests was red on that commit, but it was a flake rather than the change. Startup ran long, the service container got dropped by the boot guard in `main.py`, and every authenticated request then returned 500 while `/health_check` stayed green so the suite ran anyway. Re-ran the same commit with no changes and it passed. All 7 checks are green now.

hedhoud:
Please address these durable task-state issues before merge:

- The detached `TaskCompletionTracker` can be reused during a rolling deployment because the existing compatibility check does not identify the durable-write generation. A pre-existing actor may therefore never record settled jobs or reconcile orphans.
- If Postgres is temporarily unavailable when a task settles, the task ID is discarded after persistence fails. The durable row can remain non-terminal and later be incorrectly marked as orphaned/failed; terminal persistence needs a retry or durable retry path.
- The queue listing applies a 500-row limit before the requested status filter, so older retained jobs—and even exact status matches—can be omitted from the documented all-task view. Apply filtering before limiting or add pagination.

openrag/services/workers/task_completion.py
```diff
@@ -170,6 +181,13 @@ async def _record_finished_at(self, task_id: str) -> None:
 
         metadata = details.get("metadata")
         metadata = dict(metadata) if isinstance(metadata, dict) else {}
+        # Persist before stamping the actor: the stamp is what stops this method
+        # running again, so a failed write must leave the task unstamped for
+        # ``recover`` to pick up. Stamping regardless would strand the row at
+        # whatever non-terminal status it last held, until orphan reconciliation
+        # marked a completed job FAILED after a restart.
+        if not await self._record_settled_job(task_id, details):
```
hedhoud, line 192: When `_record_settled_job()` returns false, `track()` eventually discards the task ID and no retry is scheduled. A transient Postgres outage can leave the durable row non-terminal, after which restart reconciliation may mark a successful task as orphaned/failed. Preserve the pending settlement for retry when persistence fails.

openrag/services/orchestrators/job_service.py
```diff
@@ -154,11 +163,55 @@ async def get_task_details(self, task_id: str) -> dict | None:
...
+            logger.warning("Failed to read durable job", task_id=task_id, error=str(exc))
+            return None
+
+    async def _durable_task_info(self, *, is_admin: bool, user_id: int | None) -> dict[str, dict]:
+        if self._job_repo is None:
+            return {}
+        try:
+            jobs = await self._job_repo.list_jobs(
+                user_id=None if is_admin else user_id,
+                limit=_DURABLE_TASK_LIMIT,
```
hedhoud, line 199: This limit is applied before `list_tasks()` applies `task_status`. Because the endpoint has no pagination, a status query can miss matching jobs hidden behind 500 newer rows, and the all-task view omits older retained jobs. Filter in the repository before limiting or expose pagination.

openrag/services/workers/task_completion.py
```diff
@@ -23,6 +31,8 @@ def __init__(self, namespace: str = "openrag") -> None:
         self._logger = get_logger()
         self._tracked_task_ids: set[str] = set()
         self._recovery_lock = asyncio.Lock()
+        self._catalog_store: Any = None
+        self._catalog_lock = asyncio.Lock()
 
     def supports_cancellation_recovery(self) -> bool:
```
hedhoud, line 43: This capability check predates the durable-write and reconciliation behavior added here, so an older detached `TaskCompletionTracker` can still report compatibility during a rolling deployment. The old actor will never run `_record_settled_job()` or `reconcile_jobs()`. Add a generation/feature check that forces replacement.

openrag/services/workers/task_completion.py
EnjoyBacon7: Fixed in 93bacf40. A failed write now parks the built `IndexationJob` and a background pass retries it every 30s until Postgres takes it, then stamps the actor. Nothing re-reads task state at retry time, since the actor may have forgotten the task by then, and the stamp is skipped when it has: replaying stale details through `set_details` would put an evicted record back.

Pending ids are also passed to `fail_orphaned_jobs` as active, so a reconcile running while one waits cannot freeze the row as FAILED before the real outcome lands. The map is bounded at 1,000, oldest dropped first.

Covered by `test_a_held_settlement_is_retried_until_postgres_takes_it` and `test_a_held_settlement_is_not_reconciled_as_an_orphan`.

openrag/services/orchestrators/job_service.py
EnjoyBacon7: Fixed in 93bacf40. `list_jobs` takes `statuses` and filters in the query (`status = ANY($1::text[])`), and `_durable_task_info` passes the request down: `active` becomes QUEUED|SERIALIZING, anything else an exact match, `None` no filter.

The 500 row limit now applies to matching rows, so a status query no longer stops at the newest rows of every other status.

Covered by a parametrized test over the three cases, and the query itself was run against Postgres to check the array cast and the NULL path.

openrag/services/workers/task_completion.py
EnjoyBacon7: Fixed in 93bacf40. `_supports_task_completion_recovery` also requires `reconcile_jobs`, which only exists on a tracker that writes durable history, so an older detached one is replaced during bootstrap.

Covered by `test_task_completion_tracker_without_durable_history_is_replaced`.

openrag/core/models/catalog.py
```diff
@@ -64,12 +56,25 @@ class DocumentRecord(BaseModel):
...
+
+    ``started_at`` is stamped when the task leaves the queue and ``completed_at``
+    when it settles, so queue wait and service time are separable rather than
+    collapsed into one settle timestamp.
+    """
 
     id: str = Field(default_factory=lambda: str(uuid.uuid4()))
-    status: JobStatus = JobStatus.QUEUED
-    total_documents: int = 0
+    status: DocumentStatus = DocumentStatus.QUEUED
```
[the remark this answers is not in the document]
EnjoyBacon7: Already fixed in 300510e8. `_JOB_STATUS_CHECK` in `schema.py` is built from `DocumentStatus`, and the migration lists the same five states, so CHUNKING and INSERTING are gone from both constraints.

openrag/services/persistence/migrations/alembic/versions/c7d8e9f0a1b2_add_jobs.py
```diff
@@ -0,0 +1,85 @@
...
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from schema_helpers import index_exists, table_exists
+
+# revision identifiers, used by Alembic.
+revision: str = "c7d8e9f0a1b2"
+down_revision: str | Sequence[str] | None = "b9c0d1e2f3a4"
```
Ahmath-Gadji, line 17: **Two alembic heads after merging with `develop`, so startup fails.**

Since this branch was cut, `develop` added its own chain on `b9c0d1e2f3a4` (`c0d1e2f3a4b5` → `d1e2f3a4b5c6` → … → `f5a6b7c8d9e0`). Git merges the two cleanly, so GitHub shows no conflict, but on the merged tree:

```
ScriptDirectory.get_heads() -> ['c7d8e9f0a1b2', 'f5a6b7c8d9e0']
```

`connection.py` runs `command.upgrade(cfg, "head")` at boot, and that aborts when there is more than one head (the note in `e8f9a0b1c2d3_add_prompts_library.py` warns about this). The unit tests still pass on the merged tree, so CI won't catch it.

Fix: rebase and set `down_revision = "f5a6b7c8d9e0"`.

openrag/services/workers/task_completion.py
```diff
@@ -210,6 +250,116 @@ async def _has_unsettled_cancelled_worker(self, task_state_manager: Any, task_id
...
+            repo = await self._job_repo()
+            now = datetime.now(UTC)
+            orphaned = await repo.fail_orphaned_jobs(
+                # A settlement waiting on Postgres has an outcome already; it is
+                # not an orphan, and failing it here would freeze the wrong one.
+                active_ids=[*active_ids, *self._pending_settlements],
+                error=_ORPHANED_JOB_ERROR,
+                # A row written moments ago may belong to a dispatch that has
+                # not reached the actor yet, so leave the recent ones alone.
+                before=now - timedelta(seconds=_ORPHAN_GRACE_SECONDS),
```
Ahmath-Gadji, line 407: **Tasks cut off in the ~5 min before a restart are never cleaned up.**

`reconcile_jobs` runs only from `recover()`, and `recover()` has one caller, `bootstrap.py` (`tracker.recover.remote()`), so it runs once when an API process starts. This grace window then skips every row updated in the last 300 s.

After a full Ray restart, the `TaskStateManager` comes back empty and the API starts in well under 5 minutes. The tasks queued or started just before the crash, which are exactly the ones most likely to be orphaned, fall inside the window and are skipped. Nothing looks at them again until the next restart:
- `list_tasks` shows them as active through the durable fallback
- `get_task_state` keeps returning `QUEUED`/`SERIALIZING`, so clients polling the status URL never reach a terminal state

The grace window makes sense for a dispatch in flight during the snapshot, but reconciliation needs a second pass once it expires (a delayed re-run, or a periodic sweep), not just the startup one.

openrag/services/workers/task_completion.py
```diff
@@ -210,6 +250,116 @@ async def _has_unsettled_cancelled_worker(self, task_state_manager: Any, task_id
...
+        failure returns ``False``, because the stamp is what stops recovery
+        retrying this task.
+        """
+        try:
+            task_state_manager = self._task_state_manager()
+            state = await self._call_task_state(
+                lambda: task_state_manager.get_state.remote(task_id),
+                f"get_state({task_id}) for job history",
+            )
+            if state not in _TERMINAL_STATES:
```
Ahmath-Gadji: **A task cancelled by a delete can leave its row stuck at `SERIALIZING`.**

Returning `True` for a non-terminal state lets `_record_finished_at` stamp `finished_at`, and that stamp stops `recover` from ever revisiting the task. So "nothing to persist *yet*" actually means nothing is ever persisted.

The delete path can hit this: `cancel_active_indexing_tasks` does `ray.cancel(ref)`, waits for the ref to settle, and only then calls `set_state(task_id, "CANCELLED")`. The worker doesn't mark it either (`CancelledError` is a `BaseException`, so `process_file`'s `except Exception` doesn't run). `track()` is waiting on the same ref, so if the tracker reads the state before the delete path writes `CANCELLED`, it sees `SERIALIZING`, writes no row, and stamps the task.

The row then stays `SERIALIZING` for a file that has been deleted. Once the actor forgets the task, the job reads as active, and the next restart's reconciliation marks it `FAILED` with "interrupted by a restart" instead of `CANCELLED`.

Either don't stamp while the state is non-terminal, or have the delete path write the durable `CANCELLED` itself.

Ahmath-Gadji (commented):
Three issues the earlier rounds didn't cover. The first will stop the app from starting once this lands on `develop`.

openrag/services/persistence/migrations/alembic/versions/c7d8e9f0a1b2_add_jobs.py
EnjoyBacon7: Fixed in e4e098db. Merged develop in and added a merge-heads migration (1f53920217de) reconciling c7d8e9f0a1b2 with develop's f5a6b7c8d9e0. Ran alembic upgrade head end-to-end against Postgres, single head, jobs table lands with the expected constraints.

openrag/services/workers/task_completion.py
EnjoyBacon7: Fixed in d49b93fa. A periodic sweep (every 2 min) now runs reconcile_jobs in the background from recover(), not just once at process start, so a task orphaned inside the 300s grace window gets caught on the next pass instead of only the next restart.

openrag/services/workers/task_completion.py
EnjoyBacon7: Fixed in d49b93fa. _record_settled_job now polls get_state briefly (up to 5x50ms) instead of trusting the first read, closing the window where cancel_active_indexing_tasks's CANCELLED write hasn't landed yet when this reads the pre-cancel state.

EnjoyBacon7:
Fixed in e6aabab0. `recover_refless` had the same forever-poll gap Ahmath-Gadji flagged on #903 (discussion_r4008177895): `get_details`/`get_state`/`get_object_ref` all return `None` once the actor has evicted a task's record, so none of the loop's exit conditions can fire and it kept sleeping and polling. It now returns as soon as `get_details` comes back `None`.

Covered by `test_recover_refless_stops_once_the_actor_forgets_the_task`.

Ahmath-Gadji (approved):
Tested this end to end on a live compose stack running the branch image, and it holds up. Testing found one regression, which I fixed and pushed, plus a docs gap.

**Fixed in dd5d67f0: the tracker never wrote the final job state.** `_record_finished_at` (changed in 93bacf40) called `get_details` on a temporary `ray.get_actor()` handle. Ray holds that handle weakly, so every task that finished logged `Lost reference to actor` and wrote nothing. Completed and cancelled tasks kept their `SERIALIZING` row. After a restart the UI showed them as `SERIALIZING`, and a few minutes later the orphan sweep marked them `FAILED` ("interrupted by a restart"), even though the files were indexed. The fix stores the handle in a variable first, as `develop` does. The unit tests couldn't catch this because they share one long-lived `MagicMock`. The new test gives out a fresh, weakly held handle per `get_actor` call, and it fails without the fix.

**Added in 5fe0bddb:** `data_model.md` now documents the `jobs` table (ER diagram, columns, constraints, indexes, lifecycle).

### Tested live, after the fix

| Scenario | Result |
|---|---|
| Completed task | Row ends `COMPLETED` with `started_at` and `completed_at` set |
| Cancelled task (PDF cancelled mid-run) | Row ends `CANCELLED` once the worker stops |
| Failing task (corrupt PDF) | `FAILED`, with the traceback saved (4,750 chars, under the 8,000 cap) |
| Restart after tasks finished | Task detail and `/queue/tasks` still show `COMPLETED` / `CANCELLED`, read from Postgres |
| Restart while a task runs | Row stays `SERIALIZING` for the 5 min grace, then the periodic sweep marks it `FAILED` |
| Retention | Of three rows inserted by hand: `COMPLETED` 31 days old and `FAILED` 31 days old with no `completed_at` were purged at startup (`purged=2`); `COMPLETED` 29 days old was kept |
| Access after a restart (read from Postgres) | Owner and admin can open the task. Another non-admin user gets 403 on the task and its `/error`, and doesn't see it in `/queue/tasks` |
| Deleting a user | Their rows stay, with `user_id` set to NULL. Deleting a user while their task runs doesn't break the final job write |
| `?task_status=` filter | `COMPLETED`, `failed`, `active`, `CANCELLED` each return only matching states; counts match the table |
| Schema | Check constraint, `user_id` FK with `ON DELETE SET NULL`, the 4 indexes; alembic at `1f53920217de` |

Also checked:
- The branch is behind `develop` but merges cleanly.
- After that merge there's a single alembic head, the actor protocol version comes out as v8, and the full unit suite passes.

**Not tested live:** a failed history write that indexing survives and the tracker later retries. Testing it needs a temporary constraint that makes writes to `jobs` fail, so that path is covered only by the unit tests (`test_a_failed_job_write_does_not_fail_indexing`, `test_a_held_settlement_is_retried_until_postgres_takes_it`).

**Not caused by this PR, but found while testing:** deleting a user while their upload runs makes the upload fail. `add_file_to_partition` inserts `files.created_by` with the deleted user's id and hits `fk_files_created_by`. The job history records the failure correctly. The bug already exists on `develop` and belongs in a separate issue.

Approving.

PR #905, opened 2026-09-07
fix: /-unresolved- endpoint label and phantom error rate in HTTP metrics

### LINKED ISSUE #910: HTTP metrics: /-unresolved- swallows five unrelated cases, and browser probes inflate the error rate
## Symptom

OpenRAG HTTP Metrics dashboard showing a 6.7% error rate, service Healthy, and GET /-unresolved- at 100% of failing routes

The OpenRAG HTTP Metrics dashboard reports a **6.7 % error rate** while the service reads
**Healthy** and **5xx sits flat at 0 req/s**. Every failure is attributed to one synthetic
route, `GET /-unresolved-`, at 100 % of "Top 5 failing routes". Nothing corresponding appears
in the application logs, and the rate climbs while chatting through the Chainlit UI.

Everything here is real but benign. There are three independent causes.

## Cause 1 — `/-unresolved-` collapses five unrelated things

`InstrumentationMiddleware._get_route_template` reads `request.scope["route"]` and falls back
to the literal `"/-unresolved-"` when it is absent (`openrag/api/middleware/instrumentation.py`).

That scope key is set **only** by FastAPI's `APIRoute.matches` / `APIWebSocketRoute.matches`
(`fastapi/routing.py`). Starlette's own `Route`, `Mount` and `WebSocketRoute` never set it. So
anything not served by a FastAPI route — including everything rejected *before* routing ever
runs — shares a single label.

Measured on a live stack by issuing one request at a time and diffing `/metrics`:

| request | HTTP | recorded `endpoint` |
|---|---|---|
| `GET /version` (control) | 200 | `/version` |
| `GET /chainlit/ws/socket.io/?transport=polling` | 200 | `/-unresolved-` |
| `GET /public/favicon.svg` (unauthenticated) | 403 | `/-unresolved-` |
| `GET /public/favicon.svg` (authenticated) | 404 | `/-unresolved-` |
| `GET /no-such-path-xyz` | 404 | `/-unresolved-` |
| `GET /partition` (`redirect_slashes`) | 307 | `/-unresolved-` |
| `GET /users/info` (unauthenticated) | 403 | `/-unresolved-` |

Five distinct causes, one label: Socket.IO HTTP long-polling (a `Mount` onto a raw ASGI app),
404s, `redirect_slashes` 307s, `AuthMiddleware` short-circuits, and root `StaticFiles` mounts.

The auth case is the costly one — a request rejected before routing is exactly the one you most
want attributed to an endpoint, and it is the one that loses its name.

### Why it correlates with chatting

Socket.IO always opens on HTTP long-polling before upgrading to a WebSocket, costing ~3
unresolved hits per chat page load. The upgrade itself works through the bundled nginx front
door (the `$connection_upgrade` map in `infra/compose/nginx/openrag-admin.conf`). Behind a
front proxy that does **not** forward the `Upgrade` header, the upgrade fails and Chainlit
polls forever — one request per emitted packet — and `/-unresolved-` swamps all traffic. Worth
ruling out whenever the observed volume is far above ~3 per chat session.

## Cause 2 — sub-app routes are recorded without their mount prefix

A route inside a submounted app carries a path relative to *that* app, so Chainlit's `/user`
is recorded as `/user` — a name that looks like a parent-app route and silently merges with
one. Live series from before the fix, still in the TSDB:

```
401 GET  /user           ← Chainlit's /chainlit/user
401 POST /auth/header    ← Chainlit's /chainlit/auth/header
```

## Cause 3 — the error rate counts browser-side probes

`record_request` increments `openrag_http_request_failures_total` for every status `>= 400`,
so ordinary unauthenticated browser fetches count as failures. A normal 3-message chat session
measured **61 requests / 5 failures = 8.2 % error rate**, with zero server errors:

- **`GET /chainlit/user` → 401, twice per page load.** Chainlit's "am I logged in?" probe,
  issued before the header-auth handshake (`POST /chainlit/auth/header` → 200 then establishes
  the session). Expected, and bounded — it does not scale with chat volume.
- **`GET /public/favicon.svg` → 403, once per message sent.** This one *does* scale with chat
  volume, and is a genuine bug.

### The per-message 403 is not a missing prefix

The obvious reading — "the `ChatProfile` icon in `openrag/app_front.py` is root-absolute, so
prefix it with `CHAINLIT_ROOT_PATH`" — is wrong, and produces a second, quieter bug. Chainlit's
frontend has **two consumers of `ChatProfile.icon` with incompatible conventions**:

| consumer | rule in the bundled frontend | with `/public/favicon.svg` | with `/chainlit/public/favicon.svg` |
|---|---|---|---|
| profile-picker dropdown | `icon.includes("/public")` → `buildEndpoint(icon)` (prepends the Chainlit API base) | `/chainlit/public/favicon.svg` → **200 `image/svg+xml`** | `/chainlit/chainlit/public/favicon.svg` → **200 `text/html`** (SPA index) — broken image |
| assistant message avatar | renders `chatProfile.icon` **verbatim**, no rewrite | `/public/favicon.svg` → **403 per message** | `/chainlit/public/favicon.svg` → 200 |

So **no `/public/...` value satisfies both**: the current value breaks the avatar with a 403 on
every message, and hand-prefixing it just moves the breakage into the picker, where it returns
`200 text/html` and can never show up as an error in any dashboard.

A value containing no `/public` substring is used unmodified by both. Chainlit's own `/favicon`
route serves the identical file:

```
/chainlit/favicon                       200 image/svg+xml 1734
/chainlit/public/favicon.svg            200 image/svg+xml 1734   (same bytes)
/chainlit/chainlit/public/favicon.svg   200 text/html     1470   (the second bug)
```

Same bug *class* as the pdf.js worker already handled in `openrag/api/chainlit_assets.py`; the
`ChatProfile` icon was missed, and its fix is not the same shape.

## What a fix looks like

- [ ] **Resolve the endpoint label instead of giving up.** When nothing set `scope["route"]`,
      re-run the routing table by hand so a pre-routing rejection is attributed to the route it
      was aimed at, a mount is named after the mount, and only a genuinely unroutable URL gets a
      fixed fallback. Two traps: a mounted Starlette app **overwrites `scope["app"]` with
      itself**, so the routing table has to be captured on entry; and the label must stay
      bounded, since the whole point of the fallback is to cap Prometheus label cardinality.
- [ ] **Restore the mount prefix** on routes owned by submounted apps (`/user` → `/chainlit/user`).
- [ ] **Fix the per-message 403** with an icon URL that survives both frontend consumers.
- [ ] **Stop counting expected client errors as service failures** in the dashboard, so the tile
      can read 0 while a browser is signing in. Touches
      `infra/compose/grafana/dashboards/openrag-http.json` — **conflicts with #808**, which
      rewrites that file, so it lands after #808.

## Scope

All of this **predates #808** and reproduces on `develop`, which already carries an
"Error Percentage (%)" panel. #808 does not introduce any of it — its dashboard rewrite is
simply what made it visible.

Associated PR: #905

Ahmath-Gadji:
> [!NOTE]
> The investigation, the evidence and the reproduction table now live in **#910**.
> This PR is the implementation.
>
> **Still draft:** the dashboard item lands once #808 merges — it rewrites
> `infra/compose/grafana/dashboards/openrag-http.json`, so writing it here now would
> conflict head-on.

Closes #910

### `openrag/api/middleware/instrumentation.py`

`scope["route"]` is set **only** by FastAPI's `APIRoute.matches` / `APIWebSocketRoute.matches` —
Starlette's own `Route`, `Mount` and `WebSocketRoute` never set it. So every request not served
by a FastAPI route shared one label: Socket.IO HTTP long-polling, `StaticFiles` mounts, 404s,
`redirect_slashes` 307s, and everything `AuthMiddleware` rejected *before* routing ran.

`_get_route_template` now resolves in three steps:

1. `scope["route"]` set → its `path`, prefixed with what the mount appended to `root_path`, so
   a route owned by a submounted app keeps its browser-facing name.
2. Not set → `_match_route_table` re-runs the routing table by hand, descending into mounts. A
   pre-routing rejection is attributed to the route it was aimed at; a mount is named after the
   mount.
3. Nothing matches → `/-not-found-`.

Two traps, both covered by tests:

- **A mounted Starlette app overwrites `scope["app"]` with itself**, so reading the routing
  table after `call_next` returns gives the *sub*-app's table, not the one the request entered
  through. The table and `root_path` are captured on entry.
- **The label stays bounded.** The fallback exists to cap Prometheus label cardinality — a
  scanner walking 10k URLs must not mint 10k series — so raw request paths are never an option.

### `openrag/app_front.py`

The `ChatProfile` icon moves to Chainlit's own `/favicon` route.

`/public/favicon.svg` resolved against the origin root, outside the `/chainlit` auth bypass, and
was rejected **403 on every message sent**. Prefixing it is *not* the fix — Chainlit's frontend
has two consumers of that field with incompatible conventions:

| consumer | rule | `/public/favicon.svg` | `/chainlit/public/favicon.svg` |
|---|---|---|---|
| profile picker | `icon.includes("/public")` → `buildEndpoint(icon)` | `/chainlit/public/favicon.svg` → 200 `image/svg+xml` | `/chainlit/chainlit/public/favicon.svg` → **200 `text/html`** (SPA index) — broken image |
| message avatar | renders `icon` **verbatim** | **403 per message** | 200 |

No `/public/…` value satisfies both; a prefixed one just moves the breakage into the picker,
where it returns 200 and no dashboard can ever surface it. `/favicon` serves the identical file
(1734 bytes, `image/svg+xml`) from a path containing no `/public` substring, so both consumers
use it unmodified.

## Verified on a live stack

Endpoint labels on a running instance. Prometheus still holds both sets side by side:

| request | before | after |
|---|---|---|
| `GET /chainlit/ws/socket.io/?transport=polling` | `/-unresolved-` | `/chainlit/ws/socket.io` |
| `GET /users/info` (403, rejected pre-routing) | `/-unresolved-` | `/users/info` |
| `GET /partition` (307 slash redirect) | `/-unresolved-` | `/partition/` |
| `GET /assets/pdf.worker.mjs` | `/-unresolved-` | `/assets` |
| `GET /no-such-path-xyz` | `/-unresolved-` | `/-not-found-` |
| `GET /chainlit/user` (401) | `/user` — collided with the parent app | `/chainlit/user` |
| `POST /chainlit/auth/header` | `/auth/header` — same | `/chainlit/auth/header` |
| `GET /public/favicon.svg` (403 per message) | `/-unresolved-` | not requested at all |

Zero `/-unresolved-` series remain. 2912 unit tests pass; ruff clean.

## Not in this PR

Splitting expected client errors from service failures in the dashboard — as long as 401/403
browser probes count as failures, the error-rate tile can never read 0. Deferred to after #808.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP metric route labels for mounted, submounted, static, Socket.IO, authentication, rate-limit, redirect, and unknown-path requests.
  * Unknown routes now receive a bounded not-found label instead of an unresolved label.
  * Fixed chat profile and assistant avatar favicon URLs when the application is hosted under a path prefix.

* **Documentation**
  * Added an investigation documenting HTTP metric endpoint labeling and remaining dashboard considerations.


hedhoud (approved):
The route-label reconstruction is covered by targeted tests and the API unit suite passes. No actionable regressions were identified.

PR #898, opened 2026-09-04
chore: sync develop with main after v2.2.0 release

hedhoud:
Sync develop with main after the v2.2.0 release.

## Summary by CodeRabbit

* **New Features**
  * Released OpenRAG 2.2.0 with expanded release notes covering 13 new features.
  * Added migration guidance and schema details for the Milvus 3.0 breaking change.
  * Updated “What’s New” indicators for features introduced in version 2.2.0.

* **Updates**
  * Updated packaged deployments and container images to OpenRAG 2.2.0.

PR #906, opened 2026-09-07
fix(migrations): quarantine workspace_files orphans instead of deleting them

### LINKED ISSUE #357: Workspace FK migration silently deletes orphan rows
`openrag/scripts/migrations/alembic/versions/f1a2b3c4d5e6_add_workspace_files_file_id_fk.py` runs `DELETE FROM workspace_files WHERE file_id NOT IN (SELECT file_id FROM files)` at line 39 and a second `DELETE WHERE file_fk IS NULL` at line 59 inside `upgrade()`. The matching `downgrade()` at lines 86 to 103 only restores the column structure, not the deleted rows, so the migration is irreversibly lossy with no operator opt-in, no row-count log, and no quarantine. If the orphans were caused by an upstream bug, the data is gone before anyone can audit them.

Before each `DELETE`, query `COUNT(*)` and emit it via the alembic logger so operators can review the loss. Better, copy the targeted rows into a `workspace_files_orphans_<rev>` table inside the same transaction; the structure stays small, downgrade can restore from it, and a follow-up cleanup migration can drop the table once operators confirm.

EnjoyBacon7:
Fixes #357

The workspace_files FK migration deleted rows it could not resolve. It never said
how many, and downgrade could not bring them back.

Now each delete copies the rows into a `workspace_files_orphans_f1a2b3c4d5e6`
table in the same statement, logs the count, and `downgrade()` puts them back.
Also swapped `NOT IN` for `NOT EXISTS`.

Checked with an upgrade/downgrade round trip on a real Postgres, plus 5 new unit tests.

## Summary by CodeRabbit

* **Data Integrity**
  * Workspace file references are now validated against existing files.
  * Orphaned records are preserved during cleanup and can be restored when safely possible.
* **Reliability**
  * Migration handling is safer for already-migrated databases and avoids duplicate restorations.
* **Tests**
  * Added coverage for orphan preservation, cleanup reporting, restoration, and repeatable migration behavior.

### DIFF
--- a/openrag/services/persistence/migrations/alembic/versions/f1a2b3c4d5e6_add_workspace_files_file_id_fk.py
+++ b/openrag/services/persistence/migrations/alembic/versions/f1a2b3c4d5e6_add_workspace_files_file_id_fk.py
@@ -6,6 +6,7 @@ Create Date: 2026-03-12 10:00:00.000000
 
 """
 
+import logging
 from collections.abc import Sequence
 
 import sqlalchemy as sa
@@ -15,6 +16,7 @@ from schema_helpers import (
     column_type_is,
     fk_exists,
     index_exists,
+    table_exists,
     unique_constraint_exists,
 )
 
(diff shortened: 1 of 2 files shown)


Ahmath-Gadji:
Closing this — it's a base-branch problem, not a problem with the change.

`main` only receives release and hotfix merges ([CONTRIBUTING.md](https://github.com/linagora/openrag/blob/main/CONTRIBUTING.md#branching-model-git-flow)), and a `hotfix/*` branches **from `main`** for a production break that can't wait for the next release. This branch was cut from `develop` — it carries `c21b4649` (*Merge pull request #898 from linagora/sync-main-into-develop*) — so merging it here would pull a develop merge commit into `main` alongside the fix.

The change itself is ordinary hardening of a migration that already shipped: quarantining unresolvable `workspace_files` rows into `workspace_files_orphans_f1a2b3c4d5e6` instead of deleting them, with a `downgrade()` that restores them. Worth having, but not the "production is broken and it can't wait" case that justifies going straight to `main`.

Separately, 2.2.1 (#922) is a hotfix release scoped to exactly one fix — the Milvus v3.0.1 upgrade from #919 — so this was never going to ride along with it.

Nothing needs rewriting. Please reopen against `develop`:

```bash
gh pr create --base develop --head fix/357-quarantine-workspace-files-orphans
```

The diff is a single commit on top of `develop` (`58413c86`), and that commit currently lives **only on this branch** — please don't delete the branch before the new PR is up. #357 stays open until it lands.

andyne13:
@EnjoyBacon7 — this looks like it needs re-targeting rather than redoing.

The change never landed anywhere: `58413c86` is contained only in `fix/357-quarantine-workspace-files-orphans`, both `main` and `develop` still run the `DELETE ... WHERE file_id NOT IN (...)` at line 38 and `DELETE ... WHERE file_fk IS NULL` at line 58 of `f1a2b3c4d5e6_add_workspace_files_file_id_fk.py`, and #357 is still open. (The v2.2.1 hotfix that merged to `main` the same afternoon carried #919, not this.)

The branch is intact, so nothing needs rewriting — could you re-open it as a PR against `develop`?

EnjoyBacon7:
Reopened against `develop` as #936. Same branch, no rewrite. Also folded in the `table_exists` guard on the quarantine table that CodeRabbit flagged here (543bf8bc).

Thanks both for the pointer on the base branch.

PR #907, opened 2026-09-07
fix(vdb): prevent data loss during concurrent initialization

hedhoud:
When multiple files were indexed into a cold Milvus collection, one worker could observe an incomplete collection and fail. The fix versions the collection during creation, waits for required indexes, and handles concurrent initialization safely. Regression tests and a cold-stack verification script confirm that both files are indexed successfully.

Fix issue #862 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when creating or loading collections during concurrent operations.
  * Added validation to ensure hybrid collections include required sparse-vector support.
  * Ensured required vector indexes are available before collections are used.
  * Improved schema-version handling for newly created and existing collections.
  * Improved handling of collection creation failures and index-readiness timeouts.
  * Refined timeout tracking while checking index availability to prevent premature collection use.


openrag/services/storage/milvus_store.py
```diff
@@ -270,6 +274,56 @@ def _ensure_loaded(self) -> None:
                 collection_name=self._collection_name,
             ) from e
 
+    def _wait_for_vector_indexes(self) -> None:
+        """Wait until Milvus exposes every required vector index."""
+        required_fields = ["vector"]
+        if self._hybrid:
+            required_fields.append("sparse")
```
Ahmath-Gadji, line 301: `sparse` is only present in the schema when the collection was **created** with `hybrid_search` on. Against a dense-only collection `list_indexes(field_name="sparse")` returns `[]` forever — not "not ready yet" — so this polls until the whole `vectordb.timeout` elapses and then hard-fails `initialize()`, on every attempt, with no recovery.

Reproduced against a live `milvusdb/milvus:v3.0.0` with `timeout=5.0` (create a dense-only collection, then point a `hybrid_search=True` store at it):

```
this branch: VDBCreateOrLoadCollectionError after 5.0s
             "Timed out waiting for vector indexes on collection `pr907_scratch_dense`."
develop:     initialize -> OK in 0.0s
```

In production `vectordb.timeout` is 120 s, so that is a 120 s stall per indexing attempt. The state is reachable and supported: `VDB_HYBRID_SEARCH` is a documented env override, and `migrations/milvus/2.rebuild_text_analyzer.py` explicitly handles sparse-less collections at schema version 2 ("Collection has no `sparse` field (hybrid search disabled) — no BM25 analyzer to fix.").

Suggestion: read the field list from `describe_collection` first, and when `sparse` is absent from the schema fail immediately with a message naming the real problem (hybrid enabled against a collection that has no BM25 leg) instead of waiting on an index that can never appear.

openrag/services/storage/milvus_store.py
```diff
@@ -242,15 +242,19 @@ def _ensure_loaded(self) -> None:
...
-                    ) from e
-                self._store_schema_version()
+                    if not self._client.has_collection(self._collection_name):
+                        raise VDBCreateOrLoadCollectionError(
+                            f"Failed to create collection `{self._collection_name}`: {e!s}",
+                            collection_name=self._collection_name,
+                            operation="create_collection",
+                        ) from e
+
+                    self._check_schema_version()
```
Ahmath-Gadji: This swallows genuine creation failures, not just lost races.

pymilvus's `_create_collection_with_schema` is three steps — `conn.create_collection()` -> `create_index()` -> `load_collection()`. If `create_index` fails, the collection *does* exist, so `has_collection` returns `True`, `_check_schema_version()` passes (the properties were stamped by step one), and `_wait_for_vector_indexes()` then polls for indexes that will never be created.

Reproduced live by injecting an invalid metric type into the index params:

```
this branch: VDBCreateOrLoadCollectionError after 5.0s
             msg: "Timed out waiting for vector indexes on collection `pr907_scratch_badidx`."
             __cause__: None
develop:     VDBCreateOrLoadCollectionError after 0.0s
             msg: "Failed to create collection `...`: <MilvusException: float vector index does not
                   support metric type: NOT_A_METRIC ...>"
             __cause__: MilvusException()
```

The root cause is discarded entirely, the half-created collection is left behind, and every retry pays the full timeout again. At minimum, keep `e` and chain it onto the timeout raise so the real failure survives; better still, only swallow when the collection looks like another process's (e.g. its stamped schema version is already readable) rather than on any `has_collection` hit.

tests/unit/services/storage/test_milvus_store.py
```diff
@@ -888,3 +896,96 @@ def test_matching_version_passes_quietly(self, store: MilvusVectorStore, logs: _
...
+        store._ensure_loaded()
+
+        properties = store._client.create_collection.call_args.kwargs["properties"]
+        assert properties == {SCHEMA_VERSION_PROPERTY_KEY: "1"}
+        store._client.alter_collection_properties.assert_not_called()  # type: ignore[attr-defined]
+
+    def test_losing_creation_race_uses_winner_collection(self, store: MilvusVectorStore) -> None:
+        store._embedding_dimension = 8
+        store._client.has_collection.side_effect = [False, True]  # first call says "no", second call says "yes"
+        store._client.create_collection.side_effect = MilvusException(message="collection already exists")
```
Ahmath-Gadji: Real Milvus does not produce this exception for the case the test models. On `milvusdb/milvus:v3.0.0`, a duplicate `create_collection` with an identical schema returns **success**:

```
DUPLICATE create_collection -> NO ERROR (idempotent)
```

So in the actual two-worker race the loser never reaches the `except MilvusException` branch — what closes #862 is the `properties={SCHEMA_VERSION_PROPERTY_KEY: ...}` change on creation, which I verified does persist. The recovery branch only fires when the two racers build *divergent* schemas (different embedding dimension, different `hybrid_search`).

Not a defect, and the defensive branch is worth keeping. Flagging it because the "handles concurrent initialization safely" claim rests on a different mechanism than this test implies, and because the branch is therefore never exercised against real Milvus (see my comment on `_ensure_loaded` for how it misfires when creation genuinely fails).

Ahmath-Gadji (commented):
Reviewed for correctness, with the behavioural claims checked against a live `milvusdb/milvus:v3.0.0` rather than mocks. Full unit suite passes on the branch (2912 passed).

**The core fix is right.** Stamping the version via `create_collection(properties=...)` is atomic and genuinely closes the race — I confirmed Milvus persists the custom property (`describe_collection` -> `{'openrag.schema_version': '2', 'timezone': 'UTC', ...}`). On develop the window is seconds wide, since pymilvus's `create_collection` is `create_collection` + 5x `create_index` + `load_collection` and `_store_schema_version()` only ran after all of it.

Three things on `_wait_for_vector_indexes()`, detailed inline:

1. **A missing schema field is polled as if it were a pending index** — `hybrid_search=true` against a collection created with hybrid off stalls for the full 120 s timeout and then hard-fails `initialize()`, permanently. Confirmed against develop, which initializes fine in 0.0 s.
2. **Index-creation failures now surface as a bogus timeout with the root cause dropped** (`__cause__ is None`), after the full timeout, on every retry, leaving a half-created collection behind.
3. **The `except MilvusException` recovery branch is dead code for the race it targets** — a duplicate `create_collection` is silently idempotent on Milvus 3.0. Informational, not a defect.

One scope question: the mechanism described in #862 (an idempotency `delete` rejected with `collection not loaded`, failing the task) no longer exists on develop — that delete is now `_cleanup_indexed_vectors` in `services/workers/indexer_actor.py`, wrapped in `except Exception: return`, so it cannot fail a task. What remains on develop is the schema-version race, which this PR does fix. Worth confirming the cold-stack reproduction still fails pre-patch for the reason assumed.

openrag/services/storage/milvus_store.py
hedhoud: Fixed in commit 4452a09d. _wait_for_vector_indexes now reads the collection fields first and raises a clear schema error when hybrid search is enabled but sparse is absent, before polling indexes. Added a regression test; the storage test file passes (106 tests).

openrag/services/storage/milvus_store.py
hedhoud: Fixed in commit 4452a09d. Only collection-exists errors enter the concurrent-creation recovery path; other Milvus creation/index errors are wrapped immediately with the original exception preserved as __cause__. Added a partial-creation failure regression test; the storage test file passes (106 tests).

tests/unit/services/storage/test_milvus_store.py
hedhoud: Confirmed and kept as defensive coverage. The real Milvus 3.0 race is closed by stamping the schema version during create_collection; the recovery branch remains for divergent-schema or compatible-server cases. No additional code change is needed for this informational comment.

Ahmath-Gadji (dismissed):
Approved — both issues from my earlier review are fixed, and I re-verified the behaviour against a live `milvusdb/milvus:v3.0.0` rather than relying on the mocks.

**The fix works under real multi-process concurrency.** N separate processes, staggered arrival onto a cold collection:

| stagger | develop | this branch |
|---|---|---|
| 0.2 s, 8 workers | 1 / 8 | **8 / 8** |
| 0.4 s, 8 workers | 1 / 8 | **8 / 8** |
| 1.0 s, 8 workers | 5 / 8 | **8 / 8** |
| 0.6 s, 6 workers | 1 / 6 | **6 / 6** |

24/24 on this branch across every timing profile. Two things worth recording:

- **#862 is worse than the issue reports.** It describes "one file lost per fresh partition"; I measured up to **7 of 8**. Every late arrival fails independently, each reading schema version `0`, so the symptom is `VDBSchemaMigrationRequiredError: Collection is at schema version 0 but the application requires version 2` on a collection created seconds earlier — which reads as a config problem in logs, not a race.
- When all workers arrive *simultaneously* develop passes 6/6, because every worker sees `has_collection == False` and takes the idempotent create path. That is why this is intermittent and why a naive reproduction can miss it.

**Both earlier findings verified fixed:** hybrid-over-dense now fails in 0.01 s with a message naming the real cause, and a failed `create_index` fails in 0.03 s with the `MilvusException` chained as `__cause__`, matching develop. I also confirmed the schema-version guard cannot be bypassed via the create branch — Milvus rejects a duplicate create whose `properties` differ, so the stamped version doubles as a version fence and stays correct on disk.

---

One **non-blocking** cleanup, for a follow-up or a quick amend — your call, not a merge condition.

The `except MilvusException` block is now functionally identical to develop's, wrapped in unreachable code. I probed every path Milvus v3.0.0 can take here:

| case | message |
|---|---|
| duplicate collection, identical schema | *no error* |
| duplicate collection, divergent | `create duplicate collection with different parameters` |
| duplicate index, identical params | *no error* |
| duplicate index, divergent params | `at most one distinct index is allowed per field` |

None contain `already exist`, so `if "already exist" not in error_message: raise` always fires and the `has_collection` + `_check_schema_version()` recovery below it can never run. That is correct behaviour — it is exactly what fixed the swallowed-root-cause problem — but it leaves ~15 lines that cannot execute, and `test_losing_creation_race_uses_winner_collection` passes only by mocking `message="collection already exists"`, a string the server never emits. So the "handles concurrent initialization safely" path has no real coverage and a future refactor breaking race handling would not be caught.

Suggestion: collapse the block back to a plain `raise ... from e`, drop the unreachable branch, and rewrite that test to assert the divergent-schema case raises loudly. The actual race protection is entirely the `properties={SCHEMA_VERSION_PROPERTY_KEY: ...}` stamp.

Smaller things, all optional:
- `remaining <= 0` immediately after `deadline = time.monotonic() + self._timeout` is unreachable (`timeout` is validated `gt=0`).
- `describe_collection` in `_wait_for_vector_indexes` is not wrapped in the `MilvusException` -> `VDBCreateOrLoadCollectionError` handling that the `list_indexes` call below it gets, so it degrades to `UnexpectedVDBError`.
- Only `sparse` absence fast-fails; a missing `vector` field would still poll the full timeout, though `fields` is already in hand.
- `describe_collection` runs twice per `initialize` on the existing-collection path (once via `_check_schema_version`, once in the wait).

Note the branch is currently behind `develop` and will need an update before it can go in.

hedhoud:
@Ahmath-Gadji,  the optional cleanup has been addressed in 9e9bffbb: creation failures now surface directly with their original cause, and the replacement test covers a conflicting collection definition. The atomic schema-version stamp and index-wait behavior are unchanged. The branch already includes the latest develop. All 2,944 unit tests passed, along with lint and formatting.

Ahmath-Gadji (approved):

Issue #663, opened 2026-07-15
No admission control / backpressure on ingestion (unbounded intake, always 201)


andyne13:
## Summary

The ingestion path has no admission control or backpressure. Every uploaded file is dispatched to a worker immediately via `.remote()`, and the upload endpoint returns `201` for all of them regardless of how much work is already in flight. Excess work piles into unbounded Ray actor mailboxes and the object store, and every raw upload sits in `data_dir` until its task settles. There is no `429`/shedding and no saturation signal to the client. The Marker parse tier is bounded downstream, but nothing bounds admission ahead of it.

## Severity: high for large-batch ingestion — not a regression

This is a scale/resilience gap rather than something the hexagonal refactor broke. It matters most for the large-batch ingestion target (~2K PDFs in a burst), where the lack of an admission gate lets a single client push the whole batch into the system at once.

## Root cause / code trace

- `IndexerPool.submit` (`openrag/services/workers/indexer_pool.py:294-314`) selects the least-loaded worker (`min(range(...), key=self._inflight.__getitem__)`, line 301) and immediately fires `self._workers[idx].process_file.remote(**kwargs)` (line 304). `self._inflight` is used **only for load-balancing** — there is no total-in-flight cap and no rejection branch.
- The workers are created with `max_concurrency=max_tasks_per_worker` (`indexer_pool.py:284`). This bounds concurrent **execution** per worker, but excess `.remote()` calls queue in the Ray actor mailbox **unbounded** — admission is not bounded.
- `add_file` (`openrag/api/routers/admin/indexing.py:120`) saves the upload to disk and returns `201 CREATED` unconditionally (`indexing.py:186`). No `429`, no queue-depth check, no saturation signal.

## Impact

- A burst of N uploads (e.g. a 2K-file batch) is accepted all at once: N raw files land in `data_dir` immediately and N `process_file` calls queue into worker mailboxes / the Ray object store, which grow without bound until each settles.
- Disk pressure (2K raw uploads retained until indexed) and memory/object-store pressure with no way for the client to learn the system is saturated.
- No graceful degradation: the system accepts far more than it can promptly process instead of shedding or signalling backpressure.

## Suggested fix

- Bound total admitted-but-unsettled work (a semaphore / queue-depth limit across the pool) and return `429 Too Many Requests` (with `Retry-After`) when saturated, so clients can back off.
- Alternatively, a bounded ingestion queue with explicit shedding and a visible saturation signal.
[...]

Issue #640, opened 2026-07-08
One unparseable page fails the whole document: PDF parse stage has no fallback when PyMuPDF raises (`code=4: no font file for digest`)


aditykris:
## Summary

A single PDF with a font anomaly on **one page** causes the entire document to fail
indexing. The parse stage passes any parser exception straight through, so the file is
marked `FAILED` and silently missing from the partition — there is no retry, no
degraded extraction, and nothing in the result that tells the operator the gap exists
unless they diff corpus vs. partition counts.

## Environment

- OpenRAG hexagonal branch, PyMuPDF parser path
  (`openrag/core/indexing/parsers/pdf/pymupdf.py` → `pymupdf4llm.to_markdown`)
- Reproduced with arXiv paper `2412.01370v2.pdf` (21 pages, ~14.5 MB,
  `producer: pikepdf 8.15.1`, `creator: arXiv GenPDF (tex2pdf:)`)

## Reproduction / observed behavior

Uploading the file fails in ~380 ms:

```
File ".../openrag/core/indexing/parsers/pdf/pymupdf.py", line 61, in _extract_markdown
  chunks = pymupdf4llm.to_markdown(...)
File ".../pymupdf4llm/helpers/pymupdf_rag.py", line 969, in get_page_output
  parms.textpage = page.get_textpage(flags=textflags, clip=parms.clip)
...
RuntimeError: code=4: no font file for digest
```

Task state: `FAILED`, `n_chunks=0`. Out of a 935-file corpus, this was the only
failure — 934/935 indexed, and the gap is invisible unless you go looking.

## Diagnosis

Probing the file page by page:

- `page.get_textpage()` (default flags) succeeds on **all 21 pages**.
- `pymupdf4llm.to_markdown(path, pages=[i])` fails on **page 2 only**; all other
  pages convert fine.
- Page 2 contains three **Type3 font objects with no embedded font file**
  (unnamed, `n/a/Type3`) alongside normal embedded fonts. Type3 fonts are defined
  as glyph procedures and by design never carry a font file, so MuPDF hard-erroring
  while looking up a "font file for digest" on them looks like an upstream MuPDF
  robustness problem with unusual-but-legal input (likely worth its own upstream
  report).
- Rewriting the file with PyMuPDF itself — `doc.save(out, garbage=4, clean=True)` —
  removes the orphaned font references; `to_markdown` then converts the full
  document (~101 K chars of markdown). Content is unchanged.

[...]

PR #909, opened 2026-09-07
fix(indexing): release the failed PDF before reopening it cleaned; restart indexer actors

andyne13:
Two of the three parts of #846. The third is not what the issue thought, so this is `Refs`, not `Closes`.

**The #640 recovery path held two documents open.** The cleaned copy was opened while the failed one was still open, so MuPDF kept parsed structures for both. The failed document is now closed first.

The obvious one-line version of this — `doc.close()` inside the `except` — does not work: PyMuPDF's `close()` is not idempotent, so the enclosing `with` then raises `ValueError: document closed` on exit, turning a rare parse failure into a guaranteed one. Hence the early return. The cost is a duplicated two-line tail; a helper or a sentinel flag would remove it and both read worse.

**Detached actors default to `max_restarts=0`.** An indexer worker or the dispatcher that died — an OOM on a large document, a node fault — stayed dead and its pool slot was lost until the next deploy. Marker and Docling already set 5 on both tiers. `IndexerWorkerActor.__init__` is safe to re-run: it builds an embedder and looks up `TaskStateManager`, with catalog init deferred, so no registration or claim side effects repeat.

### Not fixed, deliberately

**The issue's headline — "the default PDF parser holds the whole document in memory" — is misattributed.** `indexer_actor.py` reads every file with `read_bytes` before any parser runs and hands the parser `document.raw_bytes` directly. The bytes are already resident; opening PyMuPDF from a path would not lower peak memory while that stands, and freeing `raw` inside the parser is impossible because the `Document` owns it. Making parsers path-based is an architectural change across the whole stack and belongs in its own issue, next to #663.

Routing large PDFs to Marker/Docling by size threshold would add a config surface and a GPU dependency to solve a memory problem. `doc_parser.py`'s `read_bytes` of the converted `.docx` is the same class, but its contract returns bytes to its caller, so changing it is a contract change rather than a one-liner.

Two further corrections for anyone reading the issue against the code: its line numbers are from v2.1.0 and have all moved, and it states Marker/Docling set `max_restarts` on their workers via decorators — they set it on the pools via decorators and on the workers via `.options()`.

### Known limitation

`max_restarts` restores the actor, not the work. A file being indexed when a worker dies is still lost; the restart means the pool slot returns instead of staying dead until redeploy.

### Verification

2907 unit tests, `ruff check`, `ruff format --check` and `check_layer_imports.py` clean. The residency test is mutation-checked — it fails against the previous code and passes against this one — and the four existing #640 fallback tests are untouched and still pass.

Refs #846

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PDF processing recovery by releasing failed document resources before retrying, reducing memory pressure during fallback processing.
  - Indexing workers and the dispatcher now automatically restart after crashes or out-of-memory failures, helping prevent interrupted indexing jobs.
  - Improved indexing service resilience during worker and dispatcher failures, allowing processing to resume more reliably without manual intervention.

### DIFF
--- a/openrag/core/indexing/parsers/pdf/pymupdf.py
+++ b/openrag/core/indexing/parsers/pdf/pymupdf.py
@@ -67,6 +67,8 @@ def _extract_markdown(raw: bytes, filename: str) -> tuple[list[str], list[ImageB
     with pymupdf.open(stream=raw, filetype="pdf") as doc:
         try:
             chunks = _to_markdown(doc)
+            pages = [(chunk.get("text") or "").strip() for chunk in chunks]
+            return pages, []
         except RuntimeError as exc:
             # MuPDF hard-errors on some legal-but-unusual object graphs — e.g.
             # Type3 fonts with no embedded font file trip "code=4: no font file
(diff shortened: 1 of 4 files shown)


hedhoud (approved):
Reviewed the PDF recovery and indexer actor restart changes. Focused tests, lint, formatting, GitHub CI, CodeRabbit, and the isolated Ray restart check all pass. I found no blocking issues. The branch being behind develop is a process item only.

PR #913, opened 2026-09-07
fix(api): preserve Milvus error status and details

### LINKED ISSUE #885: Milvus errors surface as 500 UNEXPECTED_ERROR: milvus_store raises from openrag.core.*, error_handlers catches core.*
`api/error_handlers.py:29` imports `OpenRAGError` from `core.utils.exceptions`. `services/storage/milvus_store.py:57` raises `VDBSearchError` and friends from `openrag.core.utils.exceptions`.

The container sets `PYTHONPATH=/app/openrag` and also editable-installs the `openrag` package, so both import roots are live and Python loads that one file twice, as two modules holding two distinct class objects. Verified locally: `core.utils.exceptions.VDBSearchError is not openrag.core.utils.exceptions.VDBSearchError`.

Starlette resolves handlers by walking `type(exc).__mro__` and looking each class up in a dict, so the prefixed exceptions never match the registered handler, and all 19 raise sites in `milvus_store.py` fall through to `unhandled_exception_handler`. A Milvus search failure that declares `status_code=422, code=VDB_SEARCH_ERROR` therefore reaches the client as `500 {"detail": "[UNEXPECTED_ERROR]: An unexpected error occurred"}` with the real message dropped. Nothing in the retrieval chain (`retrieval_service`, `vector_store_searcher`, `core/retrieval/searcher`, `core/retrieval/pipeline`) catches them earlier.

21 imports across 17 files use the `openrag.` prefix; 547 use the bare form. `scripts/check_layer_imports.py:44-49` treats both spellings as the same layer, so the guard cannot flag this.

Suggested fix: convert the 17 files to bare imports, then add a guard rule rejecting `openrag.`-prefixed imports so the split cannot come back.

hedhoud:
Milvus search errors were reaching clients as generic 500 responses because mixed import roots loaded separate copies of the exception classes. This uses the existing import convention consistently so the API preserves the intended status and error details, and extends the import guard to prevent a recurrence.

The regression test reproduced the 500 before the fix and now receives 422 with the original error details. All 2,930 unit tests pass, along with lint, formatting, and the import guard.

Closes #885

## Summary by CodeRabbit

- **Refactor**
  - Standardized internal module references across application services, storage integrations, workers, and tests.
  - Broadened import validation to cover project utilities and test areas.

- **Bug Fixes**
  - Improved handling of collection-inspection failures, including clearer error details and consistent error reporting.

- **Tests**
  - Expanded coverage for import rules, layer boundaries, storage errors, and collection-loading scenarios.

- **User Impact**
  - No intentional changes to core functionality or user-facing behavior.

### DIFF
--- a/scripts/check_layer_imports.py
+++ b/scripts/check_layer_imports.py
@@ -12,6 +12,9 @@ Only files inside openrag/{core,services,api,di}/ are checked. Legacy
 paths (openrag/components/, openrag/routers/, openrag/models/, ...) are
 ignored until they are migrated.
 
+Absolute application imports must use the bare layer root, not ``openrag.``:
+mixing the two loads distinct copies of classes such as exception types.
+
 Usage:
     python scripts/check_layer_imports.py
 
@@ -103,6 +106,10 @@ def check_file(path: Path) -> list[str]:
             dotted = resolve_relative(path, level, module)
         else:
             dotted = entry
+            if dotted == "openrag" or dotted.startswith("openrag."):
+                rel = path.relative_to(REPO_ROOT)
+                violations.append(f"{rel}:{lineno}  use the bare import root, not {dotted} (#885)")
+                continue
         tgt_layer = layer_of(dotted)
         if tgt_layer is None or tgt_layer == src_layer:
             continue
--- a/tests/unit/services/storage/test_milvus_store.py
+++ b/tests/unit/services/storage/test_milvus_store.py
@@ -16,16 +16,37 @@ from typing import Any
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
+from api.error_handlers import register_error_handlers
+from core.config.infrastructure import VectorDBConfig
+from core.models.chunk import Chunk, ChunkType
+from core.utils.exceptions import VDBSchemaMigrationRequiredError, VDBSearchError
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
 from pymilvus import MilvusException
-
-from openrag.core.config.infrastructure import VectorDBConfig
-from openrag.core.models.chunk import Chunk, ChunkType
-from openrag.core.utils.exceptions import VDBSchemaMigrationRequiredError, VDBSearchError
-from openrag.services.storage.milvus_store import MilvusVectorStore, analyzer_params
+from services.storage.milvus_store import MilvusVectorStore, analyzer_params
 
 # ---------------------------------------------------------------------------
 # Fixtures
 # ---------------------------------------------------------------------------
+
+
+def test_milvus_search_error_keeps_its_http_status_and_details(store: MilvusVectorStore) -> None:
+    """Mixed import roots must not turn a storage error into UNEXPECTED_ERROR (#885)."""
+    store._async_client.hybrid_search = AsyncMock(side_effect=MilvusException(1, "search unavailable"))
+    app = FastAPI()
+    register_error_handlers(app)
+
+    @app.get("/search")
+    async def search():
+        return await store.search([0.1, 0.2], query_text="test")
+
+    with TestClient(app, raise_server_exceptions=False) as client:
+        response = client.get("/search")
+
+    assert response.status_code == 422
+    assert response.json()["detail"].startswith("[VDB_SEARCH_ERROR]: Milvus hybrid search failed:")
+    assert "search unavailable" in response.json()["detail"]
+    assert response.json()["extra"] == {"collection_name": "test_collection"}
 
 
 @pytest.fixture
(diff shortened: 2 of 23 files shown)


scripts/check_layer_imports.py
```diff
@@ -103,6 +106,10 @@ def check_file(path: Path) -> list[str]:
             dotted = resolve_relative(path, level, module)
         else:
             dotted = entry
+            if dotted == "openrag" or dotted.startswith("openrag."):
```
Ahmath-Gadji, line 108: This rule needs no layer context — it only inspects a dotted string — but sitting inside `check_file` it inherits two restrictions that rule doesn't want: the early `return []` at L94 when `file_layer()` is None, and `main()` walking only `OPENRAG` at L128.

Net effect: `tests/` is unguarded, and **4 of the 23 files this PR had to fix were test files**. Also skips `openrag/app_front.py` and `chainlit_api.py`. I planted `from openrag.core.utils.exceptions import OpenRAGError` in each of those spots and the guard still printed `layer import guard: OK`.

The `tests/` hole is the load-bearing one: before this PR `test_milvus_store.py` imported the store under the `openrag.` root, so the suite exercised a different class object than the API process loads. That's how #885 shipped green.

Suggest hoisting it out into its own pass:

```python
ROOTS = (OPENRAG, REPO_ROOT / "tests", REPO_ROOT / "scripts")

def check_import_root(path: Path) -> list[str]:
    tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
    out = []
    for lineno, entry in iter_imports(tree):
        if isinstance(entry, tuple):
            continue  # relative imports are always fine
        if entry == "openrag" or entry.startswith("openrag."):
            out.append(f"{path.relative_to(REPO_ROOT)}:{lineno}  use the bare import root, not {entry} (#885)")
    return out
```

`check_file` stays scoped to `OPENRAG` as-is. No false positives today — the parametrize strings in the new `test_layer_imports.py` aren't import statements, so the AST walk won't see them.

tests/unit/services/storage/test_milvus_store.py
```diff
@@ -16,18 +16,39 @@
...
-from openrag.core.utils.exceptions import VDBSchemaMigrationRequiredError, VDBSearchError
-from openrag.services.storage.milvus_store import MilvusVectorStore, analyzer_params
+from services.storage.milvus_store import MilvusVectorStore, analyzer_params
 
 # ---------------------------------------------------------------------------
 # Fixtures
 # ---------------------------------------------------------------------------
 
 
+def test_milvus_search_error_keeps_its_http_status_and_details(store: MilvusVectorStore) -> None:
```
Ahmath-Gadji: Nit: this sits above the `# Fixtures` header and above the `store` fixture it consumes (works, just reads oddly). It's also an API-status assertion in a storage unit test — it's what drags `api.error_handlers` and `fastapi.testclient` into this module. `tests/unit/api/` would be a more natural home.

Ahmath-Gadji (commented):
Verified locally, not just read: reverting the one import in `milvus_store.py` makes the new test fail with `openrag.core.utils.exceptions.VDBSearchError` escaping the handler, so the regression test earns its keep. 2930 unit tests, the guard, and ruff all pass; the diff inside `openrag/` is purely the import root.

No bugs — good to merge. Two non-blocking notes inline. If you act on them, please `git commit --fixup a0d815af` + `git rebase -i --autosquash origin/develop` rather than stacking a third commit.

scripts/check_layer_imports.py
hedhoud: Fixed in ef2ffacf. The import-root rule now covers all Python files under openrag, tests, and scripts, while layer boundaries remain limited to application layers. Regression tests reproduce the previously missed locations and preserve allowed imports. Also resolved the develop conflict while retaining the newer Milvus initialization tests. All 2,973 unit tests, lint, formatting, and the repository guard pass.

hedhoud:
Addressed the outside-diff comment in review #5141824638 with e8b5dc9b. Schema-inspection failures now retain their collection lifecycle classification and original Milvus cause. Regression tests cover both new and existing collections and confirm loading stops after failure. All 2,975 unit tests passed; lint, formatting, and the import guard also pass.

Ahmath-Gadji (dismissed):
Approved.

Re-verified on a local checkout of e8b5dc9b:

- **Guard gap closed** (ef2ffacf) — replanted the probes in `openrag/app_front.py`, `openrag/`, `tests/unit/`, `tests/integration/` and `scripts/`; all five now fire. Folding it into `check_file` via `FORBIDDEN.get(src_layer, set())` rather than a second pass is neater than what I suggested. Confirmed the layer rule still fires on `core -> services` and `services -> api`, and the clean tree reports OK with no false positives.
- **e8b5dc9b** is a good catch off the develop merge — `describe_collection` was falling through to the outer `except Exception` at L267 and surfacing as `UnexpectedVDBError` (500); now a 422 with `operation="describe_collection"`. Same theme as the PR title.
- 2975 unit tests pass, `ruff check` and `ruff format --check` clean, `check_layer_imports.py` OK.

The test-placement nit is unaddressed, which is fine — it was a nit, not worth another round.

Ahmath-Gadji (approved):

PR #914, opened 2026-09-07
feat(metrics)!: scrape /metrics from external Prometheus on VM and k8s

paultranvan:
Goal: run OpenRAG with an external Prometheus/Grafana, on a VM (compose) or in Kubernetes (Helm), without depending on the bundled monitoring overlay.

**Access model**
`/metrics` used to require an admin bearer, so every scraper held a token that could also create users or delete partitions, and the overlay needed a root-owned token file synced from AUTH_TOKEN. Replace that with a single mechanism: the route bypasses AuthMiddleware and enforces its own `METRICS_TOKEN` (`server.metrics_token`). The scraper sends it as a bearer; any other credential, admin tokens included, gets 403. The route **fails closed**: with the token unset, every scrape gets 403 unless `METRICS_ALLOW_UNAUTHENTICATED=true` explicitly opens the endpoint to anyone reaching the API port (a configured token always wins; the API logs a warning at startup in both non-default states). The opt-in exists because port 8080 is what the Ingress / admin-ui proxy forwards, so "no token" must never silently mean "open" (review by andyne13). The token is read through the cached `load_config()` rather than the request container so a scrape keeps answering while Milvus or Postgres are down.

The admin UI's System > Metrics tab no longer reads `/metrics`: it calls `GET /monitoring/metrics`, the same exposition behind the ordinary `require_admin` gate (review by andyne13). Two audiences, two credentials: the scraper never touches the Postgres token lookup, an admin never holds the scrape secret.

**VM.**
The monitoring overlay writes `METRICS_TOKEN` from `.env` into the Prometheus container as the `credentials_file` of its `openrag` job (`configs.content`, Compose ≥ 2.23.1) and refuses to start without it, so the API and its scraper can't disagree; the `openrag_token` mount and its sudo steps are gone. The admin-ui nginx returns 404 on `/metrics`: the front door never serves it. An external Prometheus targets APP_PORT directly with the bearer, or an agent scrapes locally and remote_writes.

**Kubernetes.**
The Helm chart gains `openrag.metrics.*`: prometheus.io pod annotations by default for annotation-based discovery, and an optional ServiceMonitor for the Prometheus Operator that can read `METRICS_TOKEN` from the env Secret. `env.config.METRICS_ALLOW_UNAUTHENTICATED` (default `"false"`) is the opt-in for clusters that block `/metrics` at the Ingress and scrape the Service in-cluster. Uvicorn topology only: under Ray Serve (`ray.enabled` + `ENABLE_RAY_SERVE`, the Linagora values) the API listens on the RayCluster head Service and each replica has its own registry, so the chart renders no annotations there and refuses the ServiceMonitor with an explicit message (review by hedhoud); per-replica scraping is left for a follow-up. Nothing OpenShift-specific.

**Docs**: new prometheus_metrics page (exposed series, access control with the three states, VM and Kubernetes scrape setups, opening the endpoint without a token, agent alternative, limitations); env_vars, docker install and Grafana dashboard pages updated.

BREAKING CHANGE: GET /metrics no longer accepts admin or user tokens, and answers 403 until it is configured.
A scraper that sends the admin bearer (the v2.2.0 setup) now gets 403, and so does one that sends nothing. Any in-house client that read /metrics with an admin token (the admin UI did) must move to GET /monitoring/metrics.
To upgrade, pick one:
- Set METRICS_TOKEN in the OpenRAG .env (or env.secrets.METRICS_TOKEN in Helm) and send that value as the bearer. With the bundled compose overlay, that is the only step: delete infra/compose/prometheus/openrag_token, the overlay now hands the token to Prometheus itself. In Helm, openrag.metrics.serviceMonitor.bearerTokenFromSecret=true wires it for the Prometheus Operator.
- Block /metrics at the edge, scrape from inside the network, and set METRICS_ALLOW_UNAUTHENTICATED=true (env.config in Helm).
/metrics is no longer gated by AuthMiddleware: with METRICS_ALLOW_UNAUTHENTICATED=true and no token, anyone who can reach the API port can read the metrics.

## Summary by CodeRabbit

* **New Features**
  * Added bearer-token protection for Prometheus metrics, with explicit opt-in for anonymous access.
  * Added Helm support for Prometheus annotations and ServiceMonitor-based discovery.
  * Added a separate authenticated metrics view for administrators.
  * Added optional Docker Compose log-shipping guidance for Loki.

* **Bug Fixes**
  * Metrics access now fails closed by default, rejects admin tokens, and reports configuration warnings at startup.
  * Prevented the browser-facing admin origin from exposing scraper metrics.

* **Documentation**
  * Expanded metrics, Grafana, Docker, environment-variable, and deployment documentation.


andyne13:
Moving the scrape off the admin token is the right call, and both reasons in the description hold up: an admin credential that can also delete every partition is the wrong shape for a scraper, and the admin path hits Postgres for the token lookup, so the scrape fails exactly when the metrics matter. Reading `METRICS_TOKEN` through `load_config()` instead of the request container is a good detail.

One thing I'd change before this lands: **unset should not mean open.**

`/metrics` is served on the API port, and in the chart `8080` is the only entry in `networkPolicy.externalPorts` — the one port reachable from outside the namespace via the ingress controller. So there is no network boundary underneath this: on the default production topology, an unset `METRICS_TOKEN` means the endpoint is exposed wherever the API is exposed, which the env-var table states outright ("open to anyone reaching the API port"). The series aren't catastrophic on their own, but route templates, status-code distributions, request volumes and the circuit-breaker gauge with provider names are useful reconnaissance, and it's a default rather than a mistake someone has to make.

Two options, and they compose:

1. Fail closed — require the token when the endpoint is reachable, or put openness behind an explicit `METRICS_ALLOW_UNAUTHENTICATED=true` so it's a deliberate act rather than the default.
2. Block `/metrics` at the edge (a `location` rule returning 404 in the nginx/ingress path) and scrape the Service in-cluster. Cheapest of the two and it doesn't depend on anyone setting anything.

Separately, thanks for fixing the swapped scrape jobs in `prometheus/prometheus.yml` — the `openrag` job was pointed at 8000 (vLLM, unpublished) and `ray` at 8080 (the API), so every `openrag_http_*` panel in the bundled Grafana dashboard was querying series that never arrived.

andyne13:
One more caller to account for before this lands: **the admin UI reads `/metrics` itself**, and it will start getting 403s.

`ui/src/lib/api/system.ts:42` calls `/metrics` through the shared `request()` helper, so it sends the `or-…` bearer from localStorage, or the session cookie in OIDC mode. It backs the Metrics tab on the admin System page (`ui/src/pages/admin/system.tsx:192-197`), which polls on a 10s `refetchInterval`.

With this PR:

| `METRICS_TOKEN` | Scraper | Admin UI |
| --- | --- | --- |
| unset | works | works — nothing is checked |
| set | works | **403 every 10s** |

So the tab keeps working exactly while the endpoint is open, and breaks as soon as someone sets the token. That inverts the incentive, and it'll read as "the metrics page broke" rather than "the credential changed". The BREAKING CHANGE note covers scrapers sending the admin bearer; this in-repo caller isn't mentioned, and it's the one that fails silently from a user's point of view.

I don't think it changes the direction of the PR — a scraper credential that can't also delete partitions is still right, and keeping the token off the Postgres path is the good part. But the endpoint now has two audiences with genuinely different needs, and one credential can't serve both:

- a scraper wants something that works when the database is down
- a signed-in admin already has a session and shouldn't have to hold the scrape secret too

Cleanest split is probably to leave `/metrics` scrape-only as you have it, and give the System page an admin-gated route returning the same `get_metrics()` output. I'd avoid making `/metrics` accept an admin credential as a fallback — that puts the Postgres token lookup back on the metrics path, which is the coupling this PR is removing, on exactly the path that matters when the DB is the problem.

Third option, if it's simpler: drop the tab. It renders raw exposition text in a table every 10s and the page already links to Grafana. Your call which of the three — happy to do the work either way, just didn't want it to land as a surprise.

infra/charts/openrag-stack/values.yaml
```diff
@@ -727,6 +756,7 @@ env:
     # --set/-f) so the secret always matches the actual DB credential.
     POSTGRES_PASSWORD: "{{ .Values.postgresql.auth.password }}"
     AUTH_TOKEN: "" # API KEY for OpenRAG
+    METRICS_TOKEN: "" # Bearer for Prometheus on GET /metrics; empty = 403 unless METRICS_ALLOW_UNAUTHENTICATED
```
hedhoud, line 765: Changing this Secret does not change the Deployment pod template. After a Helm upgrade, Prometheus can read the new METRICS_TOKEN while the running OpenRAG process still expects the old value, causing every scrape to return 403 until a manual restart. Please add a rollout trigger for chart-managed secrets and define rotation behavior for existing/external secrets. In Ray Serve mode, the Ray pods or replicas that load this environment variable must also be refreshed.

infra/charts/openrag-stack/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,33 @@
...
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "openrag-stack.componentSelectorLabels" (dict "component" "openrag" "context" $) | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
+  endpoints:
+    - port: openrag
```
hedhoud, line 30: This always targets the OpenRAG Service on port 8080. In the Linagora values, Ray Serve is enabled and API traffic is routed to raycluster-head-svc on the ray-serve port (80), so this ServiceMonitor—and the pod annotations—scrapes a service that is not serving the API. Please render the target according to the selected topology and account for the NetworkPolicy and per-replica metric registry, or explicitly reject this configuration until Ray Serve metrics are supported.

hedhoud (changes requested):
Two Kubernetes deployment gaps remain before merge. Existing CodeRabbit threads already cover annotation-based discovery and cleartext bearer transport, so this review does not duplicate them.

infra/charts/openrag-stack/values.yaml
paultranvan: That's fair but I believe we have this issue on *all* our secrets, right? If this is the case, we should have a dedicated task for that, but for now I don't think we should worry about that, TBH

infra/charts/openrag-stack/templates/servicemonitor.yaml
paultranvan: Done in c8319a0a3078aecb30dae2fb2310f7ccf0a532fa

infra/charts/openrag-stack/values.yaml
paultranvan: Tracked in #985 (applies to every chart secret, not only METRICS_TOKEN).

hedhoud (approved):
Re-reviewed the current head and the resolved discussions. The unsupported Ray Serve discovery path now fails explicitly instead of rendering a broken target, while the chart-wide secret-rotation concern is correctly tracked in #985. The metrics endpoint remains fail-closed, the admin UI has its own authenticated path, and the transport limitations are documented with TLS required for external scrapes. Current-head CI is green.

PR #915, opened 2026-09-08
fix(health): check dependencies before marking the API ready

### LINKED ISSUE #845: Readiness can be green while OpenRag is unusable: /health_check verifies no dependencies but backs all three probes
## Summary

`/health_check` reports only that the web process is answering. It performs no dependency checks, and the Helm chart wires that same endpoint to **startup, readiness *and* liveness** probes. A pod whose Postgres, Milvus, Ray or model endpoints are unreachable is still marked Ready, and Kubernetes routes traffic to it.

## Code trace

`openrag/api/routers/user/health.py:23-26`:

```python
@router.get("/health_check", summary="Health check endpoint for API", dependencies=[])
async def health_check(request: Request) -> str:
    # TODO: Error reporting about llm and vlm.
    return "RAG API is up."
```

The `# TODO` is the original author flagging exactly this gap.

`infra/charts/openrag-stack/values.yaml:520-534` points all three probes at it:

```yaml
probes:
  startup:   { httpGet: { path: /health_check, port: openrag }, failureThreshold: 30, periodSeconds: 10 }
  readiness: { httpGet: { path: /health_check, port: openrag }, initialDelaySeconds: 10, periodSeconds: 10, failureThreshold: 3 }
  liveness:  { httpGet: { path: /health_check, port: openrag }, ... }
```

## Impact

- **Readiness is meaningless.** A rollout completes and the Service starts sending traffic to a pod that cannot serve a single search or chat request. Every request fails at the dependency instead of being routed to a healthy pod.
- **Startup gives a false green.** `failureThreshold: 30` × `periodSeconds: 10` allows five minutes for dependencies to come up, but nothing actually verifies they did.
- **Rolling updates lose their safety property.** Readiness gating is what makes a rolling update safe; a probe that can't fail turns it into an unguarded restart.

## Suggested fix

Split the endpoints — they have opposite requirements, and conflating them is what makes this awkward to fix:

- **`/health_check` (liveness)** — keep it dependency-free. A liveness probe that fails on a dependency outage causes restart storms precisely when the cluster is already struggling, and restarting the pod cannot fix someone else's Postgres.
- **new `/ready` (readiness + startup)** — check the dependencies actually required to serve: Postgres, Milvus, Ray, and the configured model endpoints. Return 503 with a per-dependency breakdown when any is down.

Then repoint `readinessProbe` and `startupProbe` at `/ready` in `values.yaml`, leaving `livenessProbe` on `/health_check`.

Worth keeping the readiness check cheap and cached (a second or two) so probe traffic doesn't become its own load on Postgres and Milvus.

## Found by

An independent third-party audit of OpenRag; verified against `v2.1.0`.

hedhoud:
Kubernetes can route traffic to OpenRAG before its core dependencies are available. This adds a dedicated `/ready` endpoint and uses it for readiness, while startup and liveness remain on `/health_check` to avoid restart loops during slow dependency startup.

Readiness returns 503 when PostgreSQL, Milvus, or Ray is unavailable. Default model endpoint checks are reported for diagnosis but remain informational, so a model outage does not remove API routes that do not require inference. Checks use short timeouts and a two-second cache.

Model endpoint validation now uses the selected provider and requires revalidation when that provider changes. A stalled Ray actor lookup is shared across probes and isolated from the common executor so repeated readiness timeouts cannot starve unrelated work.

Validation: 3,053 Python unit tests and 299 UI tests passed. Ruff, formatting, the layer import guard, UI lint, and the production UI build also passed. The API and integration workflows will validate the updated branch in CI.

Closes #845

## Summary by CodeRabbit

- **New Features**
  - Added an unauthenticated `/ready` endpoint reporting application and core dependency readiness.
  - Readiness responses include model status while gating availability on PostgreSQL, Milvus, and Ray.
  - Health and readiness probes bypass authentication, rate limiting, and metrics recording.
  - Startup and liveness probes use `/health_check`; readiness probes use `/ready`.
  - Improved model endpoint validation, including vendor, implementation, model availability, and health-only endpoint handling.
- **Documentation**
  - Documented probe behavior, Ray Serve overrides, model credentials, and HTTPS requirements.

### DIFF
--- a/docs/content/docs/documentation/kubernetes.md
+++ b/docs/content/docs/documentation/kubernetes.md
@@ -73,6 +73,14 @@ one before upgrading:
 
 ## Notes
 
+Startup and readiness probes use `/ready`; liveness uses `/health_check`.
+Readiness returns 503 when startup is incomplete or PostgreSQL, Milvus, Ray,
+the default chat or embedding model, or the enabled default reranker is unavailable.
+Checks use short timeouts and results are cached for two seconds. Optional VLM/STT
+and partition-specific model endpoints do not gate the whole API. Model probes
+check availability without running inference; they do not guarantee every request
+will succeed. Use an application image that includes `/ready` with these probes.
+
 - If using a public IP instead of a hostname, you can leave `ingress.host` empty in your `values.yaml`.  
   The ingress will then match all hosts.
 
--- a/docs/content/docs/documentation/kubernetes.md
+++ b/docs/content/docs/documentation/kubernetes.md
@@ -73,13 +73,20 @@ one before upgrading:
 
 ## Notes
 
-Startup and readiness probes use `/ready`; liveness uses `/health_check`.
+For the default direct-API deployment, startup and readiness probes use `/ready`;
+liveness uses `/health_check`. When `ENABLE_RAY_SERVE=true`, override the HTTP
+probes with exec probes as noted in the chart values: the Ray Serve HTTP proxy
+runs on the Ray head, not on the API pod.
 Readiness returns 503 when startup is incomplete or PostgreSQL, Milvus, Ray,
 the default chat or embedding model, or the enabled default reranker is unavailable.
 Checks use short timeouts and results are cached for two seconds. Optional VLM/STT
 and partition-specific model endpoints do not gate the whole API. Model probes
 check availability without running inference; they do not guarantee every request
 will succeed. Use an application image that includes `/ready` with these probes.
+
+Readiness uses the configured model endpoints and API keys, just like inference.
+HTTP endpoints do not encrypt those credentials; configure HTTPS when transport
+encryption is required.
 
 - If using a public IP instead of a hostname, you can leave `ingress.host` empty in your `values.yaml`.  
   The ingress will then match all hosts.
--- a/docs/content/docs/documentation/kubernetes.md
+++ b/docs/content/docs/documentation/kubernetes.md
@@ -73,16 +73,18 @@ one before upgrading:
 
 ## Notes
 
-For the default direct-API deployment, startup and readiness probes use `/ready`;
-liveness uses `/health_check`. When `ENABLE_RAY_SERVE=true`, override the HTTP
+For the default direct-API deployment, startup and liveness probes use
+`/health_check`, while the readiness probe uses `/ready`. When
+`ENABLE_RAY_SERVE=true`, override the HTTP
 probes with exec probes as noted in the chart values: the Ray Serve HTTP proxy
 runs on the Ray head, not on the API pod.
-Readiness returns 503 when startup is incomplete or PostgreSQL, Milvus, Ray,
-the default chat or embedding model, or the enabled default reranker is unavailable.
-Checks use short timeouts and results are cached for two seconds. Optional VLM/STT
-and partition-specific model endpoints do not gate the whole API. Model probes
-check availability without running inference; they do not guarantee every request
-will succeed. Use an application image that includes `/ready` with these probes.
+Readiness returns 503 when startup is incomplete or PostgreSQL, Milvus, or Ray is
+unavailable. Model checks are reported in the response but do not gate the whole
+API, so optional VLM/STT and partition-specific model endpoints do not remove
+healthy replicas from service. Checks use short timeouts and results are cached
+for two seconds. Model probes check availability without running inference; they
+do not guarantee every request will succeed. Use an application image that
+includes `/ready` with these probes.
 
 Readiness uses the configured model endpoints and API keys, just like inference.
 HTTP endpoints do not encrypt those credentials; configure HTTPS when transport
--- a/infra/charts/openrag-stack/values.yaml
+++ b/infra/charts/openrag-stack/values.yaml
@@ -529,7 +529,7 @@ openrag:
   probes:
     startup:
       httpGet:
-        path: /ready
+        path: /health_check
         port: openrag
       timeoutSeconds: 5
       failureThreshold: 30
--- a/infra/charts/openrag-stack/values.yaml
+++ b/infra/charts/openrag-stack/values.yaml
@@ -529,14 +529,16 @@ openrag:
   probes:
     startup:
       httpGet:
-        path: /health_check
+        path: /ready
         port: openrag
+      timeoutSeconds: 5
       failureThreshold: 30
       periodSeconds: 10
     readiness:
       httpGet:
-        path: /health_check
+        path: /ready
         port: openrag
+      timeoutSeconds: 5
       initialDelaySeconds: 10
       periodSeconds: 10
       failureThreshold: 3
--- a/openrag/api/routers/user/health.py
+++ b/openrag/api/routers/user/health.py
@@ -1,28 +1,29 @@
-"""Health / version endpoints (Phase 10F move).
-
-Lifts the inline ``/health_check`` and ``/version`` routes out of
-``openrag/api/main.py`` per the Phase 10A plan ("Inline health endpoint
-(3 lines) -> Dedicated api/routers/user/health.py"). Both routes are
-unauthenticated (``AuthMiddleware`` bypasses them via the bypass list)
-so they can be hit by load balancers, kubernetes probes, and ``curl``
-without a token.
-
-The actual ``app.version`` value is set in ``api/main.py`` and read off
-``request.app.version`` at call time so this router does not need its
-own access to ``importlib.metadata``.
-"""
+"""Public liveness, dependency readiness, and version endpoints."""
 
 from __future__ import annotations
 
 from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse
 
 router = APIRouter()
 
 
 @router.get("/health_check", summary="Health check endpoint for API", dependencies=[])
 async def health_check(request: Request) -> str:
-    # TODO: Error reporting about llm and vlm.
     return "RAG API is up."
+
+
+@router.get("/ready", summary="Dependency readiness", dependencies=[])
+async def ready(request: Request) -> JSONResponse:
+    container = getattr(request.app.state, "container", None)
+    if container is None or not container.is_initialized:
+        return JSONResponse({"status": "not_ready", "checks": {"startup": "unavailable"}}, status_code=503)
+    checks = await container.readiness_service.check()
+    is_ready = all(value == "ok" for value in checks.values())
+    return JSONResponse(
+        {"status": "ready" if is_ready else "not_ready", "checks": checks},
+        status_code=200 if is_ready else 503,
+    )
 
 
 @router.get("/version", summary="Get openRAG version", dependencies=[])
--- a/openrag/api/routers/user/health.py
+++ b/openrag/api/routers/user/health.py
@@ -6,6 +6,7 @@ from fastapi import APIRouter, Request
 from fastapi.responses import JSONResponse
 
 router = APIRouter()
+_CORE_READINESS_CHECKS = frozenset({"postgres", "milvus", "ray"})
 
 
 @router.get("/health_check", summary="Health check endpoint for API", dependencies=[])
@@ -19,7 +20,8 @@ async def ready(request: Request) -> JSONResponse:
     if container is None or not container.is_initialized:
         return JSONResponse({"status": "not_ready", "checks": {"startup": "unavailable"}}, status_code=503)
     checks = await container.readiness_service.check()
-    is_ready = all(value == "ok" for value in checks.values())
+    required_checks = _CORE_READINESS_CHECKS.intersection(checks)
+    is_ready = bool(required_checks) and all(checks[name] == "ok" for name in required_checks)
     return JSONResponse(
         {"status": "ready" if is_ready else "not_ready", "checks": checks},
         status_code=200 if is_ready else 503,
--- a/openrag/di/readiness.py
+++ b/openrag/di/readiness.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import asyncio
+import threading
+from concurrent.futures import Future, ThreadPoolExecutor
 from functools import partial
 from typing import TYPE_CHECKING, cast
 
@@ -13,19 +15,46 @@ from services.storage.postgres_store import PostgresStore
 from services.workers.ray_utils import call_ray_actor_method_with_timeout
 
 _ray_actor = None
+_ray_actor_lookup: Future | None = None
+_ray_actor_lookup_lock = threading.Lock()
+_ray_actor_lookup_executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="readiness-ray-lookup")
 
 if TYPE_CHECKING:
     from di.container import ServiceContainer
 
 
+async def _lookup_ray_actor():
+    global _ray_actor_lookup
+    with _ray_actor_lookup_lock:
+        lookup = _ray_actor_lookup
+        if lookup is None:
+            lookup = _ray_actor_lookup_executor.submit(ray.get_actor, "TaskStateManager", namespace="openrag")
+            _ray_actor_lookup = lookup
+    try:
+        actor = await asyncio.shield(asyncio.wrap_future(lookup))
+    except asyncio.CancelledError:
+        raise
+    except BaseException:
+        with _ray_actor_lookup_lock:
+            if _ray_actor_lookup is lookup:
+                _ray_actor_lookup = None
+        raise
+    with _ray_actor_lookup_lock:
+        if _ray_actor_lookup is lookup:
+            _ray_actor_lookup = None
+    return actor
+
+
 async def _check_ray() -> None:
-    global _ray_actor
+    global _ray_actor, _ray_actor_lookup
     if not ray.is_initialized():
         _ray_actor = None
+        with _ray_actor_lookup_lock:
+            _ray_actor_lookup = None
         raise RuntimeError("Ray is not initialized")
+    if _ray_actor is None:
+        _ray_actor = await _lookup_ray_actor()
     try:
-        if _ray_actor is None:
-            _ray_actor = await asyncio.to_thread(ray.get_actor, "TaskStateManager", namespace="openrag")
         await call_ray_actor_method_with_timeout(
             _ray_actor.get_pool_info.remote, timeout=1.0, task_description="readiness"
         )
--- a/openrag/di/readiness.py
+++ b/openrag/di/readiness.py
@@ -12,15 +12,26 @@ from services.storage.milvus_store import MilvusVectorStore
 from services.storage.postgres_store import PostgresStore
 from services.workers.ray_utils import call_ray_actor_method_with_timeout
 
+_ray_actor = None
+
 if TYPE_CHECKING:
     from di.container import ServiceContainer
 
 
 async def _check_ray() -> None:
+    global _ray_actor
     if not ray.is_initialized():
+        _ray_actor = None
         raise RuntimeError("Ray is not initialized")
-    actor = await asyncio.to_thread(ray.get_actor, "TaskStateManager", namespace="openrag")
-    await call_ray_actor_method_with_timeout(actor.get_pool_info.remote, timeout=1.0, task_description="readiness")
+    try:
+        if _ray_actor is None:
+            _ray_actor = await asyncio.to_thread(ray.get_actor, "TaskStateManager", namespace="openrag")
+        await call_ray_actor_method_with_timeout(
+            _ray_actor.get_pool_info.remote, timeout=1.0, task_description="readiness"
+        )
+    except BaseException:
+        _ray_actor = None
+        raise
 
 
 def create_readiness_service(container: ServiceContainer) -> ReadinessService:
@@ -31,7 +42,7 @@ def create_readiness_service(container: ServiceContainer) -> ReadinessService:
         config = getattr(settings.models, kind).get("default")
         if config is None:
             raise RuntimeError("Default model endpoint is not configured")
-        await check_model_endpoint(config)
+        await check_model_endpoint(config, model_type=kind)
 
     checks = {
         "postgres": cast(PostgresStore, container.catalog_store).check_health,
--- a/openrag/di/readiness.py
+++ b/openrag/di/readiness.py
@@ -1,0 +1,45 @@
+"""Wire readiness to the same stores and model configuration as API requests."""
+
+from __future__ import annotations
+
+import asyncio
+from functools import partial
+from typing import TYPE_CHECKING, cast
+
+import ray
+from services.orchestrators.readiness_service import ReadinessService, check_model_endpoint
+from services.storage.milvus_store import MilvusVectorStore
+from services.storage.postgres_store import PostgresStore
+from services.workers.ray_utils import call_ray_actor_method_with_timeout
+
+if TYPE_CHECKING:
+    from di.container import ServiceContainer
+
+
+async def _check_ray() -> None:
+    if not ray.is_initialized():
+        raise RuntimeError("Ray is not initialized")
+    actor = await asyncio.to_thread(ray.get_actor, "TaskStateManager", namespace="openrag")
+    await call_ray_actor_method_with_timeout(actor.get_pool_info.remote, timeout=1.0, task_description="readiness")
+
+
+def create_readiness_service(container: ServiceContainer) -> ReadinessService:
+    settings = container._require_settings()
+
+    async def check_model(kind: str) -> None:
+        # Resolve on every refresh so admin changes do not leave a stale target.
+        config = getattr(settings.models, kind).get("default")
+        if config is None:
+            raise RuntimeError("Default model endpoint is not configured")
+        await check_model_endpoint(config)
+
+    checks = {
+        "postgres": cast(PostgresStore, container.catalog_store).check_health,
+        "milvus": cast(MilvusVectorStore, container.vector_store).check_health,
+        "ray": _check_ray,
+        "embedder": partial(check_model, "embedder"),
+        "llm": partial(check_model, "llm"),
+    }
+    if settings.reranker.enabled:
+        checks["reranker"] = partial(check_model, "reranker")
+    return ReadinessService(checks)
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -721,6 +721,46 @@ class ModelEndpointService:
         if parsed.username or parsed.password:
             result["detail"] = "Endpoint URL must not include credentials."
             return result
+        if model_type != "stt":
+            from services.orchestrators.readiness_service import (
+                ModelEndpointProbeError,
+                ModelNotFoundError,
+                check_model_endpoint,
+            )
+
+            config = ModelEndpointConfig(
+                endpoint=url,
+                model_name=normalized_model_name,
+                timeout=timeout or 5.0,
+                extra={
+                    **(extra or {}),
+                    **({"api_key": api_key} if api_key else {}),
+                },
+            )
+            try:
+                model_ids = await check_model_endpoint(config, model_type=model_type, timeout=timeout or 5.0)
+            except httpx.TimeoutException:
+                result["detail"] = "Model list request timed out."
+            except ModelEndpointProbeError as exc:
+                result["reachable"] = True
+                result["detail"] = f"Model list returned HTTP {exc.status_code}."
+            except ModelNotFoundError as exc:
+                result["reachable"] = True
+                result["models_served"] = exc.model_ids
+                result["model_found"] = False
+            except Exception as exc:
+                result["detail"] = str(exc)
+            else:
+                result["reachable"] = True
+                result["models_served"] = model_ids
+                result["model_found"] = (
+                    None
+                    if model_ids is None
+                    else True
+                    if normalized_model_name is None
+                    else normalized_model_name in model_ids
+                )
+            return result
         base_url = url.rstrip("/")
         models_url = base_url + "/models"
         headers = {"Authorization": f"Bearer {api_key}"} if api_key else {}
--- a/openrag/services/orchestrators/model_endpoint_service.py
+++ b/openrag/services/orchestrators/model_endpoint_service.py
@@ -724,6 +724,7 @@ class ModelEndpointService:
         if model_type != "stt":
             from services.orchestrators.readiness_service import (
                 ModelEndpointProbeError,
+                ModelListUnavailableError,
                 ModelNotFoundError,
                 check_model_endpoint,
             )
@@ -744,6 +745,9 @@ class ModelEndpointService:
             except ModelEndpointProbeError as exc:
                 result["reachable"] = True
                 result["detail"] = f"Model list returned HTTP {exc.status_code}."
+            except ModelListUnavailableError as exc:
+                result["reachable"] = True
+                result["detail"] = str(exc)
             except ModelNotFoundError as exc:
                 result["reachable"] = True
                 result["models_served"] = exc.model_ids
--- a/openrag/services/orchestrators/readiness_service.py
+++ b/openrag/services/orchestrators/readiness_service.py
@@ -26,6 +26,13 @@ class ModelNotFoundError(ValueError):
         self.model_ids = model_ids
 
 
+class ModelListUnavailableError(ValueError):
+    """A reachable endpoint did not return a usable model list."""
+
+    def __init__(self) -> None:
+        super().__init__("Endpoint returned an invalid model list.")
+
+
 class ReadinessService:
     def __init__(
         self, checks: dict[str, Callable[[], Awaitable[None]]], *, timeout: float = 2.0, cache_ttl: float = 2.0
@@ -89,7 +96,10 @@ async def check_model_endpoint(
         elif response.status_code >= 400:
             raise ModelEndpointProbeError(response.status_code)
     if not health_only:
-        payload = response.json()
+        try:
+            payload = response.json()
+        except ValueError as exc:
+            raise ModelListUnavailableError from exc
         models = payload.get("data") if isinstance(payload, dict) else None
         model_ids = (
             [item["id"] for item in models if isinstance(item, dict) and isinstance(item.get("id"), str)]
@@ -98,7 +108,7 @@ async def check_model_endpoint(
         )
         if not isinstance(models, list) or (config.model_name and config.model_name not in model_ids):
             if not isinstance(models, list):
-                raise ValueError("Configured model is unavailable")
+                raise ModelListUnavailableError
             raise ModelNotFoundError(model_ids)
         return model_ids
     return None
--- a/openrag/services/orchestrators/readiness_service.py
+++ b/openrag/services/orchestrators/readiness_service.py
@@ -1,0 +1,59 @@
+"""Short, cached dependency probes for traffic readiness."""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from collections.abc import Awaitable, Callable
+
+import httpx
+from core.config.model_endpoints import ModelEndpointConfig
+
+
+class ReadinessService:
+    def __init__(
+        self, checks: dict[str, Callable[[], Awaitable[None]]], *, timeout: float = 2.0, cache_ttl: float = 2.0
+    ) -> None:
+        self._checks = checks
+        self._timeout = timeout
+        self._cache_ttl = cache_ttl
+        self._expires_at = 0.0
+        self._result: dict[str, str] = {}
+        self._lock = asyncio.Lock()
+
+    async def check(self) -> dict[str, str]:
+        async with self._lock:
+            if time.monotonic() >= self._expires_at:
+                results = await asyncio.gather(*(self._probe(check) for check in self._checks.values()))
+                self._result = dict(zip(self._checks, results, strict=True))
+                self._expires_at = time.monotonic() + self._cache_ttl
+            return dict(self._result)
+
+    async def _probe(self, check: Callable[[], Awaitable[None]]) -> str:
+        try:
+            await asyncio.wait_for(check(), timeout=self._timeout)
+            return "ok"
+        except TimeoutError:
+            return "timeout"
+        except Exception:
+            # This endpoint is public: never return connection strings or keys.
+            return "unavailable"
+
+
+async def check_model_endpoint(config: ModelEndpointConfig) -> None:
+    """Check availability without generating tokens or running embeddings."""
+    base = config.endpoint.rstrip("/")
+    implementation = config.extra.get("implementation", "vllm")
+    if implementation == "ollama" and not base.endswith("/v1"):
+        base += "/v1"
+    health_only = implementation in {"infinity", "tei"}
+    url = base + ("/health" if health_only else "/models")
+    api_key = config.extra.get("api_key")
+    headers = {"Authorization": f"Bearer {api_key}"} if api_key else {}
+    async with httpx.AsyncClient(timeout=2.0, headers=headers) as client:
+        response = await client.get(url)
+        response.raise_for_status()
+    if not health_only:
+        models = response.json()["data"]
+        if not isinstance(models, list) or (config.model_name and config.model_name not in [m["id"] for m in models]):
+            raise ValueError("Configured model is unavailable")
--- a/openrag/services/orchestrators/readiness_service.py
+++ b/openrag/services/orchestrators/readiness_service.py
@@ -7,7 +7,23 @@ import time
 from collections.abc import Awaitable, Callable
 
 import httpx
-from core.config.model_endpoints import ModelEndpointConfig
+from core.config.model_endpoints import DEFAULT_MODEL_IMPLEMENTATIONS, ModelEndpointConfig
+
+
+class ModelEndpointProbeError(RuntimeError):
+    """A model endpoint answered with an unsuccessful HTTP status."""
+
+    def __init__(self, status_code: int) -> None:
+        super().__init__(f"Model endpoint returned HTTP {status_code}")
+        self.status_code = status_code
+
+
+class ModelNotFoundError(ValueError):
+    """A reachable endpoint does not serve the configured model."""
+
+    def __init__(self, model_ids: list[str]) -> None:
+        super().__init__("Configured model is unavailable")
+        self.model_ids = model_ids
 
 
 class ReadinessService:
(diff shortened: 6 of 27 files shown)


docs/content/docs/documentation/kubernetes.md
```diff
@@ -73,6 +73,14 @@ one before upgrading:
 
 ## Notes
 
+Startup and readiness probes use `/ready`; liveness uses `/health_check`.
```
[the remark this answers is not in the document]
hedhoud: Fixed in 4ad8b26c. The documentation now limits the HTTP probe defaults to direct-API deployments and explains the required exec-probe override for Ray Serve, whose HTTP proxy runs on the Ray head.

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,59 @@
...
+    base = config.endpoint.rstrip("/")
+    implementation = config.extra.get("implementation", "vllm")
+    if implementation == "ollama" and not base.endswith("/v1"):
+        base += "/v1"
+    health_only = implementation in {"infinity", "tei"}
+    url = base + ("/health" if health_only else "/models")
+    api_key = config.extra.get("api_key")
+    headers = {"Authorization": f"Bearer {api_key}"} if api_key else {}
+    async with httpx.AsyncClient(timeout=2.0, headers=headers) as client:
+        response = await client.get(url)
```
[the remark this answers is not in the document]
hedhoud: The transport risk is real, but rejecting every HTTP endpoint with an API key would break supported deployments: the default embedder uses http://vllm:8000/v1 with api_key: EMPTY, and existing inference clients accept operator-configured HTTP endpoints with bearer authentication. Readiness needs to follow the same configuration. I have documented the lack of credential encryption over HTTP in 4ad8b26c. A mandatory HTTPS policy would be a separate compatibility change, so I have not introduced it in this readiness fix. Can this finding be dismissed for this PR?

infra/charts/openrag-stack/values.yaml
```diff
@@ -529,14 +529,16 @@ openrag:
   probes:
     startup:
       httpGet:
-        path: /health_check
+        path: /ready
```
Ahmath-Gadji: **[Note — downgraded from Blocking, see correction below]** The chart's pinned image tags predate `/ready`.

`openrag.image.tag` at line 452 of this file is `"v2.2.0"`, and `/ready` is introduced by this PR, so the pinned image 404s on the path these probes now use.

**Correction:** I originally filed this as blocking. It isn't, on two counts:

1. The `v2.2.0` pins are an artifact of this branch's base (`3440b46b`, pre-2.2.1). `develop` is already at `appVersion: 2.2.1` / `tag: "v2.2.1"` / chart `0.6.4` — the stale pins disappear on rebase.
2. Image pins are bumped as a release step, not per-feature. `chore(release): bump version to 2.2.1` (72de8fe5) moved all three, and RELEASING.md step 8 enforces it: it parses values.yaml with PyYAML, requires exactly one image block per OpenRAG repository, and fails when any tag differs from the release tag — explicitly so a stale pin can't ship. "Chart pins release N while develop carries N+1 features" is the normal steady state, not something this PR introduces.

What's left is narrow and worth a line in the release notes rather than a change here: anyone installing the interim `-dev` chart between this merge and the next release gets probes pointing at an endpoint the pinned release image lacks. That window applies to any endpoint-adding PR and the release gate closes it.

Unrelated observation while checking: nothing in CI renders this chart. `helm.yaml` runs `dependency update` -> `package` -> `push` on merge to `develop`, with no `helm template`/`lint`/`install` and no PR trigger, so chart changes reach a published artifact with no validation at all. Probably worth its own issue.

infra/charts/openrag-stack/values.yaml
```diff
@@ -529,14 +529,16 @@ openrag:
...
         port: openrag
+      timeoutSeconds: 5
```
Ahmath-Gadji, line 530: **[Blocking]** Gating the *startup* probe on the model endpoints will restart the API while vLLM is still loading.

Startup is `failureThreshold: 30` x `periodSeconds: 10` with no `initialDelaySeconds` — a 300 s budget. `/ready` now requires the default `llm` and `embedder` endpoints to answer (`di/readiness.py:40-41`), and the chart deploys vLLM in-cluster with `modelURL` pulled from HF at pod start. A cold node routinely needs well over 5 minutes for that. There is no init container making openrag wait either — `templates/openrag.yaml:29-46` only waits on the venv.

Result: the kubelet kills the openrag container mid-model-load, and every restart resets the clock (plus re-runs migrations/seeding and re-attaches the Ray actors), so the pod can take far longer to converge than its dependencies do — or not converge at all while a large model is still pulling.

Worth noting a startup probe failure *restarts the container*, which is the same restart-storm failure mode the issue correctly rejected for liveness. Suggest keeping `startupProbe` on `/health_check` and letting `readinessProbe` do the dependency gating — readiness only removes the pod from the Service, which is the behaviour #845 actually wants. If you'd rather keep startup on `/ready`, the budget needs to be much larger (`failureThreshold: 90`+).

openrag/di/readiness.py
```diff
@@ -0,0 +1,45 @@
...
+from services.workers.ray_utils import call_ray_actor_method_with_timeout
+
+if TYPE_CHECKING:
+    from di.container import ServiceContainer
+
+
+async def _check_ray() -> None:
+    if not ray.is_initialized():
+        raise RuntimeError("Ray is not initialized")
+    actor = await asyncio.to_thread(ray.get_actor, "TaskStateManager", namespace="openrag")
```
Ahmath-Gadji: **[Worth fixing]** This leaks threads from the shared default executor during a Ray outage.

`ray.get_actor` is documented as a sync call that blocks up to 60 s (`RAY_gcs_server_request_timeout_seconds`), and `asyncio.wait_for(..., 2.0)` in `_probe` cannot cancel an `asyncio.to_thread` thread — the same caveat you called out in the Milvus check ("cancelling to_thread cannot stop it"), except there you bounded it by passing `timeout=2.0` into the SDK. There's no equivalent bound here.

With a readiness probe every 10 s, a GCS outage parks ~6 threads simultaneously against a default executor of `min(32, cpu+4)` workers — 6 slots on a 2-CPU pod. That pool is shared with `MilvusVectorStore.check_health`, `collection_exists`, `_iter_query` and `_ensure_loaded`, so a Ray outage makes readiness *also* report `milvus: timeout` (misdiagnosing which dependency is down) and stalls real query paths.

Simplest fix: resolve the named handle once and cache it, re-looking-up only on failure. The actor is named + detached, and `di/workers.py:75-90` already relies on handles staying bound across restarts.

openrag/di/readiness.py
```diff
@@ -0,0 +1,45 @@
...
+
+    async def check_model(kind: str) -> None:
+        # Resolve on every refresh so admin changes do not leave a stale target.
+        config = getattr(settings.models, kind).get("default")
+        if config is None:
+            raise RuntimeError("Default model endpoint is not configured")
+        await check_model_endpoint(config)
+
+    checks = {
+        "postgres": cast(PostgresStore, container.catalog_store).check_health,
```
Ahmath-Gadji, line 77: **[Worth fixing]** This scope makes every replica go NotReady together, turning a partial degradation into a full outage.

`PostgresStore.check_health` acquires a pooled connection with `timeout=2.0` on every refresh. Under pool saturation that raises on *all* replicas at once, and with `failureThreshold: 3` / `periodSeconds: 10`, 30 s of load pulls the entire Service out of the endpoints list — the ingress then 503s everything instead of the deployment shedding load gracefully. That's the classic readiness-probe feedback loop: the probe fails *because* of load, and taking every pod out makes recovery flap rather than settle.

Same shape for the `llm`/`embedder` entries below: a brief model-endpoint blip also fails indexing-status, admin, workspace and jobs requests that never touch an LLM.

Consider gating on the stores + Ray only, and reporting the model endpoints in the response body without letting them flip the status code — that also lines up better with the PR description's "optional and partition-specific models do not gate the whole API".

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,59 @@
...
+            return "timeout"
+        except Exception:
+            # This endpoint is public: never return connection strings or keys.
+            return "unavailable"
+
+
+async def check_model_endpoint(config: ModelEndpointConfig) -> None:
+    """Check availability without generating tokens or running embeddings."""
+    base = config.endpoint.rstrip("/")
+    implementation = config.extra.get("implementation", "vllm")
```
Ahmath-Gadji: **[Worth fixing]** The `"vllm"` fallback disagrees with the reranker's actual default implementation.

The DI factory uses `default_impl="infinity"` for rerankers (`di/container.py:180-185`), and the admin UI keeps its own `DEFAULT_VENDOR_BY_TYPE` with `reranker: "infinity"` (`ui/src/lib/api/models.ts:307`). `extra` is free-form on create (`extra: dict[str, Any] = Field(default_factory=dict)` in `CreateModelEndpointRequest`), so `implementation` can legitimately be absent on an admin-created row.

Such a reranker is served by `InfinityReranker`/TEI at runtime but probed here as vllm at `{base}/models` — a 404 on TEI, so `reranker` reports `unavailable` forever and the pod never becomes Ready even though reranking works fine. (The seeded rows are safe: `model_endpoint_service.py:406` writes `implementation: s.reranker.provider`.)

Suggest passing the kind through and sharing one per-kind default map with `_wire_named_component_factories`, so this is not a third place encoding the same knowledge.

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,59 @@
...
+            await asyncio.wait_for(check(), timeout=self._timeout)
+            return "ok"
+        except TimeoutError:
+            return "timeout"
+        except Exception:
+            # This endpoint is public: never return connection strings or keys.
+            return "unavailable"
+
+
+async def check_model_endpoint(config: ModelEndpointConfig) -> None:
```
Ahmath-Gadji: **[Minor]** This largely reimplements `ModelEndpointService.validate_endpoint` (`model_endpoint_service.py:683`), which already probes `/models`, builds the same `Authorization` header, and checks `model_name` against the served ids — with sturdier parsing (`isinstance(item, dict) and isinstance(item.get("id"), str)` vs. `[m["id"] for m in models]` here) and explicit 401/403 handling.

The consequence isn't just duplication: the two now disagree about the same endpoint. `validate_endpoint` always hits `/models`; this hits `/health` for infinity/tei and appends `/v1` for ollama. So an operator can see the admin UI's "Validate" button go green while `/ready` reports that endpoint unavailable, or vice versa, with no way to tell which one is right.

Worth extracting the shared probe so both paths answer the same question.

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,59 @@
...
+                results = await asyncio.gather(*(self._probe(check) for check in self._checks.values()))
+                self._result = dict(zip(self._checks, results, strict=True))
+                self._expires_at = time.monotonic() + self._cache_ttl
+            return dict(self._result)
+
+    async def _probe(self, check: Callable[[], Awaitable[None]]) -> str:
+        try:
+            await asyncio.wait_for(check(), timeout=self._timeout)
+            return "ok"
+        except TimeoutError:
```
Ahmath-Gadji: **[Minor]** This branch never fires for the model checks. httpx raises `httpx.ReadTimeout`/`ConnectTimeout` (subclasses of `httpx.TimeoutException` -> `httpx.TransportError`), which is not a `TimeoutError`, so it falls through to `except Exception` and a hung model endpoint is labelled `unavailable` rather than `timeout`.

The inner client timeout (`httpx.AsyncClient(timeout=2.0)`) also equals `self._timeout`, so which one wins is a coin flip. Making the client timeout slightly shorter — or catching `httpx.TimeoutException` here too — would make the `timeout` label mean what it says, which matters because the two statuses point an operator at different causes.

openrag/services/storage/milvus_store.py
```diff
@@ -1326,6 +1326,11 @@ async def delete_by_filter(self, filters: dict[str, Any]) -> int:
 
         return int(result.get("delete_count", 0)) if isinstance(result, dict) else 0
 
+    async def check_health(self) -> None:
+        # A missing collection is normal before the first upload. A failed RPC
+        # is not. Bound the SDK call too: cancelling to_thread cannot stop it.
+        await asyncio.to_thread(self._client.has_collection, self._collection_name, timeout=2.0)
```
Ahmath-Gadji: **[Minor]** The store already holds an `AsyncMilvusClient` (`self._async_client`, built in `__init__`), and pymilvus 3.0.1's `AsyncMilvusClient` does expose `has_collection` — I checked against the pinned version. `await self._async_client.has_collection(self._collection_name, timeout=2.0)` drops the thread entirely and makes the "cancelling to_thread cannot stop it" caveat in the comment moot.

Also note this is now the third copy of the same call — `collection_exists()` two methods below and the inline one at line 927 both do `asyncio.to_thread(self._client.has_collection, self._collection_name)`. Reusing `collection_exists` (or having it take the timeout) would keep it to one.

Ahmath-Gadji (commented):
### Review summary

_Edited: the first finding below was originally filed as blocking and has been downgraded to a note — the chart's stale image pins are branch-base drift plus a release-time step, not a defect in this PR. Corrected in the thread on `values.yaml:532`._

The endpoint split is the right shape and the implementation is clean — liveness stays dependency-free, `/ready` is cached, bypasses auth, is exempt from rate limiting, and the probes never run inference. Lint, format, the layer guard and the 25 new unit tests all pass on the branch (verified locally).

The problems are all in what this does in a real cluster rather than in the code's logic:

**Blocking (1)**
1. `startupProbe` on `/ready` has a 300 s budget and now gates on the vLLM endpoints, so the API container gets killed while the models are still loading their weights. A startup probe failure restarts the container — the same restart-storm mode the issue correctly rejected for liveness.

**Worth fixing (3)**
2. `_check_ray` parks uncancellable 60 s threads in the shared default executor during a Ray outage, starving every other `asyncio.to_thread` caller — including the Milvus check right next to it.
3. The `implementation` fallback is `"vllm"`, but rerankers default to `"infinity"` everywhere else in the codebase.
4. Readiness is scoped so that Postgres pool saturation or a brief LLM blip takes every replica out of the Service at once.

**Minor (3)** — duplicate probe logic vs. `validate_endpoint`, an `except TimeoutError` that never matches httpx, and a Milvus check that can drop its thread entirely.

**Note (1)** — the chart's `v2.2.0` pins vs. `/ready`; see the thread for why this resolves itself at release time.

Details inline. Nothing here questions the approach — the remaining blocker is probe configuration, and the rest is about blast radius under failure.

One process note: since these touch four files that are already small, please fold the fixes in with `git commit --fixup` + `git rebase --autosquash` (or just amend) rather than stacking a series of one-line fix commits — it keeps the two-commit history of this branch readable.

infra/charts/openrag-stack/values.yaml
hedhoud: Thanks for the correction. I agree this is not a blocker for the feature branch because the release bump and release gate update the image tag. The interim dev-chart window and missing chart-render validation are worth tracking separately, but no change is required here.

infra/charts/openrag-stack/values.yaml
hedhoud: Agreed. The startup probe should remain on `/health_check`; `/ready` should only control readiness while dependencies, especially in-cluster models, are still starting. This avoids restart loops during normal model loading.

openrag/services/orchestrators/readiness_service.py
hedhoud: Agreed. `httpx.TimeoutException` should be handled explicitly so connection/read timeouts are reported as `timeout` rather than `unavailable`; the outer and client timeout boundaries should also be deterministic.

openrag/services/orchestrators/readiness_service.py
hedhoud: Agreed. This duplicates endpoint validation and can produce conflicting results between the admin validation flow and `/ready`. Shared probe/validation logic should be extracted so both paths use the same URL, authentication, response parsing, and model checks.

openrag/services/storage/milvus_store.py
hedhoud: Agreed. Since the store already owns an async Milvus client, the health check should use it or reuse the existing collection-health helper instead of adding another `to_thread` call and duplicating the same RPC.

openrag/di/readiness.py
hedhoud: Agreed. `ray.get_actor` can block beyond the readiness timeout because `to_thread` cancellation does not stop the underlying call. The probe should avoid accumulating executor threads, for example by caching the detached actor handle and only re-resolving it after a failure.

openrag/services/orchestrators/readiness_service.py
hedhoud: Agreed. The fallback must match the reranker’s actual default implementation. The readiness probe should reuse the same per-kind implementation defaults as dependency wiring so admin-created endpoints do not remain permanently unavailable.

openrag/di/readiness.py
hedhoud: Agreed. Readiness should not remove every API replica from service because an optional model or temporary pool-saturation event affects a narrower feature. The status policy should gate on dependencies required for core API traffic and report model-specific checks without making the whole deployment NotReady.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -721,6 +721,46 @@ async def validate_endpoint(
...
+                endpoint=url,
+                model_name=normalized_model_name,
+                timeout=timeout or 5.0,
+                extra={
+                    **(extra or {}),
+                    **({"api_key": api_key} if api_key else {}),
+                },
+            )
+            try:
+                model_ids = await check_model_endpoint(config, model_type=model_type, timeout=timeout or 5.0)
```
Ahmath-Gadji, line 742: `validate_endpoint` can never see `implementation` here, so `model_implementation()` always falls back to `DEFAULT_MODEL_IMPLEMENTATIONS["reranker"] = "infinity"` and **every** reranker gets probed at `/health` instead of `/models`.

Both validate routes drop `extra` for non-STT types:

```python
# api/routers/admin/model_endpoints.py:255 and :273
extra=body.extra if body.model_type == "stt" else None,
extra=endpoint.extra if model_type == "stt" else None,
```

So the `ModelEndpointConfig` built just above always has an `extra` without `implementation` (only `api_key`), regardless of the endpoint's real provider.

For a reranker served by an OpenAI-compatible backend — the `openai` provider documented at `conf/config.yaml:105` — where `/health` is not routed under the `/v1` base:

```jsonc
// develop
{"reachable": true, "model_found": true, "models_served": [...], "detail": null}
// this branch
{"reachable": true, "model_found": null, "models_served": null, "detail": "Model list returned HTTP 404."}
```

Reproduced against a running instance of this branch, on both `POST /model-endpoints/{model_type}/{name}/validate` and the draft route `POST /model-endpoints/validate` — the draft call sent `"extra": {"implementation": "openai"}` explicitly in the request body and got the identical result, because the router discards it before `validate_endpoint` is reached. The same instance's `/ready` reports `"reranker": "ok"`, since readiness reads the stored row's `extra` and does resolve the provider correctly — the divergence is specific to this path.

Downstream in the admin UI this lands in the **success** branch (`ui/src/pages/admin/models.tsx:697`): a green toast whose text is `Model list returned HTTP 404.`. And the `model_found === false` guard at `:681` — the one whose comment says it exists because "a green 'reachable' hides a wrong model name" — is now dead for every reranker, infinity included, since infinity does serve `/models`.

The new test can't catch this: `test_validate_health_only_endpoint_keeps_model_presence_unknown` passes `extra={"implementation": "infinity"}`, a shape neither route can produce.

Suggested fix: forward `extra` for all model types in both routes (`extra=body.extra` / `extra=endpoint.extra`). The STT-only gate is there to keep STT request options out of the multipart transcription body, and that concern is already handled separately in `_stt_validation_form_data`. Alternatively, keep `validate_endpoint` on the plain `/models` probe and leave the `health_only` branch to the readiness caller.

EnjoyBacon7 (commented):
_(Retracted — posted from the wrong account by mistake. Superseded; disregard.)_

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -721,6 +721,46 @@ async def validate_endpoint(
...
+            except httpx.TimeoutException:
+                result["detail"] = "Model list request timed out."
+            except ModelEndpointProbeError as exc:
+                result["reachable"] = True
+                result["detail"] = f"Model list returned HTTP {exc.status_code}."
+            except ModelNotFoundError as exc:
+                result["reachable"] = True
+                result["models_served"] = exc.model_ids
+                result["model_found"] = False
+            except Exception as exc:
```
Ahmath-Gadji, line 755: **Regression: a reachable endpoint whose `/models` body isn't OpenAI-shaped now reports `reachable: false`, which blocks save in the admin UI.**

On `develop`, reachability was decided by one fact — the server answered — and set *before* parsing, so a parse failure could only downgrade the model fields:

```python
result["reachable"] = True          # we got an HTTP response
if resp.status_code == 200:
    try:
        ...parse the list...
    except (TypeError, ValueError):
        result["detail"] = "Endpoint returned an invalid model list."   # never touches reachable
```

Here everything is decided by a single try/except, so both `check_model_endpoint`'s bare `ValueError` (non-list `data`) and an unparseable body (`json.JSONDecodeError`, also a `ValueError`) land on this line with `reachable` still `False`:

| endpoint answers | develop | this branch |
|---|---|---|
| `200 {"models": [...]}` | `reachable: true`, `"Endpoint returned an invalid model list."` | `reachable: false`, `"Configured model is unavailable"` |
| `200`, non-JSON body | `reachable: true`, same message | `reachable: false`, `"Expecting value: line 1 column 1 (char 0)"` |

`ui/src/pages/admin/models.tsx:1022` disables Save unless `validated === true`, and `:676` sets it `false` on `!reachable` — so an endpoint that was registrable before this PR can no longer be registered, with a raw parser message as the stated reason.

Suggested fix: derive `reachable` from *transport success* rather than from the absence of an exception, and let the parse outcome decide only `model_found` / `models_served`. That needs the probe to distinguish "never answered" from "answered, but the list was unusable" — see my note on the exception types in `readiness_service.py`.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -721,6 +721,46 @@ async def validate_endpoint(
...
+                result["detail"] = str(exc)
+            else:
+                result["reachable"] = True
+                result["models_served"] = model_ids
+                result["model_found"] = (
```
Ahmath-Gadji, line 760: `model_found` is now `True` when no model name was supplied.

`True if normalized_model_name is None` — but "the caller didn't ask about a model" isn't the same as "the model is served". `develop` left it `None` (unknown) and only set a boolean when `normalized_model_name is not None`.

Visible effect: `ui/src/pages/admin/models.tsx:697` picks its success message off this field, so a nameless probe renders `Reachable — "" is served.`

`None` is the honest value here.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -721,6 +721,46 @@ async def validate_endpoint(
...
+                model_name=normalized_model_name,
+                timeout=timeout or 5.0,
+                extra={
+                    **(extra or {}),
+                    **({"api_key": api_key} if api_key else {}),
+                },
+            )
+            try:
+                model_ids = await check_model_endpoint(config, model_type=model_type, timeout=timeout or 5.0)
+            except httpx.TimeoutException:
```
Ahmath-Gadji, line 743: Two `detail` strings got coarser on this path, and one can come back empty.

- **Timeout:** `develop` said `"Connection timed out"`; this branch says `"Model list request timed out."` for any `httpx.TimeoutException`, including a *connect* timeout where no request was ever sent.
- **Connection error** (`:751`): `develop` had a dedicated `except httpx.ConnectError` producing `f"Connection error: {exc}"`. That's now folded into the catch-all as a bare `str(exc)`. `httpx.ConnectError` frequently stringifies to `""`, so `detail` can be an empty string — at which point `ui/src/pages/admin/models.tsx:678` falls back to the generic `"Endpoint is unreachable."` and the operator loses the cause.

Keeping a distinct `except httpx.ConnectError` with the prefix restores both.

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,104 @@
...
+        payload = response.json()
+        models = payload.get("data") if isinstance(payload, dict) else None
+        model_ids = (
+            [item["id"] for item in models if isinstance(item, dict) and isinstance(item.get("id"), str)]
+            if isinstance(models, list)
+            else []
+        )
+        if not isinstance(models, list) or (config.model_name and config.model_name not in model_ids):
+            if not isinstance(models, list):
+                raise ValueError("Configured model is unavailable")
```
Ahmath-Gadji: **These two conditions are indistinguishable to a caller, and that's what produces the `reachable` regression in `validate_endpoint`.**

`ModelNotFoundError` subclasses `ValueError`, and this line raises a bare `ValueError` carrying the *same* message (`"Configured model is unavailable"`). So "answered 200 but the payload wasn't a model list" and "answered 200 with a valid list that doesn't contain the model" can't be told apart by type or by message — yet they need opposite `model_found` values (`None` vs `False`), and both need `reachable: true`.

Separately: both exception classes are declared in this module and inherit from builtins. `core/utils/exceptions.py` is the project's exception home — everything rooted at `OpenRAGError` with a `code`, `status_code` and `extra`, with the tree drawn out in the module docstring; CLAUDE.md states the convention. `InferenceError` (503), `InferenceConnectionError` and `InferenceTimeoutError` already live there, and are already what the inference adapters raise (`ollama_client.py:92-94`, `reranker_clients.py:51`).

Suggested shape — one slot per outcome, in `core/utils/exceptions.py`:

```
InferenceError
├── InferenceConnectionError    never answered        → reachable=False
├── InferenceTimeoutError       never answered        → reachable=False
├── EndpointProbeError          answered non-2xx      → reachable=True
├── ModelListUnavailableError   200, unparseable      → reachable=True,  model_found=None
└── ModelNotServedError         200, parsed, absent   → reachable=True,  model_found=False
```

`validate_endpoint` then maps type → fields instead of falling through to a catch-all, and `ReadinessService._probe` stops needing `import httpx` for its `except httpx.TimeoutException` — at which point it carries no I/O vocabulary at all and could reasonably live in `core/`.

openrag/di/readiness.py
```diff
@@ -0,0 +1,56 @@
...
+
+    async def check_model(kind: str) -> None:
+        # Resolve on every refresh so admin changes do not leave a stale target.
+        config = getattr(settings.models, kind).get("default")
+        if config is None:
+            raise RuntimeError("Default model endpoint is not configured")
+        await check_model_endpoint(config, model_type=kind)
+
+    checks = {
+        "postgres": cast(PostgresStore, container.catalog_store).check_health,
```
Ahmath-Gadji, line 77: **The `cast()`s are a sign the capability is missing from the ports.**

`check_health` was added to the concrete `MilvusVectorStore` and `PostgresStore`, but not to the `VectorStore` ABC (`core/vector_stores/vector_store.py`) or the `CatalogStore` port (`core/ports/catalog_store.py`). So wiring code has to narrow a port back to its adapter to reach it — which type-checks only because of the cast, and would fail at runtime with `AttributeError` if the adapter were swapped. Declaring `check_health` on both ports removes both casts and both concrete imports.

The inference side has the same gap but pays more for it. `readiness_service.py` re-derives provider knowledge that already exists in the adapters:

- `implementation` resolution — already at `di/factories.py:99`, which is what picks the class in the first place
- "ollama needs `/v1` appended" — already at `ollama_client.py:71-73` and `:169-171`, duplicated verbatim
- "infinity/tei use `/health`, everything else `/models`" — already implicit in `reranker_clients.py`

That's a fourth copy, and nothing fails when the copies drift — which is exactly the bug in [my other comment](https://github.com/linagora/openrag/pull/915#discussion_r3968107777): readiness resolves the provider correctly, `validate_endpoint` doesn't.

`LLM`, `Embedder` and `Reranker` are already clean ABCs in `core/`. A non-abstract default there costs nothing:

```python
class Reranker(ABC):
    async def check_health(self) -> None:
        """Raise InferenceError if this endpoint can't serve requests."""
        return None
```

Every adapter already holds exactly what a probe needs — no new state:

| adapter | `self._endpoint` as stored | probe |
|---|---|---|
| `VLLMClient` / `VLLMEmbedder` | includes `/v1` by convention | `GET {_endpoint}/models` |
| `OllamaClient` / `OllamaEmbedder` | `__init__` already appends `/v1` | `GET {_endpoint}/models` |
| `InfinityReranker` / `TEIReranker` | server root (they POST `{_endpoint}/rerank`) | `GET {_endpoint}/health` |
| `OpenAIReranker` | server root, OpenAI-shaped | `GET {_endpoint}/models` |

Four of those seven are identical, so the bodies should delegate to two shared helpers rather than being written out seven times — the point isn't bespoke logic per provider, it's that **which helper applies is expressed by the class** instead of by an `if implementation in {...}` in a service that first has to reconstruct `implementation` from config.

Reusing `self._client` also gives the probe the same credentials real traffic uses (note `OllamaEmbedder` deliberately builds its client with no auth — a config-driven prober has to re-guess that; a method inherits it). One caveat: don't decorate `check_health` with `@with_circuit_breaker` — a probe has to observe real state, not be short-circuited.

Wiring then reduces to `container.llm_factory("default").check_health`, with no `model_implementation()`, no `DEFAULT_MODEL_IMPLEMENTATIONS` lookup and no URL rules in this file. And `validate_endpoint` can build a throwaway adapter from the draft config through the same registry and call the same method — at which point the Test button, `/ready` and real inference share one resolution path by construction.

openrag/services/orchestrators/readiness_service.py
```diff
@@ -0,0 +1,104 @@
...
+
+def model_implementation(config: ModelEndpointConfig, model_type: str | None = None) -> str:
+    """Resolve the configured provider, matching the DI factory defaults."""
+    configured = config.extra.get("implementation")
+    if isinstance(configured, str) and configured:
+        return configured
+    return DEFAULT_MODEL_IMPLEMENTATIONS.get(model_type or "", "vllm")
+
+
+async def check_model_endpoint(
```
Ahmath-Gadji, line 74: This probe already exists in the codebase: `openrag/services/inference/healthcheck.py` has `check_endpoint_health()` (OpenAI-compatible `/v1/models`), `check_infinity()` (`/health`), and `check_model_available()` (probe + verify a specific model is served).

It has **no production callers** — grepping `openrag/` for `healthcheck`, `check_endpoint_health`, `check_infinity`, `check_model_available`, `HealthResult` and `EndpointStatus` turns up only `tests/unit/services/inference/test_healthcheck.py`. Its module docstring's claim that it's "used at container startup (fail-fast) and by the `/health_check` route" is stale.

So this PR makes a third implementation of the same probe. Worth resolving either way — revive it or delete it in this change, rather than leaving two dead variants beside a live one.

It's also worth a look before rewriting, because its result type already encodes the distinction that got lost:

```python
class EndpointStatus(str, Enum):
    HEALTHY      # answered, list parsed, model present
    UNHEALTHY    # answered, but not usable   → reachable=True
    UNREACHABLE  # never answered             → reachable=False
```

`UNHEALTHY` vs `UNREACHABLE` is exactly the `reachable` distinction behind the regression I flagged in `model_endpoint_service.py`, and `check_model_available()` already returns `UNHEALTHY` *with `models` populated* for "reachable but the model isn't served" — verbatim `{reachable: true, model_found: false, models_served: [...]}`.

If you go the value-object route, one field needs widening: `models: list[str]` defaults to `[]`, which conflates "advertises nothing" with "couldn't parse the list". `validate_endpoint` needs those to differ (`model_found` `False` vs `None`), so it should be `list[str] | None`.

andyne13:
One more on the model probes, separate from the gating discussion above — that part looks settled in the right place.

`create_readiness_service` builds `check_model(kind)` per model *type*, and it raises "Default model endpoint is not configured", so what gets probed is the **default** endpoint for each kind. But a partition can name a different registry row through seven fields:

- `PartitionRow.embedder`, `PartitionRow.chat_llm`
- `RetrievalPipelineConfig.reranker`, `RetrievalPipelineConfig.llm` (multiQuery / hyde)
- `IndexationPipelineConfig.vlm`, `.stt`, `.contextualization_llm`

So a healthy default says nothing about the endpoint a given partition's queries actually hit. And the failure is quiet: `_resolve_llm_endpoint_name` (`api/routers/user/chat.py:302-308`) documents that a `chat_llm` naming a deleted or renamed endpoint silently falls back to `"default"`, because nothing cascades the preset. A tenant configured for a large-context model gets answered by the default one and nothing anywhere reports it.

Suggestion, keeping the current non-gating scope:

- Probe the **distinct endpoint URLs actually referenced** — the per-kind defaults plus whatever partition presets name — deduplicated by URL rather than by registry name. Registry rows that no preset names and that aren't a default are dead configuration; probing those produces alerts nobody can action. Dedup by URL also keeps the set bounded by how many inference services exist rather than how many partitions do.
- Report an **unresolvable preset reference** as its own condition rather than folding it into "endpoint down". It isn't an unreachable endpoint, it's a config error with a silent fallback, and the fix is reassigning the preset rather than restarting anything.

Both are additive to what's here — the probe machinery (concurrent `gather`, 2 s timeout, 2 s cache) already carries it.

openrag/api/routers/user/health.py
```diff
@@ -1,30 +1,33 @@
...
-"""
+"""Public liveness, dependency readiness, and version endpoints."""
 
 from __future__ import annotations
 
 from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse
 
 router = APIRouter()
+_CORE_READINESS_CHECKS = frozenset({"postgres", "milvus", "ray"})
```
Ahmath-Gadji, line 9: Excluding the model endpoints from the gating set is right — they're a shared dependency, so gating on them would flip every replica NotReady at once and take down routes that need no LLM. As-is it surfaces as a `503 INFERENCE_ERROR` on the chat route only, which is the better blast radius.

Given they're informational, two things worth weighing:

**They're narrower than they look.** `di/readiness.py:41` probes `models.<kind>["default"]` — the default alias, and only 3 of the 5 types. But selection is per-partition and per-request (`retrieval_service.py:206`, `:263`, `chat.py:413`), so a multi-tenant deployment can report `"llm": "ok"` for an endpoint nobody uses while the one serving that partition is down.

**They aren't free.** `cache_ttl=2.0` < `periodSeconds: 10`, so every probe forces a full refresh: ~26k outbound `/models` calls per pod per day, for a value the handler discards.

So either keep them and make the contract explicit — separate the gating checks from the informational ones in the payload, and give the model checks their own longer TTL — or drop them, since `_circuit_breaker.py` already tracks this from real traffic, for the endpoint each partition actually resolved to.

I'd lean drop, since the per-partition gap makes the value misleading rather than merely incomplete. Your call either way.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Withdrawing this one.

`"Model list request timed out."` is the more precise of the two strings, so that bullet is a non-issue. And I overstated the second: `str(httpx.ConnectError)` is `'All connection attempts failed'` for a refused connection and `'[Errno -2] Name or service not known'` for a DNS failure — I checked against the pinned httpx, it isn't empty on the paths that matter. What's actually lost is the `Connection error: ` prefix, which the operator doesn't need.

Not worth a change. Resolving.

openrag/services/orchestrators/readiness_service.py
Ahmath-Gadji: Resolving this one — it isn't a defect in this PR.

`services/inference/healthcheck.py` is pre-existing dead code (still test-only at `b0616666`, I re-checked), so "revive it or delete it" is cleanup that belongs in its own issue, not in a readiness fix.

The half that is actionable here — separating "answered but unusable" from "never answered", and `models: list[str] | None` — is already covered by the exception taxonomy in https://github.com/linagora/openrag/pull/915#discussion_r3968836910 and the port-level `check_health` in https://github.com/linagora/openrag/pull/915#discussion_r3968837134. Those two are the shape I'd argue for over reviving the value object, so this comment only adds a third competing proposal. Closing it, leaving those open.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Minor enough to close — the whole consequence is a cosmetic toast, nothing is blocked.

Leaving the fix here for whoever next touches this expression: `develop` only assigned `model_found` when a name was actually supplied, so `None if normalized_model_name is None` restores that and stops `ui/src/pages/admin/models.tsx:697` rendering `Reachable — "" is served.` on a nameless probe. Worth folding in if you're already editing these lines for the `reachable` regression above; not worth a round-trip on its own.

andyne13:
One follow-on worth doing in this PR while `/ready` is being introduced, related to #937.

`.github/workflows/api_tests.yml:52` gates the suite on `/health_check`, which returns a hardcoded string and never touches the container (`api/routers/user/health.py:22-25`). So the gate passes on a degraded boot and the suite runs against a dead application — that is the run #937 cites, roughly a hundred `[UNEXPECTED_ERROR]` failures whose actual cause was a startup timeout nothing reported.

Repointing that line at `/ready` makes the gate mean what it is there to mean, and it is a one-line change that only becomes possible once this PR lands. Doing it here keeps the endpoint and the thing that depends on it in the same commit.

openrag/di/readiness.py
hedhoud: This is related to readiness design, but the port-wide refactor is not required to close #845. The runtime currently has one supported catalog adapter and one supported vector adapter, so the casts do not hide an active backend substitution path. Moving health probing into every storage and inference port would widen several interfaces and touch all providers. The endpoint-probing regressions identified in the other threads should be fixed in this PR; adapter-level probe unification should be handled separately.

hedhoud:
Regarding https://github.com/linagora/openrag/pull/915#issuecomment-5602462310: no change is required in #915. Global readiness is not intended to monitor every partition-specific model endpoint. Those checks are informational and deliberately non-gating, because one tenant-specific outage must not remove every API replica from service. Discovering preset references, deduplicating endpoints, and reporting stale references is useful catalog-aware monitoring, but it is separate from the Kubernetes readiness failure in #845.

hedhoud:
Regarding https://github.com/linagora/openrag/pull/915#issuecomment-5614789703: no change is required in #915. Repointing the API-test gate addresses the degraded-boot diagnosis tracked by #937, rather than the Kubernetes readiness behavior tracked by #845. It also depends on /ready being available first. Keeping that workflow change with #937 allows the readiness gate and the authenticated-request 503 behavior to be validated together.

openrag/services/orchestrators/model_endpoint_service.py
hedhoud: Fixed in d734a7a3. Provider metadata now reaches the shared probe for both saved and draft validation, and the dashboard includes its selected provider in the validation request. The regression tests cover an OpenAI-compatible reranker so it is no longer probed with the Infinity default.

openrag/services/orchestrators/model_endpoint_service.py
hedhoud: Fixed in d734a7a3. Receiving an HTTP response now keeps the endpoint reachable even when its model-list payload is unusable. Model presence remains unknown and the admin receives a stable validation message, restoring the pre-PR behavior instead of blocking the endpoint as unreachable.

openrag/services/orchestrators/readiness_service.py
hedhoud: Fixed in d734a7a3. An invalid or unparseable model list now has a dedicated outcome, distinct from a valid list that does not contain the configured model. The exceptions remain local because they are internal probe outcomes; moving the complete inference exception hierarchy is not required for this readiness fix.

ui/src/pages/admin/models.tsx
```diff
@@ -362,6 +362,7 @@ function EndpointDialog({
     modelName,
     apiKey,
     extraJson,
+    vendor,
```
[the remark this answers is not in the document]
hedhoud: Fixed in da7e2bdc. Changing the provider now invalidates a completed validation, and edit mode restores the validated state only when the selected provider matches the stored implementation. Regression coverage checks both new and edited endpoints.

openrag/di/readiness.py
hedhoud: Completed in da7e2bdc. Timed-out probes now share one in-flight actor lookup on a dedicated executor. This bounds stalled lookups and keeps the shared executor available for unrelated work; the regression test exercises repeated timeouts before releasing the lookup.

Ahmath-Gadji (approved):
Approving.

I re-checked every thread against the code at `da7e2bdc` rather than trusting its resolved state. All of mine are either fixed or ones I withdrew myself, and the blocking one — the startup probe killing the API mid-model-load — is fixed the way we agreed: startup and liveness on `/health_check`, readiness alone on `/ready` (`infra/charts/openrag-stack/values.yaml:528-547`). The `extra`-dropping bug is fixed at both routes, with router-level tests that the earlier test shape couldn't reach; the `reachable: false` regression is fixed by deriving reachability from transport success; and `DEFAULT_MODEL_IMPLEMENTATIONS` is now a single map shared by the DI wiring and readiness, so the reranker default isn't encoded twice on the Python side. All seven checks are green, and I re-ran the seven affected unit files at that SHA locally — 267 passed.

Four residuals, none blocking the merge.

**Tracked in #945** — the two worth real follow-up, filed there rather than left to disappear with these threads:

1. **The model checks probe `"default"`, not the endpoint a partition actually uses.** `di/readiness.py:71` reads `settings.models.<kind>["default"]`, while the LLM comes from `_agreed_partition_chat_llm` → `_llm_factory(chat_llm)` (`query_service.py:241-244`), the embedder from `embedder_factory(embedder_name)` (`di/container.py:537`), and the reranker from `_reranker_factory(reranker_name)` (`retrieval_service.py:206`). `vlm` and `stt` aren't probed at all. Both error directions are live — `"llm": "ok"` while the endpoint serving a partition is down, and `"llm": "unavailable"` while every partition's preset is healthy. It can't pull a pod out of service, since only `postgres`/`milvus`/`ray` gate; the cost is that `/ready` misinforms about exactly the endpoints a multi-tenant deployment cares about most. `_circuit_breaker.py` already tracks this correctly, per endpoint, from real traffic.

2. **The model checks share the gating checks' 2 s TTL.** `cache_ttl=2.0` sits below the chart's `periodSeconds: 10`, so every probe is a guaranteed cache miss: ~26k outbound `/models` calls per pod per day for fields `ready()` reads and discards. The 2 s TTL is right for Postgres/Milvus/Ray; the model checks don't gate anything and don't need per-probe freshness.

The issue lays out the options and leans drop — since the per-partition gap makes the value misleading rather than merely incomplete, and `_circuit_breaker.py` already covers it. That was also the thread resolved here without a reply, so the decision is genuinely still open.

**3. The chart deviates from #845's text on purpose.** CodeRabbit's Linked Issues check still flags it — it wants startup on `/ready` and the model endpoints gating readiness, both of which we deliberately rejected in this review. The final behavior is documented in `docs/content/docs/documentation/kubernetes.md`, so this just needs #845 amended or a note at merge, so nobody "fixes" it back later.

**4. Two cosmetic leftovers**, for whoever next edits these lines: `model_found` is still `True` on a nameless probe (`model_endpoint_service.py:760-766`), so the admin UI renders `Reachable — "" is served.`; and `hasattr(response, "raise_for_status")` at `readiness_service.py:91` is a test-mock accommodation living in production code.

Separately, the three things we agreed to park still have no follow-up issues — the port-level `check_health`, chart rendering never being validated in CI, and `services/inference/healthcheck.py` being dead code. Happy to file those too if you'd like them tracked.

PR #916, opened 2026-09-08
Feat/loki logs

paultranvan:
## Summary

Ships OpenRAG's application logs (API process and Ray indexing workers) to an external Loki, from a VM (Compose) or Kubernetes (Helm), with the container's stdout/stderr as the single log output.

- `LOG_FORMAT=json` (default `text`, unchanged): one flat JSON object per line, no ANSI — `ts`, `level`, `logger`, `function`, `line`, `msg`, optional `exception`, then every bound extra (`request_id`, `task_id`, …). Stdlib loggers (uvicorn, ray, …) are routed through the same sink; `asyncio`, `httpcore`, `httpx`, `urllib3` are capped at WARNING.
- `request_id` is bound into every log line of a request (inbound `X-Request-ID` validated, otherwise minted).
- Opt-in Compose overlay `infra/compose/logging.docker-compose.yaml`: Grafana Alloy on the Docker socket strips Ray's colorized `(Actor pid=N) ` relay prefix, parses the JSON, promotes `level` to a label and `request_id`/`task_id`/`ray_actor` to structured metadata, and pushes to `LOKI_URL` (optional basic auth and tenant).
- Ray's relay of worker output is made JSON-safe wherever `LOG_FORMAT=json` is set (`RAY_DEDUP_LOGS=0`, `RAY_COLOR_PREFIX=0`); with `ray.enabled=true` the chart sets `RAY_LOG_TO_STDERR=1` so workers write to their own pod's stderr (the driver relay only carries the current job's actors and goes silent for the detached workers after an API restart).
- Helm defaults `LOG_FORMAT=json` and relies on the platform DaemonSet; `docs/documentation/loki_logs.md` gives the Alloy stages, LogQL starters, requirements and limitations.

## Breaking changes and prerequisites

- `GET /indexer/task/{task_id}/logs` and the MCP tool `get_task_logs` are removed. Use `GET /indexer/task/{task_id}` (and `.../error`) for the final state, and `{app="openrag"} | json | task_id="<id>"` in Loki for the detail.
- `logs/app.json` is no longer written. `LOG_DIR`, `LOG_VOLUME` and the Helm `logs` PVC are removed; leftovers in a `.env`/values file are ignored.
- **Kubernetes prerequisite.** The chart used to mount one `ReadWriteMany` logs volume into the API and every Ray node so the task-log route could read a shared `app.json`; dropping the file sink removes that shared-volume dependency. In exchange, a log collector DaemonSet (Alloy, Fluent Bit, …) shipping pod stdout to Loki is now required: without one, logs exist only in the container runtime (`kubectl logs`) and are lost with the pod. Stated in `NOTES.txt`, in the `LOG_FORMAT` values comment and in the guide. A kept `<release>-logs` PVC (`helm.sh/resource-policy: keep`) survives the upgrade unused and can be deleted by hand.
- **Ordering with #904.** After this PR, `GET /indexer/task/{task_id}/error` is the only in-product answer to "why did this job fail", and it is not durable: `FAILED` is not in `RECOVERABLE_TASK_STATES`, so the traceback is gone when the `TaskStateManager` actor restarts. #904 mirrors job state to Postgres; until it lands, Loki is the only durable record of a failure.
- Structured metadata needs Loki ≥ 3.0 with `allow_structured_metadata: true` (fallback documented).

## Verification

- Unit suite: 2933 passed; the 10 remaining failures are the known local-`.env` baseline (`tests/unit/core/config/*_env.py`, `test_token_validation.py`) and fail identically on `develop`.
- `ruff check`, `ruff format --check`, `scripts/check_layer_imports.py`: clean. `helm lint` and `helm template` clean for `ray.enabled` true and false.
- Alloy pipeline re-verified against Loki 3.7.7 with synthetic API, colorized-prefix, plain-prefix and banner lines after the review fixes: API entries carry no `ray_actor`, worker entries carry `ray_actor` + `task_id`, labels stay `app/service/container/level`.
- Earlier end to end against a local Loki 3.7.7 with the real API: labels exactly `app/service/container/level`; `request_id` and `ray_actor` stored as structured metadata; the Ray line stored starting at `{` (ANSI prefix stripped); Loki timestamp equals the app `ts`; basic-auth path verified through an nginx `htpasswd` proxy (204 as the configured user), and an empty `basic_auth` block sends no `Authorization` header (401 against the proxy with empty credentials).

## Known limitations (documented)

- With `ENABLE_RAY_SERVE=true` the `ray.serve` replica logger is reconfigured after the JSON interception, so those lines (and, with Chainlit, its second uvicorn's) come out as text; same for the `python -m api.main` dev path. The container entrypoint's single-uvicorn path is not affected.
- uvicorn's own access line carries no `request_id` (emitted after the middleware returns); the unhandled-500 line binds it explicitly.
- In the relay path, a worker line above ~8 KiB can interleave with an API line (no shared lock on stderr); `RAY_LOG_TO_STDERR=1` avoids it.
- Ray's internal logs and the other services (Milvus, Postgres, vLLM, reranker, nginx) are not in the shipped pipeline.
- Two OpenRAG stacks on one host share `app`/`service` labels in the compose overlay; only `container` tells them apart.

## Summary by CodeRabbit

* **New Features**
  * Added text and JSON logging modes with request and task correlation metadata.
  * Added optional Docker Compose and Kubernetes log shipping to Grafana Loki.
  * Added validation for incoming request IDs, generating safe IDs when needed.
* **Changes**
  * Application logs now write to stderr instead of a dedicated log file or persistent logs volume.
  * Removed task-log retrieval through the API, MCP, and UI.
* **Documentation**
  * Added guidance for logging configuration, Loki integration, and deployment behavior.


andyne13:
Architecture looks right to me — stderr as the only sink, and keeping `/error` while dropping `/logs`, is the split I'd want. Worth noting in the description that the chart mounted one `ReadWriteMany` logs volume into the API and every Ray node so `log_tail` could read a shared `app.json`; dropping the file sink removes that shared-volume dependency too.

Two things before it leaves draft.

**The Kubernetes path removes the old mechanism and depends on one the chart doesn't ship.** The logs PersistentVolumeClaim and every `/app/logs` mount go, and `LOG_FORMAT=json` assumes a collector DaemonSet is already shipping pod logs. On a cluster without one, logs are only in the container runtime's output and go with the pod. Since this is marked breaking and deletes a PersistentVolumeClaim and its mounts across three templates, could the prerequisite land somewhere an operator will hit it — `NOTES.txt`, the values comment, and the PR description — rather than only in the Loki guide? If it helps, the format, `request_id` and the Alloy overlay stand on their own and could go in ahead of the volume and endpoint removals.

**After this, `/error` is the only in-product answer to "why did this fail".** Today it isn't durable — `FAILED` isn't in `RECOVERABLE_TASK_STATES`, so the transition deletes the Ray key-value entry and the traceback is gone when the actor restarts. #904 mirrors it to Postgres, so ordering the two is worth a line in the description: until that lands, this PR leaves no durable record of why a job failed.

paultranvan:
> Architecture looks right to me — stderr as the only sink, and keeping `/error` while dropping `/logs`, is the split I'd want. Worth noting in the description that the chart mounted one `ReadWriteMany` logs volume into the API and every Ray node so `log_tail` could read a shared `app.json`; dropping the file sink removes that shared-volume dependency too.
> 
> Two things before it leaves draft.
> 
> **The Kubernetes path removes the old mechanism and depends on one the chart doesn't ship.** The logs PersistentVolumeClaim and every `/app/logs` mount go, and `LOG_FORMAT=json` assumes a collector DaemonSet is already shipping pod logs. On a cluster without one, logs are only in the container runtime's output and go with the pod. Since this is marked breaking and deletes a PersistentVolumeClaim and its mounts across three templates, could the prerequisite land somewhere an operator will hit it — `NOTES.txt`, the values comment, and the PR description — rather than only in the Loki guide? If it helps, the format, `request_id` and the Alloy overlay stand on their own and could go in ahead of the volume and endpoint removals.
> 
> **After this, `/error` is the only in-product answer to "why did this fail".** Today it isn't durable — `FAILED` isn't in `RECOVERABLE_TASK_STATES`, so the transition deletes the Ray key-value entry and the traceback is gone when the actor restarts. #904 mirrors it to Postgres, so ordering the two is worth a line in the description: until that lands, this PR leaves no durable record of why a job failed.

Fair points, this is integrated , and PR is now ready to review

EnjoyBacon7 (approved):

PR #921, opened 2026-09-08
fix(workspaces): preserve independently indexed files on deletion

### LINKED ISSUE #275: Data loss: workspace deletion removes independently-indexed files
## Summary

Deleting a workspace can permanently delete files that were independently indexed into the partition (not added via workspace), causing irreversible data loss.

## Root cause

`delete_workspace()` in `openrag/components/indexer/vectordb/utils.py:636` determines which files are \"orphaned\" (and therefore should be deleted from Milvus) by checking whether a file appears in **any other workspace**. It does not check whether the file was independently indexed into the partition via the regular `files` table.

```python
# utils.py:640 — only checks other workspaces, ignores files table
subq = select(WorkspaceFile.file_id).where(WorkspaceFile.workspace_id != workspace_id).subquery()
result = session.execute(
    select(WorkspaceFile.file_id)
    .where(WorkspaceFile.workspace_id == workspace_id)
    .where(WorkspaceFile.file_id.notin_(select(subq.c.file_id)))
)
orphaned_file_ids = [r[0] for r in result.all()]
```

## Steps to reproduce

1. Upload a file to a partition normally (without workspace) — file is in the `files` table.
2. Create a workspace and add that file to it via `POST /partition/{p}/workspaces/{ws}/files`.
3. Delete the workspace via `DELETE /partition/{p}/workspaces/{ws}`.
4. The file is now gone from Milvus and the partition, even though it was independently indexed.

## Expected behaviour

Only files that **exist exclusively in workspace storage** (i.e. are not present in the `files` table at all) should be deleted when a workspace is removed. Files that were independently indexed must be preserved.

## Fix

The orphan query should additionally exclude any `file_id` that exists in the `files` table:

```python
from openrag.components.indexer.vectordb.models import File  # adjust import as needed

subq_other_ws = select(WorkspaceFile.file_id).where(WorkspaceFile.workspace_id != workspace_id).subquery()
subq_files    = select(File.file_id).subquery()
result = session.execute(
    select(WorkspaceFile.file_id)
    .where(WorkspaceFile.workspace_id == workspace_id)
    .where(WorkspaceFile.file_id.notin_(select(subq_other_ws.c.file_id)))
    .where(WorkspaceFile.file_id.notin_(select(subq_files.c.file_id)))
)
```

## Affected file

`openrag/components/indexer/vectordb/utils.py`, line 640

### LINKED ISSUE #490: Workspace-delete test masks #275: it exercises a stale `DeleteWorkspaceHelper`, not production `delete_workspace`
## Summary

`openrag/components/indexer/vectordb/test_delete_workspace.py` does not test the production deletion path. It defines its own `DeleteWorkspaceHelper` — a hand-copied "Mirror of `PartitionFileManager.delete_workspace`" (its own docstring, line 52/73) — and asserts against **that**, while the real code has since diverged. The result: the test stays green while production silently regressed into the data loss tracked in #275.

## Details

- The test helper still contains the **correct** orphan-detection logic — it excludes files that were *independently indexed* into the partition via `subq_indexed = select(FileModel.file_id).where(FileModel.partition_name == partition)` (`test_delete_workspace.py:87-92`).
- Production `delete_workspace` (`openrag/components/indexer/vectordb/utils.py:736`) only excludes files that belong to *another* workspace (`subq_other_ws`); it no longer excludes independently-indexed files. That exclusion was **removed** in `1b154df4` ("fix(workspaces): use files.id (int) FK in workspace_files…") — see the deleted `subq_indexed` block in that commit's diff.
- The unit test never imports or calls `PartitionFileManager.delete_workspace` (nor the `vectordb.py:1488` actor wrapper), so the regression shipped undetected.

## Impact

A file that was independently indexed into a partition **and** referenced by a single workspace is reported as an orphan and destructively deleted on workspace deletion (the #275 data loss). The existing test provides false confidence that this path is safe.

## Suggested fix

1. Restore the independently-indexed exclusion in the production `delete_workspace` (resolves #275).
2. Replace `DeleteWorkspaceHelper` with a test that drives the real `PartitionFileManager.delete_workspace` against an in-memory session, so production and test cannot diverge again.

_Found during a triage of open issues against `main` (217a883b)._

hedhoud:
Deleting a workspace could permanently remove files that had been indexed independently and later attached to it. The existing regression test exercised copied deletion logic, so it stayed green while production lost data.

Preserve independently indexed files and existing files whose origin is unknown. New workspace uploads are cleaned up only after their last workspace is deleted; retained files stay protected on later workspace deletions. Regression tests now exercise the production deletion path against PostgreSQL and run in a dedicated CI job.

Validation: 313 focused tests passed, plus Ruff and the layer import guard. A broader run found three unrelated failures, all reproduced on unchanged develop. Full API tests were not run against a deployed server.

Fixes #275
Fixes #490

## Summary by CodeRabbit

* **New Features**
  * Added ownership tracking for workspace-indexed files.
  * Workspace deletion now preserves independently indexed and shared files.
  * Added the `keep_files` option to retain eligible workspace files after deletion.
  * Failed file cleanup can now be retried safely.

* **Bug Fixes**
  * Improved cleanup reliability and recovery for interrupted or concurrent deletions.
  * Prevented files undergoing destructive cleanup from being reattached.
  * Uploads now reject workspace ID values that are not strings.
  * Files remain independently indexed when some workspace attachments fail.

* **Documentation**
  * Clarified workspace file retention, deletion behavior, and ownership rules.


openrag/services/persistence/workspace_repo.py
```diff
@@ -136,7 +171,9 @@ async def add_files_to_workspace(
                 resolved = await conn.fetch(
                     """
                     SELECT file_id, id FROM files
-                    WHERE file_id = ANY($1::text[]) AND partition_name = $2
+                    WHERE file_id = ANY($1::text[])
+                      AND partition_name = $2
+                      AND NOT workspace_cleanup_claimed
```
EnjoyBacon7: **Blocking: this does not serialize against the claim.**

Under READ COMMITTED this SELECT takes no row lock, so it reads the pre-claim row version and does not block. The INSERT below *does* block, on the FK KEY SHARE lock that conflicts with the claim's `FOR UPDATE`, but once the claim transaction commits the insert simply proceeds. The `NOT workspace_cleanup_claimed` predicate is never re-evaluated.

I reproduced the data loss against Postgres with production code on both sides, parking `delete_workspace` mid-transaction with `LOCK TABLE workspaces IN SHARE MODE`:

```
add_files_to_workspace('ws2') missing=[]          # caller told the attach succeeded
delete result={'orphaned_files_deleted': 1, ...}
ws2 file list after=[]                            # file is gone
file still in partition=False
```

Adding `FOR UPDATE` fixes it. The attach then blocks, re-reads the updated row after the claim commits, filters it out, and correctly reports the file as not found. Verified: repro flips to `missing=['exclusive']` and all 38 integration tests stay green.

```suggestion
                      AND NOT workspace_cleanup_claimed
                    FOR UPDATE
```

openrag/services/persistence/workspace_repo.py
```diff
@@ -92,19 +92,54 @@ async def delete_workspace(self, workspace_id: str) -> list[str]:
...
+                          )
+                        """,
+                        workspace_id,
+                    )
+                    await conn.execute(
+                        """
+                        UPDATE files SET independently_indexed = TRUE
+                        WHERE id IN (
+                            SELECT file_id FROM workspace_files WHERE workspace_id = $1
+                        )
```
EnjoyBacon7: This flips ownership for files the workspace does not exclusively own. The `SELECT` above correctly excludes files that are still in another workspace, but this UPDATE hits every row in `workspace_files` for the workspace.

Verified: file in both ws1 and ws2, delete ws1 with `keep_files=true`. It reports `kept_files=0`, which is right, but still marks the file independent, so deleting ws2 later cleans up nothing. A `keep_files` call on one workspace permanently disables cleanup for another workspace's file, and the reported count says otherwise.

```suggestion
                        UPDATE files SET independently_indexed = TRUE
                        WHERE id IN (
                            SELECT wf.file_id FROM workspace_files wf
                            WHERE wf.workspace_id = $1
                              AND wf.file_id NOT IN (
                                  SELECT file_id FROM workspace_files WHERE workspace_id <> $1
                              )
                        )
```

openrag/services/persistence/workspace_repo.py
```diff
@@ -92,19 +92,54 @@ async def delete_workspace(self, workspace_id: str) -> list[str]:
...
+                              AND NOT f.independently_indexed
+                              AND NOT f.workspace_cleanup_claimed
+                              AND wf.file_id NOT IN (
+                                  SELECT file_id FROM workspace_files
+                                  WHERE workspace_id <> $1
+                              )
+                            FOR UPDATE OF f
+                        )
+                        UPDATE files f
+                        SET workspace_cleanup_claimed = TRUE
```
EnjoyBacon7: The claim commits here, but only `finalize_claimed_file_cleanup` and `release_claimed_file_cleanup` clear it, and both run in-process right after. A crash or pod eviction between the two leaves `workspace_cleanup_claimed = TRUE` with no way back.

Verified end state: the file stays in the catalog, `add_files_to_workspace` reports it as not found to the user forever, and no later `delete_workspace` can reclaim it. Nothing else in the codebase resets the column, so there is no sweeper and no startup pass.

`file_content_claims` already solves this shape with `expires_at` plus a `claim_token`. Worth reusing that here, or at least adding a claimed_at timestamp so a stale claim can be reclaimed.

EnjoyBacon7 (commented):
Reviewed the branch against a real PostgreSQL instance. The core fix is right and the PR body is accurate that the old SQLite based unit test could never have caught the bug, so replacing it with real database tests is the right call.

One blocking issue: the second commit does not achieve what it says. The claim protocol still lets a concurrent attach lose its file, and I reproduced it with production code on both sides. Details inline, with a one line fix that I verified closes it without breaking the existing tests.

Also flagged: `keep_files` marking files it does not own, no recovery path for a stuck claim, and a release that runs after the vectors are already deleted. The rest is minor.

Verified clean on the branch: ruff check and format, the layer import guard, 2930 unit tests, and 38 integration repo tests against PostgreSQL. Migration chain is single headed and reversible. Caveat: my local server was PG14, CI uses PG16. The locking semantics at issue are the same on both.

openrag/services/orchestrators/workspace_service.py
```diff
@@ -181,17 +182,22 @@ async def _delete_file(self, file_id: str, partition: str) -> None:
...
+            ids = await self._vector_store.query_ids_by_filter(
+                self._collection,
+                {"partition": partition, "file_id": file_id},
+            )
+            if ids:
+                await self._vector_store.delete(ids, self._collection)
+            if not await self._workspace_repo.finalize_claimed_file_cleanup(file_id, partition):
+                raise RuntimeError(f"Workspace cleanup claim disappeared for {file_id}")
+        except Exception:
+            await self._workspace_repo.release_claimed_file_cleanup(file_id, partition)
```
EnjoyBacon7: The `try` covers the vector delete as well as the finalize, so a DB error in `finalize_claimed_file_cleanup` releases the claim after the chunks are already gone. That makes the file attachable again with no content behind it, which is worse than the old behaviour of just failing and leaving the row alone.

Suggest narrowing the release to failures that happen before the vector delete, and leaving the claim set once the vectors are destroyed.

tests/integration/repos/test_workspace_deletion.py
```diff
@@ -0,0 +1,192 @@
...
+    vectors.query_ids_by_filter.side_effect = query_ids
+    svc = WorkspaceService(
+        workspace_repo=store.workspace_repo,
+        document_repo=store.document_repo,
+        vector_store=vectors,
+        collection="test",
+    )
+
+    deletion = asyncio.create_task(svc.delete_workspace("p", "ws1"))
+    await cleanup_started.wait()
```
EnjoyBacon7: This only exercises the safe ordering. `cleanup_started` fires inside `query_ids_by_filter`, which runs after `delete_workspace` has already committed the claim, so the attach on line 145 sees the committed flag and correctly reports the file as missing.

The interleaving that loses data is the attach landing while the claim transaction is still open. That case is untested and currently fails. One way to park the delete deterministically inside its transaction is to hold `LOCK TABLE workspaces IN SHARE MODE` from a second connection. It conflicts with the `DELETE FROM workspaces` at the end of `delete_workspace` but not with the attach path's plain SELECT, so the claim is written and left uncommitted.

tests/integration/repos/test_workspace_deletion.py
```diff
@@ -0,0 +1,192 @@
...
+@pytest.mark.parametrize("shared", [False, True])
+async def test_keep_files_preserves_upload_after_later_workspace_deletion(postgres_store, shared):
+    store = postgres_store
+    await setup_workspace(store)
+    await store.workspace_repo.create_workspace(Workspace(workspace_id="ws2", partition="p"))
+    await upload(store, "kept", workspace_ids=["ws1", "ws2"] if shared else ["ws1"])
+    svc, vectors = service(store)
+
+    result = await svc.delete_workspace("p", "ws1", keep_files=True)
+    assert result["kept_files"] == (0 if shared else 1)
```
EnjoyBacon7, line 81: The `shared=True` case asserts the over-broad update from `workspace_repo.py` as intended behaviour. Here `kept_files` is 0, so the API told the caller nothing was retained, yet the file is silently marked independent and ws2 can never clean it up.

Once the UPDATE is scoped to exclusive files, this parametrisation should assert the opposite for `shared=True`: the file stays eligible and is deleted when ws2 goes.

tests/integration/api/test_workspaces.py
```diff
@@ -190,7 +199,7 @@ def test_delete_workspace_keep_files(self, api_client, workspace_partition, work
         body = response.json()
         assert body["status"] == "deleted"
         assert body["orphaned_files_deleted"] == 0
-        assert body["kept_files"] == 1
+        assert body["kept_files"] == 0
```
EnjoyBacon7: With this change the `keep_files=true` test now asserts `0` and `0`, exactly the same numbers as the default-delete test above it. Nothing at the API level distinguishes `keep_files` any more.

To keep the coverage, upload the file with `workspace_ids` set so it is genuinely workspace owned, then `kept_files` is 1 again and the flag means something.

tests/integration/repos/test_workspace_repo.py
```diff
@@ -161,8 +160,7 @@ async def test_files_shared_with_other_workspaces_are_not_orphaned(
         await repo.add_files_to_workspace("ws1", ["f1", "f2"])
         await repo.add_files_to_workspace("ws2", ["f1"])  # f1 shared
         orphans = await repo.delete_workspace("ws1")
-        # f1 is still in ws2 so it's not orphaned. f2 only lived in ws1.
-        assert set(orphans) == {"f2"}
+        assert orphans == []
```
EnjoyBacon7, line 171: This now passes because the seeded files are independently indexed, not because f1 is shared. The test would pass identically with the `add_files_to_workspace("ws2", ...)` line removed, so the name no longer describes what it checks.

Same for `test_independently_indexed_files_are_preserved` above: both assert `[]` for the same reason. The real shared-file coverage moved to `test_workspace_deletion.py`, so these two could either seed workspace owned files or be dropped.

openrag/services/persistence/migrations/alembic/versions/c0d1e2f3a4b5_add_file_indexing_ownership.py
```diff
@@ -0,0 +1,27 @@
+"""Preserve files whose workspace ownership is unknown.
+
+Revision ID: c0d1e2f3a4b5
+Revises: b9c0d1e2f3a4
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from services.persistence.migrations.alembic.schema_helpers import column_exists, table_exists
```
EnjoyBacon7, line 9: Both new migrations use the package qualified path while all 18 existing ones use `from schema_helpers import ...`, which works because `env.py` puts the alembic directory on `sys.path`. Both resolve at runtime, so this is cosmetic, but it is the only pair that differs.

Same file: the other migrations annotate `revision: str = ...` and `down_revision: str | Sequence[str] | None = ...`. Worth matching for consistency.

.github/workflows/integration_tests.yml
```diff
@@ -7,14 +7,55 @@ on:
...
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+      - run: uv sync
```
EnjoyBacon7, line 41: The `milvus-integration` job below caches uv with `actions/cache` keyed on `pyproject.toml` and `uv.lock`. This job does a cold `uv sync` on every run. Copying the cache step over would cut most of the runtime.

openrag/services/persistence/workspace_repo.py
hedhoud: Confirmed actionable. The attachment query does not lock the candidate file row, so it can read the row before the cleanup claim commits. The attachment can then succeed while cleanup continues. This is in scope for PR #921 because it can still delete a file newly attached to another workspace. A code change is still required.

openrag/services/persistence/workspace_repo.py
hedhoud: Confirmed actionable. keep_files must preserve only files exclusively owned by the deleted workspace; marking shared files independent changes their later cleanup behavior. This is in scope because it changes the retention semantics introduced by this PR.

openrag/services/persistence/workspace_repo.py
hedhoud: Confirmed actionable. A boolean cleanup claim without expiry or recovery can leave a file permanently unavailable after a worker crash. This affects the cleanup mechanism introduced here and needs a recovery path before merge.

openrag/services/orchestrators/workspace_service.py
hedhoud: Confirmed actionable. Releasing the catalog claim after vector deletion can leave an attachable catalog row with no vectors if finalization fails. Error handling must distinguish failures before and after destructive vector deletion.

tests/integration/repos/test_workspace_deletion.py
hedhoud: Confirmed actionable. The existing test covers only the post-claim ordering; it does not exercise an attachment while the claim transaction is still open. A deterministic concurrency test is needed.

tests/integration/repos/test_workspace_deletion.py
hedhoud: Confirmed actionable. The shared=True case should verify that a shared file remains eligible and is deleted only after the last workspace is removed. The current assertion does not cover that behavior.

tests/integration/api/test_workspaces.py
hedhoud: Confirmed actionable as test coverage. The current fixture does not create a genuinely workspace-owned file, so the keep_files behavior is not being exercised. The test data needs to establish workspace ownership before deletion.

tests/integration/repos/test_workspace_repo.py
hedhoud: Confirmed. The current setup allows this test to pass even without the second workspace attachment, so it does not validate shared-file behavior. The test needs a genuinely shared scenario or should be replaced.

openrag/services/persistence/migrations/alembic/versions/c0d1e2f3a4b5_add_file_indexing_ownership.py
hedhoud: Reviewed. No functional change is required here: the package-qualified import is intentional because the migration is imported directly by the PostgreSQL integration test, while Alembic also supports it. Matching annotation style with older migrations is cosmetic and outside this PR's behavioral scope.

.github/workflows/integration_tests.yml
hedhoud: Reviewed. This is a valid CI optimization, but it is independent of workspace file-retention correctness and does not affect the workflow's behavior. I am leaving it out of PR #921 so the performance improvement can be handled separately.

hedhoud:
The docstring-coverage warning is noted, but it is not a functional defect in the workspace-retention change. Adding broad docstrings solely to satisfy the review threshold would expand this PR beyond its behavioral scope, so it can be handled separately.

.github/workflows/integration_tests.yml
```diff
@@ -8,6 +8,40 @@ on:
   pull_request:
 
 jobs:
+  workspace-integration:
```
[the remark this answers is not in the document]
hedhoud: Fixed in 63706b15. keep_files now marks only files with no membership in another workspace; the shared-file integration test now verifies that the file is deleted when the last workspace is removed.

openrag/services/persistence/workspace_repo.py
hedhoud: Fixed in 63706b15. add_files_to_workspace now locks candidate file rows with SELECT FOR UPDATE, so an attachment waits for the cleanup transaction and is re-evaluated after the claim commits. The integration test now holds the claim transaction open to verify this exact race.

openrag/services/persistence/workspace_repo.py
hedhoud: Fixed in 63706b15. keep_files now marks a file independent only when the deleted workspace is its last workspace membership. The shared-file integration test now verifies it is deleted when the remaining workspace is removed.

openrag/services/persistence/workspace_repo.py
hedhoud: Fixed in 63706b15. Workspace cleanup claims now have a timestamp via a new migration; claims older than one hour are recoverable, and attachment clears a stale claim while holding the row lock. The integration suite covers reattaching a stale-claimed file.

openrag/services/orchestrators/workspace_service.py
hedhoud: Fixed in 63706b15. The claim is released only when cleanup fails before vector deletion starts. If vector deletion has started, the claim remains in place so a catalog failure cannot make a vector-less file attachable again; the unit test covers this path.

tests/integration/repos/test_workspace_deletion.py
hedhoud: Fixed in 63706b15. The race regression test now uses two database connections, keeps the cleanup claim transaction open, asserts that attachment blocks, and verifies the attachment is rejected after the cleanup transaction commits.

tests/integration/api/test_workspaces.py
hedhoud: Fixed in 63706b15. The API keep_files test now uploads with workspace_ids, so it creates a genuinely workspace-owned file and expects kept_files=1. The existing independent-file test remains separate.

tests/integration/repos/test_workspace_repo.py
hedhoud: Fixed in 63706b15. The repository fixture now distinguishes independent and workspace-owned files, and the shared-file case attaches the owned file to both workspaces before deleting each one.

EnjoyBacon7:
### Code review

Found 2 issues:

1. The attach-first / cleanup-second interleaving still loses a file. `FOR UPDATE OF f` locks the `files` row, but the orphan test is a `NOT IN` subquery over `workspace_files`, a different and unlocked table. If `add_files_to_workspace` locks the row, inserts its `workspace_files` row and commits while `delete_workspace`'s statement is already waiting on that lock, the waiter never requalifies: a plain `SELECT ... FOR UPDATE` that commits without modifying the tuple creates no new row version, so EvalPlanQual does not fire, and the `NOT IN` subquery is still evaluated against the pre-wait statement snapshot. The file is claimed as an orphan and deleted, and because `workspace_files.file_id` is `ON DELETE CASCADE` the new membership disappears with no error. `test_workspace_attachment_cannot_race_orphan_cleanup` only covers the opposite direction: it sets the claim first, then starts the attach.

https://github.com/linagora/openrag/blob/f3ff845c63fe849dc678ef7b2859110abcd6b16f/openrag/services/persistence/workspace_repo.py#L141-L154

2. A stale claim is reclaimed on wall-clock age alone, with no record of whether vector deletion already ran. Once `vector_cleanup_started` is true, `_delete_file` deliberately skips `release_claimed_file_cleanup`, so a transient `finalize_claimed_file_cleanup` failure leaves the row claimed with its original timestamp. There is no retry or reconciliation path (`finalize_claimed_file_cleanup` has a single production call site). An hour later `add_files_to_workspace` clears the claim and reattaches the file, whose Milvus chunks are already gone, so it is searchable in the catalog and returns nothing. This is the same shape flagged on `file_content_claims` in #699: renew or verify the work state, do not expire on elapsed time alone.

https://github.com/linagora/openrag/blob/f3ff845c63fe849dc678ef7b2859110abcd6b16f/openrag/services/orchestrators/workspace_service.py#L194-L204

https://github.com/linagora/openrag/blob/f3ff845c63fe849dc678ef7b2859110abcd6b16f/openrag/services/persistence/workspace_repo.py#L186-L219

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

openrag/services/persistence/workspace_repo.py
```diff
@@ -136,14 +185,38 @@ async def add_files_to_workspace(
                 resolved = await conn.fetch(
                     """
                     SELECT file_id, id FROM files
-                    WHERE file_id = ANY($1::text[]) AND partition_name = $2
+                    WHERE file_id = ANY($1::text[])
+                      AND partition_name = $2
+                      AND (
+                          NOT workspace_cleanup_claimed
+                          OR workspace_cleanup_claimed_at IS NULL
+                          OR workspace_cleanup_claimed_at < NOW() - INTERVAL '1 hour'
```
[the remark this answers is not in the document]
hedhoud: Fixed in c7b3b0fa. Cleanup claims now enter an explicit destructive phase before Milvus deletion. Time-based recovery applies only before that phase, so a file cannot be reattached once vector deletion may have started.

openrag/services/persistence/workspace_repo.py
```diff
@@ -92,19 +92,68 @@ async def delete_workspace(self, workspace_id: str) -> list[str]:
...
+                              AND NOT f.independently_indexed
+                              AND (
+                                  NOT f.workspace_cleanup_claimed
+                                  OR f.workspace_cleanup_claimed_at IS NULL
+                                  OR f.workspace_cleanup_claimed_at < NOW() - INTERVAL '1 hour'
+                              )
+                              AND wf.file_id NOT IN (
+                                  SELECT file_id FROM workspace_files
+                                  WHERE workspace_id <> $1
+                              )
```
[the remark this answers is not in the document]
hedhoud: Fixed in c7b3b0fa. Workspace deletion now locks every attached file in a stable order before running orphan evaluation in a fresh statement. This serializes concurrent deletions and observes memberships committed while waiting.

tests/integration/repos/test_workspace_deletion.py
```diff
@@ -0,0 +1,226 @@
...
+            """,
+        )
+        await conn.execute(
+            """
+            UPDATE files
+            SET workspace_cleanup_claimed = TRUE,
+                workspace_cleanup_claimed_at = NOW()
+            WHERE file_id = 'exclusive'
+            """,
+        )
```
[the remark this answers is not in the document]
hedhoud: Fixed in c7b3b0fa. The integration test now invokes the production delete_workspace method while PostgreSQL holds a conflicting lock. It covers cleanup-first, attach-first, and concurrent last-workspace deletion orderings.

hedhoud:
Addressed both issues from the latest review in c7b3b0fa. Orphan evaluation now runs only after an ordered pre-lock, giving it a fresh snapshot after a concurrent attachment commits. Cleanup claims also record when destructive vector deletion begins and cannot be recovered by age after that point. The regression suite covers both interleavings and stale claims before and after the destructive boundary.

hedhoud:
Addressed the outside-diff CodeRabbit comment in c6cb6865. The fake now records both the workspace ID and keep_files value, and the unit tests assert True for the retention path and False for the default path.

EnjoyBacon7 (commented):
Inline notes for the two findings from the review above.

openrag/services/persistence/workspace_repo.py
```diff
@@ -92,19 +92,68 @@ async def delete_workspace(self, workspace_id: str) -> list[str]:
...
+                            FOR UPDATE OF f
```
EnjoyBacon7, line 206: `FOR UPDATE OF f` locks the `files` row, but the orphan test on the lines just above is a `NOT IN` subquery over `workspace_files`, which is never locked. That leaves the attach-first ordering unguarded:

1. `add_files_to_workspace(ws2, [F])` takes its own `SELECT ... FOR UPDATE` on F's `files` row, inserts the `workspace_files` row for ws2, and commits. F carried no claim, so the conditional claim-clearing UPDATE matches zero rows and the tuple is never actually modified, only locked.
2. `delete_workspace(ws1)` has already taken its statement snapshot and is waiting on that row lock. When the lock is released it simply acquires it: a `SELECT ... FOR UPDATE` that commits without modifying the tuple creates no new row version, so EvalPlanQual never fires. And even when EPQ does fire it only substitutes the locked row itself, never re-evaluating subqueries against other tables.
3. The `NOT IN` subquery therefore still reads the pre-wait snapshot, does not see ws2's membership, and F is claimed as an orphan and deleted. Because `workspace_files.file_id` is `ON DELETE CASCADE`, ws2's membership row goes with it and no error surfaces.

`test_workspace_attachment_cannot_race_orphan_cleanup` covers only the opposite ordering: it sets `workspace_cleanup_claimed = TRUE` by hand before starting the attach, which is the direction where the blocked attach genuinely re-reads an updated row. The attach-first direction has no coverage.

Locking the membership rows as well, or re-checking `workspace_files` inside `finalize_claimed_file_cleanup` before the delete, would close it.

openrag/services/orchestrators/workspace_service.py
```diff
@@ -181,17 +182,25 @@ async def _delete_file(self, file_id: str, partition: str) -> None:
...
+            )
+            if ids:
+                vector_cleanup_started = True
+                await self._vector_store.delete(ids, self._collection)
+            if not await self._workspace_repo.finalize_claimed_file_cleanup(file_id, partition):
+                raise RuntimeError(f"Workspace cleanup claim disappeared for {file_id}")
+        except Exception:
+            if not vector_cleanup_started:
+                await self._workspace_repo.release_claimed_file_cleanup(file_id, partition)
+            raise
```
EnjoyBacon7: Once `vector_cleanup_started` is true the release is deliberately skipped, which is right on its own, but nothing retries afterwards: `finalize_claimed_file_cleanup` has exactly one production call site. A transient failure here (pool exhaustion, a network blip) leaves the `files` row claimed with its original `workspace_cleanup_claimed_at`, vectors already gone, and no path back.

openrag/services/persistence/workspace_repo.py
```diff
@@ -136,14 +185,38 @@ async def add_files_to_workspace(
...
                     """
                     SELECT file_id, id FROM files
-                    WHERE file_id = ANY($1::text[]) AND partition_name = $2
+                    WHERE file_id = ANY($1::text[])
+                      AND partition_name = $2
+                      AND (
+                          NOT workspace_cleanup_claimed
+                          OR workspace_cleanup_claimed_at IS NULL
+                          OR workspace_cleanup_claimed_at < NOW() - INTERVAL '1 hour'
+                      )
```
EnjoyBacon7, line 265: This reclaim is on wall-clock age alone, with nothing recording whether the vector deletion for that claim already ran. Combined with the skipped release in `workspace_service.py`, a claim stranded after a successful Milvus delete becomes reclaimable an hour later: the claim is cleared here, the file is attached to a new workspace, and it is live in the catalog with no chunks behind it, so searches silently return nothing for it.

This is the pattern reviewers rejected on `file_content_claims` in #699 ("Renew the claim for active tasks, or only expire it after checking that its associated task is no longer active"). A phase marker on the claim, or a reconciliation pass, would distinguish "nobody got to it" from "vectors are already deleted".

`test_stale_cleanup_claim_can_be_attached_again` sets the claim columns directly and never runs a vector deletion first, so it does not exercise this combination.

openrag/services/persistence/workspace_repo.py
hedhoud: Addressed in c7b3b0fa. This review was submitted against f3ff845c. The current transaction waits for the file lock before evaluating workspace membership in a fresh statement, so an attachment committed while deletion waits is visible. The PostgreSQL regression test covers this attach-first ordering.

openrag/services/orchestrators/workspace_service.py
hedhoud: This is related to cleanup failure handling, but it does not require another change for #275 or #490. Once vector deletion starts, retaining the claim is intentional: releasing it could make a vectorless file attachable. The workspace-deletion response reports the file under orphaned_files_failed, and the existing partition file-deletion operation provides an explicit retry path. Automatic background reconciliation would be a separate reliability feature.

openrag/services/persistence/workspace_repo.py
hedhoud: Addressed in c7b3b0fa. This review was submitted against f3ff845c. Cleanup claims now record whether destructive vector cleanup started, and elapsed-time recovery applies only before that phase. A stale destructive claim therefore cannot be cleared and reattached; the PostgreSQL regression test covers that case.

openrag/services/orchestrators/workspace_service.py
```diff
@@ -181,17 +182,27 @@ async def _delete_file(self, file_id: str, partition: str) -> None:
...
-        await self._document_repo.remove_file_from_partition(file_id=file_id, partition=partition)
+        vector_cleanup_started = False
+        try:
+            ids = await self._vector_store.query_ids_by_filter(
+                self._collection,
+                {"partition": partition, "file_id": file_id},
+            )
+            if ids:
+                vector_cleanup_started = True
+                if not await self._workspace_repo.start_claimed_file_cleanup(file_id, partition):
```
Ahmath-Gadji: **Major: a transient vector-store failure permanently bricks the file.**

`start_claimed_file_cleanup` sets `workspace_cleanup_started = TRUE`, and nothing ever resets it — all four recovery clauses in `workspace_repo.py` (lines 210, 237, 288, 307) require `NOT workspace_cleanup_started`.

I reproduced this against a real Postgres by making `vector_store.delete` raise once (a Milvus blip, which `orphaned_files_failed` exists to report):

```
STATE AFTER FAILURE: {'workspace_cleanup_claimed': True, 'workspace_cleanup_started': True, 'independently_indexed': False}
ATTACH AFTER FAILURE,   missing = ['f']
ATTACH 30 DAYS LATER,   missing = ['f']
```

Aftermath: the catalog row survives, still counts against the uploader's quota, and is still listed by `GET /partition/{p}/files` — but `POST /partition/{p}/workspaces/{ws}/files` returns `404 "File IDs not found in partition"` forever, and no later workspace deletion can ever reclaim it. The workspace is already gone so the operation can't be retried, and `orphaned_files_failed` is returned once in the HTTP response and never persisted, so nobody knows to reach for the `DELETE /indexer/partition/{p}/file/{id}` escape hatch.

Migration `d2e3f4a5b6c7` deliberately gave abandoned claims a 1-hour expiry; `d3e4f5a6b7c8` gave `workspace_cleanup_started` no equivalent. Some recovery route is needed — an expiry on `started` too, a startup sweeper, or an operator endpoint. This path currently has no test.

---

**Minor, same two lines:** `vector_cleanup_started = True` is set on line 196, *before* `start_claimed_file_cleanup` runs. A DB error inside that call destroys nothing but still suppresses the release in the `except` block:

```
STATE AFTER start_claim DB ERROR: {'workspace_cleanup_claimed': True, 'workspace_cleanup_started': False}
```

The file is then un-attachable for the full hour instead of being freed immediately. Moving the assignment to after the call returns `True` fixes it.

tests/integration/api/test_workspaces.py
```diff
@@ -165,22 +178,32 @@ def test_delete_workspace_default_purges_orphaned_files(self, api_client, worksp
         assert response.status_code == 200
         body = response.json()
         assert body["status"] == "deleted"
-        assert body["orphaned_files_deleted"] == 1
+        assert body["orphaned_files_deleted"] == 0
```
Ahmath-Gadji, line 181: **Major: the default purge path lost its only end-to-end test.**

This was `test_delete_workspace_default_purges_orphaned_files` asserting `orphaned_files_deleted == 1`. After the rename it asserts `0`, and grepping every occurrence of `orphaned_files_deleted` across `tests/` shows nothing in `tests/integration/api/` asserts a positive value any more.

The replacement coverage in `tests/integration/repos/test_workspace_deletion.py` uses an `AsyncMock` vector store, so the real Milvus deletion inside `_delete_file` — and the `start_claimed_file_cleanup` ordering around it — is never exercised against a live stack. This file *does* run against a real deployment in CI (`api_tests.yml`), so that reach is available and currently unused.

Given the PR description's own diagnosis — "the existing regression test exercised copied deletion logic, so it stayed green while production lost data" — the mirror risk is that the deletion path is now not exercised end-to-end at all. Suggest adding a case here that uploads with `workspace_ids=[workspace_id]`, deletes the workspace with the default `keep_files`, and asserts `orphaned_files_deleted == 1` plus the file being gone from the partition.

Ahmath-Gadji (commented):
Reviewed the full diff and ran the suites: 3008 unit tests pass, the 42 repo integration tests (incl. the 13 new ones) pass against a real PostgreSQL 15, and `ruff check` / `ruff format --check` / the layer-import guard are all clean. The migration chain is linear and single-headed (`b9c0d1e2f3a4 -> c0d1e2f3a4b5 -> d1e2f3a4b5c6 -> d2e3f4a5b6c7 -> d3e4f5a6b7c8`) and all four migrations are idempotent.

The core serialization is genuinely sound — I traced both interleavings against Postgres READ COMMITTED `FOR UPDATE` re-check (EvalPlanQual) semantics and confirmed by probe that attach-then-delete and delete-then-attach both resolve correctly, and that the fresh statement snapshot in the candidates CTE makes the cleanup re-see a membership that landed while it was blocked. `process_file`'s remote contract is unchanged, so no `_INDEXER_ACTOR_PROTOCOL_VERSION` bump is needed.

Two major findings left, inline. Also worth stating explicitly even though it looks intended and is documented: every client using the two-step flow (upload, then `POST .../workspaces/{ws}/files`) now gets `independently_indexed = TRUE`, so workspace deletion will never clean up its files — only uploads that pass `workspace_ids` are ever purged. The fix for "deletes too much" swings to "deletes nothing" for that flow, which is the change most likely to surprise existing integrations.

Ahmath-Gadji (approved):
Both findings I raised are fixed, and I re-verified them against a real PostgreSQL rather than relying on the green CI.

**Finding 1 — permanently bricked cleanup: fixed.** Resolved better than what I suggested: instead of just correcting the bound constant, `claim_failed_file_cleanup` now keeps the row in `CLEANUP_STARTED` with consistent booleans, and `_delete_owned_file` takes `cleanup_started=True` so the retry skips the `CLAIMED → STARTED` transition rather than faking it. The `cleanup_session` advisory lock added in 37fe7267 closes the rest: every claim mutator goes through `_owned_connection`, so two workers can no longer interleave on the same file.

Probe results at this head:

- failure leaves a durable `CLEANUP_FAILED` marker
- retry re-attempts the vector delete and removes the catalog row
- three consecutive failed retries preserve the marker; a fourth succeeds once the vector store recovers
- attachment stays fenced while `CLEANUP_FAILED`, including 30 days later (no claim-age escape hatch)
- concurrent retries yield `[True, False]` — the advisory lock holds

That also covers the re-attachment hazard @EnjoyBacon7 raised: a file whose vectors may already be gone is never re-attachable.

**Finding 2 — test coverage: fixed.** The new test now creates the workspace first, and the vector assertion is stronger than what I asked for — it reads `/partition/{p}/chunks` before and after, asserting chunks exist then are empty.

**Independent verification at 37fe7267:** 3025 unit tests pass; ruff check/format and the layer-import guard are clean; 34 workspace repo/deletion tests pass; migration chain is linear with a single head. The 3 repo-integration failures (`test_partition_repo` x2, `test_prompt_service_integration` x1) reproduce identically on unchanged `develop` at 84b757c2 — I checked develop out and ran them, so the "unrelated" claim holds.

Also worth calling out: always writing `independently_indexed=True` at catalog-write time and only flipping it after every attachment succeeds closes a window I hadn't flagged, where a file marked workspace-owned but never attached would have been unreachable by any cleanup.

Non-blocking, for a follow-up:

1. `retry_failed_file_cleanup` has no API route — it is reachable only from tests, so an operator cannot trigger recovery. The file is safely fenced rather than lost and `DELETE /indexer/partition/{p}/file/{id}` remains an escape hatch, so this is a convenience gap, not a correctness one.
2. `WorkspaceService.__init__` still stores `document_repo`, but the `remove_file_from_partition` call it fed was replaced by `finalize_claimed_file_cleanup` — nothing reads it now.
3. The `keep_files` note in `docs/.../workspaces.md` still says such files "must never be deleted as a side effect of removing a workspace", which overclaims for a file shared with another workspace.

The branch is behind `develop`; leaving the update to you.

PR #939, opened 2026-09-10
fix(partitions): validate the embedder endpoint on create and update (1/2, #762)

### LINKED ISSUE #762: Partition embedder/endpoint changes have no guardrails — can silently corrupt retrieval quality for already-indexed partitions
## Problem

## Scope

Relates to #708, #735.

Ahmath-Gadji:
Part 1 of 2 addressing #762 — partition embedder/endpoint changes have no guardrails.

The issue is not one bug but a set of them, with different blast radii. This PR closes the hard-fail paths, the one correctness bug, the missing record that makes the rest diagnosable, and the unguarded in-place edit; the remainder follows in a second PR.

### Solved here

- **A — the embedder name is never validated, on create or update.** `partitions.embedder` was accepted as any non-empty string. A typo passed `POST /partition/{name}` and `PATCH /partition/{name}` unchallenged and only surfaced later as a bare `KeyError` out of `_build_embedder_factory`, taking down every upload *and* every query on that partition. `_validate_preset_refs` and `_validate_chat_llm_ref` already existed; there was simply no `_validate_embedder_ref`.

- **B — deleting an endpoint still strands `partitions.embedder`.** `_PARTITION_COLUMN_BY_TYPE` was consulted by `rename()` only, so `delete_and_promote_default` cleared preset references and never looked at the partitions table. Deleting an in-use embedder now returns 409 — clearing the reference the way preset selections are cleared would silently repoint indexed vectors at a different model, so there is no safe fallback to clear *to*. A `chat_llm` reference is cleared instead, since that one really does fall back to the default LLM at request time. Deleting the default embedder counts as a reference too: promotion moves every partition on the `default` alias to another model. The delete dialog's static "Partitions referencing it will break" is now a real `used_by_partitions` count.

- **C — editing an endpoint's URL or model name is entirely unguarded.** The edit never touches the partition row, so neither A nor B covers it, and `update_model_endpoint` invalidates the client cache — the new model serves the very next query, with no restart to act as a speed bump. Repointing `endpoint` or `model_name` therefore swaps the vector space under every file already indexed through it, silently. Three steps now sit in front of that: clicking edit on an embedder explains that an embedder owns its vector space and offers a prefilled create form instead, so standing up a second endpoint for a second model is the path of least resistance; Update shows what changed, from what to what; and a material change states the consequences and will not submit until the operator ticks that they understand.

  Material means `endpoint` or `model_name` — the two fields that decide which model serves a request. A rename is a label (#770 cascades it) and `batch_size`/`timeout` are throughput knobs, so those get the summary without the gate. `extra` is diffed by neither: a stored API key comes back redacted, so comparing it would report a change on every save. D is the prerequisite for the numbers — a new `GET /model-endpoints/{model_type}/{name}/indexed-usage` counts indexed files per partition for one endpoint, resolving the `default` alias the way the rest of the resolution path does, so the dialog says how many files in which partitions are about to be stranded instead of issuing a generic caution. Model types that store no vectors return empty without touching the database.

- **D — nothing recorded which embedder produced a vector.** Changing a partition's embedder rewrites nothing, so after a swap there was no way to tell which files were left behind, or that anything had happened at all. Each indexing run now writes what actually ran — endpoint reference, resolved model name and base URL, measured dimension — into the per-file `files.indexation_config` snapshot it already persists, so there is no migration and no extra write. Written once at index time and never rewritten: the record going stale against the partition's current setting *is* the drift signal. Surfaced on partition detail (a panel, escalating to a warning with the affected file count) and on Documents (a sortable Embedder column, a partition-wide toolbar summary, and a CSV field). Files indexed before this landed report `null` and are shown as unknown, never as drifted.

- **Drift is announced, not just recorded.** D leaves the mismatch discoverable but silent: a reader learns of it only by opening the partition page and reading the panel. Everywhere else a swap looks like nothing happening — queries keep returning results, just ranked in a vector space the files were never built in, so nothing errors and nothing prompts anyone to look. Partition detail now interrupts once with a modal naming the drifted models, their file counts and the model queries run today. Once per browser session per drift state, keyed on *which* models drifted rather than on the partition, so re-opening does not nag but a new embedder appearing later does.

  Drift is judged on the **model**, never the endpoint name. An endpoint name is a renameable label: #770 cascades a rename to `partitions.embedder`, but a file's recorded provenance is a historical fact and is deliberately never rewritten — so after any rename the two labels necessarily disagree, and comparing them reports every healthy partition that was ever renamed. Only the model fixes the vector space. This replaced the label comparison in D's own panel and column, where it was a live false positive: on a partition whose endpoint had been renamed, ten healthy files read as drifted and one model rendered as two rows both marked "current". Groups merge by model, so two endpoints running one model are one entry. The label survives only as a fallback for a reference that resolves to no model at all, and files with no provenance stay unknown rather than drifted. Per-file rows carry `embedder_model_name` from the snapshot they already store, so the column names the model that actually ran instead of resolving a reference to whatever its endpoint points at today.

- **E — the API returns a fabricated dimension.** `partitions.dimension` is never written by any code path, so it sits at its `server_default` of 1024 forever — and both partition endpoints returned it verbatim. A client reading it got a *wrong* number, not a missing one, whenever the embedder was anything but 1024-d. The API now reports the dense-vector dimension the live collection actually holds, or `null` when that is genuinely unknown. The column stays: with `collection_name` it is the hook a per-partition-collection topology would need, and it is now documented as such rather than reported as fact.

### Continued in #952

- **F** — one vector field per embedder, so a swap no longer mixes two models' vectors in one index.

## New Features
- Admin model endpoints show partition usage and indexed-file counts.
- Document lists display the embedder model used for indexing, CSV exports, and drift indicators.
- Partition details show indexed embedder breakdowns, live vector dimensions, and drift alerts.
- Embedder edits include impact summaries, safeguards, and copy-as-new options.
- Default embedders are preserved for partitions containing indexed files.

## Bug Fixes
- Partition embedder references are validated, including the default embedder.
- Unknown vector dimensions now display as unavailable.
- Endpoint deletion prevents unsafe embedder removal and resets chat LLM references appropriately.

## Documentation
- Partition configuration guidance now explains embedder validation and default resolution.


ui/src/pages/admin/models.tsx
```diff
@@ -1036,6 +1204,176 @@ function EndpointDialog({
...
+      {editing && pendingUpdate && (
+        <EndpointUpdateConfirmDialog
+          editing={editing}
+          changes={pendingUpdate.changes}
+          loading={loading}
+          onCancel={() => setPendingUpdate(null)}
+          onConfirm={() => {
+            const { data } = pendingUpdate;
+            setPendingUpdate(null);
+            onUpdate(editing.model_type, editing.name, data);
```
hedhoud: This confirmation exists only in the browser. A direct model-endpoint update still changes the URL or model immediately, so API clients can bypass the safeguard and silently strand every indexed vector. Please enforce the material-change guard in the service/API as well, with an explicit server-validated acknowledgement when indexed data is affected.

openrag/services/persistence/model_endpoint_repo.py
```diff
@@ -39,6 +39,64 @@
...
+_CLEARABLE_PARTITION_COLUMN_BY_TYPE = {"llm": "chat_llm"}
+
+# Partitions whose *resolved* embedder is this endpoint: the ones naming it
+# outright, plus — when it is the row being deleted and that row is the default
+# — the ones riding the `default` alias, which promotion would silently move to
+# another model. Split so the error can say which is which.
+_EMBEDDER_USAGE_SQL = """
+    SELECT
+        COUNT(*) FILTER (WHERE embedder = $1)::int AS direct,
+        COUNT(*) FILTER (WHERE $2::boolean AND embedder = $3)::int AS via_default
```
hedhoud: This protects alias-backed partitions when the default embedder is deleted, but ordinary default promotion remains unguarded. The set-default route and create/update with is_default can still move every partition using the default alias to another vector space without reindexing. Please apply the same protection or explicit acknowledgement to every embedder default change in the shared backend transaction.

ui/src/pages/admin/partitions/embedder-drift.ts
```diff
@@ -0,0 +1,116 @@
...
+        id,
+        key: name,
+        // When either side's model is unknown, the labels are all that is left,
+        // and an unresolved alias on either side leaves nothing to compare.
+        drifted:
+          recorded &&
+          (widthDrifted ||
+            (model !== null && currentModel !== null
+              ? model !== currentModel
+              : !unresolvedAlias(configured) && !unresolvedAlias(r.embedder) && shown !== currentName)),
```
hedhoud: For non-admin users the endpoint registry is unavailable by design. With an explicit configured endpoint, currentModel is then unknown while the row model is known, and this fallback compares a model name with an endpoint label, marking a healthy partition as drifted. Please treat an unresolved model comparison as unknown; a known dimension mismatch can still prove drift independently.

ui/src/pages/admin/documents/list.tsx
```diff
@@ -280,6 +293,59 @@ export default function DocumentListPage() {
...
+  // endpoint label, or every file indexed before an endpoint rename reads as
+  // drifted when the same model produced it.
+  const driftedFrom = (file: PartitionFile): string | null => {
+    const recorded = file.embedder;
+    if (typeof recorded !== "string" || !recorded) return null;
+    const model = file.embedder_model_name ?? resolveEmbedderModel(recorded, embedderEndpoints);
+    if (model !== null && currentModel !== null) {
+      return model === currentModel ? null : model;
+    }
+    // No model on one side or the other: the labels are all that is left.
```
hedhoud: The endpoint query is admin-only, so a partition member cannot resolve the current model. For a partition using the default alias, a file recorded under the concrete endpoint is then compared against the literal default label and shown as drifted even when it is healthy. Please keep the state unknown when the current side cannot be resolved instead of falling back to this label comparison.

ui/src/pages/admin/models.tsx
```diff
@@ -394,48 +545,56 @@ function EndpointDialog({
...
         setExtraJson(JSON.stringify(displayExtra, null, 2));
-        setValidated(true);
+        setValidated(editing ? true : null);
         setValidationMsg(
-          editing.has_api_key
+          editing?.has_api_key
             ? "API key is stored server-side. Leave it unchanged to keep it, or type a new key to rotate it."
-            : null,
+            : seed?.from.has_api_key
+              ? "Copied from " + seed.from.name + ". Its API key is stored server-side and cannot be copied — enter one for the new endpoint."
```
hedhoud: This copy warning shares validationMsg with the validation-reset effect below. Because a seeded create has no editing endpoint, that effect clears the message as the seeded fields settle, so the user loses the explanation that the API key was not copied. Please keep this notice in separate state from endpoint validation feedback.

hedhoud (changes requested):
Two backend guardrails are still bypassable, and two non-admin views can report healthy data as drifted. Details are inline. These need changes because the API still permits the silent vector-space switch this PR is intended to prevent.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -621,6 +696,37 @@ async def update_model_endpoint(self, name: str, model_type: str, **fields: obje
             self._invalidate_client_cache(model_type, "default")
         return await self._repo.get(effective_name, model_type) or (updated or existing)
 
+    async def _refuse_unacknowledged_repoint(self, existing: ModelEndpointRow, fields: Mapping[str, object]) -> None:
+        """Raise unless this embedder edit leaves every indexed file's vectors valid."""
+        changed = _material_embedder_changes(existing, fields)
+        if not changed:
+            return
+        usage = await self._repo.indexed_file_usage(existing.name, existing.model_type)
```
hedhoud: This check is not serialized with the full indexing operation. A job can be admitted while usage is empty, keep the old embedder configuration, and persist its vectors after this check and after the endpoint update. Please use a shared fence for material endpoint edits and in-flight indexing, then recheck usage under that fence before committing the edit. Otherwise this race can still create the vector-space mismatch the guard is meant to prevent.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -607,6 +681,7 @@ async def update_model_endpoint(self, name: str, model_type: str, **fields: obje
             # Clears any prior default and sets this row in one transaction, then
             # re-points the 'default' alias to it — never leaves two defaults.
             await self._repo.set_default(model_type, effective_name)
+            await self._reload_partitions_after_default_change(model_type)
```
hedhoud, line 701: Please refresh the endpoint registry before reloading partitions here. The database promotion leaves empty partitions on the default alias, but this reload still resolves that alias to the old endpoint. The later registry refresh does not reload those partitions again, so they can keep the old embedder in memory until another reload or restart.

hedhoud (changes requested):
The five issues from my previous review are fixed. Two backend consistency gaps remain: a material endpoint edit can race with an already admitted indexing job, and update-based default promotion reloads partitions before the alias registry. Both can leave a partition querying with a different embedder than the one that produced its vectors.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: This reload doesn't capture the alias. A loaded partition keeps only the name `default` (`PartitionConfig.embedder`), and each request resolves it through `config.models.embedder` (`searcher_factory(partition_cfg.embedder)`). The `load_all()` right after it repoints `default` and then evicts the cached `default` client, so once `update_model_endpoint` returns, an empty alias partition embeds with the new default. Nothing stays on the old model until a restart.

The order only matters while the partition reload awaits its DB read, and there the proposed order is the unsafe one:

- Partitions first (this path): an indexed partition already reads its pinned name, and an empty one still reads the old default. No query hits the wrong vector space.
- Registry first (what `set_default` does today): an indexed partition the transaction just pinned still reads `default` in memory while `default` already points at the new model, so a query in that window embeds with a model its vectors weren't built with.

I checked both against the service, with a stub partition service that resolves an indexed alias partition during the reload: `set_default` returned the new model, the update path the pinned one. So the change goes the other way: `set_default` should reload partitions before `load_all()`, like this path.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: The window is real: the usage check counts catalog rows, and a file's row is written only after its vectors are in Milvus, so an edit made while an embedder's first files are still indexing passes without acknowledgement.

A shared fence with a recheck at edit time wouldn't close it, though. Indexer actors don't read the endpoint registry per job: they refresh `cfg.models` on a 60 s TTL (`_MODEL_REGISTRY_TTL_SECONDS`), and when it expires the refresh runs in the background while the file that triggered it keeps the old copy. A job admitted after the edit has committed, with nothing in flight during the recheck, can still embed with the pre-edit URL or model. That lag already exists on develop for every endpoint edit.

The check that holds runs at write time: in the catalog write, lock the endpoint row, compare it with the endpoint the worker actually embedded with, and fail the file on a mismatch (the failure path already removes the vectors when the catalog row isn't written). The edit locks the same row and rechecks usage under it. That changes the worker's write path, so I'd track it as a follow-up in #958 rather than add it here.

Until then a lost race is visible rather than silent: each file records the model and base URL that actually ran, so a model change shows up as drift on the partition page and in the documents list.

hedhoud (changes requested):
The current head still has one correctness issue that needs an explicit disposition before approval: a material embedder edit can race with an admitted or active indexing job, allowing vectors to be written with stale endpoint configuration after the edit. The existing inline thread describes the write-time fingerprint/revision check needed; #958 now tracks that work, but the guard added here remains bypassable until it lands. The branch also conflicts with develop, so please resolve the conflicts and rerun CI before the next review. I did not raise the two minor automated suggestions: the test is genuinely async, and the document embedder summary is intentionally partition-wide.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -567,6 +639,9 @@ async def update_model_endpoint(self, name: str, model_type: str, **fields: obje
         if isinstance(fields.get("extra"), dict):
             fields["extra"] = preserve_existing_secrets(existing.extra, fields["extra"])  # type: ignore[arg-type]
 
+        if model_type == "embedder" and not acknowledge_indexed_data:
+            await self._refuse_unacknowledged_repoint(existing, fields)
```
hedhoud: This call still allows a material edit to race with an admitted or active indexing job using stale endpoint configuration. Please carry a vector-producing configuration fingerprint or revision with the job and validate it under the endpoint lock when writing the catalog row, while rechecking usage under the same serialization boundary for the edit. That closes both the in-flight window and the worker registry TTL window described in #958.

tests/unit/services/persistence/test_model_endpoint_repo.py
```diff
@@ -544,6 +668,204 @@ async def test_delete_and_promote_default_promotes_survivor_under_lock():
...
+    repo = PgModelEndpointRepository(pool_getter=lambda: pool)
+
+    with pytest.raises(ConflictError) as exc:
+        await repo.delete_and_promote_default("jina", "embedder")
+
+    assert "'default' alias" in exc.value.message
+    assert not any("DELETE FROM model_endpoints" in q for q, _ in pool.conn.executed)
+
+
+@pytest.mark.asyncio
```
hedhoud: This test contains no asynchronous operation. Please make it a regular `def` and remove `pytest.mark.asyncio`; that keeps the test independent of the async runner and makes its synchronous nature explicit.

ui/src/pages/admin/documents/list.tsx
```diff
@@ -280,6 +293,58 @@ export default function DocumentListPage() {
...
+    if (typeof recorded !== "string" || !recorded) return null;
+    const model = file.embedder_model_name ?? resolveEmbedderModel(recorded, embedderEndpoints);
+    if (model === null || currentModel === null) return null;
+    return model === currentModel ? null : model;
+  };
+
+  // Distinct embedders across the listed files, for the toolbar summary. Built
+  // from the rows themselves, so it always describes what is on screen, and
+  // grouped by model so two endpoints running one model read as one entry.
+  const indexedEmbedders = (() => {
```
hedhoud, line 338: Using `fileRows` is consistent with the partition-wide summary rendered below, but the preceding comment says this describes what is currently on screen. Please update that comment to say it intentionally covers the whole partition, including rows hidden by the active filters, so a future cleanup does not incorrectly switch this to `filteredFileRows`.

hedhoud (commented):
Separated follow-up comments are attached to the relevant lines. The indexing race remains blocking; the other two are small clarity and test-cleanup items. The branch conflict with develop still needs resolution before another approval pass.

openrag/api/schemas/admin/model_endpoint_schemas.py
```diff
@@ -235,6 +265,11 @@ class ModelEndpointResponse(BaseModel):
     extra: dict[str, Any]
     has_api_key: bool = False
     is_default: bool
+    # Partitions whose resolved reference is this endpoint — those naming it
+    # plus, for the default endpoint, those riding the `default` alias. Zero
+    # for types with no partition column (reranker/vlm/stt), which are
+    # referenced through presets instead. Only the list view populates it.
+    used_by_partitions: int = 0
```
hedhoud, line 273: Non-blocking: the single-endpoint GET currently returns the schema default `used_by_partitions: 0`, while the list endpoint returns the actual partition count. That can make the same endpoint appear unused depending on which route a client reads. Please populate this count consistently, while preserving zero for endpoint types without partition references. This does not affect the embedder integrity guard, so it does not block approval.

hedhoud (dismissed):
The previously blocking endpoint-edit/indexing race is closed: the vector-producing configuration is validated at the catalog boundary, and stale vectors are removed when the configuration changed in flight. The current head has green CI, 319 targeted tests pass, and Ruff passes. One response-consistency issue remains as a non-blocking inline comment.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Fixed in 4dab6689, following the plan in #958:

- The worker stamps each embedder client with a fingerprint of the config it was built from: URL, model, `implementation` and `max_model_len`. The catalog write compares that fingerprint with the partition's current endpoint row, in the transaction that writes the file row and with the row locked FOR SHARE. On a mismatch the file fails with `EMBEDDER_CHANGED_DURING_INDEXING` and its vectors are removed.
- The endpoint update locks the same row FOR UPDATE and runs the acknowledgement guard's file count and the write in one transaction. A file being recorded either commits first and gets counted, or waits and then sees the edit.
- `partitions` is locked before the endpoint row, as in `set_default`. Without that lock a re-index deadlocks with a change of default, and one of the new tests reproduces the deadlock.
- Before embedding, a worker compares its loaded config with the stored row and reloads its registry when they differ, instead of waiting for the 60 s TTL. Only files that were already embedding during the edit get refused.

New Postgres tests use two connections. They cover an edit and a file write in both orders, and a change of default against both a new file and a re-index.

tests/unit/services/persistence/test_model_endpoint_repo.py
Ahmath-Gadji: Done. It's a plain `def` now, folded into 7656db4d.

ui/src/pages/admin/documents/list.tsx
Ahmath-Gadji: Reworded. The comment now says the summary covers the whole partition, including rows the filters hide, and should stay on `fileRows`. Folded into 8b32681c.

openrag/api/schemas/admin/model_endpoint_schemas.py
Ahmath-Gadji: Fixed, folded into 730d759f, the commit that added the field. GET, create, update and set-default now fill in the count with the same query as the list (`ModelEndpointService.with_partition_usage`), so the two can't disagree. Reranker, vlm and stt endpoints still report 0. A parametrized route test covers all four routes.

hedhoud (approved):

PR #942, opened 2026-09-10
fix(rag): hand the query contextualizer pre-computed calendar anchors

paultranvan:
The query with temporality such as "Résume mes échanges de la semaine dernière" reached Milvus with the wrong `created_at` range.
Tested on staging with Mistral-Small-3.2: the model resolved "last week" to the past seven days and "this week" to a window starting yesterday (a Wednesday), which matched nothing and fell through to the filterless fallback, so the answer confidently summarised documents from other weeks. The prompt said "Week starts Monday" and asked the model to do the calendar arithmetic itself; it doesn't.

What changed:
- `core/prompts/calendar_anchors.py` computes today/tomorrow/yesterday, half-open last/this week, month and year, recent/latest (90 days) and a worked "e.g. past 2 weeks" range in Python, from the UTC clock (document `created_at` is UTC). The bundled template renders the block through a new `{calendar_anchors}` placeholder, allow-listed for that prompt type. Layout is load-bearing: one anchor per line, "last X" listed before "this X", formula line last; with both on one line or "this" first, the model still answered "last week" with the current week's range.
- Only the today line is spelled out as full `T00:00:00+00:00` timestamps, as the format to reproduce; the other boundaries are bare dates, because the suffix costs as many tokens as the date itself (spelling out all 19 boundaries adds ~190 tokens to a prompt that is already ~1.5k). `Query.to_milvus_filter()` now reads a value without a timezone as UTC instead of dropping it, so a boundary copied bare still filters exactly like a completed one.
- The bundled template points its date-resolution rules at those anchors and gains a rule for explicit dates ("January 2024", "2023").
- The contextualizer call is pinned to temperature 0: identical requests flipped between a date filter and none across requests.

Verified by replaying the exact system prompt against the same LLM as staging (temperature 0, 20 cases fr + en × 4 runs): last/this week, yesterday, today, last/this month, last year, since two months, January 2024, in 2023, past 7 days, past 2 weeks, past 3 weeks, latest/recent and the no-filter/no-retrieval controls are exact on 80/80 runs. Known residual, unchanged from before: "since X" still gets an upper bound of today although the prompt asks for a single predicate. Cost: the anchors block is 252 tokens (`get_num_tokens`); the whole French system prompt goes from ~1340 to ~1620 tokens.

**Deployment note.** `seed_defaults()` never overwrites an existing library default, so an instance that has already booted keeps its previous `default_query_contextualizer` and gets no anchors (an unused format argument is not an error). Until #982 lands, update that prompt through `PATCH /prompts/{id}` with the new template content after deploying.

## Summary by CodeRabbit

* **New Features**
  * Added calendar-aware handling for relative date queries, including today, yesterday, tomorrow, weeks, months, years, and recent periods.
  * Query prompts now use consistent UTC-based calendar boundaries to improve date interpretation.

* **Bug Fixes**
  * Improved temporal filtering for bare dates, naive timestamps, UTC values, and timestamps with offsets.
  * Reduced date mismatches near midnight across different time zones.
  * Query-generation responses are now more consistent through deterministic processing.


openrag/core/prompts/calendar_anchors.py
```diff
@@ -0,0 +1,65 @@
...
+    return (
+        "Calendar anchors (use verbatim, do not recompute):\n"
+        f"- today {today}, tomorrow {tomorrow}, yesterday {yesterday}\n"
+        f"- last week [{last_monday}, {this_monday})\n"
+        f"- this week [{this_monday}, {next_monday})\n"
+        f"- last month [{last_month}, {this_month})\n"
+        f"- this month [{this_month}, {next_month})\n"
+        f"- last year [{last_year}, {this_year})\n"
+        f"- this year [{this_year}, {next_year})\n"
+        "- past N days/months: [today − N, tomorrow)"
```
hedhoud: Please cover every relative range promised by the prompt. The current block only gives a generic past N days/months formula, while [the template](https://github.com/linagora/openrag/blob/eccd6e5d2a04db93e912eef00fdbee599ad648d2/openrag/prompts/templates/query_contextualizer_tmpl.txt#L33-L40) also promises past N weeks, past N years, and recent/latest (90 days). Those requests have no concrete range for the model to copy, so the filter can be missing or wrong. Add matching formulas or anchors for these cases and tests for each supported form.

openrag/services/orchestrators/prompt_service.py
```diff
@@ -54,7 +54,7 @@
     # when a request sets metadata.spoken_style_answer), so it takes the same
     # placeholders and must be validated identically.
```
hedhoud, line 55: Please handle existing database defaults when this template changes. [seed_defaults()](https://github.com/linagora/openrag/blob/eccd6e5d2a04db93e912eef00fdbee599ad648d2/openrag/services/orchestrators/prompt_service.py#L191-L200) skips the type when a default row already exists, so upgraded deployments can keep the old query_contextualizer prompt and never receive the new {calendar_anchors} value. Add a migration or safe backfill for the bundled default while preserving administrator-edited prompts, and cover an existing-row upgrade in a test.

openrag/core/prompts/calendar_anchors.py
```diff
@@ -0,0 +1,65 @@
...
+    this_month = today.replace(day=1)
+    next_month = _first_of_next_month(today)
+    last_month = _first_of_previous_month(today)
+    this_year = today.replace(month=1, day=1)
+    next_year = this_year.replace(year=this_year.year + 1)
+    last_year = this_year.replace(year=this_year.year - 1)
+    return (
+        "Calendar anchors (use verbatim, do not recompute):\n"
+        f"- today {today}, tomorrow {tomorrow}, yesterday {yesterday}\n"
+        f"- last week [{last_monday}, {this_monday})\n"
```
hedhoud, line 81: Please render every anchor boundary as a timezone-aware ISO-8601 UTC timestamp. These values are copied into temporal_filters, and [Query.to_milvus_filter()](https://github.com/linagora/openrag/blob/eccd6e5d2a04db93e912eef00fdbee599ad648d2/openrag/core/models/query.py#L64-L96) drops timezone-naive values. With the current output, a valid anchored response can lose all date predicates and search unfiltered documents. Keep the same ranges, but emit values like 2026-04-15T00:00:00+00:00 for every boundary, and add a test for the generated anchor text.

openrag/services/orchestrators/query_service.py
```diff
@@ -365,9 +366,13 @@ async def generate_query(
             "query_contextualizer",
             names=[self._retrieval_prompt_name("query_contextualizer_prompt_name", partition)],
         )
+        now = datetime.now()
```
hedhoud: Please compute the anchors from UTC time. datetime.now() uses the host local clock, but the document timestamps and prompt output are UTC. Around midnight this can shift today, week, month, or year by one day. Use one timezone-aware UTC value for both current_date and calendar_anchors, such as datetime.now(UTC), and add a test with a fixed UTC instant. See [the current code](https://github.com/linagora/openrag/blob/eccd6e5d2a04db93e912eef00fdbee599ad648d2/openrag/services/orchestrators/query_service.py#L369-L375).

openrag/core/prompts/calendar_anchors.py
paultranvan: Adding "T00:00:00+00:00" cost ~19 tokens. And we repeat it for each anchor, so 19 x 10 = 190 additional tokens. So I propose the following: we add the ISO-8601 UTC for today / tomorrow / yerstarday anchor since it can matter, but not for the others.
And we deal with `milvus_filter` by adding the timezone when we call it, not in the prompt. 
Does that sound fair to you?

openrag/core/prompts/calendar_anchors.py
paultranvan: Fair point, fixed in 1b592fb9

openrag/services/orchestrators/prompt_service.py
paultranvan: That's a fair point, but I suggest we create a dedicated issue since this is not really in the PR scope to me.? It's also kind of a product decision to decide whether or not we force new prompts for new deployment. Honestly I'm not sure prompts running in prod should be migrated automatically ; at least it might be good for the admin to have control on it

hedhoud: Yes, that makes sense. Let’s create a dedicated issue to define how bundled prompt updates should be applied without overwriting administrator customizations. Could you open the issue and link it here?

openrag/core/prompts/calendar_anchors.py
hedhoud: Yes, that sounds fair to me. Adding the timezone in to_milvus_filter() is more robust and avoids adding too many tokens to the prompt. Just make sure it applies to all temporal boundaries and is covered by tests.

hedhoud: The backend normalization we agreed on has not landed on the current PR head. Query.to_milvus_filter() still drops timezone-naive values; passing the date-only bounds produced by these anchors logs both predicates as dropped and returns no filter. I am reopening this until the normalization applies to every boundary and is covered end to end.

openrag/core/prompts/calendar_anchors.py
hedhoud: I checked the current PR head (eccd6e5d). The referenced commit is not part of this branch: the anchor block still covers only past N days/months, and recent/latest still has no concrete 90-day range. I am reopening this until the fix and its tests are present on the PR.

openrag/services/orchestrators/prompt_service.py
hedhoud: This follow-up is now tracked in #982. I agree that preserving administrator-edited prompts needs its own design, so this point is not blocking this PR.

openrag/services/orchestrators/query_service.py
hedhoud: This is still unchanged on the current PR head: the anchors are derived from the host-local datetime.now(), and there is no fixed-UTC regression test. I am reopening it because deployments outside UTC can select the wrong calendar day near midnight.

hedhoud (changes requested):
Re-reviewed the current head (eccd6e5d) together with the discussion. These are not new issues, and the proposed direction is agreed; the corresponding changes are simply not present on the PR yet.

Pending from the existing threads:
- Apply the agreed timezone normalization for all temporal boundaries and cover the complete path with tests.
- Derive the current date and calendar boundaries from UTC.
- Push the relative-range fix referenced in 1b592fb9, including the promised coverage for recent/latest and past N weeks/years.

The stored-prompt migration is tracked separately in #982 and does not block this PR. I am keeping changes requested only until the agreed fixes are pushed, then this can be reviewed again.

openrag/core/prompts/calendar_anchors.py
```diff
@@ -0,0 +1,83 @@
...
+        f"- this week [{this_monday}, {next_monday})\n"
+        f"- last month [{last_month}, {this_month})\n"
+        f"- this month [{this_month}, {next_month})\n"
+        f"- last year [{last_year}, {this_year})\n"
+        f"- this year [{this_year}, {next_year})\n"
+        "- past N days/months: [today − N, tomorrow)"
```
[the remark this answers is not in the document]
paultranvan: Added in 1b592fb9: the block now carries a concrete `recent / latest (past 90 days) [start, tomorrow)` line, and the formula names every unit (`past N days/weeks/months/years`). The template's own "recent / latest → past 90 days" rule was dropped as redundant with the anchor. Replayed against Mistral Small: "Latest safety bulletins" / "documents récents" went from 0/3 (three calendar months ending today) to 4/4. Pinned by `test_recent_anchor_is_the_past_90_days_up_to_tomorrow`.

openrag/services/orchestrators/query_service.py
```diff
@@ -367,7 +368,9 @@ async def generate_query(
         )
         prompt = contextualizer.format(
             query_language=detect_language(last_user),
-            current_date=datetime.now().strftime("%A, %B %d, %Y, %H:%M:%S"),
+            # Stamp + pre-computed week/month/year ranges: the model must not
+            # do the calendar arithmetic itself (see core.prompts.calendar_anchors).
+            current_date=current_date_with_anchors(datetime.now()),
```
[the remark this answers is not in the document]
paultranvan: Done in 1b592fb9: `generate_query()` reads `datetime.now(UTC)` and `calendar_anchors()` converts an aware value to UTC (a naive one is taken as UTC). `test_generate_query_reads_the_clock_in_utc` freezes the clock at 23:30 UTC and checks the anchors stay on that UTC day.

openrag/prompts/templates/query_contextualizer_tmpl.txt
```diff
@@ -29,14 +30,10 @@ Emit `temporal_filters` when the user refers to WHEN a document was created / pu
...
-Date resolution (vs Current date):
-- today / now → `[today, tomorrow)`
-- yesterday → `[yesterday, today)`
-- this / last week → Monday-to-Monday
-- this / last month → 1st-to-1st
-- this / last year → Jan 1-to-Jan 1
-- past N days/weeks/months/years → `[Current − N, tomorrow)`
-- recent / latest → past 90 days (~3 months, NOT a year): `[Current − 90 days, tomorrow)`
+Date resolution:
+- Relative periods (today, yesterday, this / last week, month, year, past N, recent) → copy the matching anchor under Current date verbatim; never recompute it.
```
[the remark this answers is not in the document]
paultranvan: Resolved the other way round in 1b592fb9: rather than removing "recent" from the copy rule, the injected block now has a concrete `recent / latest` anchor, so the rule is consistent, and the separate "past 90 days" rule (which had no anchor to copy) was removed.

openrag/core/prompts/calendar_anchors.py
paultranvan: Covered in 1b592fb9. The block now carries a `recent / latest (past 90 days)` anchor, a worked `e.g. past 2 weeks` instance, and the formula names `days/weeks/months/years`. Arbitrary N cannot be enumerated, so `past N` stays a formula; the worked instance is what makes the model apply it. Replay against Mistral Small (temperature 0, 4 runs): "past 2 weeks" fr/en 0/4 → 4/4, "past 3 weeks" 4/4, "latest" / "récents" 0/3 → 4/4, "past 7 days" 4/4, and every case that passed before still passes (80/80 over 20 cases). Note the "e.g." label matters: as a plain anchor, "past 3 weeks" copied the 2-weeks range instead of applying the formula. Tests: `test_recent_anchor_is_the_past_90_days_up_to_tomorrow`, `test_past_n_formula_names_every_unit_the_prompt_accepts`, `test_past_two_weeks_is_given_as_a_worked_example_of_the_formula`.

openrag/services/orchestrators/prompt_service.py
paultranvan: Agreed on the gap, but it is a change to the lifecycle of every seeded prompt type, not just this one, so it gets its own issue and PR: #982. An implementation is ready (SHA-256 allow-list of superseded bundled templates, refreshed in place at boot, edited defaults and custom `PROMPTS_DIR` templates untouched, tested with the v2.2 template as a fixture). Until it lands, the deployment step is `PATCH /prompts/{id}` with the new content, stated in the PR description.

openrag/core/prompts/calendar_anchors.py
paultranvan: The risk is real (a verbatim copy of a bare date was dropped by `to_milvus_filter()` and the search ran unfiltered), but spelling out every boundary is expensive: the suffix `T00:00:00+00:00` tokenizes to 10 tokens, as much as the date itself, and there are 19 boundaries, so +190 tokens on a prompt already at ~1.5k. Two changes instead: (1) a125ce6c makes `to_milvus_filter()` read a value without a timezone as UTC (the prompt's own "Timestamps: UTC" contract) and render the normalised aware timestamp, so `2026-09-07` reaches Milvus as `2026-09-07T00:00:00+00:00`; (2) in 1b592fb9 only the today line is spelled out as full timestamps, as the format to reproduce, the other boundaries stay bare dates. `test_every_boundary_survives_the_milvus_filter_whether_copied_bare_or_completed` checks both forms of every anchor produce the same filter; `tests/unit/core/models/test_query.py` covers bare date, naive datetime, `Z`, non-UTC offset and unparseable values. In the replay (80 runs) the model completed the suffix itself every time.

openrag/services/orchestrators/query_service.py
paultranvan: Done in 1b592fb9: one aware UTC value from `datetime.now(UTC)` feeds both `current_date` and `calendar_anchors`, and `calendar_anchors()` converts any aware input to UTC. `test_generate_query_reads_the_clock_in_utc` uses a fixed instant (2026-09-16 23:30 UTC, already the 17th in Paris) and checks the today/tomorrow/last-week anchors stay on the UTC day; `test_now_is_read_in_utc_not_in_its_own_zone` covers the conversion itself.

hedhoud (approved):
Re-reviewed the current head and the resolved discussion. The requested UTC clock handling, temporal-boundary normalization, and relative-range coverage are now present with regression tests. The stored-prompt lifecycle remains correctly tracked in #982 and does not block this change. Current-head CI and the GitHub CodeRabbit review are green.

hedhoud (approved):

PR #943, opened 2026-09-11
fix(storage): report and repair catalog-vector divergence

### LINKED ISSUE #924: No way to detect or repair Postgres↔Milvus divergence; two code paths already defer to a reconciliation that does not exist
## Summary

Postgres and Milvus hold the same documents and are written without a shared transaction, and nothing detects when they disagree. Two code sites already defer to a reconciliation process that does not exist.

This is the residual after #658 / #671, not a re-file: that work fixed the *ordering* (vectors deleted before catalog rows) and narrowed the window. What it did not add is a way to detect or repair a divergence once one happens.

## Evidence

**Two sites log a deferral to reconciliation.** On the re-index/replace path, when deleting the superseded chunks fails after the catalog has been committed, both of these swallow the exception and log:

- `openrag/services/workers/indexer_actor.py:283-287` — `"re-index: catalog was updated but failed to delete N stale chunk(s); duplicates remain until reconciliation"`
- `openrag/services/workers/pipeline_builder.py:309-313` — `"re-index: stored new chunks but failed to delete N stale chunk(s); duplicates remain until reconciliation"`

Both are marked best-effort in the code (`# noqa: BLE001 - cleanup is best-effort after catalog commit`). Nothing consumes those log lines.

**Nothing reconciles.** `grep -rn "def reconcile\|reconciliation" openrag/ scripts/` returns only those two log strings.

**The delete path still has a residual window.** `services/workers/dispatcher.py:522-547`: Milvus `delete_by_filter` at 534, the Postgres rows at 535-536, then a second Milvus sweep at 539 which re-raises on failure. A process killed between 536 and 539 leaves any chunks an in-flight indexer wrote between 534 and 536 with no catalog row.

**The replace path cannot self-heal.** `indexer_actor.py:270-287` deletes superseded chunks by explicit ids taken from the task row (`REPLACE_OLD_CHUNK_IDS_ROW_KEY`). If that delete fails, the ids go with the row and nothing on the surviving chunks marks them as superseded.

## What the two stores key on

- Postgres `files`: `partition_name`, `file_id` — indexed by `ix_partition_file` (`services/persistence/schema.py:228`)
- Milvus: `partition` (partition_key) and `file_id` (`services/storage/milvus_store.py:306,312`)

## Proposal

A read-only reconciliation report by default, with repair behind an explicit flag:

1. **Milvus chunks with no `files` row** — safe to delete; this is the direction that keeps deleted content queryable.
2. **`files` row with no chunks** — needs a human decision (re-index or remove the row), so no automatic repair.
3. **Duplicate chunk sets for one `file_id`** — the failed-replace case above. Detection is the hard part: nothing currently stamps a chunk with the indexing run that produced it, so this may need a schema addition to be reliable. Worth scoping separately if so.

Requirements:

- **Per-partition and paged.** An unbounded scan of the collection is not viable at the corpus sizes this is meant for.
- **A grace window.** A document mid-indexing is legitimately in Postgres and not yet in Milvus. Without a minimum age, every run reports drift that is not drift, and the output gets ignored within a fortnight. `TaskStateManager` uses the same pattern for orphaned tasks.
- **Machine-readable output**, so it can be run after a restore (where the two backups are necessarily taken at different instants and cannot agree) and after a bulk import to prove completeness.

## Related

- #658 — deleted files/partitions leaving orphaned queryable chunks (fixed by #671, ordering)
- #379, #378 — earlier delete/replace divergence, both closed
- #693 — indexing dispatch and delete lack a shared admission fence

hedhoud:
Interrupted indexing, deletion and restores can leave Postgres and Milvus disagreeing without giving operators a way to inspect the drift. A paged, per-partition command now reports it as JSON, with an indexing grace window and read-only behavior by default.

Explicit repair removes only aged orphan chunks after a fresh catalog check. Writers must be paused and drained first because the stores have no shared transaction. Missing chunks, unknown ages and indexing timestamp mismatches remain report-only; reliable duplicate-set repair needs a durable generation identity and is outside this scope, as allowed in #924.

This PR addresses #924 separately from the retrieval fix in #944, which is now merged.

Review follow-ups reject wildcard scans, protect newly copied documents with a shared indexing timestamp, and preserve iterator cleanup during cancellation. The operator safety guide is now included in the published documentation.

Closes #924.

## Summary by CodeRabbit

* **New Features**
  * Added catalog reconciliation between PostgreSQL and Milvus.
  * Added a command-line scan that reports orphaned chunks, missing vectors, timestamp mismatches, and unaged records as JSON.
  * Added optional repair mode to remove aged orphan chunks after safety checks.
  * File copies now receive consistent indexing timestamps across catalog and vector records.

* **Documentation**
  * Added usage guidance, options, output formats, exit codes, safeguards, and repair behavior for catalog reconciliation.


openrag/services/storage/milvus_store.py
```diff
@@ -1448,6 +1455,32 @@ async def query_ids_by_filter(
...
+            raise ValueError("A partition and a batch size between 1 and 1000 are required")
+        if file_ids == []:
+            return
+        filters: dict[str, Any] = {"partition": partition}
+        if file_ids is not None:
+            filters["file_id"] = file_ids
+        iterator = await asyncio.to_thread(
+            self._client.query_iterator,
+            collection_name=self._collection_name,
+            filter=self._build_filter_expr(filters),
```
Ahmath-Gadji: **Critical: `partition="all"` silently widens this scan to the entire collection.**

`_build_filter_expr` treats `"all"` as `_PARTITION_WILDCARDS` and skips the partition clause, so this produces `filter=""` — an unfiltered, cross-tenant scan. I reproduced it:

```python
store._client.query_iterator.return_value = it
async for page in store.iter_chunk_metadata("c", partition="all"): pass
print(repr(store._client.query_iterator.call_args.kwargs["filter"]))
# -> ''
```

`reconcile_partition` then resolves every row against the key `("all", file_id)`, which matches no catalog row, so **every aged chunk in every partition is reported as an orphan** — and with `--repair` the recheck agrees and deletes them:

```
summary: {'scanned_chunks': 2, 'orphan_chunks': 2, 'deleted_chunks': 2}
DELETED chunk ids (healthy rows from every tenant): ['1', '2']
```

(Both tenants above were fully catalogued and healthy.)

Neither `reconcile_partition`'s validation nor the CLI's `parser.error` rejects the wildcard — `not partition` only catches the empty string. And `all` is the product's established wildcard everywhere else (`?partitions=all`, `openrag-all`, `partitions.py:78`), so an operator sweeping a whole deployment after a restore — the exact use case the new doc recommends — would plausibly reach for it, and it *looks* like it works. It's also reachable without any operator mistake, since nothing reserves `all` as a partition name at creation time.

Worth noting `delete_by_filter` already guards this exact case ("Guarded so an accidental empty / wildcard filter does NOT nuke the entire collection"). This method is the new path that ends in a delete without that guard.

Suggest rejecting wildcards here, where the wildcard semantics live:

```python
if not partition or partition in self._PARTITION_WILDCARDS or not 1 <= batch_size <= 1000:
    raise ValueError("A concrete partition and a batch size between 1 and 1000 are required")
```

plus a matching `parser.error` in `reconcile_catalog.py` so it fails before connecting.

Ahmath-Gadji (commented):
## Review: 4 findings (1 critical)

Checked out the branch, ran the full unit suite (3120 passed), `ruff check` / `ruff format --check` (clean), and reproduced each issue below against the real code.

---

### 1. Critical — `--partition all` becomes a cross-tenant wildcard, and `--repair` then deletes the whole collection

`iter_chunk_metadata` routes the operator-supplied partition straight through `_build_filter_expr`, where `"all"` is a documented wildcard (`_PARTITION_WILDCARDS`) that **drops the partition clause entirely**:

```
filter expr passed to Milvus: ''     # partition="all" -> full-collection scan
```

`reconcile_partition` then looks up `("all", file_id)` in Postgres — no such partition — so **every aged chunk of every tenant is classified as an orphan**, and the repair recheck confirms the same absence. Reproduced end to end with two healthy, fully-catalogued tenants:

```
summary: {'scanned_chunks': 2, 'orphan_chunks': 2, 'deleted_chunks': 2}
DELETED chunk ids (healthy rows from every tenant): ['1', '2']
```

Reachable two ways: an operator typing the product's established wildcard (`?partitions=all`, `openrag-all`, `partitions.py:78`) to sweep a deployment — exactly the "after a restore or bulk import" case the new doc describes — or a partition genuinely named `all`, since nothing reserves the name at creation. Without `--repair` it still produces a catastrophically wrong report.

The codebase already knows about this hazard: `delete_by_filter` explicitly rejects a filter that is "only a wildcard partition" so it can't nuke the collection. `iter_chunk_metadata` reintroduces it on a path that ends in a delete. Details inline.

### 2. Medium-high — the grace window gives zero protection on the file-copy path

The grace window is the safety story for orphan repair, and it works for normal indexing: `stages/store.py:46` stamps chunks with `datetime.now(UTC)` before the catalog row is written, so an in-flight index is always "recent".

The copy path inverts this. `dispatcher.py:606-635` copies source rows verbatim (`strip_internal_metadata` only drops `_openrag*` keys, so **copied chunks keep the source's original `indexed_at`**), inserts them, and only then calls `add_file_to_partition(...)` *without* `indexed_at`, so the catalog row gets `now()`. Details inline.

### 3. Low — large finite `--grace-seconds` crashes instead of failing validation

### 4. Low — the operator doc won't be published

---

### Checked and fine

`aclose()` matches pymilvus 3.0.1 (`AsyncMilvusClient.close` is a coroutine, `MilvusClient.close` exists); `delete(ids, collection)` signature and the `str(_id)` <-> `_str_id_to_milvus` round-trip; scalar-only `output_fields` with `batch_size <= 1000` stays well under the `_safe_batch_size` ceiling, so the oversized-page guard isn't needed; `files.indexed_at` is `nullable=False` timestamptz, so `existing[key] < cutoff` can't hit `None` or a naive datetime; keyset pagination is collation-consistent and can't loop; the CLI's `rdb` copy correctly disables `auto_create_database` and skips migrations; `aclosing` correctly closes the Milvus iterator on early `break` and on error; `_timestamp` safely handles missing/undated dynamic fields; the local-import monkeypatching in the CLI test is sound.

Two nits not worth their own threads: pass 2 always issues one extra `list_indexed_documents` query (the loop only terminates on an empty page, never on a short one), and the new integration test calls `aclose()` on the function-scoped `dense_only_store` before its fixture teardown touches `store._client` (swallowed by `except Exception: pass`).

openrag/services/storage/reconciliation.py
```diff
@@ -0,0 +1,124 @@
...
+    now: datetime | None = None,
+) -> AsyncIterator[dict[str, Any]]:
+    """Yield NDJSON-ready findings followed by a summary on successful completion.
+
+    Repair is for maintenance windows with writers paused. A fresh catalog
+    lookup reduces races but cannot substitute for a shared admission fence.
+    Only explicit, aged orphan IDs can be deleted. Timestamp mismatches are
+    diagnostic: legacy imports and copies need not share an indexing timestamp.
+    """
+    if not partition or not 1 <= page_size <= 1000 or not 0 <= grace_seconds < float("inf"):
```
Ahmath-Gadji: This validation rejects the empty string but not the wildcard partition name `"all"`, which the Milvus filter builder expands to "every partition" — see the thread on `milvus_store.py`. Since this module is backend-agnostic the primary guard belongs in `iter_chunk_metadata`, but a defensive check here (and in the CLI's `parser.error`) would stop it before any I/O happens.

openrag/services/storage/reconciliation.py
```diff
@@ -0,0 +1,124 @@
...
+                summary["timestamp_mismatches"] += len(mismatches)
+                yield {"type": "indexing_timestamp_mismatch", "partition": partition, "chunk_ids": mismatches}
+            if orphans:
+                summary["orphan_chunks"] += len(orphans)
+                deleted = 0
+                if repair:
+                    current = await catalog.get_indexed_documents({(partition, r["file_id"]) for r in orphans})
+                    ids = [str(r["_id"]) for r in orphans if (partition, r["file_id"]) not in current]
+                    if ids:
+                        deleted = await vectors.delete(ids, collection)
```
Ahmath-Gadji, line 107: **The grace window provides no protection at all for chunks created by the file-copy path.**

For normal indexing the window works as intended: `stages/store.py:46` sets `indexed_at = datetime.now(UTC)` and stamps it on the chunks *before* the catalog row is written, so an interrupted index always leaves chunks that are "recent" and therefore skipped at line 72.

The copy path inverts that. In `dispatcher.py:606-635`:

1. rows are read with `output_fields=["*", "vector"]` and passed through `strip_internal_metadata`, which only drops `_openrag*`-prefixed keys — so **the copied entities keep the source document's original `indexed_at`**;
2. `_insert_entities(entities)` writes them;
3. only *then* does `add_file_to_partition(...)` run, and it is called **without** `indexed_at`, so the catalog row gets the `files.indexed_at` server default (`now()`).

Two consequences:

**Data-loss window.** Copying a document that was indexed longer ago than `--grace-seconds` produces chunks that are *immediately* past the cutoff. A repair running in the gap between steps 2 and 3 sees no catalog row on the recheck at line 90, deletes the chunks here, and the catalog row then lands — leaving a file that exists in Postgres and returns nothing. The docs do say to pause copy operations first, but the grace window is presented as the backstop for exactly this class of race, and here it contributes nothing. Unlike the indexing path, there is no window size that makes copies safe.

**Permanent report noise.** Every *successfully* copied file also emits `indexing_timestamp_mismatch` for all of its chunks, forever, because the catalog timestamp (copy time) can never equal the chunk timestamp (source index time). So on any deployment that uses copy the summary and exit code 1 never go quiet. The docstring acknowledges "legacy imports and copies need not share an indexing timestamp", but the doc also frames mismatches as a failed-replacement-cleanup signal, and the output gives an operator no way to tell the two causes apart.

Stamping a fresh `indexed_at` on the copied entities in `dispatcher.py` would fix both at once, and would make the copy path's chunk timestamps mean the same thing as everywhere else.

scripts/reconcile_catalog.py
```diff
@@ -0,0 +1,97 @@
...
+    parser.add_argument("--grace-seconds", type=float, default=3600, help="Ignore indexing times younger than this age")
+    parser.add_argument(
+        "--repair",
+        action="store_true",
+        help="Delete aged orphan chunk IDs. Pause and drain all writers before using this flag.",
+    )
+    args = parser.parse_args(argv)
+    if not args.partition or not 1 <= args.page_size <= 1000:
+        parser.error("A nonempty partition and page size between 1 and 1000 are required")
+    if not math.isfinite(args.grace_seconds) or args.grace_seconds < 0:
```
Ahmath-Gadji: `isfinite` and `>= 0` aren't sufficient — `reconciliation.py:42` feeds this into `timedelta(seconds=grace_seconds)`, which overflows well below infinity:

```
$ ... --grace-seconds 1e18
OverflowError: Python int too large to convert to C int
```

No damage (`_run`'s `except Exception` catches it and exits 2 with an error event), but it surfaces as a storage-failure-shaped event rather than the clean `parser.error` the other invalid inputs get. An upper bound here would keep argument validation in one place.

docs/catalog-reconciliation.md
```diff
@@ -0,0 +1,11 @@
+# Catalog and vector reconciliation
```
Ahmath-Gadji: This is the only top-level `.md` under `docs/` — every other operator-facing doc lives in `docs/content/docs/documentation/` and is what the Astro site actually publishes. As placed, the sole safety guidance for a destructive flag ("pause new indexing, copy, replace and delete operations and drain in-flight work") won't appear anywhere an operator browsing the docs would find it.

Moving it to `docs/content/docs/documentation/catalog_reconciliation.md` would match `backup_restore.md`, `milvus_migration.mdx` and the rest.

openrag/services/storage/milvus_store.py
hedhoud: Confirmed and fixed in b3425e7c. Metadata scans now reject the `all` wildcard, so this path cannot widen into a cross-tenant scan. The regression covers two healthy tenants and verifies that neither querying nor deletion starts.

openrag/services/storage/reconciliation.py
hedhoud: Fixed in b3425e7c. The reconciliation service and CLI both reject `all` before any storage I/O, in addition to the Milvus guard. Each run must name one concrete partition.

openrag/services/storage/reconciliation.py
hedhoud: Fixed in b3425e7c. Each copy now gets one fresh UTC indexing timestamp shared by every copied chunk and its catalog entry, overriding any inherited or caller-supplied indexing timestamp. Tests cover the gap before the catalog write and a clean report after the grace window. Repair still requires paused, drained writers; older copies remain report-only.

scripts/reconcile_catalog.py
hedhoud: Fixed in b3425e7c. The CLI and service now share validation that checks whether the cutoff is representable. Oversized finite values are rejected before opening storage, including both timedelta overflow and date-subtraction overflow.

docs/catalog-reconciliation.md
hedhoud: Moved into the published operator documentation in b3425e7c, with the required title metadata. The Astro build passes and generates the reconciliation page, including the maintenance-window safety guidance.

hedhoud:
Response to the [outside-diff cancellation finding](https://github.com/linagora/openrag/pull/943#pullrequestreview-5179156021): confirmed and fixed in b3425e7c. Cancellation retains ownership of a late-created iterator and waits for any active read before closing it. Regressions cover creation and reading, including repeated consumer cancellation, and verify closure happens exactly once.

The independent follow-up review also caught a shutdown edge case: cancelling the cleanup task itself could cause an infinite retry loop. Fixed in 9f299270 with regressions for both creation and reading during task-wide shutdown. The final focused suite passes all 209 tests; the independent reviewer reports no remaining actionable findings.

hedhoud:
Response to the [outside-diff catalog interface finding](https://github.com/linagora/openrag/pull/943#pullrequestreview-5179156021): fixed in b3425e7c. `list_indexed_documents` is now abstract, so an incomplete repository fails at construction instead of halfway through reconciliation. An interface regression verifies this.

hedhoud:
Response to the [outside-diff vector interface finding](https://github.com/linagora/openrag/pull/943#pullrequestreview-5179156021): fixed in b3425e7c. `iter_chunk_metadata` is now abstract and incomplete implementations cannot instantiate. Existing test doubles explicitly declare that reconciliation is outside their scope, and an interface regression checks the required contract.

hedhoud:
Also addressed both minor notes in the [review summary](https://github.com/linagora/openrag/pull/943#pullrequestreview-5179268506), in b3425e7c: a short catalog page now ends the scan without an extra empty-page query, and the integration test leaves collection cleanup to its owning fixture instead of closing its client prematurely.

Validation: 3,133 unit tests passed, plus 207 final focused tests and 20 live Postgres/Milvus integration tests. Lint, formatting and the documentation build pass. The local CodeRabbit follow-up raised 0 issues.

openrag/services/workers/dispatcher.py
```diff
@@ -613,17 +613,20 @@ async def copy_file(
                 return
 
             public_metadata = strip_internal_metadata(metadata)
+            indexed_at = datetime.now(UTC)
             entities = []
             for row in rows:
                 entity = strip_internal_metadata(row)
                 entity.pop("_id", None)
                 entity.update(public_metadata)
+                entity["indexed_at"] = indexed_at.isoformat()
```
Ahmath-Gadji, line 622: This is the right fix for `copy_file`, and it also closes a spoofing gap as a side effect: the server stamp is applied *after* `entity.update(public_metadata)`, so a caller-supplied `indexed_at` in the copy metadata can no longer win. Good.

Worth noting the sibling path a few lines up didn't get the same treatment. `update_file_metadata` (line ~568) does:

```python
entity = dict(row)
entity.update(public_metadata)       # caller-supplied
entity.update(internal_metadata)     # only _openrag* keys
```

and `indexed_at` is not in `PROTECTED_METADATA_KEYS`, so `strip_protected_metadata` in `indexing_service.py:309` doesn't drop it. A partition editor can therefore set a chunk's `indexed_at` to any value they like through the metadata PATCH path. Verified:

```
chunk indexed_at after caller PATCH: 2099-01-01T00:00:00+00:00
```

This is pre-existing — but this PR changes what the field *means*. Before, `indexed_at` was descriptive metadata; now it is the input that decides whether a chunk is old enough for `--repair` to delete. The practical consequence is the "hiding" direction: an editor who stamps a far-future value makes their partition's chunks permanently `recent_chunks_skipped`, so real drift there is never reported and never repairable. (The reverse direction isn't straightforwardly weaponizable — you'd need the catalog row gone while the chunks survive, which `delete_file` doesn't leave behind.) It will also generate permanent report-only `indexing_timestamp_mismatch` noise for any file whose metadata has been patched this way.

Adding `indexed_at` to `PROTECTED_METADATA_KEYS` would make it server-authoritative on every path and cost nothing — no legitimate caller sets it. Happy for it to be a follow-up issue instead, since it predates this PR; flagging it because this PR is what makes the field safety-relevant.

Ahmath-Gadji (commented):
## Re-review of `b3425e7c` + `9f299270` — all four findings verified fixed

Re-checked out at `9f299270`. Full unit suite **3136 passed** (up from 3120), `ruff check` / `ruff format --check` clean. I re-ran each original reproduction against the new head; all four are now blocked, most of them at two layers.

| # | Finding | Status |
|---|---|---|
| 1 | `--partition all` → cross-tenant scan + repair deletes everything | ✅ blocked in `iter_chunk_metadata`, `validate_scan_options` **and** the CLI |
| 2 | Grace window inert on the file-copy path | ✅ fresh shared timestamp in `copy_file`, with a race-window regression test |
| 3 | `--grace-seconds` overflow past validation | ✅ `OverflowError` → `ValueError` → clean `parser.error` |
| 4 | Doc outside the published tree | ✅ moved, frontmatter matches siblings, picked up by `autogenerate: { directory: 'documentation' }` |

Both nits are gone too: the short-page `break` in pass 2 removes the extra `list_indexed_documents` query, and the integration test no longer closes the fixture's client out from under its own teardown.

**Verification output:**

```
$ store.iter_chunk_metadata("c", partition="all")
ValueError: A concrete partition (not 'all') and a batch size between 1 and 1000 are required

$ reconcile_partition(..., "all", repair=True)
ValueError: A concrete partition is required; the 'all' wildcard is not supported

$ reconcile_catalog.py --partition all
error: A concrete partition is required; the 'all' wildcard is not supported

$ reconcile_catalog.py --partition a --grace-seconds 1e18
error: Grace seconds exceeds the supported datetime range
```

`test_copy_receives_grace_before_catalog_write_and_matches_afterward` is a nicely targeted regression — running `reconcile_partition(repair=True)` from *inside* `insert_entities` reproduces the exact insert/catalog-write window and asserts `recent_chunks_skipped == 2`, `orphan_chunks == 0`, `delete` not awaited. That is the failure mode, encoded.

---

## I stress-tested the new `iter_chunk_metadata` cleanup machinery — it holds

`9f299270` adds ~40 lines of fairly intricate cancellation handling, so I wrote nine adversarial probes against it rather than reasoning alone. **All nine pass:**

- normal drain → `close()` exactly once
- early `break` + `aclose()` → exactly once
- **cancel while a `next()` thread is blocked** → cancellation is correctly *deferred*; `close()` never overlaps an in-flight `next()` (I instrumented the fake iterator to flag any overlap), then exactly once
- **cancel during iterator creation** → the iterator born after the consumer gave up is still closed exactly once (the leak this commit targets — confirmed fixed)
- `next()` raises → closed exactly once, exception propagates
- 50 × `task.cancel()` hammered during cleanup → terminates, no spin (the retry loop your comment guards)
- generator abandoned without `aclose()` → `loop.shutdown_asyncgens` finalizes it, closed exactly once
- task still running at `asyncio.run` teardown → closed exactly once, no hang
- no pending tasks left behind after a clean drain

One consequence worth being deliberate about rather than a defect: because `to_thread` calls can't be cancelled and cleanup now retains ownership until they finish, `aclose()` is effectively **uninterruptible** while a `next()` is in flight — and `next()` gets no timeout (`self._timeout` is passed only to `query_iterator` creation). My third probe asserts this directly (`assert not t.done()` after `cancel()`). If Milvus ever stalls mid-scan, `Ctrl-C` on the CLI won't land until it returns. That's the right trade for pymilvus thread-safety, but the operator doc's "a failed or interrupted run is incomplete; rerun it" might be worth one clause noting an interrupt can take a page to take effect.

---

## One new low-severity observation

`indexed_at` is now load-bearing for a *destructive* decision, but it is still caller-writable on one path — see the inline comment.

Also trivial: `validate_scan_options` hardcodes `partition == "all"` while `iter_chunk_metadata` uses `self._PARTITION_WILDCARDS`. If that frozenset ever grows, the orchestration-layer guard drifts and the rejection moves from "before any I/O" to "at first page". A comment tying the two together would be enough — the Milvus-side guard still catches it, so this is cosmetic.

Nothing here blocks. Nice, thorough turnaround.

openrag/services/storage/reconciliation.py
```diff
@@ -0,0 +1,140 @@
...
+) -> datetime:
+    """Validate before opening storage and return a representable cutoff."""
+    if not partition or partition == "all":
+        raise ValueError("A concrete partition is required; the 'all' wildcard is not supported")
+    if not 1 <= page_size <= 1000:
+        raise ValueError("Page size must be between 1 and 1000")
+    if not 0 <= grace_seconds < float("inf"):
+        raise ValueError("Grace seconds must be finite and nonnegative")
+    try:
+        return (now or datetime.now(UTC)) - timedelta(seconds=grace_seconds)
```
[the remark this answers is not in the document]
hedhoud: Verified and fixed in the amended review-fix commit 57699f63. A naive `now` now fails validation with `ValueError`, while aware values are normalized to UTC before the grace subtraction. Focused regressions cover both cases.

openrag/services/workers/dispatcher.py
hedhoud: Verified and fixed in the amended review-fix commit 57699f63. `indexed_at` is now protected caller metadata, so REST, MCP, upload and copy entry paths share the same server-authoritative rule. The update regression proves a future caller timestamp is dropped before dispatch.

hedhoud:
Verified and fixed in the amended review-fix commit 57699f63. If asynchronous Milvus client setup fails after the synchronous client is created, constructor-owned cleanup closes the partial resource while preserving the original connection error.

The regression failed against the previous head and passes with the fix. Verification: 3,139 unit tests, 904 affected tests, repository-wide lint/format, the documentation build, and a local CodeRabbit follow-up with 0 issues.

openrag/services/storage/reconciliation.py
```diff
@@ -0,0 +1,143 @@
...
+                if timestamp is None:
+                    unaged.append(str(row["_id"]))
+                    continue
+                if timestamp >= cutoff:
+                    summary["recent_chunks_skipped"] += 1
+                    continue
+                key = (partition, row["file_id"])
+                if key not in existing:
+                    orphans.append(row)
+                elif existing[key] < cutoff and timestamp != existing[key]:
```
Ahmath-Gadji, line 121: **Every file indexed before the `files.indexed_at` migration will be reported as a timestamp mismatch.**

This check is the first code to compare the chunk timestamp with the catalog timestamp, and for older deployments the catalog side is not the real indexing time:

- Chunks have carried their true `indexed_at` (`datetime.now(UTC)`) since `53291d7d` (Feb 2026).
- `files.indexed_at` was added later by migration `b7c1d2e3f4a5`, whose `server_default=now()` backfilled every existing row with **the time the migration ran** (its docstring says so).
- Postgres has no reliable copy of the original time to fix this with: the `file_metadata` JSON only carried it between Feb and mid-May 2026, until the storage refactor stopped writing it.

So a file indexed on Mar 1 has chunks at Mar 1 and a catalog row at the migration date, and every one of its chunks lands in `indexing_timestamp_mismatch`. I reproduced it through `reconcile_partition`: 3 files × 4 chunks, catalog at the migration date, chunks at their original date → `timestamp_mismatches: 12`, `deleted_chunks: 0` (also with `repair=True`).

Nothing is deleted, so it's safe. But on any deployment older than the migration, the report is flooded, the exit code is always 1, and the real signal this category exists for (failed replacement cleanup) gets buried. Copies made before this PR fall into the same bucket.

### Proposed fix: split the category on whether the file has a current set of chunks

A failed replacement and a legacy row look different:

| File | Catalog | Chunks |
|---|---|---|
| Failed replacement cleanup | Aug 1 | 10 × Aug 1 **+** 10 × Mar 1 → some chunks match the catalog |
| Pre-migration / pre-PR copy | Jul 10 (migration) | 12 × Mar 1 → **no** chunk matches the catalog |

So instead of reporting mismatched chunks directly, collect them per file. After each page, ask one question about those files: does at least one chunk carry the catalog's timestamp?

- **Yes:** keep reporting the non-matching chunks as `indexing_timestamp_mismatch` (unchanged, exit 1).
- **No:** report the file as a new report-only `unverified_catalog_timestamp` (file IDs rather than chunk IDs), with a summary counter that the CLI leaves out of the exit-code keys.

It needs no new storage API, just `iter_chunk_metadata(file_ids=...)`:

```python
async def _files_with_current_generation(vectors, collection, partition, expected, page_size) -> set[str]:
    """File IDs having at least one chunk stamped with the catalog's timestamp."""
    found: set[str] = set()
    pages = vectors.iter_chunk_metadata(collection, partition=partition, file_ids=list(expected), batch_size=page_size)
    async with aclosing(pages):
        async for page in pages:
            found.update(r["file_id"] for r in page if _timestamp(r.get("indexed_at")) == expected[r["file_id"]])
            if len(found) == len(expected):
                break
    return found
```

```python
elif existing[key] < cutoff and timestamp != existing[key]:
    suspects.setdefault(row["file_id"], []).append(str(row["_id"]))
...
if suspects:
    current = await _files_with_current_generation(
        vectors, collection, partition, {f: existing[partition, f] for f in suspects}, page_size
    )
    mismatches = [cid for f, ids in suspects.items() if f in current for cid in ids]
    unverified = [f for f in suspects if f not in current]
```

Cost and limits:

- On a healthy deployment it adds no reads, because the lookup only runs for files that already mismatch. On an old one, it re-reads the scalar fields of pre-migration files once, and memory stays bounded to one page.
- A file whose chunks span two pages would be looked up and counted twice. Remembering the previous page's verdicts should cover it, since one file's chunks are written in one batch.
- One rare case is misclassified: a re-index that produced 0 chunks *and* failed to delete the old ones would land in "unverified" instead of "mismatch". It is still reported and still never repaired.
- Repair is untouched: neither category is ever deleted.

Correcting the old catalog rows themselves (setting `files.indexed_at` from the chunks) is a separate, write-side fix. It would also correct the admin UI's "Indexed" column, which currently shows the migration date for those files. I'd keep it out of this PR as a follow-up.

What do you think?

Ahmath-Gadji: Pushed the fix in https://github.com/linagora/openrag/commit/b4e7209decfeac7010b64cf8c764a23bf1c7cc01 so you can check it directly.

It follows the proposal above:

- A timestamp mismatch is now reported only when the file also has chunks matching its catalog timestamp, i.e. the stale-chunk-set case.
- Files where no chunk matches are reported once as `unverified_catalog_timestamp` (file IDs), with a `unverified_catalog_timestamps` summary counter that the CLI leaves out of the exit-code keys.
- Verdicts carry over to the next page, so a file spanning a page boundary is looked up and reported once.
- The operator doc now describes both categories, the migration backfill and the zero-chunk re-index edge case.

Tests: stale chunks on an earlier page than the current set are still a mismatch; pre-migration files and old copies (including one spanning pages) come out as unverified, once each; the CLI returns 0 when only unverified timestamps are found. Both reconciliation tests fail against `57699f63`. The earlier reproduction goes from `timestamp_mismatches: 12` to `0`, nothing deleted. Full unit suite passes locally (3142).

Nothing in the repair path changed. Happy to adjust naming or anything else if you'd prefer a different shape.

Ahmath-Gadji (approved):
All earlier findings are verified fixed, and the legacy timestamp-mismatch issue is addressed in b4e7209d. Tested against develop merged in: unit suite, ruff and layer guard all clean. Approving.

PR #944, opened 2026-09-11
fix(retrieval): exclude chunks absent from the catalog

### LINKED ISSUE #925: Retrieval can cite a chunk whose catalog row is gone: nothing on the read path checks the file still exists
## Summary

Retrieval reads Milvus and never consults the catalog, so a chunk that outlives its `files` row stays fully citable. #658 / #671 narrowed the window that produces such chunks by deleting vectors before catalog rows; nothing on the read path prevents one from being returned if it exists anyway.

Adding the check makes the failure mode impossible rather than rare, and it does so in one place.

## Evidence

**Nothing in the retrieval path touches the catalog.** `grep -rn "document_repo\|_repo\." openrag/core/retrieval/` returns nothing. Chunks are matched on the Milvus `partition` and `file_id` fields only (`services/storage/milvus_store.py:306,312`).

**The residual window after #671.** `services/workers/dispatcher.py:522-547` — Milvus `delete_by_filter` at 534, Postgres rows at 535-536, a second Milvus sweep at 539 that re-raises on failure. A process killed between 536 and 539 leaves chunks written by an in-flight indexer between 534 and 536 with no catalog row. The second sweep exists precisely because that interval is known to be raceable.

**The re-index path leaves stale chunks on failure**, logged as `"duplicates remain until reconciliation"` (`indexer_actor.py:283-287`, `pipeline_builder.py:309-313`) — a second source of chunks whose catalog state no longer matches.

## Proposal

A decorator on the searcher port, wired in DI, so no call site changes.

`RetrievalSearcher` (`core/retrieval/searcher.py`) is an ABC with four abstract methods — `search`, `multi_query_search`, `get_related_chunks`, `get_ancestor_chunks` — all four reached from `core/retrieval/retriever.py` (lines 95, 155, 205, 256, 265), and it has a single implementation, `VectorStoreSearcher` (`services/storage/vector_store_searcher.py:42`). Wrapping the port covers the chat pipeline, the raw `/search` route and the related/ancestor expansion in one place; the related and ancestor methods matter as much as `search`, since expansion can surface an orphan the initial query did not.

The check is one indexed lookup per call: collect the distinct `(partition, file_id)` pairs from the returned chunks and test them against `files`, which already carries `Index("ix_partition_file", "partition_name", "file_id")` (`services/persistence/schema.py:228`). Result sets are top-k sized, so this is a small `IN` over an existing index. Note the column names differ between the stores — Milvus `partition`, Postgres `partition_name`.

Three decisions worth settling in review:

- **Return fewer than `top_k` rather than over-fetching.** A silently padded result set is harder to reason about than a short one.
- **Fail closed when the catalog is unreachable.** Search already depends on Postgres — `AuthMiddleware` resolves partition membership from it on every request — so this adds no new dependency, and failing open would serve deleted documents exactly when the system is unhealthy.
- **No caching beyond the request.** A cache is staleness, and staleness is the bug being fixed.

A counter of dropped chunks would also give a live signal of how often the two stores actually disagree, which nothing currently measures.

## Related

- #658 — deleted files/partitions leaving orphaned queryable chunks (fixed by #671, delete ordering)
- #924 — detection and repair of the divergence; this issue is the read-path complement

hedhoud:
Chunks can outlive their catalog rows after interrupted writes, leaving deleted documents available to retrieval. Search and expansion now verify catalog membership before returning results, fail closed when the catalog cannot be checked, and count discarded orphan chunks. Results may contain fewer matches than requested.

This PR addresses retrieval protection only. Reconciliation and operator repair are in #943, which is based on this branch to reuse the catalog lookup.

Validated with 131 focused tests, lint and formatting checks.

Closes #925.

## Summary by CodeRabbit

- **New Features**
  - Search results now exclude chunks linked to documents no longer present in the catalog.
  - Filtering applies consistently across standard, multi-query, related-chunk, and ancestor-chunk searches.
  - Added monitoring for chunks removed because their documents are absent from the catalog.
- **Bug Fixes**
  - Deleted or unindexed document content is no longer exposed through retrieval results.
  - Catalog lookup failures are surfaced instead of silently returning potentially stale content.


openrag/services/storage/catalog_searcher.py
```diff
@@ -0,0 +1,34 @@
...
+        return await self._filter(await self._searcher.search(*args, **kwargs))
+
+    async def multi_query_search(self, *args, **kwargs) -> list[Chunk]:
+        return await self._filter(await self._searcher.multi_query_search(*args, **kwargs))
+
+    async def get_related_chunks(self, *args, **kwargs) -> list[Chunk]:
+        return await self._filter(await self._searcher.get_related_chunks(*args, **kwargs))
+
+    async def get_ancestor_chunks(self, *args, **kwargs) -> list[Chunk]:
+        return await self._filter(await self._searcher.get_ancestor_chunks(*args, **kwargs))
```
Ahmath-Gadji: **The two expansion methods are wrapped redundantly, and it doubles DB fan-out on the default chat path.**

`get_related_chunks` and `get_ancestor_chunks` already resolve their file ids from the catalog before they ever touch Milvus:

- `vector_store_searcher.py:148-154` -> `get_file_ids_by_relationship` -> `SELECT file_id FROM files WHERE partition_name=$1 AND relationship_id=$2` (`document_repo.py:703`)
- `vector_store_searcher.py:170-176` -> `get_ancestor_file_ids` -> recursive CTE over `files`

Both then call `query_chunks_by_filter({"partition": partition, "file_id": file_ids})`, so every chunk they return carries a `(partition, file_id)` that came out of `files` milliseconds earlier. `_filter` can only drop something inside a TOCTOU window -- not the durable-orphan case #925 describes.

That redundancy is not free. With the shipped defaults (`include_related: true`, `include_ancestors: true`, `reranker.top_k: 10`), `_expand_with_related_chunks` gathers up to 20 expansion calls per partition group, each of which now issues a second Postgres query. Multiply by `max_partition_concurrency: 16`, and on the multiQuery path by `k_queries: 3`, against `pool_max_size: 20`.

The pool has no acquire timeout: `asyncpg.Pool.fetch` is `async with self.acquire() as con` with nothing passed, and `command_timeout: 30` only starts ticking once a connection is held. So saturation surfaces as unbounded retrieval latency rather than an error.

Suggested fix: drop these two overrides and let them pass through. The path that genuinely surfaces an uncataloged chunk is `_fetch_surrounding` inside `search` / `multi_query_search`, which the decorator already covers and which `test_surrounding_orphan_is_removed_after_real_search_expansion` already proves.

openrag/services/storage/catalog_searcher.py
```diff
@@ -0,0 +1,34 @@
...
+        self._document_repo = document_repo
+
+    async def _filter(self, chunks: list[Chunk]) -> list[Chunk]:
+        if not chunks:
+            return []
+        keys = {(chunk.partition, chunk.document_id) for chunk in chunks}
+        # No fallback or cache: a catalog outage must never expose deleted files.
+        existing = await self._document_repo.get_indexed_documents(keys)
+        result = [chunk for chunk in chunks if (chunk.partition, chunk.document_id) in existing]
+        ORPHAN_CHUNKS_DROPPED.inc(len(chunks) - len(result))
```
Ahmath-Gadji: **Orphans are dropped with no log line -- only a global counter.**

This is the only place in the system that observes a live catalog/Milvus divergence, and it discards *which* `(partition, file_id)` diverged. An operator watching the counter tick learns that something is orphaned, not what, and #943's reconciliation has to rediscover it by full scan. A bounded, deduplicated `logger.warning` carrying the dropped keys would make the counter actionable at the moment the evidence is in hand.

Worth reflecting in the metric help text too: the same orphan is counted once per multiQuery sub-query, once per partition group, and again on the `allow_filterless_fallback` retry in `pipeline.py:118-125`. It measures drop *events*, not distinct orphan chunks.

openrag/core/ports/document_repo.py
```diff
@@ -24,6 +25,13 @@ class ContentClaimLease:
 class DocumentRepository(ABC):
     """CRUD operations for documents."""
 
+    async def get_indexed_documents(self, keys: Collection[tuple[str, str]]) -> dict[tuple[str, str], datetime]:
+        """Return existing (partition, file_id) keys and their indexing times.
+
+        Implementations must perform a fresh, batched lookup and propagate errors.
+        """
+        raise NotImplementedError
```
Ahmath-Gadji, line 34: **This is the only non-abstract method on the port.**

Every sibling here is `@abstractmethod`; this one raises `NotImplementedError` instead. Now that `CatalogSearcher` sits on every search, a repository implementation that misses the method fails at query time with a 500 rather than at construction. Making it abstract moves that failure to where the other twenty-odd port methods already put it.

openrag/services/storage/catalog_searcher.py
```diff
@@ -0,0 +1,34 @@
...
+        return result
+
+    async def search(self, *args, **kwargs) -> list[Chunk]:
+        return await self._filter(await self._searcher.search(*args, **kwargs))
+
+    async def multi_query_search(self, *args, **kwargs) -> list[Chunk]:
+        return await self._filter(await self._searcher.multi_query_search(*args, **kwargs))
```
Ahmath-Gadji: **`*args, **kwargs` on all four overrides erases the port's typed signature.**

`RetrievalSearcher` spells out four precise signatures; forwarding them untyped means a signature drift or a typo'd kwarg at a call site fails deep inside `VectorStoreSearcher` rather than at this boundary. This is the only searcher the app ever constructs, so it is the boundary worth keeping typed.

Two module conventions are also skipped here: `from __future__ import annotations` and a closing `__all__`, both present in every other module under `services/storage/`.

Ahmath-Gadji (commented):
Reviewed for critical bugs -- **I did not find one**. The mechanism is correct and the wiring is complete; four follow-ups below, one of which I would fix before merge.

What I verified as sound, so it does not need re-checking:

- **The `unnest` SQL pairs positionally.** Ran it against a live Postgres with a deliberately mismatched `('nope', '777f09a9-...')` pair, where that `file_id` exists under a different partition. Correctly excluded -- no cross-product bug, which is the failure mode this query shape usually has.
- **No hot-path scan regression.** On a 2M-row synthetic `files` table the planner picks a nested loop over `ix_partition_file` (~1 ms for 20 keys), not the seq scan that shows up on small tables.
- **Wiring covers every entry point.** Both `_searcher` and `_searcher_factory` are wrapped, `di/providers.py` delegates to the container, and nothing else in the tree constructs a `RetrievalSearcher`.
- **Indexing ordering is right.** `indexer_actor.py:127-141` stores chunks then writes the catalog row, so the drop window during indexing covers uncommitted chunks by design.
- **Fail-closed propagation is intact.** No `return_exceptions=True` on the search path, and the broad `except Exception` in `_expand_with_related_chunks` cannot mask an outage because the initial `search` raises first.
- 3024 unit tests pass; `ruff check`, `ruff format --check` and the layer-import guard are all clean.

Finding 1 is the only one I would hold the merge for, and it is an availability concern rather than a correctness one -- the fix is deletions, not new logic, so it is cheap to do now rather than later. One caveat on that call: I derived the pool arithmetic from config defaults and the asyncpg source, not from a loaded instance. If production runs `openrag-all` across many partitions with `retriever.type: multiQuery`, the headroom is thinner than my estimate.

openrag/services/storage/catalog_searcher.py
hedhoud: The extra database round trips and lack of an acquire timeout are real concerns, though we have not measured saturation under load. I would not remove the final checks without agreeing on the change in coverage.

I reproduced the deletion window with a controlled test: the relationship lookup returns a file ID, its catalog row disappears before the vector lookup returns, and the expansion then returns an orphan. The final check drops it. This does not make retrieval atomic with deletion, but it catches a case that the earlier lookup cannot. Issue #925 explicitly includes related and ancestor expansion in the final validation.

My recommendation is to retain that coverage and investigate batching or bounded waiting. Accepting the earlier catalog lookup as sufficient needs an explicit decision on that tradeoff.

openrag/services/storage/catalog_searcher.py
hedhoud: This fits #925: a counter shows that drift exists, while bounded logging of deduplicated dropped keys would let operators identify the affected documents. The log should cap the keys included and avoid flooding logs on repeated hits.

The metric wording should also clarify that it counts chunk-drop occurrences, not distinct orphan chunks; the same chunk can be counted again on another retrieval or retry. Both changes can be addressed directly in this PR without expanding its scope.

openrag/core/ports/document_repo.py
hedhoud: This should be abstract. Catalog validation is now a required capability on the retrieval path, so an incomplete repository should fail at construction rather than on a user's search request. This is an interface correction introduced by this PR and can be fixed here directly, with the repository contract tests updated as needed.

openrag/services/storage/catalog_searcher.py
hedhoud: Explicit signatures would preserve the port's parameter contract and make invalid arguments easier to catch at this boundary. This can be corrected directly in this PR without changing retrieval behavior. The annotations import and explicit export are also reasonable consistency cleanups; I would treat those as nonblocking.

openrag/services/storage/catalog_searcher.py
hedhoud: Fixed in 28525f8b. Warnings now include a bounded sample of deduplicated partition/file keys and are limited to once per minute per process. The counter still records every drop, and its help text now makes repeated counting explicit. Tests cover payload bounds and continued catalog checks and counting while logs are throttled.

openrag/core/ports/document_repo.py
hedhoud: Fixed in 28525f8b. The lookup is now abstract, with a regression test confirming that a repository without it cannot be constructed.

openrag/services/storage/catalog_searcher.py
hedhoud: Fixed in 28525f8b. All four overrides now match the port's explicit signatures and retain the requested arguments when forwarding. The annotations import and explicit export are included. The full unit suite passed: 3,028 tests; lint, formatting and the layer import guard also passed.

EnjoyBacon7 (approved):

openrag/services/storage/catalog_searcher.py
Ahmath-Gadji: Agreed on keeping the coverage — the deletion window you describe is real, and #925 does name related and ancestor expansion explicitly. Accepting the earlier lookup as sufficient was my suggestion, not a defect, and it is your call to make; nothing here blocks the merge.

One consequence of keeping those two overrides, for the record rather than as a change request. The catalog round trip they now perform lands inside `_safe_related` / `_safe_ancestors` (`retriever.py:256-275`), which catch `Exception` and return `[]`. I reproduced it with the inner expansion returning two valid, cataloged chunks and `get_indexed_documents` raising a statement timeout: both chunks disappear from the answer and `openrag_retrieval_orphan_chunks_dropped_total` moves by 0, because the exception short-circuits before `inc()`. The only trace left is the pre-existing generic `get_related_chunks failed` warning.

The direction is safe — context is lost, never a deleted file — and `search` / `multi_query_search` do not share it, since they propagate. But it does mean the failure mode that pool saturation would actually produce is invisible on the very metric this PR adds. Worth pairing a signal with the batching or bounded-wait work whenever that happens.

openrag/services/storage/catalog_searcher.py
Ahmath-Gadji: Verified on 28525f8b. The sample is capped at 10 deduplicated keys with both fields truncated to 128 chars, the throttle is process-wide so a second instance cannot reset the budget, and the catalog check and counter both keep running while logs are suppressed.

Since `_next_orphan_warning_at` is module state, I also checked it for test-order flakiness: the fixture pins `monotonic` and monkeypatch-restores the global, and repeated plus partial runs of `tests/unit/services/storage/` came out identical every time. The help text reads correctly as occurrences now.

openrag/core/ports/document_repo.py
Ahmath-Gadji: Verified on 28525f8b. `PgDocumentRepository` is the only concrete `DocumentRepository` on this branch and on current develop (d9b462b2), so nothing else loses the ability to be constructed, and `merge-tree` against develop is conflict-free.

The regression test is self-guarding as well — it builds its method list from `__abstractmethods__`, so dropping the decorator makes the class fully concrete and the test fails, rather than silently passing.

openrag/services/storage/catalog_searcher.py
Ahmath-Gadji: Verified on 28525f8b. I diffed all four overrides against `core/retrieval/searcher.py` parameter by parameter: names, order and defaults match the port exactly, and forwarding is by keyword into `VectorStoreSearcher`, whose parameter names match as well.

Every call site already passes keywords (`retriever.py:95,155,205,256,265` and `retrieval_service.py:353`), so no caller changes behaviour. The `__future__` import and `__all__` are in.

Ahmath-Gadji:
**Final round on `28525f8b` — nothing blocking. All four threads answered and resolved.**

The three follow-ups check out, each verified independently rather than taken on the commit message:

- **Abstract port method** — `PgDocumentRepository` is the only concrete `DocumentRepository` on this branch *and* on current develop (`d9b462b2`), so making it abstract breaks no other implementation, and `merge-tree` against develop is conflict-free.
- **Explicit signatures** — diffed all four overrides against `core/retrieval/searcher.py` parameter by parameter; names, order and defaults match the port, and every call site already passes keywords.
- **Bounded orphan logging** — sample capped and truncated as described, throttle survives new instances, catalog check and counter unaffected by suppression. Checked the module-global throttle for test-order flakiness; repeated and partial runs are deterministic.

Verification on the head commit: **3028 unit tests pass**, `ruff check` clean, 529 files already formatted, layer import guard OK, and all six CI checks green. No docs or Grafana dashboards reference the metric whose help text changed, so nothing is left stale.

The redundant-expansion thread stays a deliberate tradeoff rather than a fix, which is the maintainer's call to make — I've accepted it and recorded the one downstream consequence in that thread: because those two overrides are kept, a catalog timeout during expansion is swallowed by `_safe_related` / `_safe_ancestors` and the new counter stays at zero, so the failure mode pool saturation would actually produce is invisible on the metric this PR adds. Follow-up material for the batching investigation, not a merge hold.

PR #950, opened 2026-09-11
fix: retrieve non-casual chat messages by default

### LINKED ISSUE #949: SciFact: chat returns no retrieved sources for 71 claims despite available search candidates
### Context

SciFact sends scientific claims to OpenRAG to test whether the system can find supporting evidence. In the September 10, 2026 evaluation, chat returned no retrieved sources for 71 of 300 claims (23.7%). The requests completed successfully, so this was a silent retrieval failure.

### Exact cause

The chat request is processed in two stages. First, the query contextualizer classifies the message and creates search queries. Then OpenRAG searches the selected partition only when that result requires retrieval. For many claim-style messages, the contextualizer returned:

```json
{"requires_retrieval": false, "query_list": []}
```

That result is intended for greetings and casual conversation. It is wrong for an evidence claim. Because the query list was empty, the chat pipeline never called vector search or reranking and generated an answer with empty context.

The evidence is reproducible on `develop` (`044cc33e`): 70 of the 71 affected claims were classified this way. Directly sending the original claim to retrieval returned 10 chunks in every skipped attempt. One historical case was not reproduced, so the report does not claim that every original request had the same runtime decision.

### Simple example

The claim `Albendazole is used to treat lymphatic filariasis.` has expected document `1215116`.

- Direct search found document `1215116` at rank 1.
- Chat returned no retrieved sources and no error.
- The replay showed that chat skipped retrieval before search was called.

This is why the problem is not “the document is absent from the index.” The chat router decided not to search.

### Solution

PR #950 adds the opt-in metadata flag `require_retrieval: true` for evidence-verification and evaluation requests. If the router says retrieval is unnecessary, OpenRAG now searches the original input instead. The fallback is available for chat, streaming chat, and text completions and preserves existing partition, workspace, attachment, and temporal-filter restrictions.

The flag is disabled by default so greetings and casual conversation remain inexpensive. The contextualizer prompt is also clearer about treating factual and scientific claims as retrieval requests, but the explicit flag is the deterministic guarantee.

Example request:

```json
{
  "model": "openrag-scifact",
  "messages": [
    {"role": "user", "content": "Albendazole is used to treat lymphatic filariasis."}
  ],
  "metadata": {"require_retrieval": true}
}
```

The flag guarantees that retrieval is attempted. It does not guarantee that a relevant document exists, that it ranks highly, or that the LLM cites it.

### Validation

With the flag enabled, a 300-claim replay attempted retrieval for all 300 requests, returned sources for all 300, and recovered sources for all 71 previously empty cases. The fix passes 722 relevant tests. Retrieval relevance and citation quality are reported separately because they are distinct problems.

The implementation is available in draft PR #950.

### Expected behavior

Evidence-verification requests must not silently bypass retrieval. Responses should distinguish an attempted search with no matching evidence from a request that intentionally skipped retrieval.

hedhoud:
## Context

Factual chat messages are often statements rather than questions. Partition-backed chat must retrieve evidence by default instead of relying on a contextualizer to opt in.

## Problem

The previous behavior could skip retrieval for factual statements. A strict casual allowlist also made natural greetings retrieve irrelevant documents, while Chainlit could display retrieval candidates the answer had not cited.

## Expected behavior

Standard chat, streaming chat, and Chainlit now share one retrieval policy. Common casual messages are recognized immediately; other social-only phrasing may be classified conservatively by the contextualizer. Factual, mixed, ambiguous, and malformed results retrieve by default, with the original message used when no usable query is returned.

Explicit retrieval, attachments, web search, and map-reduce still take priority. Casual replies use a dedicated OpenRAG response in the detected language, with English as the low-confidence or empty-message fallback. Only greetings introduce OpenRAG.

Chainlit displays strict cited sources for current responses. Text completions retain their contextualizer-controlled retrieval default.

## Upgrade note

Existing deployments keep the contextualizer prompt already stored in their database. Retrieval remains safe because the default is enforced in code, but operators should refresh that stored prompt from the bundled template to receive the improved greeting, gratitude, farewell, and capability responses.

## Validation

- 3,346 unit tests pass.
- Lint, formatting, and architecture checks pass.
- CodeRabbit reports no findings on the final diff.

Fixes #949.

## Summary by CodeRabbit

- **New Features**
  - Added a `require_retrieval` option to explicitly force retrieval for supported chat and completion requests.
  - Casual greetings, thanks, farewells, and capability questions now receive concise responses without unnecessary retrieval.
  - Casual responses can follow the detected language, including English and French.
  - Retrieval decisions better distinguish casual messages from factual requests.

- **Bug Fixes**
  - Source displays now prioritize cited sources while supporting legacy source data.

- **Documentation**
  - Updated API guidance and examples for retrieval behavior and metadata options.


openrag/core/models/query.py
```diff
@@ -103,6 +103,10 @@ class SearchQueries(BaseModel):
     """Collection of sub-queries produced by query decomposition."""
 
     query_list: list[Query] = Field(..., description="Search sub-queries to retrieve relevant documents.")
+    intent: Literal["greeting", "gratitude", "farewell", "capability", "other"] = Field(
+        default="other",
+        description="Conservative classification of the complete latest user message.",
+    )
```
Ahmath-Gadji: When asking the LLM to generate queries. We ask it to do a lot of task as the same time. See the prompt [file](https://github.com/linagora/openrag/blob/develop/openrag/prompts/templates/query_contextualizer_tmpl.txt). I wonder another isn't gonna derail the LLM further

openrag/app_front.py
```diff
@@ -733,13 +733,7 @@ async def on_message(message: cl.Message):
...
                     extra = json.loads(extra)
                 if extra:
-                    if "cited_sources" in extra:
-                        # Strictly what the model cited; fall back to
-                        # everything it was shown when nothing was cited
-                        # (e.g. it didn't report citations at all).
-                        sources = extra["cited_sources"] or extra.get("presented_sources")
-                    elif "sources" in extra:
-                        sources = extra["sources"]
+                    sources = extra.get("cited_sources")
```
Ahmath-Gadji, line 736: Agreed — and `extra` already carries the field that makes this decidable, so we don't have to choose between the two failure modes.

`cited_sources` is `[]` in two different situations:

- the model emitted a `[Sources: ...]` tag and genuinely cited nothing — hiding the sources here is correct, and is what this PR is fixing;
- the model never emitted the tag at all — here `[]` means *"no report"*, not *"no sources"*, and blanking the panel throws away real information.

`_build_extra_payload` already separates them (and `stream_with_source_filtering` mirrors it):

```python
"cited_sources": filtered if citations is not None else [],
"citations_reported": citations is not None,
```

and `filter_sources_by_citations` falls back for exactly the second case — *"if no `[Sources: ...]` tag is found at all, every presented source is kept instead (a missing tag means the model didn't report citations, not that it used none)"*.

So rather than the old `extra["cited_sources"] or extra.get("presented_sources")` (which also falls back in the **first** case, restoring the uncited-candidates bug this PR is fixing) or the current strict `extra.get("cited_sources")` (which loses the second), gate on the flag:

```python
sources = extra["cited_sources"] if extra.get("citations_reported") else extra.get("presented_sources")
```

That keeps the fix and keeps the safety net.

Worth noting `sys_prompt` is admin-editable through the prompt library, so an operator edit that drops the citation instruction would silently empty the Chainlit source panel under the current code. On a live stack running this branch, Mistral-Small-3.2-24B emitted the tag on every request I sent (`citations_reported: true` each time), so this is latent rather than live — but it's a one-line difference either way.

Separately: `CLAUDE.md` still documents the old contract — *"Chainlit is expected to move to this field, falling back to `presented_sources` in its UI when `cited_sources` is empty"* — worth updating so the doc matches whichever behaviour lands.

Ahmath-Gadji (commented):
I deployed this branch on a live stack to check it (`7dd2fb18` built into the GPU compose stack, Mistral-Small-3.2-24B, real Milvus partition). The retrieval fix itself works — factual statements retrieve now, including the `Albendazole` case from #949, and that holds even against a stale contextualizer prompt because `intent` defaults to `"other"`. Nice.

One thing to sort out before merge, plus a note inline on `app_front.py`.

## The new contextualizer template never reaches an existing deployment

`PromptService.seed_defaults` is create-only:

```python
for prompt_type in _TYPE_TO_CONFIG_KEY:
    if await self._repo.get_default(prompt_type) is not None:
        continue
```

> Idempotent per type: if a default already exists for a type it is left untouched, so an admin's edits survive restarts.

So on any database that has booted before, the `query_contextualizer` default stays at the **old** template and the new one ships in the image unused. After deploying this branch onto a stack whose DB was seeded by the previous build:

```
prompt.resolve query_contextualizer <- default:default_query_contextualizer  length=5339
system[6113]: ... Output shape (return this JSON object, nothing else):
              {"requires_retrieval": <true|false>, "query_list": [...]}
```

5339 bytes and no `intent` in the output shape — while the bundled template sitting on the same image is 6448 bytes with 15 `intent` occurrences.

`_QUERY_JSON_HINT` lives in code, so it still gets the model to emit an `intent` and retrieval gating survives. This does **not** reintroduce #949. What breaks is the *classification*, because the stale template's examples contradict the hint:

| message | stale DB template | bundled PR template |
|---|---|---|
| `thanks a lot!` | `intent: greeting` | `intent: gratitude` |
| `What can you do?` | `intent: greeting` | `intent: capability` |

And `intent` doesn't only gate retrieval — it selects the reply through `build_casual_response_prompt`, where `greeting` is the only intent that emits the full introduction. Actual replies from the deployed branch:

- `thanks a lot!` → *"Hello! I'm OpenRAG, a RAG assistant developed by LINAGORA. You can index many document and media formats like PDFs, text files, office documents, images, audio, video, and more…"*
- `What can you do?` → the same full introduction
- `merci` → *"Avec plaisir ! Je suis ravi d'avoir pu vous aider…"* ✅

`merci` is the control: it's in `CASUAL_MESSAGE_INTENTS`, so it skips the contextualizer entirely and behaves correctly. Same intent, two code paths, two outcomes — which is what points at the seeding gap rather than at the model.

Both `_INTENT_INSTRUCTIONS["gratitude"]` and `["capability"]` say *"Do not repeat the OpenRAG introduction or list its capabilities"*, and both get overridden on every deployment that has booted before. Fresh installs are fine, which is probably why CI and local testing didn't surface it.

Options, roughly in order of preference:

1. **Move the intent examples into `_QUERY_JSON_HINT`.** It's in code, so it's always current, and it's already where the JSON contract is stated. Smallest change, no migration story.
2. **Version the seeded defaults** — store a content hash and re-seed when the bundled template changes *and* the row hasn't been edited by an admin.
3. At minimum, document that operators must delete the `query_contextualizer` default so it re-seeds. That's a silent, easy-to-miss upgrade step for a user-visible regression, so I'd rather not rely on it.

Related: CodeRabbit's note on the `Hello, how can you helpe me?` example is the same fault line — the bundled template teaches `greeting` for a message whose intent is `capability`, so even a freshly seeded deployment inherits that ambiguity.

Ahmath-Gadji:
Correction to my review above: **the seeding behaviour is not an issue, and I was wrong to frame it as something to sort out before merge.** Prompts are data, not code — `seed_defaults` leaving an existing default untouched is the intended contract, and a stored prompt outliving an image upgrade is the whole point of the prompt library. Please disregard the three "options" I listed; none of them should happen, and this isn't a blocker on the PR.

What survives from that investigation is a **release-note item**, not a code change. The bundled `query_contextualizer` template genuinely did improve here, and by design an existing deployment keeps serving the default already stored in its database — so operators who want the new intent classification have to refresh it deliberately.

Measured on a stack upgraded to this branch, where the DB was seeded by an earlier build:

| message | stored default | bundled template |
|---|---|---|
| `thanks a lot!` | `intent: greeting` | `intent: gratitude` |
| `What can you do?` | `intent: greeting` | `intent: capability` |

Retrieval gating is unaffected either way — `_QUERY_JSON_HINT` is in code and `intent` defaults to `"other"` — so #949 stays fixed on both. The only difference is which casual reply gets selected: a stale default answers `thanks a lot!` with the full OpenRAG introduction instead of a short acknowledgement.

Suggested block to nest under this PR's bullet in the release notes:

> [!NOTE]
> This release improves the bundled `query_contextualizer` prompt template, which now classifies casual messages by intent (greeting / gratitude / farewell / capability). Prompts are data: an existing deployment keeps the `query_contextualizer` default already stored in its database and will **not** pick up the new template automatically. Retrieval behaviour is correct either way. To get the improved classification, replace the content of the stored `default_query_contextualizer` prompt with the bundled template (`openrag/prompts/templates/query_contextualizer_tmpl.txt`, also at `/app/openrag/prompts/templates/` inside the image) — via the Admin Console's Prompts screen, or `PATCH /prompts/{id}`. Note that deleting the default is refused, so update it in place rather than expecting it to re-seed.

Happy to fold that into the PR description, or leave it here for whoever cuts the release.

openrag/prompts/templates/query_contextualizer_tmpl.txt
```diff
@@ -47,25 +52,34 @@ For exclusions, split into two sub-queries covering each remaining range. Never
...
 
 User: "How can you help me?"
-{{"requires_retrieval":false,"query_list":[]}}
+{{"intent":"capability","requires_retrieval":false,"query_list":[]}}
+
+User: "Thanks"
+{{"intent":"gratitude","requires_retrieval":false,"query_list":[]}}
+
+User: "Hello, how can you helpe me?"
+{{"intent":"greeting","requires_retrieval":false,"query_list":[]}}
```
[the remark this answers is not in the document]
hedhoud: Fixed in 63e236c4. The typo-tolerant help request is now classified as capability, and the regression test verifies that it selects the concise capability prompt without the greeting introduction.

openrag/app_front.py
hedhoud: Kept the strict cited_sources display intentionally. The reported Chainlit regression was that an answer without cited evidence still displayed retrieval candidates; falling back when citations_reported is false would restore that behavior. CLAUDE.md now documents this contract. Clients that need consulted candidates can still use presented_sources, while Chainlit shows cited evidence only.

hedhoud:
Review follow-up

The typo-tolerant help request is now classified as a capability question instead of a greeting. This selects the concise capability response and avoids showing the full OpenRAG introduction.

Chainlit intentionally continues to display cited evidence only. Falling back to every retrieved candidate when the model reports no citation would restore the original regression by showing unrelated documents. The internal documentation now states this contract clearly.

Existing deployments keep their stored contextualizer prompt. Retrieval remains safe because the default is enforced in code, but operators should refresh that prompt to receive the improved intent-specific casual responses.

The review correction is isolated in the second commit. All 3,346 unit tests and the static checks pass, and all review threads are resolved.

Ahmath-Gadji:
## Empty replies to `ok` / `Cool`, fixed in 4107a7b3

On `63e236c4`, `ok` came back empty 8/8 and `Cool` 6/8 (develop: 0/8 and 3/8). Neither is in `CASUAL_MESSAGE_INTENTS`, so both went to retrieval. With nothing relevant indexed, `sys_prompt` ran on an empty context and the model replied only `[Sources: none]`, which strips to `""`. `great` was classified as `greeting` and got the full introduction.

4107a7b3 is a template-only change: short acknowledgements now map to `gratitude`, and an `ok` that accepts an offer from the assistant still retrieves. On a fresh stack:

- `ok`, `Cool`, `great`: 21/21 short replies. Acknowledgements not listed in the template (`awesome`, `sounds good`, `très bien`, `noted`): 20/20.
- Factual and mixed messages, plus `ok` after *"Would you like the recommended dosages?"*: 20/20 retrieve.
- `hello`, `Hello, how can you helpe me?`, `What can you do?`, `thanks a lot!`, `bye`: unchanged.

Ahmath-Gadji (dismissed):
Approving. I checked `63e236c4` on a live stack: the #949 retrieval fix, casual intents, `require_retrieval`, and streaming citations all behave as described. Please apply the contextualizer change from my comment above so `ok` / `Cool` don't get empty replies.

Ahmath-Gadji (dismissed):

openrag/prompts/templates/query_contextualizer_tmpl.txt
```diff
@@ -1,15 +1,22 @@
...
 
+# Intent and retrieval rules
+- Retrieval is the default. If there is any uncertainty, use `intent: "other"`, set `requires_retrieval: true`, and return a search query.
+- Use `intent: "greeting"`, `"gratitude"`, `"farewell"`, or `"capability"` only when the complete latest user message consists exclusively of that social intent. For those messages, set `requires_retrieval: false` and return an empty `query_list`.
+- Greetings and salutations include brief wellbeing exchanges. Capability means a question about what the assistant can do. Understand common wording variations and obvious typing mistakes conservatively.
+- Gratitude also covers brief acknowledgements and positive reactions with no request attached, in any language: "ok", "okay", "cool", "great", "nice", "perfect", "got it", "d'accord", "super", "parfait", "👍". These are not greetings and not uncertain: use `intent: "gratitude"`.
+- Exception: when such a reply accepts an offer or answers a question from the assistant's previous turn (for example "ok" after "Would you like the dosage details?"), it is a request. Use `intent: "other"`, set `requires_retrieval: true`, and rewrite the accepted request as the query.
```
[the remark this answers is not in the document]
hedhoud: Fixed in bfd17e6f. The exception now applies only when an acknowledgement accepts factual, informational, analytical, or document-backed continuation. Replies to social, wellbeing, and feedback questions remain gratitude and skip retrieval. The same rule is present in the always-appended query hint so deployments with an older stored contextualizer are covered. The two-sided regression tests, full unit suite, lint, and formatting checks pass.

Ahmath-Gadji (approved):

EnjoyBacon7 (approved):

PR #951, opened 2026-09-13
fix(ci): pull minio from quay.io instead of Docker Hub

Ahmath-Gadji:
## The problem

MinIO removed the `minio/minio` repository from Docker Hub. Every CI job that
starts the Milvus stack now fails at image pull, **before a single test runs**:

```
pull access denied for minio/minio, repository does not exist or may require 'docker login'
```

This is not branch-specific. `api-tests` and `milvus-integration` fail on every
branch, develop included — develop just has not re-run CI since the removal
(its last run was 2026-09-11, and it was green).

## The fix

The same releases are published on quay.io, so the pinned tag is unchanged and
only the registry moves:

```diff
-  image: minio/minio:RELEASE.2024-12-18T13-15-44Z
+  image: quay.io/minio/minio:RELEASE.2024-12-18T13-15-44Z
```

Five compose files carry that image — the two under `tests/integration/` are the
ones the failing jobs use:

| file | used by |
|---|---|
| `tests/integration/api/api_run/docker-compose.yaml` | `api-tests` |
| `tests/integration/repos/docker-compose.yaml` | `milvus-integration` |
| `infra/compose/milvus/milvus.yaml` | local + deploy stacks |
| `infra/compose/milvus/milvus.named-volumes.yaml` | local + deploy stacks |
| `tests/load/workspace/docker-compose.yml` | load bench |

## Why this is safe

The quay image is byte-identical to the Docker Hub one it replaces. Pulled and
compared against a pre-removal Docker Hub copy still cached locally:

```
quay.io/minio/minio  sha256:1dce27c494a16bae114774f1cec295493f3613142713130c2d22dd5696be6ad3
      minio/minio    sha256:1dce27c494a16bae114774f1cec295493f3613142713130c2d22dd5696be6ad3
```

Same digest, same pinned release. No version bump, no configuration change —
CI going green is the whole of the verification.

## Summary by CodeRabbit

- **Chores**
  - Updated the MinIO container image source to Quay.io across deployment and test environments.
  - Kept the existing MinIO release version unchanged.


EnjoyBacon7:
Follow-up: the compose fix missed the Helm-deployed path. The bundled Milvus subchart's minio still defaulted to Docker Hub's minio/minio at the same pinned tag, with no override in values.yaml — so a Helm install/upgrade would hit the same pull error.

Added `milvus.minio.image.repository: quay.io/minio/minio` in values.yaml and fixed a stale README reference. Verified with `helm template` that the rendered image is `quay.io/minio/minio:RELEASE.2024-12-18T13-15-44Z`.

EnjoyBacon7 (approved):

PR #952, opened 2026-09-13
feat(partitions): give each embedder its own vector field (2/2, #762)

### LINKED ISSUE #762: Partition embedder/endpoint changes have no guardrails — can silently corrupt retrieval quality for already-indexed partitions

Ahmath-Gadji:
> [!IMPORTANT]
> **Review this after #939 merges.** Part 2 of 2 for #762, stacked on #939. It targets `develop`, so until #939 is in, the diff shown here carries #939's commits too; once it merges, what remains is this PR's own work.

Closes #762; follow-ups are in #958.

#939 makes an embedder swap visible; this makes it safe. Every embedder wrote into one shared `vector` field, so a partition that changed model searched two vector spaces in one index. Now each embedder owns a field and a search reads only its own, and a partition that changes embedder re-embeds its files into the new field before switching to it.

### Changes

- **A field per embedder.** `model_endpoints.vector_field` (`vector_<name>`, suffixed on collision) is allocated server-side when an embedder is created, is never settable by a client, and survives renames. SQL revision `c4e8f2a6b913` allocates one for each existing embedder and pins partitions on the `default` alias to the embedder that is default at upgrade time.
- **Fields are added to a live collection.** `ensure_vector_field` adds a nullable, HNSW-indexed field on an embedder's first write. Search skips null rows instead of scoring them as zero.
  *Milvus 3.0.1: 100+ concurrent searches with zero errors while a field was added. A null costs ~4 bytes a row; a zero placeholder costs a full vector plus an extra index.*
- **Reads and writes use the partition's embedder field (schema v3).** Indexing writes into it and every searcher reads from it; BM25 stays shared. The `vector` field is gone, with no fallback: an embedder without a field raises.
- **Migration 3** moves existing vectors into their embedder's field in place (partial upsert, no re-embedding), verifies, then drops `vector`. It refuses up front if any partition can't be routed, aborts if rows change mid-run, and is safe to re-run.
  *Milvus 3.0.1: a synchronous `create_index` on a new field over existing data never finished, so indexes build asynchronously; `IS NULL` is rejected on vector fields, so verification scans.*
- **A partition can move to another embedder.** `POST /partition/{partition}/embedder-swap` walks the partition file by file, embeds each chunk's stored text with the new embedder and writes it into that embedder's field. Nothing is re-parsed or re-chunked; chunk ids, text, metadata and BM25 vectors are untouched. The old field keeps serving searches until every file is done — the partition switches in the transaction that marks the swap complete. Progress is a row in `partition_embedder_swaps`, updated once per file, which doubles as the cancellation check and lets a resumed swap skip what it did. Uploads, metadata updates and copies into the partition are fenced out while it runs, and the target endpoint cannot be deleted or renamed under it.
  *Milvus 3.0.1: 500 rows partial-upserted into a field added to an already-sealed collection were all searchable 100 ms after the last write, exact top hit — the index is created on the swap's first write, and rows not yet in a built segment are brute-forced, so completion needs no waiting period.*
- **Vector fields stay out of chunk metadata**, and are refused as metadata keys so a metadata update can't overwrite them.
- **Deleting an embedder drops its field**, freeing one of Milvus's 10 vector-field slots.
  *Milvus 3.0.1: the drop works on a loaded collection without a reload. A write to a dropped field is silently stored in the dynamic field, so `ensure_vector_field` no longer caches which fields exist.*

### Breaking change

Collections below schema v3 are refused for search and indexing (`VDB_SCHEMA_MIGRATION_REQUIRED`), and tools that read Milvus's `vector` field directly stop working. To upgrade:

1. Run the SQL migrations (on startup, or `python -m services.persistence.migrations.run`).
2. Stop OpenRAG and, with `rdb` and `milvus` up, run `services/persistence/migrations/milvus/migrate.py --dry-run`, then without `--dry-run`.
3. Start OpenRAG.

Changing `embedder` through `PATCH /partition/{partition}` on a partition that holds indexed files now returns `409 EMBEDDER_SWAP_REQUIRED` — repointing the field without re-embedding would hide every indexed file. Use the swap endpoint, which re-embeds first; the edit is still accepted on an empty partition.

### Still open (#958)

`/search` and MCP search ignore the partition's embedder; new partitions still follow set-default; no dimension check on an existing field; docs for migration 3.

## Summary by CodeRabbit

* **New Features**
  * Added support for different embedders per partition with dedicated vector storage.
  * Added resumable embedder swaps with progress tracking and cancellation.
  * Added partition and model usage details, including indexed-file counts and embedder provenance.
  * Added UI indicators for embedder drift, re-embedding progress, upload pauses, and affected files.
  * Added safeguards and confirmation prompts when changing or deleting embedders with indexed data.

* **Bug Fixes**
  * Protected stored vector metadata from user-supplied values.
  * Improved embedder resolution and search routing across renamed endpoints.


Ahmath-Gadji:
Replaced by #974: the same work rebuilt on the current #939 (this branch still carried #939's commits from before its review fixes), plus fixes for the CodeRabbit findings left open here.

Issue #894, opened 2026-09-03
parse_timeout and marker_timeout share a default, so the #659 pool recycle never runs


andyne13:
## Problem

`parse_timeout` and `marker_timeout` both default to **3600**, and the parse-stage
bound starts *earlier* than Marker's own. The outer therefore always expires first,
so the only code path that recycles a wedged Marker pool (#659) is unreachable.

Three nested timeouts, all `3600`:

| # | Where | Value | Clock starts | Recycles the pool? |
|---|---|---|---|---|
| 1 | `services/workers/stages/parse.py:43` | `loader.parse_timeout` | parse-stage entry | no |
| 2 | `services/workers/parsers/marker_workers.py:326` (`_run_chunk`) | `loader.marker_timeout` | after slot acquisition | no |
| 3 | `services/workers/parsers/marker_workers.py:210` (`process_pdf`) | `loader.marker_timeout` | at child `submit()` | **yes** |

Only #3 recycles:

```python
except FuturesTimeoutError:
    # ... recycle the pool to reclaim the wedged worker's slot (#659)
    self.setup_mp(old_executor=current_executor)
```

Because #1 starts first and all three share a duration, #1 wins for any non-zero
slot wait, cancelling the chain before #3's handler runs.

This contradicts the documented intent in `conf/config.yaml`:

> Outer wall-clock bound (seconds) for a single file's parse stage. **marker and
> docling self-limit at marker_timeout/docling_timeout**; pymupdf has no internal
> timeout […]

`parse_timeout` is meant to be a backstop for PyMuPDF, with Marker self-limiting
first. The shared default inverts that.

## Impact

A wedged Marker worker is never reclaimed. It keeps its pool slot and the GPU, and
every subsequent document fails at the outer timeout. Only a manual restart recovers.

Observed in a production deployment on `v2.1.1`: a batch of PDFs (165 KB–2.2 MB)
submitted within ~400 ms of each other **all failed at exactly 3600 s**
(`duration_ms` 3600031–3600218). After a plain stack restart with unchanged
configuration, the same 40 documents indexed in **4 minutes** (~6 s/document) —
confirming the documents were fine and the pool was stalled.

The traceback shows the inner call was **cancelled from outside**, not timed out —
`asyncio.exceptions.CancelledError` propagating into
`TimeoutError: parse timed out after 3600s`, with no
`"Resetting ProcessPoolExecutor"` log line. The recycle never ran.

## Fix

Enforce the ordering invariant rather than documenting it:

[...]

PR #954, opened 2026-09-14
fix(marker): recycle worker pool before releasing slot on cancel/timeout

### LINKED ISSUE #723: Marker pool releases its worker slot while the child process is still running (cancellation path)
## Problem

`MarkerPool._process_chunk` returns the worker to the queue in a `finally:`, but the work it is guarding is not cancellable — so on cancellation the slot is freed while the child process is still busy.

`services/workers/parsers/marker_workers.py:339-346`:
```python
async def attempt(_i: int):
    worker = await self._queue.get()
    try:
        await self.ensure_worker_pool_healthy(worker)
        return await self._run_chunk(worker, file_path, page_range, label)
    finally:
        await self._queue.put(worker)          # slot released here
```
The actual parse runs at `:236`:
```python
result = await loop.run_in_executor(None, run_with_timeout)
```
`run_in_executor` cannot be cancelled — cancelling the awaiting coroutine raises in the caller while the OS thread keeps blocking and the child process keeps the GPU.

## Impact

The pool oversubscribes: N queue tokens but fewer than N actually-free processes. Subsequent chunks are dispatched to a worker still executing the previous parse, so throughput degrades and GPU memory pressure rises with no error surfaced.

This is the hole remaining **after** #659. That fix addressed the wedged-on-a-PDF case (timeout kills the child); this is the *caller-cancelled* case, which became materially more reachable when bulk job cancellation landed in #685.

## Fix

Don't release the slot until the underlying work has actually stopped: either run the parse under a cancellation-aware wrapper that awaits child termination before `queue.put`, or mark the worker dirty on cancellation and recycle it (`setup_mp`) instead of returning it to the pool.

EnjoyBacon7:
## Summary
- `_process_chunk` released the queue slot unconditionally, even when `_run_chunk` was cancelled or timed out.
- `run_in_executor` isn't cancellable, so the child could still be parsing when the slot went back, letting the next chunk dispatch onto a still-busy worker.
- Now the slot is only released after the worker's pool is recycled (`setup_mp`), which kills the stale child first.

Fixes #723

## Test plan
- [x] Added a unit test reproducing the race (fails on old code, passes with the fix)
- [x] Full unit suite passes (3122 tests)
- [x] ruff clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document processing reliability by isolating recovery to the affected worker slot.
  * Prevented timeouts or cancellations in one processing slot from disrupting other active parses.
  * Improved worker recovery after interruptions, including retries when reset operations are cancelled.
  * Workers with ordinary processing errors can return to service without unnecessary resets.
  * Prevented workers from being reused when recovery fails.

* **Tests**
  * Added coverage for slot-specific recovery and worker release scenarios.


EnjoyBacon7:
Addressed: _recycle_and_release now retries the reset with backoff and drops the slot instead of returning it if recycling never confirms (9311c54).

The two failing checks (api-tests, milvus-integration) are pre-existing infra breakage unrelated to this PR — MinIO pulled its image from Docker Hub, tracked separately in #951.

openrag/services/workers/parsers/marker_workers.py
```diff
@@ -339,13 +365,20 @@ async def _process_chunk(self, file_path: str, page_range: list[int] | None, lab
...
+                return result
             finally:
-                await self._queue.put(worker)
-                self.logger.debug(f"MarkerWorker returned to pool for {label}")
+                if completed:
+                    await self._queue.put(worker)
+                    self.logger.debug(f"MarkerWorker returned to pool for {label}")
+                else:
+                    self.logger.warning(f"MarkerWorker for {label} did not complete cleanly; recycling before reuse")
+                    asyncio.create_task(self._recycle_and_release(worker, label))
```
Ahmath-Gadji, line 402: **Critical: any failed chunk kills the parse running in the other slots.** *(Updated: reproduced on a live stack; one earlier claim corrected below.)*

`completed` is false for *every* failure, not only cancel or timeout. So an ordinary exception also triggers a full `setup_mp()`. That includes a parse error or a CUDA OOM, where the child has already stopped and there is nothing to reclaim. `_reset_worker_pool` also calls `setup_mp()` with no argument, which skips the `old_executor` guard from #894. With `marker_max_processes` slots sharing one `ProcessPoolExecutor`, each such failure kills whatever else is parsing on that actor.

**Seen live** (PR head `9311c548`, one L4, 1 actor × 3 slots, three image-only PDFs parsed together):

```
(8p)  torch.OutOfMemoryError: CUDA out of memory ...
      MarkerWorker for (8p) did not complete cleanly; recycling before reuse
      setup_mp: Resetting ProcessPoolExecutor (killing worker processes)
(9p)  concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
      MarkerPool PDF (9p) (...) failed (attempt 1/4) ... Retrying
```

The 9-page file was killed mid-parse by the 8-page file's recycle and used up a retry. In that run, 3 PDFs caused 4 failed attempts and 4 pool resets, and took 41–65 s instead of about 20 s each. I'll be upfront that OOM confounds this run: the 9-page parse also hit OOM on its own later. OOM is still exactly the ordinary exception that shouldn't recycle the neighbours.

**Simulation** (real `_process_chunk`, fake actor with a shared executor; file A always fails, file B is healthy):

| | B failed | pool recycles per run |
|---|---|---|
| this PR | **18 / 30 runs** | 7–8 |
| `develop` | 0 / 30 runs | 0 |

**Correction to my first version:** I said that when a chunk fails, `process_pdf` cancels the file's other chunks and each one recycles. Only chunks *holding a slot* at that moment recycle. Chunks waiting in `queue.get()` or sleeping in the retry backoff are cancelled outside the `try`. On a job cancel, those recycles can also be cancelled by Ray itself (see the comment on line 353).

**Suggestion:**
- Recycle only when the child may still be running (`CancelledError` / `TimeoutError`). Return the slot directly on an ordinary exception.
- When you do recycle, use the guarded form (`setup_mp(old_executor=...)` or a per-worker generation counter), so a burst of failures on one actor causes one reset, not one per failed chunk.

openrag/services/workers/parsers/marker_workers.py
```diff
@@ -329,6 +329,32 @@ async def _run_chunk(self, worker, file_path: str, page_range: list[int] | None,
...
+        try:
+            await retry_with_backoff(
+                lambda _i: self._reset_worker_pool(worker),
+                max_retries=self.config.loader.marker_max_task_retry,
+                base_delay=self.config.loader.marker_retry_base_delay,
+                task_description=f"MarkerWorker recycle after {label}",
+            )
+        except Exception:
+            self.logger.exception(f"MarkerWorker recycle after {label} failed; dropping its slot")
+            return
```
Ahmath-Gadji: **Critical: cancelling an indexing job can permanently drop a Marker slot. Enough cancels and PDF parsing stops completely.** *(Updated: reproduced on a live stack.)*

**How it happens:**
1. `DELETE /indexer/task/{id}` reaches `MarkerPool.process_pdf` as `ray.cancel(..., recursive=True)`.
2. The `_recycle_and_release` task is started from inside that cancelled Ray task, so its `setup_mp.remote()` call is cancelled along with it.
3. `retry_with_backoff` re-raises `TaskCancelledError` without retrying, and the `except Exception` here drops the slot.

**Live setup:** PR head `9311c548`, Ray 2.47.1, one NVIDIA L4, `marker_pool_size: 1`, `MARKER_MAX_PROCESSES=3` (3 slots), `marker_chunk_size: 10`. A partition uses an indexation preset with `parsing_strategy: marker`. I upload a 40-page PDF (4 chunks) and send `DELETE /indexer/task/{id}` as soon as a chunk gets a slot.

```
MarkerPool PDF [p0-9] (/tmp/tmp62k1a0sh.pdf) cancelled, cancelling Ray task
MarkerWorker for [p0-9] did not complete cleanly; recycling before reuse
MarkerWorker pool reset Ray task was cancelled
ray.exceptions.TaskCancelledError: The task ... is canceled from an actor ... before it executes.
MarkerWorker recycle after [p0-9] failed; dropping its slot
```

**Results:**
- It's a race, not every time: **3 of 12** cancelled chunks that held a slot lost it. The others recycled and were returned.
- Loss was cumulative. After the first drop, 3 PDFs uploaded together never ran more than **2** at a time.
- After the third drop the pool was empty. A new 4-page PDF (the same file parsed in 1.45 s earlier) **got no slot within 180 s**, and its task stayed in `SERIALIZING`. `_process_chunk` blocks forever on `self._queue.get()`, and only `marker_timeout` (3600 s) or a restart of the detached `MarkerPool` ends it.
- In the first run, the recycle didn't run at all, so the cancelled parse's child wasn't killed either. It only stopped with `FileNotFoundError` because the indexer had deleted its temp file. On this path the #723 fix doesn't reclaim the child, and it loses the slot on top.

**A second way to lose slots (not reproduced live):** a `MarkerWorker` restart. On Ray 2.47.1, a call to an actor still running `__init__` raises `ActorUnavailableError` after about 2 s. With the defaults this recycle gives up after roughly 22–28 s, and Marker's model reload in `init_resources` can take longer than that.

**Suggestion:**
- Don't drop a slot for good. Return it once a health check (`is_pool_broken`) succeeds, and keep retrying until then.
- Run the recycle outside the cancelled task's lineage, so a job cancel doesn't cancel it. For example, route it through a long-lived background loop on the pool instead of `create_task` from inside `attempt`.
- Treat `TaskCancelledError` from the reset call as "retry", not "failed".

The new tests use a single slot and never cancel through Ray, which is why CI doesn't catch this.

Ahmath-Gadji (commented):
This closes the #723 race, but it opens two worse problems. Both come from the pool layout: `marker_max_processes` slots on one actor share one `ProcessPoolExecutor`, so `setup_mp()` kills the child processes of **every** slot on that actor.

**Update:** both problems are now reproduced on a live stack (PR head `9311c548`, Ray 2.47.1, one L4, 1 actor × 3 slots, a partition using `parsing_strategy: marker`), not only in a simulation:
- **Line 353:** job cancellation dropped all 3 Marker slots, one at a time. Ray cancels the recycle along with the cancelled task. A new PDF then couldn't get a slot.
- **Line 381:** a CUDA OOM in one file's chunk recycled the pool and killed another file's parse mid-run (`BrokenProcessPool`).

Details, logs and one correction to my first version are inline.

openrag/services/workers/parsers/marker_workers.py
EnjoyBacon7: Fixed in 12cd4492. Ordinary exceptions (parse error, OOM) now return the slot directly instead of recycling `setup_mp()` unguarded; only cancel/timeout still recycle.

openrag/services/workers/parsers/marker_workers.py
EnjoyBacon7: Fixed in 12cd4492. `_recycle_and_release` now retries the reset a bounded number of times when its own call is cancelled, instead of treating that as a failed recycle and dropping the slot on the first hit.

Ahmath-Gadji (approved):
Approving. `12cd4492` fixes the slot loss, and I pushed `5d88db97` to fix the remaining collateral kill on cancel and timeout.

**What `5d88db97` changes:** each Marker slot now has its own single-process `ProcessPoolExecutor`. `MarkerPool` queues `(actor, slot)` pairs, so a cancel, timeout or broken-pool reset only kills the child that ran that slot's chunk. Before, every slot shared one executor, so a recycle killed every parse on the actor, including other files'. The number of child processes and GPU memory use are unchanged, and so is the external `MarkerPool.process_pdf(file_path)` call.

**Live results** (one L4, 1 actor × 3 slots, `parsing_strategy: marker`):

| | `12cd4492` | `5d88db97` |
|---|---|---|
| 12 cancels of a 40-page PDF | 0 slots lost (Ray-cancelled resets were retried) | 0 slots lost, 3 concurrent slots afterwards |
| Two 1-slot files parsing together, cancel one (3 rounds) | both children killed; the other file lost an attempt (37 s) | only the cancelled file's child killed; the other kept its PID and finished with no retry (25 s) |
| Other files cancelled during one file's parse | 4 cancels → that file `FAILED` | 6 cancels → it completed |

The unit tests in `test_marker_workers.py` now cover slot-scoped reset, health check, child timeout and pool calls.

**Not covered:**
- **Dropped slots:** a reset that keeps failing for real (not a Ray cancel), for example a worker restart slower than the retry budget, still drops the slot for good.
- **OOM:** the "return the slot without recycling on ordinary errors" path has a unit test but wasn't triggered live, because I couldn't force an OOM.

PR #955, opened 2026-09-14
fix: security headers on statically served admin UI

### LINKED ISSUE #927: Security headers are set by the API middleware but not on the statically served admin UI
## Summary

The API sends four security headers; the statically served admin UI sends none. They arrive by different paths, and only one of them goes through the middleware.

## Evidence

`SecurityHeadersMiddleware` is registered at `openrag/api/main.py:316` and sets `X-Content-Type-Options: nosniff`, `X-Frame-Options: SAMEORIGIN`, `Referrer-Policy: strict-origin-when-cross-origin`, and `Strict-Transport-Security` once the request is HTTPS.

`infra/compose/nginx/openrag-admin.conf` proxies `location /` to the API (line 68), but serves `/app/` and `/app/assets/` straight from disk with `alias`/`try_files` (lines 49, 56). Those responses never reach the middleware, and the file contains exactly **one** `add_header` — `Cache-Control`, at line 52.

So the single-page application's own document and its bundles carry no security headers at all.

The file path is misleading in the other direction: this config is baked into the image (`infra/docker/ui.Dockerfile:43`), so it ships to Kubernetes as well as Compose.

## Proposal

1. **Mirror the four headers in the nginx server block** so the statically served files match what the API already returns.
2. **Add a Content-Security-Policy.** The middleware omits one deliberately — its module docstring notes that the mounted admin UI and Chainlit need per-UI tuning — so this is new policy rather than a copied line, and it is where most of the effort sits.

EnjoyBacon7:
Fixes #927

nginx serves /app/ and /app/assets/ straight from disk, bypassing
SecurityHeadersMiddleware, so the SPA and its bundles carried no
security headers.

- Added a shared security-headers.conf with the same four headers
  as the middleware, plus a CSP tuned for the SPA (no inline
  scripts, inline style attributes used so style-src needs
  'unsafe-inline').
- HSTS only sent when the request is HTTPS, mirroring the
  middleware's behavior, via the existing forwarded-proto map.
- Included from both /app/ and /app/assets/ locations, and copied
  into the image in ui.Dockerfile (same file ships to Compose and k8s).

Verified with a local nginx container and the built image: headers
present on index.html and hashed assets, HSTS absent over plain
HTTP and present with X-Forwarded-Proto: https, config passes
`nginx -t`.

## Summary by CodeRabbit

- **Security**
  - Added consistent security headers to the admin web interface, including content protection, framing restrictions, referrer controls, and MIME-type protections.
  - Added HTTPS Strict Transport Security for secure connections, including forwarded HTTPS requests.
  - Added a Content Security Policy to restrict permitted content sources and improve browser-side protection.
  - Configured the policy to permit connections to the configured API origin.


infra/compose/nginx/security-headers.conf
```diff
@@ -0,0 +1,12 @@
...
+# SecurityHeadersMiddleware; these are served by nginx and never hit it).
+add_header X-Content-Type-Options "nosniff" always;
+add_header X-Frame-Options "SAMEORIGIN" always;
+add_header Referrer-Policy "strict-origin-when-cross-origin" always;
+add_header Strict-Transport-Security $hsts_header always;
+
+# The SPA build has no inline scripts (module script tag with an external
+# src) but does use inline style attributes, so style-src needs
+# 'unsafe-inline'. Everything else stays same-origin only.
+add_header Content-Security-Policy "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src 'self'; object-src 'none'; base-uri 'self'; frame-ancestors 'self'" always;
```
Ahmath-Gadji, line 12: **`connect-src 'self'` breaks the documented browser-direct build**

An absolute `VITE_API_BASE_URL` is a supported mode — `docs/content/docs/documentation/env_vars.md:702` says "Set to an absolute URL only for a browser-direct build — then list the UI's origin in `CORS_EXTRA_ORIGINS`", and `ui/src/lib/api/client.ts:77` carries `credentials: "include"` specifically for it. That SPA is still served by *this* nginx, so the CSP it ships blocks every API call: CORS configured exactly as documented, and the requests never leave the browser.

Measured with the real bundle (`vite build` with `VITE_API_BASE_URL=http://127.0.0.1:18090`), this branch's two config files in `nginxinc/nginx-unprivileged:1.27-alpine`, a mock API on the other origin sending `Access-Control-Allow-Origin` + `Access-Control-Allow-Credentials` and answering preflight, loaded in headless Chromium:

| | requests reaching the API | result |
|---|---|---|
| this branch | **0** | stuck on the login screen |
| same bundle, parent commit | **19** | logged in, console loaded |

```
[error] Connecting to 'http://127.0.0.1:18090/users/info' violates the following
        Content Security Policy directive: "connect-src 'self'". The action has been blocked.
```

The failure mode is the expensive part: signing in with a *valid* token renders **"That token was rejected. Check it and try again."** while zero requests leave the browser. That sends the operator into `AUTH_TOKEN`, `CORS_EXTRA_ORIGINS` and the IdP, not into a CSP baked into the admin-ui image.

The default same-origin build is unaffected — same check with `VITE_API_BASE_URL=""` and nginx proxying `/`: zero CSP violations, 10 API calls through, console rendered, fonts and inlined `data:` assets fine. So this is only about that one documented mode.

Options, cheapest first:

- document it beside the `VITE_API_BASE_URL` note — an operator can already bind-mount over `/etc/nginx/security-headers.conf` without rebuilding the image (verified);
- widen `connect-src` from the same build arg that sets `VITE_API_BASE_URL`;
- move the file to `/etc/nginx/templates/` so the base image's envsubst fills a `CSP_CONNECT_SRC` at container start.

infra/compose/nginx/openrag-admin.conf
```diff
@@ -14,6 +14,14 @@ map $http_x_forwarded_proto $forwarded_scheme {
     ''      $scheme;
 }
 
+# HSTS only makes sense over HTTPS (browsers ignore it on plain HTTP anyway),
+# matching api.middleware.security_headers. An empty add_header value is
+# dropped by nginx, so this keeps local HTTP access header-free.
+map $forwarded_scheme $hsts_header {
+    https   "max-age=63072000; includeSubDomains";
+    default "";
+}
```
Ahmath-Gadji, line 25: **HSTS diverges from the middleware on a chained `X-Forwarded-Proto`**

`apply_security_headers` takes the *first* token (`xfp.split(",", 1)[0]`, `openrag/api/middleware/security_headers.py:34`); this map matches the whole header value. Behind two proxies the two disagree, so the API sends HSTS and the SPA doesn't. Measured against this branch's config:

| `X-Forwarded-Proto` | nginx (this branch) | middleware |
|---|---|---|
| `https` | HSTS | HSTS |
| `HTTPS` | HSTS | HSTS |
| **`https, http`** | **no HSTS** | **HSTS** |
| `http, https` | no HSTS | no HSTS |
| `http` | no HSTS | no HSTS |

Using `~*^https(,|$)` as the key matches the middleware on all five (verified in the same container), and the `~*` keeps the case-insensitivity the plain key already had:

```nginx
map $forwarded_scheme $hsts_header {
    ~*^https(,|$)  "max-age=63072000; includeSubDomains";
    default        "";
}
```

Minor either way — it only shows up behind more than one proxy.

Ahmath-Gadji:
Reviewed this end to end and ran the config in a real `nginx-unprivileged:1.27-alpine` container plus headless Chromium against the actual SPA bundle. Two inline comments on the substantive points (`connect-src` and the HSTS map); the rest is collected here.

First, the part that's easy to miss: #927 proposed mirroring the headers **in the server block**, and this PR deliberately didn't. I tested that variant — `location /app/assets/` has its own `add_header Cache-Control`, so nginx drops every inherited `add_header` and all five headers vanish on every hashed bundle, silently. The per-location `include` is the only thing that works here. Also confirmed: headers present on `/app/`, on hashed assets and on the `try_files` fallback; HSTS absent over plain HTTP and present with `X-Forwarded-Proto: https`; proxied paths (`/v1`, `/auth`, `/chainlit`) unaffected, so Chainlit keeps its own headers.

**No test, though the guard file already exists.** `tests/unit/infra/test_admin_ui_compose.py` exists precisely to assert nginx-conf/Dockerfile invariants — upload limit, `Host` header, websocket upgrade, `USER 10001:0`. Two regex asserts (both static locations carry the include; `ui.Dockerfile` copies the file) would fit that pattern and catch the silent regression: a static location added later without the include serves no headers and nothing complains.

**`security-headers.conf` depends on `$hsts_header` without saying so.** Including it from a server block that doesn't define the map aborts startup:

```
nginx: [emerg] unknown "hsts_header" variable
```

Loud rather than silent, so just worth a line in the file's header comment.

**Remote images in rendered markdown are now blocked.** `img-src 'self' data:` stops third-party image loads; `src/pages/admin/documents/detail.tsx:94` renders chunk content through `react-markdown`, so a markdown document carrying `` renders a broken image. Verified in the browser (remote blocked, `data:` and same-origin fine, nothing reached the remote host). Probably what we want — no tracking pixels from uploaded content — but worth a line in the PR description or the docs.

**Nits.** The `= /` and `= /app` 301s carry no headers (harmless, browsers always follow them). Pre-existing and adjacent: `expires 1y` plus `add_header Cache-Control` emits two `Cache-Control` headers on every asset.

**CI.** Both failures are stale-base fallout, not this PR — the branch is 59 commits behind `develop` and the jobs die on `pull access denied for minio/minio`. `develop` already moved to `quay.io/minio/minio`. A rebase should green both and also unblock CodeRabbit, which skipped the review because the stale base makes the diff look like 180 files.

infra/compose/nginx/security-headers.conf
EnjoyBacon7: Fixed in cc328f71: connect-src now allows the API origin extracted from VITE_API_BASE_URL at build time when it's absolute. Verified with the actual bundle and both HTTP status paths.

infra/compose/nginx/openrag-admin.conf
EnjoyBacon7: Fixed in cc328f71: switched to the ~*^https(,|$) regex you suggested. Verified against https, https/http and http/https chains, now matches the middleware on all of them.

infra/compose/nginx/security-headers.conf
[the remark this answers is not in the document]
EnjoyBacon7: Not changing this: blocking remote images in ingested-document previews is intended, not an oversight, since it stops arbitrary third-party origins (tracking pixels, exfiltration) from loading inside the admin UI. No approved external source is currently in use.

Ahmath-Gadji (approved):

PR #956, opened 2026-09-14
feat(config)!: refuse published default credentials at startup

andyne13:
Values this project publishes in its examples, chart and test stacks are now refused at
startup and at `helm template` time. The canonical list lives in
`openrag/core/config/secrets_guard.py`, and a test fails the build if the chart's copy
disagrees.

**Upgrade:** `uv run python scripts/gen_env.py` generates the credentials, or set
`ALLOW_INSECURE_SECRETS=true` for a disposable development or CI stack.

Two things worth knowing when reviewing:

- `cp .env.example .env` no longer produces a working stack. That is the point — the
  example ships `__GENERATE_ME__` placeholders, the generator fills them, and the marker
  is itself refused so a copied template fails at startup naming what is unset.
- The API test stack sets `ALLOW_INSECURE_SECRETS=true`, since its credentials are
  published in its own compose file. The guard's tests assert it still trips.

## Summary by CodeRabbit

- **New Features**
  - Added a tool to generate secure environment files with randomized credentials.
  - Added startup validation that rejects published, placeholder, or too-short credentials.
  - Added an optional warning mode for insecure credentials.

- **Improvements**
  - Updated setup instructions and templates to require generated secrets.
  - Documented secret requirements, validation rules, metrics authentication, and generation options.
  - Improved Helm validation for unsafe and insufficiently long credentials.

- **Tests**
  - Added coverage for credential generation, validation, warning behavior, configuration precedence, and Helm policy consistency.


scripts/gen_env.py
```diff
@@ -0,0 +1,181 @@
...
+    if args.output.exists() and not args.force:
+        # Fill the gaps in what is already there rather than rebuilding from
+        # the template: an existing .env holds endpoints and tuning the
+        # operator set by hand, and regenerating a credential that data was
+        # already written under (a database password, say) breaks the stack.
+        text, filled = fill(args.output.read_text())
+        _write_private(args.output, text)
+        print(f"Updated {args.output}; generated {len(filled)} missing value(s): {', '.join(filled) or 'none'}")
+
+        missing = sorted(set(_assignment_keys(args.template.read_text())) - set(_assignment_keys(text)))
```
hedhoud: This detects missing template variables but only warns and exits successfully. I reproduced it by removing `AUTH_TOKEN` from a generated environment: the rerun left it absent, then `--check` also returned success. That breaks the documented upgrade path and may leave a newly required credential unset. Please preserve existing values while adding and generating missing template assignments, make `--check` validate template/output parity, and cover both paths with a regression test.

scripts/gen_env.py
```diff
@@ -0,0 +1,181 @@
...
+    """Write ``text`` to ``path`` so the credentials are never briefly readable.
+
+    ``write_text`` would create the file at 0644 under a 022 umask and only
+    narrow it afterwards, and on an overwrite it would put credentials into an
+    already-permissive file. Write a 0600 temporary file alongside the target
+    and rename it over instead.
+    """
+    path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = path.with_name(f".{path.name}.tmp")
+    fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, stat.S_IRUSR | stat.S_IWUSR)
```
hedhoud: The predictable temporary path can already exist as a symlink, and `O_TRUNC` follows it. I reproduced generated secrets being written into the symlink target with mode `0644`, while the final output remained a symlink. Please create a unique or exclusive same-directory temporary file that cannot follow links, then keep the atomic replacement and cleanup behavior. A regression test should cover a pre-existing temporary path.

hedhoud (changes requested):
The runtime guard is well centralized, and the targeted tests pass, but the credential setup workflow still has fail-open paths. A normal rerun and `--check` can both succeed while a credential newly added to the template is absent, and the temporary-file write can follow a pre-existing symlink and expose generated secrets with broader permissions. I reproduced both cases locally. The previously reported Helm issue accepting whitespace-only required secrets is also still unresolved.

infra/charts/openrag-stack/templates/secrets-env.yaml
```diff
@@ -37,8 +55,14 @@ stringData:
...
   {{- end }}
-  {{- if has $stringValue $placeholderSecrets }}
-  {{- fail (printf "env.secrets.%s still uses an unsafe placeholder value" $key) }}
+  {{- if has (lower (trim $stringValue)) $placeholderSecrets }}
+  {{- fail (printf "env.secrets.%s is set to a value published by this project as an example or development default; supply a generated value" $key) }}
+  {{- end }}
+  {{- /* Trimmed, and blank reads as unset, so this matches the runtime guard
+     exactly: requiredness is the check above's job, and an optional secret
+     left blank is not a length failure. */}}
+  {{- if and (not (empty (trim $stringValue))) (has $key $lengthCheckedSecrets) (lt (len (trim $stringValue)) $minSecretLength) }}
```
hedhoud, line 67: The required-secret check above still uses the untrimmed value, while this condition skips whitespace-only values after trimming. As a result, whitespace-only `AUTH_TOKEN` and `POSTGRES_PASSWORD` values pass chart rendering. Please trim values for the requiredness check too and add a regression test.

scripts/gen_env.py
andyne13: Confirmed and fixed in `395641cd`.

Both halves were real, and both reported success:

- **Re-run:** filled placeholders in the existing file, warned on stderr about template variables the file lacked, `return 0`. The documented "add a secret to the example, re-run" path left the credential unset while the exit code said otherwise.
- **`--check`:** scanned only for leftover placeholders. A variable absent outright holds none, so it passed too — the blind spot from the other side.

`add_missing()` now appends the template assignments the file lacks, generating placeholder values, under a comment marking where they came from. Existing assignments are never rewritten, so an operator's endpoints and tuning survive and a credential the data was written under is not regenerated. `--check` now validates template/output parity as well as placeholders, and says so in its success line.

Six tests in `tests/unit/scripts/test_gen_env.py`, mutation-checked: disabling the parity check or the add-missing call each fails a different one.

scripts/gen_env.py
andyne13: Confirmed and fixed in `395641cd`. Good catch — this was the serious one.

`O_CREAT` without `O_EXCL` opens an existing path and follows a symlink, and the `0600` mode argument applies only when `open` actually creates the file. So a pre-planted `.env.tmp` meant the credentials were written through the link into a target that kept its own permissions, exactly as you reproduced.

Now `tempfile.mkstemp(dir=path.parent, ...)`: a unique name opened `O_EXCL` at 0600, so the open cannot land on a path someone else prepared. The atomic `os.replace` and the cleanup-on-failure are unchanged — and worth noting `os.replace` swaps the *name*, so a symlink at the destination is replaced rather than written through, which the second new test pins.

Three tests cover it: a planted symlink at the temp path, a symlinked destination, and no temp file left behind. Mutation-checked — restoring the old `os.open` fails the first.

infra/charts/openrag-stack/templates/secrets-env.yaml
andyne13: Confirmed and fixed in `dd06bbdc`. The requiredness check now trims.

Worth recording why nothing else caught it: the runtime guard has the same blind spot from the other direction — `secrets_guard.py` does `value = _resolve(...).strip()` then `if not value: continue`, so it treats whitespace as unset and moves on without complaint. Chart rendering succeeded, boot succeeded, and the credential was absent at every layer.

To be precise about the blast radius, since "auth token unset" sounds worse than it is: this does **not** silently disable authentication — the no-auth bypass needs an explicit `ALLOW_NO_AUTH=true`. What breaks is the chart's own stated contract, "must be set before installing the chart", for a value nothing downstream will honour. A CI substitution rendering to spaces is the plausible route in.

Regression test added in `test_helm_security_hardening.py`; it fails if the check reverts to the untrimmed value.

openrag/core/config/secrets_guard.py
```diff
@@ -0,0 +1,274 @@
...
+    environ = os.environ if env is None else env
+    problems = find_insecure_secrets(settings, environ)
+    if not problems:
+        return
+
+    detail = " ".join(problems)
+    if environ.get(ALLOW_INSECURE_SECRETS_ENV_VAR, "").strip().lower() == "true":
+        # Warned on every boot, deliberately: a stack running with this flag
+        # should never be able to claim it did not know.
+        logger.warning("%s=true: starting with insecure configuration. %s", ALLOW_INSECURE_SECRETS_ENV_VAR, detail)
```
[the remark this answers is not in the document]
andyne13: False positive, now pinned by tests rather than argued.

`find_insecure_secrets` reads each value to decide *whether* to complain, then builds the message from the variable **name** and a static purpose string — the value is never interpolated:

```python
problems.append(
    f"{spec.env_var} is shorter than {MIN_SECRET_LENGTH} characters. "
    f"Supply a generated value ({spec.purpose})."
)
```

So the dependence between the secret and `detail` is control, not data, which is what the taint tracking over-approximates here.

That said, the alert landed on the one path where the invariant was not asserted. `test_enforce_never_echoes_the_value` covered the `raise`; the warning path only checked that the flag's name appeared in the log. That path is the one that matters more — it logs on **every boot**, into whatever ships the logs, rather than once into a crash message.

`dfc13c35` adds two tests: one on the warning path, one on `find_insecure_secrets` itself so the invariant holds for every consumer rather than the two call sites below it. Both are mutation-checked — interpolating the value into either message fails both.

Worth recording one trap for anyone writing a similar assertion: the first version used `"short"` as a test value and failed, because `"short"` is found inside `"shorter than"`. Canary values, not words.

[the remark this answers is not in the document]
andyne13: False positive, now pinned by tests rather than argued.

`find_insecure_secrets` reads each value to decide *whether* to complain, then builds the message from the variable **name** and a static purpose string — the value is never interpolated:

```python
problems.append(
    f"{spec.env_var} is shorter than {MIN_SECRET_LENGTH} characters. "
    f"Supply a generated value ({spec.purpose})."
)
```

So the dependence between the secret and `detail` is control, not data, which is what the taint tracking over-approximates here.

That said, the alert landed on the one path where the invariant was not asserted. `test_enforce_never_echoes_the_value` covered the `raise`; the warning path only checked that the flag's name appeared in the log. That path is the one that matters more — it logs on **every boot**, into whatever ships the logs, rather than once into a crash message.

`dfc13c35` adds two tests: one on the warning path, one on `find_insecure_secrets` itself so the invariant holds for every consumer rather than the two call sites below it. Both are mutation-checked — interpolating the value into either message fails both.

Worth recording one trap for anyone writing a similar assertion: the first version used `"short"` as a test value and failed, because `"short"` is found inside `"shorter than"`. Canary values, not words.

scripts/gen_env.py
```diff
@@ -0,0 +1,241 @@
...
+            absent = sorted(set(_assignment_keys(args.template.read_text())) - set(_assignment_keys(output_text)))
+            if absent:
+                print(
+                    f"{args.output} is missing variables the template defines: {', '.join(absent)}. "
+                    f"Re-run without --check to add them.",
+                    file=sys.stderr,
+                )
+                problems = True
+        else:
+            print(f"Template not found: {args.template}; checked placeholders only.", file=sys.stderr)
```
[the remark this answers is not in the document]
andyne13: Right, and fixed in `7e4409e4`.

This is the same defect the commit above it fixed — a verification path reporting success for a check it did not perform — reintroduced inside that fix. `--check` was handed a template it could not read and still exited 0, having skipped the parity half entirely.

It now fails, with a message that says what could not be verified rather than what was. That also makes it consistent with the generate path, which already refuses a missing template.

Mutation-checked: dropping the flag fails the new test.

hedhoud (dismissed):
Re-reviewed the current head against the requested fixes. Generator reruns now add missing template variables without replacing existing credentials,  validates template parity and fails when the template is unavailable, temporary credential writes use unique exclusive files before atomic replacement, and Helm rejects whitespace-only required secrets. The new regression coverage passes, as do Ruff and current-head CI. The remaining test-isolation suggestion is non-blocking and does not affect production behavior.

tests/unit/scripts/test_gen_env.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_generated_file_is_private(tmp_path: Path) -> None:
+    template = _write(tmp_path / ".env.example", "AUTH_TOKEN=__GENERATE_ME__\n")
+    output = tmp_path / ".env"
+
+    gen_env.main(["-t", str(template), "-o", str(output)])
+
+    assert stat.S_IMODE(output.stat().st_mode) == 0o600
+    assert os.environ.get("AUTH_TOKEN") is None  # the script writes, it does not export
```
hedhoud: Non-blocking: this assertion depends on `AUTH_TOKEN` already being absent from the caller environment. Please inject a sentinel value with `monkeypatch` before invoking the generator and assert that the sentinel is unchanged. That proves the script does not export or mutate process state while keeping the test reliable on developer machines and CI. This does not affect production behavior, so my approval stands.

andyne13: Fixed in `c8974c2c` — and it was worse than non-blocking, so thanks for catching it.

Reproduced first: `AUTH_TOKEN=or-devmachinetoken uv run pytest -k generated_file_is_private` **fails**. `AUTH_TOKEN` is set in any shell that runs this stack, so the test would fail for most of the team; CI passed only because CI happens not to set it. And the assertion proved nothing anyway — absence is just absence, not evidence the script does not export.

Now a sentinel is injected with `monkeypatch.setenv` and asserted unchanged, plus that the ambient value never reaches the generated file. Verified passing with `AUTH_TOKEN` both set and unset, and mutation-checked: making `_write_private` write to `os.environ` fails it.

Approval was dismissed by the push — sorry for the extra round trip. Nothing else changed.

hedhoud (approved):
The non-export test is now isolated from the caller environment: it injects a sentinel, verifies that the value remains unchanged, and confirms it is not copied into the generated file. The targeted tests and Ruff pass, and all GitHub checks are green.

PR #957, opened 2026-09-14
feat(indexing): verify uploaded content against its declared extension

andyne13:
A file's extension alone decides which parser it reaches, so a file renamed `.pdf` is
handed to the PDF backend whatever it contains. This checks the content against the
declared extension wherever bytes meet a parser chosen by a caller-supplied name — the
upload routes, `index_url`, and EML attachments — by signature for the formats that carry
one, and by reading the package for the OOXML formats, which do not.

Extensions with nothing reliable to check pass through untouched. Text formats have no
signature, and `.doc`, `.wma` and the audio/video containers were tested against the
bundled matchers and are not reliably detected — enforcing those would refuse valid
uploads. The reasoning is recorded next to the map so it is not "fixed" later by mistake.

Worth knowing when reviewing:

- **Signature formats are read from the head, then rewound.** A test pins the rewind:
  without it every accepted upload would be silently truncated by 8 KiB.
- **`docx`/`pptx` are settled by reading the package instead.** A signature matcher looks
  for an entry *named* `word/` or `ppt/` near the head of the ZIP, and fails both ways: it
  accepts an archive holding `word/not-a-document.txt`, and misses a genuine document once
  `customXml` parts push `word/document.xml` past the entries it checks. They get
  `validate_ooxml_package`, which requires both mandatory OPC parts plus the main part and
  reads only the archive's central directory — no member is decompressed, and the stream is
  rewound on rejection too. The tests pin `filetype`'s verdict on both crafted archives, so
  an upgrade that changes it is visible rather than silent.
- **`index_url` previously verified nothing** — it takes its filename from the URL path.
  It now rejects and deletes the temp file on a mismatch.
- **A mismatched EML attachment is skipped and logged**, not fatal; the rest of the
  message still parses.
- **`filetype` was already resolved transitively** and is now declared directly.
- The MCP fixtures used `.pdf` URLs with placeholder payloads and now carry a PDF
  signature, which leaves what they assert untouched.

## Summary by CodeRabbit

* **Security & Validation**
  * Added content-signature checks for uploaded files, downloaded URL content, and email attachments.
  * Files whose contents do not match their declared extensions are rejected or skipped.
  * OOXML documents are checked for valid package structure before processing.
  * Supported file types are validated more reliably, including common document and image formats.
  * Upload streams are preserved so accepted files can be saved in full.


hedhoud:
Please validate the ZIP as a complete OOXML package before accepting it as a DOCX or PPTX file. Checking only for a `word/` or `ppt/` entry can accept unrelated ZIP files and reject valid Office files. Verify the required OOXML parts before passing the file to the parser.

EnjoyBacon7 (dismissed):

andyne13:
You were right on both counts, and I reproduced them against the bundled matcher
before changing anything:

- a ZIP whose first entry is `word/not-a-document.txt` is reported as a docx — no
  content-type map, no document in it;
- a genuine document is reported as a plain zip once `customXml` parts push
  `word/document.xml` past the few local file headers the matcher looks at. A
  LibreOffice-produced docx clears that window by one entry.

So docx/pptx no longer go through the signature map at all. `validate_ooxml_package`
opens the archive and reads its central directory, requiring both mandatory OPC parts
— `[Content_Types].xml` and `_rels/.rels` — plus the main part, `word/document.xml`
or `ppt/presentation.xml`. A pptx renamed `.docx` still fails, on the main part.

Only the directory is parsed; no member is decompressed. The tests assert `filetype`'s
actual verdict on both crafted archives, so an upgrade that changes it is visible
rather than silent, and they cover real python-docx, python-pptx and LibreOffice
layouts — including one that writes `[Content_Types].xml` last.

5aa083f4

EnjoyBacon7 (approved):

Issue #929, opened 2026-09-09
Only HTTP transport metrics exist: nothing measures indexing, queue depth, or inference health


andyne13:
## Summary

Every metric OpenRag exposes describes HTTP transport. Nothing measures whether documents are actually being indexed, where parse time goes, how deep the queue is, or whether the external inference endpoints are answering. The pipeline can be completely stalled with all current metrics green.

## What exists today

`openrag/core/observability/monitoring.py` defines three: `openrag_http_requests_total`, `openrag_http_request_failures_total`, `openrag_http_request_duration_seconds`. Plus `CIRCUIT_BREAKER_STATE` (`openrag/services/inference/_circuit_breaker.py:17`). Four in total, and the first two are redundant with each other — failures are derivable from `requests_total{status_code=~"5.."}`.

## Proposed Tier-1 set

| Metric | Type | Labels | Answers |
| ------ | ---- | ------ | ------- |
| `openrag_ingest_documents_total` | counter | `status` | throughput and failure rate |
| `openrag_ingest_stage_duration_seconds` | histogram | `stage` | where parse time goes |
| `openrag_ingest_tasks` | gauge | `state` | queue depth / backlog |
| `openrag_inference_requests_total` | counter | `provider`, `operation`, `outcome` | are the endpoints healthy |
| `openrag_inference_duration_seconds` | histogram | `provider`, `operation` | are they slow |
| `openrag_llm_tokens_total` | counter | `operation`, `kind` | aggregate token burn |

Plus queue wait (admission → parse start) and time-since-last-parse, which is the cheapest stall detector available.

## No unbounded labels — enforced by a test, not a convention

None of these may carry `partition`. Partitions are created on write by callers (`services/orchestrators/indexing_service.py`, `_ensure_partition_exists`), so the label is bounded by user behaviour rather than by anything we control — the same property that rules out `user_id` and `file_id`.

A histogram costs roughly one series per bucket per label combination, so putting `partition` on `openrag_ingest_stage_duration_seconds` scales with the number of tenants multiplied by the number of stages. On a shared platform Prometheus that is somebody else's capacity problem, and the remedy they reach for is dropping the scrape.

The codebase already has the right instinct one layer up: `api/middleware/instrumentation.py` records the **route template** with a fixed `/-unresolved-` fallback, pinned by `tests/unit/api/middleware/test_instrumentation.py:38-41` whose comment reads *"otherwise Prometheus label cardinality explodes"*.

[...]

PR #961, opened 2026-09-14
feat(observability): S3-2 — six Tier-1 metrics, queue wait, and time-since-last-parse

andyne13:
Implements S3-2 (#929): the six Tier-1 metrics, plus queue wait and the parse watchdog. Draft — open for CodeRabbit and reviewer comments.

## The gap this closes is not "six metrics are missing"

Four of the six are produced inside Ray actors, and `/metrics` is served from the API process. There is no shared registry, so a `prometheus_client` counter incremented in a worker is invisible to a scrape — and under `ENABLE_RAY_SERVE=true` even the API runs as Serve replicas that HTTP cannot address individually.

`openrag_circuit_breaker_state` is the existing instance of that bug, and S3-4 plans an alert on it. It is fixed here (see below).

So the substance of this PR is the bridge, not the counters. Writing six `Counter()` calls and incrementing them in the stages would have looked complete, passed review, and produced empty dashboards.

## What was verified rather than assumed

A spike (Phase 0) settled two load-bearing questions before any code:

- `ray.util.metrics` **does** export under embedded `ray.init()`, so this covers the standalone compose product, not just Kubernetes.
- Ray injects `WorkerId` / `SessionName` / `NodeAddress`, and the first two churn. That churn is **bounded**: dead workers' series expire after `RAY_WORKER_TIMEOUT_S` (default 120 s), measured at ~130 s end to end. Worst case is restart-rate × window — roughly 2 400 series at ten indexer actors.

## Where each metric is recorded, and why it matters

| Metric | Recorded in | Reason |
|---|---|---|
| `ingest_documents_total` | TaskStateManager | The worker only sees documents that reached it. Rejected submissions, stale ref-less tasks and queued cancellations are the systemic failures the alert needs |
| `ingest_queue_wait_seconds` | TaskStateManager | The only place holding both halves — the worker never receives `created_at` |
| `ingest_stage_duration_seconds` | `pipeline_builder._timed` | Timings already existed; recorded in `finally` so a stage hanging to its timeout is still measured |
| `ingest_last_parse_completion_*` | `ParserDispatcher._parse_with` | The one point knowing both which backend ran and that it finished |
| `ingest_tasks` | `/metrics` handler | Read at scrape time; a hand-maintained gauge would drift from the actor on every restart |
| `inference_*`, `llm_tokens_total` | both sides | Backend chosen per process on actor identity |

Terminal counts guard on the *previous* state: the TSM setters are re-entrant, and counting writes rather than transitions would inflate the very ratio `OpenRagIngestFailureRate` thresholds on.

## Cardinality

No metric carries `partition`. A unit test fails the build if any metric declares `partition`, `user_id`, `file_id`, `task_id`, `request_id` or `filename` — on either backend, so it covers future metrics rather than these ten. The integration test asserts the same against the exposition text, which also covers labels Ray attaches on its own.

`provider` is the admin-configured endpoint name, never the model or base URL: both are client-controllable via `metadata.llm_override`, which would make the label unbounded from the *value* side. Endpoint-override requests go to a fixed `client_override` bucket so a third-party endpoint cannot corrupt an operator endpoint's error rate — the same line `_circuit_breaker` already draws with `skip_if=_targets_client_endpoint`.

## Two behaviour changes outside the metrics

**`stream_chat` now sends `stream_options: {"include_usage": true}`.** Without it a streamed response carries no usage block, so the primary user-facing path contributed nothing to the token metric. The extra `"choices": []` chunk is already handled by `core/utils/source_filtering`, so no consumer changed.

**`openrag_circuit_breaker_state` moved onto the dual backend.** The `llm` breaker was visible; `embedder` and `vlm`, which trip during indexing, could never produce a sample — and a tripped embedder stops ingestion entirely.

## A defect found by self-audit, before review

`stream_chat` recorded `outcome="error"` on every **successful** chat. `stream_with_source_filtering` breaks on `[DONE]` then closes the iterator, raising `GeneratorExit` at the yield, which was being recorded as a failure. The error-rate panel would have read ~100% while everything worked.

Fixed by taking the semantics from code that already owns them: `source_filtering` reports a stream ending without `[DONE]` as truncated, so success is now *proven* by `[DONE]`, never assumed. Verified by running the new tests against the pre-fix commit.

## Testing

- 64 new unit tests; suite at **3229 passing**
- `tests/integration/test_ray_metrics_export.py` — starts embedded Ray, drives every worker-side recorder inside an actor, scrapes the node's metrics agent over HTTP, asserts the series are on the wire. This is the only test here that could have caught the circuit-breaker defect. Marked `integration`/`slow`; runs in ~7 s
- Each guard was mutation-tested rather than trusted for passing

## Notes for S3-3 / S3-4 / S3-5

Under Ray Serve the inference metrics land on the **Ray** scrape target, not `/metrics`. Scraping only `/metrics` yields HTTP metrics and no inference metrics, with nothing to indicate it.

Every worker-side query needs `sum without(WorkerId, SessionName, NodeAddress, Component, Version)(rate(...))` — `rate()` detects resets only within a series, so `rate(sum(...))` silently loses them. Recording rules, not per-panel. `WorkerId` must never be dropped at scrape time: collapsing concurrent workers produces duplicate samples and Prometheus rejects the whole scrape.

`docs/content/docs/documentation/metrics_reference.md` covers all of this; its names and label sets are cross-checked against `metric_specs` programmatically.

## Open for review

- `openrag_ingest_clock_skew_events_total` is a 9th metric beyond the brief — the honest handling of a cross-node measurement, but scope I added.
- `generate` posts to `/completions` yet is labelled `operation="chat"`; the enum has no "completion" value.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive metrics for ingestion progress, processing times, queue waits, inference requests, token usage, and circuit-breaker status.
  - In-flight ingestion task counts are now available through the metrics endpoint.
  - Metrics are supported across worker and API environments.
- **Bug Fixes**
  - Monitoring failures no longer interrupt ingestion or inference requests.
  - Circuit-breaker metrics now correctly report open and closed states.
- **Documentation**
  - Added a Tier-1 metrics reference covering scrape targets, querying guidance, labels, and availability notes.


openrag/services/inference/_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+            if has_override(kwargs):
+                return CLIENT_OVERRIDE_PROVIDER
+        except Exception:  # noqa: BLE001 - never let label resolution fail a call
+            pass
+    name = getattr(instance, PROVIDER_NAME_ATTR, None)
+    return name if isinstance(name, str) and name else _UNKNOWN_PROVIDER
+
+
+def _outcome_for(exc: BaseException) -> str:
+    if isinstance(exc, CircuitBreakerError):
```
hedhoud: Blocking: the real circuit-breaker wrapper converts `CircuitBreakerError` to `InferenceConnectionError` before this function sees it, so an open circuit is recorded as `error` and `circuit_open` is effectively unreachable. Please preserve a dedicated circuit-open signal (the existing `CircuitBreakerOpenError` would fit) and add a regression test using the complete decorator stack rather than raising `CircuitBreakerError` directly.

openrag/services/inference/_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+        return "circuit_open"
+    if isinstance(exc, InferenceTimeoutError):
```
hedhoud: Embedding clients raise `EmbeddingTimeoutError`, which is not an `InferenceTimeoutError`; vLLM embedding timeouts therefore land in the generic `error` bucket. That makes the timeout ratio incomplete. Please classify both timeout families and verify it through a decorated embed call.

openrag/services/inference/ollama_client.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         )
         logger.bind(model=self._model, endpoint=self._endpoint, timeout=timeout).debug("OllamaClient ready")
 
+    @with_inference_metrics("chat", capture_usage=True)
```
hedhoud, line 84: The Ollama instrumentation stops at synchronous `generate` and `chat`. `stream_chat` and `OllamaEmbedder.embed` still emit no request or duration metrics, so a supported Ollama deployment remains partly invisible. Please instrument both paths; the stream needs lifetime-aware handling like the vLLM generator, plus focused coverage for success and failure.

openrag/api/routers/admin/monitoring.py
```diff
@@ -7,15 +7,52 @@
...
+    except Exception as exc:  # noqa: BLE001 - the scrape must survive any of this
+        clear_ingest_task_counts()
+        logger.debug(f"ingest task gauge not sampled for this scrape: {exc}")
+
 
 @router.get("/metrics", summary="Prometheus metrics endpoint", dependencies=[Depends(require_admin)])
-async def prometheus_metrics():
+async def prometheus_metrics(request: Request):
     """Return all metrics in Prometheus text exposition format."""
+    await _refresh_ingest_tasks(request)
```
hedhoud: Concurrent scrapes can interleave between refreshing this process-global gauge and generating the exposition: one request may clear or overwrite another request's snapshot. Please serialize the complete refresh-and-collect section with one shared async lock so each response exposes the state it sampled.

hedhoud (changes requested):
The Ray/Prometheus bridge and the ingestion coverage are solid, but the inference metrics still produce incomplete or misleading signals in normal supported paths. I am requesting changes for the outcome classification and Ollama coverage gaps below. I also added one scrape-consistency issue. Please cover the fixes with tests that exercise the real decorator stack and each supported Ollama path.

openrag/services/inference/_metrics.py
andyne13: Fixed in `b449a50d`. `with_circuit_breaker` now raises a dedicated `CircuitBreakerOpenError` (a 503, where the old `InferenceConnectionError` was not), and `_classify_outcome` matches on it.

`test_open_circuit_is_its_own_outcome` drives the **complete decorator stack** — real `@with_circuit_breaker`, tripped by real failures — rather than raising by hand, as you asked. Mutation-verified: removing the `circuit_open` branch fails it.

openrag/services/inference/_metrics.py
andyne13: Fixed in `b449a50d`. `_classify_outcome` now matches `(InferenceTimeoutError, EmbeddingTimeoutError)`.

`test_embedding_timeouts_are_counted_as_timeouts` verifies it through a decorated embed call. Mutation-verified: narrowing the tuple back to `InferenceTimeoutError` alone fails it.

openrag/services/inference/ollama_client.py
andyne13: Fixed in `b449a50d`. `OllamaEmbedder.embed` is decorated; `stream_chat` is hand-instrumented because `@with_inference_metrics` would time only the creation of the async generator, not the transfer — so it mirrors the vLLM generator, with the outcome pessimistically `error` until `[DONE]` so an abandoned stream is never counted as a finished chat.

Coverage for success and failure on both paths. Mutation-verified: removing the decorator and the stream `record_inference` fails three tests.

openrag/api/routers/admin/monitoring.py
andyne13: Fixed in `b449a50d` — `_scrape_lock` serialises refresh **and** collect, since guarding only the sampling would leave the race intact.

You were right to flag this one twice over: the fix shipped **without a test**. I mutation-tested it just now — deleting `_scrape_lock` entirely left the full unit suite green at 3370 passed. `fb9bfb3d` adds `test_concurrent_scrapes_do_not_interleave`, which asserts the event order across two overlapping scrapes rather than the exposed numbers (a value assertion could pass on scheduling luck; an interleaved sequence is the race itself). It now fails when the lock is removed.

andyne13:
All four review findings addressed and **mutation-verified** — for each, I removed the fix and confirmed a test fails.

| finding | fix | guarded by |
|---|---|---|
| `circuit_open` unreachable | `CircuitBreakerOpenError` raised by `with_circuit_breaker`, matched in `_classify_outcome` | `test_open_circuit_is_its_own_outcome` — drives the complete decorator stack, not a hand-raised error |
| embedding timeouts in the generic bucket | `(InferenceTimeoutError, EmbeddingTimeoutError)` | `test_embedding_timeouts_are_counted_as_timeouts`, through a decorated embed call |
| Ollama `stream_chat` / `embed` uninstrumented | decorator on `embed`; hand-instrumented stream with a pessimistic outcome until `[DONE]` | 3 tests, success and failure on both paths |
| concurrent scrapes interleave | `_scrape_lock` over refresh **and** collect | `test_concurrent_scrapes_do_not_interleave` (new, `fb9bfb3d`) |

Worth calling out on the last one: the lock landed in `b449a50d` **without a test**. Deleting `_scrape_lock` outright left the whole suite green at 3370 passed, so the fix was unprotected — the same failure mode as the original `circuit_open` bug you caught. `fb9bfb3d` closes it.

`develop` merged in. Full suite 3423 passed; ruff, format and layer-import guard clean.

andyne13:
Pushed `e354a21b` after re-auditing the branch from scratch. It changes what there is to review, so flagging it rather than leaving it in the diff.

**`openrag_circuit_breaker_state` was always `-1`.** `_STATE_VALUES` keyed on `"OpenState"` / `"ClosedState"` / `"HalfOpenState"`, but aiobreaker's classes are `CircuitOpenState` / `CircuitClosedState` / `CircuitHalfOpenState`. Every transition fell through to `_UNKNOWN_STATE`, so the gauge never held `1` and S3-4's `OpenRagCircuitBreakerOpen` (`== 1`, severity critical) could never fire.

Pre-existing on `develop`, where it was harmless because the gauge never reached a scrape — exporting it is what turns it into a permanently-wrong exported metric. The integration test passed because it calls `record_circuit_breaker_state("embedder", 1)` directly: it proves the transport, not the content. Now keyed on aiobreaker's own `CircuitBreakerState` enum, so a rename breaks the import instead of silently emitting `-1`.

**Two recorder call sites had no guard.** Deleting `record_parse_completion(backend)` from `ParserDispatcher._parse_with`, or `record_circuit_breaker_state(...)` from the listener, left the whole suite green at 3423 passed. Both had recorder-level and export-level tests, but nothing asserting production calls them. `stage_duration` already had `test_pipeline_builder_still_calls_the_recorder` — the right shape existed, just not applied to these two. `OpenRagIngestStalled` reads `time() - max(openrag_ingest_last_parse_completion_timestamp_seconds)`, so a missing stamp makes that critical alert unfireable too.

Four tests added, each mutation-checked: dropping either call site fails one, and stamping before the parse rather than after fails a different one.

Also verified and unchanged: all 10 declared metrics have real production call sites (no orphans), the other six instrumentation sites are guarded, and the Ray/prometheus split holds.

Full suite 3427 passed; ruff, format and layer-import guard clean.

andyne13:
Pushed `bf61d67a` — a build-time guard so the two gaps found in `e354a21b` cannot recur silently.

The cardinality guard next door asks whether a metric is *safe to export*. This asks whether it is exported **at all**. Three links, each a way the chain breaks without anything noticing:

```
MetricSpec ──▶ in ALL_SPECS ──▶ instrument ──▶ recorder ──▶ production call site
```

- a spec left out of `ALL_SPECS` is exempt from the forbidden-label check, silently — that guard iterates the tuple;
- a spec nothing instruments passes the cardinality guard *trivially* (safe labels on a metric that never exists);
- a recorder nothing calls is the case found by hand twice here.

Two things that make it catch the shape that fooled us:

- **Static analysis, not imports.** It cannot initialise Ray, cannot need a live registry, and cannot be satisfied by a test calling the recorder — production callers are found by walking the package, so `tests/` does not count. Both recorders in `e354a21b` *were* called: from their unit tests and from the export harness. That is exactly why the suite stayed green at 3423.
- **Reachability is transitive.** `record_tokens` is only ever reached through `record_usage_from_response`; a direct-caller check would fail it legitimately.

Mutation-verified against all four failure modes, including both real findings from this branch: deleting the parse-completion stamp or the circuit-breaker state call each fails it.

Full suite 3439 passed; ruff, format and layer-import guard clean.

Note on placement, in case it looks like scope creep: this could not be a follow-up PR against `develop`, because `metric_specs.py`, `ray_metrics.py` and `inference_metrics.py` only exist on this branch — `develop` has just `monitoring.py`. It guards this PR's own invariant, so it lands with it.

tests/unit/core/observability/test_metric_emission.py
```diff
@@ -0,0 +1,181 @@
...
+    changed = True
+    while changed:
+        changed = False
+        for caller in list(reached):
+            for callee in recorders[caller] & recorders.keys():
+                if callee not in reached:
+                    reached.add(callee)
+                    changed = True
+
+    orphaned = sorted(set(recorders) - reached)
```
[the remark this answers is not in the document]
andyne13: Correct, and fixed in `3d98136c`. I verified the hole before fixing it: a spec declared, added to `ALL_SPECS`, with an instrument built from it and no function writing it, **passed all 13 tests**. The spec check sees the instrument and passes; the recorder check only inspects recorders that exist, so there was nothing to fail.

Each spec now has to have a recorder that production reaches. The spec→recorder link is found two ways, because the three modules build instruments differently:

- module-level `_X = _counter(SPEC)` bindings, which covers `ray_metrics`;
- propagation along the call graph, which is the only way `inference_metrics` reaches its specs — they are bound inside the `_instruments()` factory rather than to named module globals.

`monitoring` declares some metrics without a spec at all (`ORPHAN_CHUNKS_DROPPED`, the HTTP ones), so they fall outside this iteration by construction; the cardinality guard covers those through the live registry.

Mutation-checked three ways: the instrument-without-recorder shape now fails, and both real findings from `e354a21b` still fail — dropping the parse-completion stamp now fails *two* tests rather than one, since `INGEST_LAST_PARSE_TIMESTAMP` loses its writer as well as the recorder losing its caller.

Full suite 3449 passed.

hedhoud (commented):
One additional issue remains outside the existing review threads.

tests/integration/test_ray_metrics_export.py
```diff
@@ -0,0 +1,170 @@
...
+
+import os
+import pathlib
+import re
+import time
+import urllib.request
+
+import pytest
+
+pytestmark = [pytest.mark.integration, pytest.mark.slow]
```
hedhoud, line 34: This is the only test that proves worker metrics reach Ray's scrape endpoint, but no workflow runs it: unit CI stops at tests/unit/, and integration CI only runs tests/integration/repos/. That leaves the load-bearing export path outside the merge gate. Please add this test as a dedicated CI step or job. It should run in a clean Ray process, since it currently fails when Ray was initialized earlier and its runtime_env is ignored.

openrag/services/inference/ollama_client.py
hedhoud: The current implementation does not preserve the behavior described here: it sets success whenever the HTTP iterator ends, even if no [DONE] marker was received, so a truncated stream is reported as healthy. It also never records a final usage block, which leaves normal Ollama streaming chat out of openrag_llm_tokens_total. Please keep the outcome as error until [DONE] is observed and record any top-level usage payload before completing the stream.

andyne13: Fixed in #1007 (`e6889b3f`, merged into this branch): the outcome stays `error` until `[DONE]` is seen, the usage block is recorded before the stream completes, and both are covered by tests (truncated stream → error; usage chunk → counted).

tests/unit/core/observability/test_metric_emission.py
```diff
@@ -0,0 +1,270 @@
...
+    for path in _python_files(_OBSERVABILITY_DIR):
+        for node in ast.walk(_parse(path)):
+            if isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
+                bodies.setdefault(node.name, node)
+
+    touches: dict[str, set[str]] = {}
+    for fn_name, fn in bodies.items():
+        names = {n.id for n in ast.walk(fn) if isinstance(n, ast.Name)}
+        for spec in _spec_variable_names():
+            if spec in names or (spec_vars.get(spec, set()) & names):
```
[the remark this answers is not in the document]
andyne13: Fixed in #1007 (`31eec9d3`, `3746d860`): a spec is linked to a function only through a value write (`inc`/`dec`/`observe`/`set`), including through a `.labels(...)` receiver; calling the factory alone links nothing. Mutation-checked.

tests/integration/test_ray_metrics_export.py
andyne13: Fixed in #1007 (`8c2848a5`): `test_ray_metrics_export.py` runs as its own `ray-metrics-export` job in its own pytest process, and the fixture shuts down any existing Ray session so its `runtime_env` applies.

openrag/core/observability/metric_specs.py
```diff
@@ -0,0 +1,353 @@
+"""Declarative specifications for every OpenRag metric — one source of truth.
+
+**Why specs rather than metric objects.** S3-2's metrics are produced on two
+sides of a process boundary. The HTTP and chat-path metrics live in the API
+process and use ``prometheus_client``; every ``ingest_*`` metric is produced
+inside a Ray actor (``services/workers/indexer_actor.py``, driven by
+``indexer_pool.py``) and must use ``ray.util.metrics``, because a
```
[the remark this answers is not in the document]
andyne13: Fixed in `2468dc83`: only the per-document `ingest_*` metrics are described as Ray-produced; `openrag_ingest_tasks` is named as the API-process gauge served on `/metrics`.

openrag/services/inference/vllm_client.py
```diff
@@ -41,11 +44,39 @@
...
+    Called for every SSE line — hundreds per answer — so the substring test
+    comes first. Content deltas also begin with ``data: ``, and parsing each of
+    them would duplicate, on the primary user-facing path, work that
+    ``stream_with_source_filtering`` already does downstream.
+
+    A delta whose *content* happens to contain the word "usage" is parsed and
+    then discarded by ``record_usage_from_response``, which requires ``usage``
+    to be a top-level object.
+    """
+    if '"usage"' not in line or not line.startswith("data: "):
```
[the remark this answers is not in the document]
andyne13: Fixed in `2468dc83`: `data:` is accepted with or without the optional space, with tests for both forms and a control that a content delta without `usage` is not parsed. This also covers the outside-diff comment on `metric_specs.py` (add `completion` to the operation values): done in the same commit, with a test that fails if an emitted operation isn't declared.

openrag/services/orchestrators/job_service.py
```diff
@@ -103,6 +103,16 @@ async def get_queue_info(self) -> dict:
         worker_info = await self._call(lambda: self._tsm.get_pool_info.remote(), "get_pool_info")
         return {"workers": self._format_pool_info(worker_info), "tasks": task_summary}
 
+    async def get_active_task_counts(self) -> dict[str, int]:
+        """The ``active_statuses`` counts alone, without the worker-pool round trip.
+
+        ``get_queue_info`` always fetches pool info too, which the metrics scrape
+        never uses; that second Ray call would just eat into its timeout budget.
+        """
+        all_states: dict = await self._call(lambda: self._tsm.get_all_states.remote(), "get_all_states")
```
hedhoud: Blocking: this gauge bypasses the durable job state that #1023 made authoritative. After a TaskStateManager/Ray actor restart, `jobs` can still contain QUEUED/SERIALIZING rows, but this method reads only `get_all_states()` and publishes zero. The `/metrics` backlog is therefore blind during the restart/recovery window, so stalled-ingest alerting can stay green. Please make this count use the same durable-first/reconciliation path as `get_queue_info()` (with actor fallback only when Postgres is unavailable) and add a regression test for durable active rows missing from the actor.

openrag/core/observability/metric_specs.py
```diff
@@ -0,0 +1,355 @@
...
+    kind="counter",
+    labels=(),
+)
+
+
+# ---------------------------------------------------------------------------
+# Tier 1 — produced in the API process (exported via prometheus_client)
+# ---------------------------------------------------------------------------
+
+#: Sampled at scrape time from ``job_service.get_queue_info()`` rather than
```
hedhoud: Minor but now stale: this documentation says the gauge samples `get_queue_info()` and treats the TaskStateManager as authoritative. Since #1023 made durable job history authoritative and this PR adds `get_active_task_counts()`, please update the description together with the durable-first fix so the metric/runbook does not document actor-only behavior.

openrag/services/orchestrators/job_service.py
andyne13: Fixed in `76380fbf`: `get_active_task_counts()` now goes through the same durable-first, reconciled counting as `get_queue_info()` (shared `_task_status_counts()`); the actor alone is used only when the durable count fails. Regression test: durable QUEUED/SERIALIZING rows with an empty (restarted) actor are counted; plus a parity test against `get_queue_info()` and the Postgres-down fallback. Reverting to the actor-only read fails both new tests.

openrag/core/observability/metric_specs.py
andyne13: Updated in the same commit: the spec comment and `metrics_reference.md` now describe the durable-first source and when it falls back to the actor.

hedhoud (approved):
Re-reviewed the current head. The previously requested fixes are addressed, CI is green, and the focused observability suite passes. Approving.

Issue #967, opened 2026-09-15
The semaphore starvation fix does not reach a detached actor on a persistent Ray cluster until it restarts


andyne13:
## Summary

#969 fixes release starvation in `DistributedSemaphoreActor`: `release` gets its own Ray
concurrency group, and because that group runs on its own event loop it hands the wake-up
to the acquire loop with `call_soon_threadsafe`, under a lock covering the window before
the first `acquire` has captured that loop. That is the server-side guarantee this issue
was opened for, so its original scope is **done**.

What remains is the rollout. The actor is detached and looked up with `get_if_exists=True`,
so on a persistent Ray cluster the previous release's actor survives a rolling deploy and
keeps serving the old code until it is restarted or the cluster is recreated. #969
documents this; it does not change it. On compose, where Ray is embedded in the API
container, a container restart is enough and the fix applies immediately.

The gap matters most exactly where the bug is easiest to hit. On Kubernetes the Ray
cluster is long-lived, and the chart defaults are already past the threshold:
`RAY_POOL_SIZE: 3` x `RAY_MAX_TASKS_PER_WORKER: 50` x `max_concurrent_contextualization`
10 = 1500 outstanding acquires against a 1000-slot group (see #965). So the deployment
most exposed to the wedge is also the one the fix reaches last.

## It is not only the code that goes stale

The same mechanism strands *configuration*, which is easier to hit and quieter when it
does. `DistributedSemaphore._get_or_create_actor` passes `max_concurrent_ops` only on the
create path, and bootstrap has always created the actor first, so raising `VLM_SEMAPHORE`
or `LLM_SEMAPHORE` and rolling the indexer pods does not raise the live budget — the actor
keeps whatever capacity it was born with until it is recreated. (Noticed by a reviewer on
#971.)

That interacts badly with the per-document caption cap in #971, which reads
`config.semaphore.vlm_semaphore` at worker construction. After a config bump the workers
cap at the *new* value while the actor still admits the *old* one, so the local cap sits
above the real budget and quietly stops bounding anything. Nothing fails; the fan-out
bound just becomes decorative.

Versioning the actor name fixes both at once: a deploy gets an actor built from current
config running current code, instead of inheriting either from the previous release.

## Direction

Version the actor name so a deploy creates a new generation instead of attaching to the
previous release's actor — the same move `_INDEXER_ACTOR_PROTOCOL_VERSION` makes for the
indexer tier after #846. Retiring the old generation needs care: it may still hold permits
[...]

PR #969, opened 2026-09-15
fix(inference): stop semaphore release starving behind pending acquires

### LINKED ISSUE #965: `DistributedSemaphoreActor` has no `max_concurrency`/`concurrency_groups` — blocked `acquire` calls can starve `release` and deadlock the LLM/VLM gate
## Summary

`DistributedSemaphoreActor` is created without `max_concurrency` or `concurrency_groups`, so it runs at Ray's asyncio default of **1000 concurrent calls**. A waiting `acquire` holds one of those slots for the entire wait. If the number of blocked `acquire` calls reaches that limit, every slot is held by a waiter and a subsequently submitted `release` is queued behind them and never dispatched — so no permit is returned and no waiter can wake.

The consequence would be a permanent cluster-wide stall of the LLM/VLM gate: in-flight files stay in `SERIALIZING`, queued files never start, counters stop moving. The actor is detached, so it would outlive an API restart and clear only on a Ray cluster restart.

This came out of analysing an indexing freeze, but it is **not** presented as the confirmed cause of that freeze — the state that would have confirmed it was discarded by the restart used to recover. The defect itself is verifiable from the source and stands independently of that incident.

## Relationship to #630

#630 described the same symptom via a different mechanism, and its fix (`cfd9ad86`, shipped v2.0.1 onward) does not cover this path:

| | #630 | this |
|---|---|---|
| Mechanism | permit **leak** — a cancelled `__aenter__` consumed a permit that was never released | slot **exhaustion** — no permit is lost; none can be returned, because `release` is never dispatched |
| Trigger | caption/VLM task cancellation after a failure | enrichment fan-out reaching the actor's concurrency limit; no cancellation required |
| Addressed by | cancellation-safe `__aenter__` | nothing today |

## Root cause / code trace

`openrag/services/inference/distributed_semaphore.py:20-39`

```python
@ray.remote(max_restarts=5)          # no max_concurrency, no concurrency_groups
class DistributedSemaphoreActor:
    async def acquire(self) -> str:
        await self.semaphore.acquire()   # holds its actor slot for the whole wait
        return self.incarnation

    def release(self, incarnation=None) -> None:   # same default concurrency group
        self.semaphore.release()
```

- Ray defaults asyncio actors to `DEFAULT_MAX_CONCURRENCY_ASYNC = 1000` (`ray/actor.py:1564-1568`, ray 2.53).
- Queued actor calls are dispatched in arrival order. Once blocked `acquire` calls occupy every slot, each slot that frees goes to the next queued `acquire`, which blocks in turn; a `release` submitted afterwards never runs.
- `DistributedSemaphore.__aenter__` (`:86-92`) awaits `acquire.remote()` with no timeout, so callers wait indefinitely.

**Callers that can reach the limit:**

- `stages/caption.py:68` creates **one task per image for the whole document**, each awaiting the semaphore (`:61`). Unbounded in document size.
- `core/indexing/contextualize.py:112-125` issues `max_concurrent_contextualization` concurrent acquires per file (default **10**, `core/config/chunking.py:24`).

For contextualization the worst case is `pool_size × max_tasks_per_worker × max_concurrent_contextualization`:

| config | outstanding acquires | vs. 1000 |
|---|---|---|
| `conf/config.yaml` (1 × 50 × 10) | 500 | under |
| `RAY_MAX_TASKS_PER_WORKER=100` (1 × 100 × 10) | 1000 | at the limit |
| `infra/charts/.../values.yaml:709-710` (3 × 50 × 10) | **1500** | **over** |

So the chart defaults already exceed the threshold whenever contextualization is enabled. Captioning has no equivalent bound at all — it scales with images per document.

Two things would stop it from self-clearing:

- `_build_pipeline_timeouts` (`indexer_pool.py:1048`) bounds only the **parse** stage; `caption` and `contextualize` are unbounded, on the rationale that "their backends self-limit" — which does not hold when the wait is *in front of* the backend.
- The actor is detached, so it outlives the API process.

`enable_contextualization` defaults to `False` (`core/config/indexation_pipeline.py:43`), so batches indexed with the default preset never touch `llmSemaphore` — which is likely why this has not been hit more often.

## Impact

- Indexing would stop cluster-wide and not recover without a Ray restart.
- The map-reduce answer path is exposed too: `RAGMapReduce`'s per-document relevancy sub-calls share `llmSemaphore` (`services/orchestrators/query_service.py:405`). Ordinary `/v1/chat/completions` does not use it.
- Restarting to recover discards queued tasks — their Postgres rows are swept to `FAILED` by orphan reconciliation, but the files are not indexed and no callback fires.

## Direction

The actor needs `release` to be dispatchable independently of waiting `acquire` calls, and the number of outstanding remote acquires needs to be bounded by construction rather than by config arithmetic. Design details belong in the PR.

Related: #663 (no admission control on ingestion) covers the intake side of the same pressure; this is the enrichment-stage side.

## Interim mitigation (no code)

- Keep `pool_size × max_tasks_per_worker × max_concurrent_contextualization` well under 1000. Lowering `RAY_MAX_TASKS_PER_WORKER` is the cheapest lever and costs no throughput, since the ceiling is the inference endpoint's parallel capacity, not the slot count.
- Keep `MAX_CONCURRENT_CONTEXTUALIZATION ≤ LLM_SEMAPHORE`; both default to 10, so one file's batch can take the entire global budget.
- Raising `MAX_CONCURRENT_CONTEXTUALIZATION` to speed up contextualization makes this more likely and adds no throughput.

paultranvan:
Prod (v2.2.0) stopped indexing for five days, from 2026-09-10 16:18 UTC until the vlmSemaphore actor was restarted by hand on 2026-09-15. A bulk upload of image-heavy PPTX/PDF files dispatched one acquire() per image up-front, which is more than the 1000 in-flight calls Ray allows an async actor by default. Every acquire on a full semaphore stays suspended inside the actor and keeps its slot, so once the waiters alone filled the cap every release() queued behind them and never ran: holders could not hand permits back, waiters never got one, and 100 tasks sat in the caption stage (ms_caption ~= 4.96 days) until the actor was recreated.

release() now runs in its own Ray concurrency group, so it always has slots of its own. In an async actor that group runs on a separate event loop and thread, and asyncio.Semaphore is not thread-safe, so the wake-up is handed to the acquire loop with call_soon_threadsafe; a lock covers the window before the first acquire has captured that loop.

The regression test fills the actor with 1000 blocked acquires and asserts a release still lands and reaches a waiter; it timed out on the previous actor. Reproduced against Ray 2.47.1.

Deploy note: the actor is detached and looked up with get_if_exists, so on a persistent Ray cluster the old, vulnerable actor survives a rolling deploy until it is restarted or the cluster is recreated. On the compose deployment a container restart is enough. Versioning the actor name so a deploy creates a new generation (as `_INDEXER_ACTOR_PROTOCOL_VERSION` does for the indexer tier) is tracked in #967 and kept out of this PR.

Closes #965

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved distributed inference reliability under heavy concurrent load.
  - Permit releases are no longer delayed when many requests are waiting, helping prevent request stalls and improving recovery when capacity becomes available.

- **Tests**
  - Added coverage for high-concurrency scenarios to verify that waiting requests resume correctly after a permit is released.


andyne13:
Confirming the mechanism from a different angle: I filed #965 earlier today after an indexing freeze on another deployment, and it is the same defect — blocked `acquire` calls occupying the async actor's default concurrency group until `release` can no longer be dispatched. Your production evidence settles what that write-up could only infer from the source, and the caption path being the trigger matches: it fans out one `acquire` per image with no per-document bound.

Worth adding `Closes #965` to the description so the two stay linked — that issue carries the code trace and the threshold arithmetic, including that the chart defaults (`RAY_POOL_SIZE: 3` × `RAY_MAX_TASKS_PER_WORKER: 50` × `max_concurrent_contextualization` 10 = 1500 outstanding acquires against a 1000-slot group) are already past the line.

On your deploy note: I would suggest versioning the actor name so a deploy creates a new generation instead of attaching to the previous release's actor, the way `_INDEXER_ACTOR_PROTOCOL_VERSION` does for the indexer tier after #846. Without it the fix reaches last exactly the deployment most exposed to the bug — a long-lived Ray cluster on K8s, where the defaults above apply. Tracked separately as #967 if you would rather keep it out of this PR; retiring the old generation needs care, since it may still hold permits for in-flight callers.

Ahmath-Gadji (approved):
LGTM. Verified locally against Ray 2.47.1:

- The new regression test times out against the pre-fix actor and passes with the fix.
- `release` in its own group really does run on a separate thread and event loop from `acquire`, so the `call_soon_threadsafe` hand-off is needed.
- Stress run: 2500 concurrent callers on a 3-permit semaphore (past the 1000-call cap), 200 of them cancelled while waiting. All the others finished, no more than 3 ever held a permit at once, and capacity was exactly 3 afterwards (no leak, no over-count).
- Mixed versions are safe: a plain `.remote()` sends no concurrency-group name, so the actor picks the group from its own class. Old drivers reach the new actor's release group, and new drivers still work against an old actor.

Non-blocking:
1. CodeRabbit's `sleep(1.0)` point is fair but only works one way: if the acquires haven't all reached the actor, the test can pass on regressed code, but it can't fail on this fix. A waiter count would make it deterministic, if you want that.
2. Until #967 lands, the release notes should tell operators on persistent Ray clusters to restart `llmSemaphore` / `vlmSemaphore` / `audioSemaphore` after deploying, or they keep running the vulnerable actor.

paultranvan:
@andyne13 @Ahmath-Gadji I suggest we close https://github.com/linagora/openrag/issues/967 , since a ray restart is enough

Issue #930, opened 2026-09-09
No alert rules exist anywhere: metrics are collected but nothing notifies


andyne13:
## Summary

There are no alert rules anywhere in the repository — `grep -rl "alert:" infra/` returns nothing. Metrics exist to be looked at; nothing tells anyone to look.

## Proposal

A `PrometheusRule` with eight named rules, each carrying a `runbook_url` annotation:

| Alert | Fires on |
| ----- | -------- |
| `OpenRagIngestStalled` | queued tasks non-zero with no completion for N minutes |
| `OpenRagIngestFailureRate` | failure ratio over the aggregate `status` label |
| `OpenRagBacklogGrowing` | queue depth rising monotonically over the window |
| `OpenRagInferenceProviderDown` | inference error ratio, or the circuit-breaker gauge open |
| `OpenRagNotReady` | readiness failing or flapping |
| `OpenRagGpuSaturated` | GPU memory or utilisation at ceiling |
| `OpenRagCatalogDriftDetected` | dropped-hit rate above zero (#925) |
| `OpenRagCanaryFailing` | the synthetic canary not passing |

Depends on #929 for the series most of these query, and on the GPU and canary work for the last two.

## Two requirements that decide whether this works

**Every rule needs a `runbook_url`** pointing at a page under `docs/deployment/runbooks/<AlertName>.md`. At three in the morning an alert without a link is an alert without a runbook, however well the page was written.

**Labels must match the receiving Alertmanager's routing conventions.** A `PrometheusRule` whose labels do not match fires into a void that is indistinguishable from "no alerts". Where the alerting stack belongs to a platform team rather than to us, their severity/team/service labels are an input, not something to guess — so make the label set configurable from chart values rather than baked into the rule file.

Eight is deliberately small: a first alerting set should contain only things worth waking someone for. Rules nobody acts on are how a team learns to mute the channel.

Issue #931, opened 2026-09-09
Nothing scrapes Milvus, Postgres or Ray on Kubernetes — Ray already exports and is simply not collected


andyne13:
## Summary

Even with the application's own `/metrics` scraped, the three things it depends on are unobserved on Kubernetes. Every incident then starts at "the application is slow" with no way to tell which layer is at fault. The three are **not** in the same state.

## Ray — already exports, is not scraped

`infra/charts/openrag-stack/templates/raycluster.yaml:72` sets `--metrics-export-port=8090` with a named `ray-metrics` container port, and a comment recording that 8090 is deliberately kept off `networkPolicy.externalPorts`' 8080 because *these Ray metrics are unauthenticated*.

So Ray needs a **`PodMonitor`, not an exporter** — metrics are per-pod across head and workers — and the target must stay namespace-internal. It must not be routed through the ingress.

*Worth watching:* pending tasks, actor restarts (which is how a healthy restart is distinguished from a crash loop), node resources, GPU allocation.

## Postgres — exporter not enabled

The bitnami `postgresql` subchart ships a metrics switch that runs a `postgres_exporter` sidecar. Nothing is configured: `grep -cE "^\s+metrics:" infra/charts/openrag-stack/values.yaml` returns 0.

**Do this one first.** The review on #915 identified a concrete failure it would have caught: `PostgresStore.check_health` takes a pooled connection with a 2-second timeout, so under pool saturation readiness fails on *every replica at once* and the ingress 503s everything. Connection-count-against-max and pool-wait metrics are how that is seen coming rather than reconstructed afterwards.

*Worth watching:* connections against max, pool wait, slow queries, disk.

## Milvus — exporter not enabled, and the key names need confirming

Milvus components export metrics and the subchart has a metrics block, but the chart pins milvus-helm 5.0.25 with the image overridden to Milvus 3.0.0 (`values.yaml`). Value keys must be checked against that pair rather than assumed, and Milvus 3 reorganised components — Woodpecker replaced Pulsar — so the 2.x metric documentation does not transfer.

*Worth watching:* search latency, insert throughput, whether the collection is loaded, compaction and segment backlog, memory.

## Applies to all three

Each `ServiceMonitor` or `PodMonitor` must carry whatever label the receiving Prometheus selects on, or it is created and silently ignored — indistinguishable from working. And Milvus and Ray are both chatty; where the Prometheus belongs to a platform team, the added series volume is a conversation to have up front.

[...]

PR #976, opened 2026-09-16
feat(observability): S3-4 — alert rules with runbooks, delivered to chart and Compose

andyne13:
Six alert rules as plain Prometheus rule groups, delivered two ways from one definition: wrapped in a `PrometheusRule` by the chart, loaded via `rule_files` by the Compose overlay. Each carries `severity`, `summary`, `description` and a `runbook_url`, and every runbook page exists.

**Draft — do not merge before #961.** Five of the six reference metrics that #961 introduces, so on `develop` today they match nothing, and an inert alert rule is indistinguishable from a healthy system.

Six rather than the nine named in the observability plan:

| Not shipped | Blocked on |
| --- | --- |
| `OpenRagNotReady` | Nothing exports readiness. `up == 0` is not a substitute — `/metrics` keeps answering while the container is degraded, so a degraded instance is `up == 1` and not ready at once. `OpenRagTargetDown` ships instead, scoped to what it actually measures. |
| `OpenRagGpuSaturated` | A GPU exporter on Kubernetes |
| `OpenRagCanaryFailing` | The synthetic canary |

All three are listed in the runbook index so their absence is visible rather than assumed.

**Known limitation, documented in the rules file and the index:** every expression reads per-process counters, so under `ENABLE_RAY_SERVE=true` with `num_replicas > 1` a scrape reaches one replica at random and these will both miss and misfire.

Verified: `promtool check rules` (6 rules) and `check config` on the Compose path; `helm template` on both render paths including `commonLabels` precedence; `helm lint`; full unit suite. The metric-name, forbidden-label, runbook-exists and `jobMatcher` guards are each mutation-tested.

Refs #930 — partial: three of the named alerts remain.


andyne13:
**Blocked on #931 — do not merge yet.** Checked these expressions against what is actually on the wire, and 4 of the 7 cannot fire as deployed.

Two stacked causes, both outside this PR:

1. **No Ray metrics target is scraped.** `infra/compose/prometheus/prometheus.yml` has four jobs and none for Ray; the chart has no ServiceMonitor/PodMonitor template at all. S3-2 emits the whole `ingest_*` set through `ray.util.metrics`, which lands on Ray's own endpoint, not on `/metrics`.
2. **The exported name carries a `ray_` prefix.** `tests/integration/test_ray_metrics_export.py` asserts `_PREFIX = "ray_openrag_"`. These rules use the bare `openrag_*` names, so they would not match even once the target is scraped.

| alert | severity | verdict |
| --- | --- | --- |
| `OpenRagIngestStalled` | critical | **never fires** — `and on()` against an empty series |
| `OpenRagIngestFailureRate` | warning | **never fires** |
| `OpenRagCircuitBreakerOpen` | critical | **half-blind** — sees `llm`/`reranker` (API process), blind to `embedder`/`vlm` (workers) |
| `OpenRagInferenceProviderDown` | critical | **half-blind** — sees chat/rerank, blind to embed/vlm |
| `OpenRagBacklogGrowing` | warning | works — `ingest_tasks` is sampled in the API process |
| `OpenRagCatalogDriftDetected` | critical | works — incremented on the API's retrieval path |
| `OpenRagTargetDown` | critical | works — `up{}` |

The two half-blind ones come from `inference_metrics._use_ray_backend()` choosing per process: the same metric name lands on both targets, so scraping one makes them *partial* rather than absent.

`metric_specs.py` already called this out — *"A deployment that scrapes only `/metrics` sees HTTP metrics and no inference metrics; both targets must be scraped for the set to be complete."* The design documented the requirement; the deployment config never met it.

Merging now would install four alerts that sit permanently green, which is the exact condition these rules exist to detect. Details and the prefix decision (relabel at scrape time vs. rename in the expressions) are on #931.

Nothing here needs changing yet — the expressions are right for the names `metric_specs.py` declares. What changes is whichever way #931 resolves the prefix.

Ahmath-Gadji:
Follow-up to the blocking note above, on where its two causes stand now:

1. **No Ray target scraped.**
   - Kubernetes with `ray.enabled=true`: #979 adds the Ray PodMonitor.
   - Compose: #1011 pins the embedded Ray's metrics port (`RAY_METRICS_EXPORT_PORT`, 8091 in the monitoring overlay) and adds a `ray` job.
   - Still uncovered: Kubernetes with `ray.enabled=false`, the chart default. Ray runs embedded in the API pod on a random port, and nothing scrapes it.
2. **The `ray_` prefix.** #979 relabels at scrape time (`ray_(openrag_.+)` → `$1` on `__name__`), and #1011's Compose job does the same. On both paths the stored names are the bare ones these rules use. Measured on Compose: Prometheus stores `openrag_ingest_documents_total{job="ray", WorkerId=…}` and holds no `ray_openrag_*` series.

With #979 and #1011 in, the four rules in your table match on both deployments, except under `ray.enabled=false`.

If you would rather not depend on the relabel being present, for example on a platform Prometheus that scrapes Ray through its own PodMonitor, `{__name__=~"(ray_)?openrag_…"}` matches either way. The S3-5 dashboard uses that form. `sum by (provider)` then adds the API and Ray halves, which come from different targets, so there is no duplicate-labelset error (checked on Prometheus 2.54.1).

Issue #932, opened 2026-09-09
No GPU metrics on Kubernetes, and gpu-metrics.json queries an exporter namespace that does not exist there


andyne13:
## Summary

There is no GPU observability on Kubernetes, and the existing GPU dashboard cannot be reused there because it queries a different exporter's metric names.

## Evidence

Compose has an exporter: `infra/compose/monitoring.docker-compose.yaml:57` runs `utkuozdemir/nvidia_gpu_exporter:1.2.0`. The chart has nothing — `grep -ric dcgm infra/charts/` returns 0 across every file.

`infra/compose/grafana/dashboards/gpu-metrics.json` queries `nvidia_gpu_duty_cycle`, `nvidia_gpu_memory_used_bytes`, `nvidia_gpu_memory_total_bytes`, `nvidia_gpu_power_draw_watts`, `nvidia_gpu_temperature_celsius`, `nvidia_gpu_fan_speed_percent` — that exporter's namespace. The Kubernetes-standard NVIDIA DCGM exporter emits `DCGM_FI_DEV_*` instead.

So this dashboard does not merely need its datasource rebinding, as the other two do: it queries **metric names that do not exist on the other deployment**.

## Why it matters here

A single GPU is shared by the Marker parser pool, the embedder and the reranker. GPU utilisation and memory headroom are therefore the saturation signal for the whole ingestion pipeline — the number that says whether the parser pool is sized correctly, and the one you need before deciding a second card is required.

## Proposal

1. **An exporter on Kubernetes.** If the cluster already runs the NVIDIA GPU Operator or a DCGM exporter DaemonSet, this reduces to a scrape target and a dashboard; if not, the DaemonSet is part of the work. Worth establishing before estimating.
2. **A GPU dashboard that works on both**, via either two variants or recording rules normalising one metric namespace onto the other. Whichever is chosen, it should be decided once rather than diverging per deployment.

Related: #852 (chart ships no monitoring), #931 (datastore metrics), #846 (parser memory residency).

PR #979, opened 2026-09-16
feat(helm): S3-7 — scrape Ray, Postgres and Milvus with the Prometheus Operator

### LINKED ISSUE #931: Nothing scrapes Milvus, Postgres or Ray on Kubernetes — Ray already exports and is simply not collected

Ahmath-Gadji:
Closes #931

## Summary

- **Ray, Postgres and Milvus can now be scraped by an existing Prometheus Operator.** Everything is off by default. With default values the rendered manifests match `develop` exactly, apart from the Postgres password the bitnami sub-chart regenerates on every render.
- **Found along the way: Ray workers were exporting their metrics on the public port 8080.** Only the head was on 8090. Fixed in its own commit; see below.
- **The NetworkPolicy was the real blocker.** The default-deny policy only admits the release namespace, so a Prometheus anywhere else saw every Milvus target down. It now gets a narrow, opt-in way in.

| | Switch | What it does |
|---|---|---|
| Ray | `ray.metrics.podMonitor.*` (new) | Chart-owned `PodMonitor` over the head and every worker, on port `metrics` (8090). Relabels `ray_io_cluster` / `ray_io_node_type` for Ray's Grafana dashboards. |
| Postgres | `postgresql.metrics.enabled`, `postgresql.metrics.serviceMonitor.*` | The bitnami sub-chart's `postgres_exporter` sidecar and ServiceMonitor, now documented in `values.yaml`. Enabling the sidecar restarts Postgres. |
| Milvus | `milvus.metrics.serviceMonitor.*` | milvus-helm's ServiceMonitor. All five components (proxy, mixcoord, datanode, querynode, streamingnode) already exported on 9091; only the monitor was missing. |
| All three | `networkPolicy.metricsFrom` (new) | One NetworkPolicy per exporter, opening only its metrics port, only on its own pods, only to the listed peers. Nothing goes through `externalPorts` or the Ingress. |

A monitor switched on while its exporter is off renders nothing, so the render now fails with a message instead. `NOTES.txt` warns when a monitor has no selector labels or when `metricsFrom` is empty. `kubernetes.md` gains a section covering setup, the three silent failure modes, verified metric names to watch, what is not covered, and measured series volume.

On the issue's note about chart versions: `develop` now pins milvus-helm **5.0.27** (Milvus 3.0.1, no image override). The value keys were checked against that pair and bitnami postgresql 18.10.0 by rendering and deploying them, not taken from upstream docs.

## Ray workers exported metrics on 8080

The issue's starting point was that Ray already exports on 8090, off the public port. That was true only of the head. Workers start from `$KUBERAY_GEN_RAY_START_CMD`, and KubeRay builds that command with its default `--metrics-export-port=8080`. 8080 is the port `networkPolicy.externalPorts` opens to every source. KubeRay also appends a `metrics: 8080` port to any Ray container that lacks a port of that name. So the head and its Service advertised a port nothing listened on.

The fix sets `rayStartParams.metrics-export-port` on the worker groups. It also renames the port from `ray-metrics` to `metrics` on both head and workers. Nothing in the repo referenced the old name.

Measured on kind with KubeRay 1.4.0, curling `/metrics` from a pod in an unrelated namespace:

| Target | `develop` | this branch |
|---|---|---|
| worker :8080 | **200, 152 `ray_` samples** | blocked |
| worker :8090 | blocked | blocked |
| head :8090 | blocked | blocked |

> [!WARNING]
> **Upgrading an existing RayCluster:** KubeRay does not recreate Ray pods when the `RayCluster` spec changes. After `helm upgrade` the pod UIDs were unchanged, the workers were still on 8080, and the head target was down. Delete the pods once:
> `kubectl delete pod -l ray.io/cluster=openrag-raycluster`.
> This applies to any deployment with `ray.enabled=true`, including `values-linagora.yaml`.

## Verification

Test setup:

- kind (Kubernetes 1.31), where kindnet enforces NetworkPolicy.
- kube-prometheus-stack 91.4.1 in a separate `monitoring` namespace, with selectors left at their default (`release: kps`).
- KubeRay operator 1.4.0.
- This chart with Postgres, Milvus in cluster mode with Woodpecker, and a RayCluster. vLLM, the reranker, the admin UI and the API were off.
- Ray pods used a Ray 2.47.1 stand-in image with the same uv entrypoints and UID. The real `openrag-ray` image runs `uv sync` over the full app dependencies on first start.

Scrape targets, first with `develop` plus the sub-charts' stock monitor switches labelled `release: kps`, then with this branch:

| Target | `develop` | this branch |
|---|---|---|
| Milvus ×5 | down, `context deadline exceeded` (default-deny) | up |
| Ray head | no monitor | up |
| Ray worker | no monitor | up |

The metric names in the docs table were read from that Prometheus after a small workload: an actor killed and restarted, three GPU tasks left pending on the GPU-less node, 5 000 inserts and 50 searches. For example, `ray_tasks{State="PENDING_NODE_ASSIGNMENT"}` = 3, `sum(pg_stat_activity_count) / scalar(pg_settings_max_connections)` = 0.01, and `milvus_querynode_entity_num{segment_state="Sealed"}` = 5000. The kind node has no GPU, so the `Name="GPU"` series of `ray_resources` could not appear; only its `CPU` series were observed.

Samples per scrape, for the volume conversation the issue asks to have up front (one 5 000-row collection):

| Target | Samples |
|---|---|
| Milvus, all five components | ~45 000 (streaming node ~20 000) |
| Ray, head and one worker | ~1 150 |
| Postgres | ~550 |

Also run:

- `helm lint`.
- Render diff against `develop`: identical with default values; with `values-linagora.yaml`, only the RayCluster lines change.
- Each of the three `fail` guards, and the `NOTES.txt` warnings on a dry run.
- `tests/unit/infra`: 25 passed. The new tests fail against `develop`'s `raycluster.yaml`, and against the port renamed back to `ray-metrics`.

## Not in this PR

- **The API's connection-pool wait.** The pool lives inside the OpenRAG process, so Postgres cannot see a request queued for a connection. Covering it needs an application metric.
- **Slow queries.** These need `pg_stat_statements`, which means a `shared_preload_libraries` change on the server. Without it, `pg_stat_activity_max_tx_duration` still shows long-running transactions.
- **Embedded Ray** (`ray.enabled=false`): it exports on no fixed port. **MinIO and etcd** are not wired either. **GPU metrics** are #932 / #978.

## Overlap with open PRs

Test-merged against each PR's current head:

- #914 (`openrag.metrics.serviceMonitor`): the new keys follow its `<component>.metrics.<monitor>` shape. **Conflicts in `_helpers.tpl`**: both append a helper at the end of the file, so keep both.
- #976 (`monitoring.prometheusRule`): merges cleanly.
- #978: **conflicts in `kubernetes.md`**: both append a section at the end of the page, so keep both.


andyne13 (approved):
Rendered both sides with Helm rather than reading the templates.

**The headline claim holds exactly.** With default values the chart renders **identically** to the merge base — the only differing line is the Postgres password the bitnami sub-chart regenerates per render, which is the caveat you stated. That is the claim I most wanted to check on a 379-line chart change, and it is true.

Also verified:

- **The port fix is real.** Rendered `RayCluster` contains **zero** occurrences of 8080; head and workers both expose `metrics: 8090`, the head passes `--metrics-export-port=8090` in its args and the workers get it through `rayStartParams`. Unauthenticated Ray metrics on the port `externalPorts` opens to every source was a genuine exposure — good catch, and worth its own commit.
- **The PodMonitor** selects head and workers together via `ray.io/is-ray-node`, so one endpoint covers both.
- **The NetworkPolicy** is as narrow as advertised: one policy per exporter, each selecting only that exporter's pods and opening only its port (8090 Ray, 9091 Milvus).
- **The `ray.enabled` guard** fails with a precise, explanatory message rather than rendering something that selects nothing — I hit it by accident and it told me exactly what was wrong.

**One gap, and it is about what happens after S3-2 lands rather than anything wrong here.**

`ray.util.metrics` series arrive with Ray's own `ray_` prefix — `ray_openrag_ingest_documents_total`, and so on. S3-2's integration test pins this (`_PREFIX = "ray_openrag_"`), and `core/observability/ray_metrics.py` writes its worked PromQL with it. The S3-4 alert rules use the bare `openrag_*` names.

The PodMonitor here has `relabelings`, but those act on target labels, and the two present are for Ray's own Grafana dashboards. Nothing rewrites the metric names, so once S3-2 merges the series will be collected and the alerts still will not match — targets up, data in Prometheus, four rules permanently green.

Not a defect in this PR: `ray_metrics.py` does not exist on `develop` yet, so there is nothing here to relabel today. But whoever lands second inherits it, and the cheapest place to fix it is a `metricRelabelings` entry on this PodMonitor — one name across alerts, dashboards and `metric_specs.py` — rather than writing `ray_openrag_*` into every expression. Either way it needs deciding once; more detail on #931.

Adjacent and out of scope: `infra/compose/prometheus/prometheus.yml` still has no Ray job, and Compose is what loads the generated alert-rules file.

Approving — the chart work is sound and the default-render equivalence is the strong evidence for it. The branch is 23 commits behind `develop`.

docs/content/docs/documentation/kubernetes.md
```diff
@@ -108,3 +108,107 @@ In `values.yaml`, disable the bundled PostgreSQL chart, set `postgresProvisionin
...
+  The Postgres sub-chart calls the key `labels`; the Milvus one calls it `additionalLabels`.
+- **NetworkPolicy.** The default-deny policy admits only same-namespace traffic.
+  Until Prometheus's namespace is listed in `networkPolicy.metricsFrom`, its
+  targets report `up == 0`. Each entry opens only the metrics port, and only on
+  the pods that export it.
+- **Ray pods created before this change.** KubeRay does not recreate Ray pods
+  when the `RayCluster` changes, so after upgrading an existing release the
+  workers keep exporting on 8080, which `networkPolicy.externalPorts` opens to
+  every source, and the head target is down. Recreate them once:
+  `kubectl delete pod -l ray.io/cluster=openrag-raycluster` (the cluster is
```
hedhoud: This command only works with the default `fullnameOverride`. A deployment using the custom name documented earlier will select no pods, so its existing Ray workers can remain on the old metrics port after the upgrade. Please make the selector use the rendered `<fullnameOverride>-raycluster` name, or state clearly that operators must substitute their configured value.

hedhoud (commented):
One minor documentation correction is needed before approval.

docs/content/docs/documentation/kubernetes.md
Ahmath-Gadji: Fixed in 89a4e90f. The command now takes the name as a placeholder: `kubectl delete pod -n <release namespace> -l ray.io/cluster=<RayCluster name>`. The doc gives `openrag-raycluster` as the name with the default `fullnameOverride`. Otherwise `kubectl get raycluster -n <release namespace>` prints it. The upgrade warning in the PR description has the same fix.

Ahmath-Gadji:
@andyne13 I went with the relabel. The Ray PodMonitor now has one `metricRelabelings` rule on `__name__`, `ray_(openrag_.+)` → `$1`. It keeps `WorkerId`, and Ray's own `ray_*` metrics keep their names. I checked it on Prometheus 2.54.1 and 3.5.0 against a mock Ray exposition. The results are in the PR description.

Still open:
- #961's docs query `ray_openrag_*`. Whichever of #961 and this PR merges second needs to update them.
- Compose still has no Ray scrape job. When one is added, it should use the same rule.

I also rebased onto `develop`. The only conflict was in `_helpers.tpl`, and I kept both sides.

docs/content/docs/documentation/kubernetes.md
```diff
@@ -114,3 +114,114 @@ In `values.yaml`, disable the bundled PostgreSQL chart, set `postgresProvisionin
...
+  metrics:
+    enabled: true                                    # adds the exporter sidecar: restarts Postgres
+    serviceMonitor:
+      enabled: true
+      labels: { release: kube-prometheus-stack }
+milvus:
+  metrics:
+    serviceMonitor:
+      enabled: true
+      additionalLabels: { release: kube-prometheus-stack }
```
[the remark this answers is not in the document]
Ahmath-Gadji: Done in 293b1e1e: the three examples now use `release: <Prometheus release name>`, and the selector-labels warning says the example assumes kube-prometheus-stack, which selects on `release: <its Helm release name>` by default.

docs/content/docs/documentation/kubernetes.md
```diff
@@ -114,3 +114,114 @@ In `values.yaml`, disable the bundled PostgreSQL chart, set `postgresProvisionin
...
+
+### What to watch
+
+| Dependency | Query | Signal |
+|---|---|---|
+| Ray | `ray_tasks{State="PENDING_NODE_ASSIGNMENT"}` | Tasks no node has the resources to run |
+| Ray | `ray_actors{State="RESTARTING"}` | Actors being restarted. This gauge is sampled, so it catches a crash loop but can miss a single fast restart |
+| Ray | `ray_resources{Name="GPU"}` by `State` (`USED`, `AVAILABLE`) | GPU allocation per node |
+| Ray | `ray_node_mem_used`, `ray_node_cpu_utilization`, `ray_object_store_memory` | Node resources |
+| Postgres | `sum(pg_stat_activity_count) / scalar(pg_settings_max_connections)` | Connections against the server limit |
```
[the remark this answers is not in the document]
Ahmath-Gadji: Done in 293b1e1e: now `sum by (instance) (pg_stat_activity_count{namespace="<release namespace>"}) / sum by (instance) (pg_settings_max_connections{namespace="<release namespace>"})`. I left out the `job` selector and `on (instance)`: grouping by instance already keeps servers apart, and both sides end up with the same labels. Checked on Prometheus 3.5.0 with two mock exporters in two namespaces: the old query returns NaN, the new one 0.13 for the selected namespace.

hedhoud (dismissed):
The upgrade guidance now handles custom RayCluster names, the monitoring examples are safe for shared Prometheus deployments, and all checks pass. No blocking issues remain.

docs/content/docs/documentation/kubernetes.md
```diff
@@ -156,3 +156,116 @@ scrape it and Grafana to show it.
...
+|---|---|---|
+| Ray | `ray_tasks{State="PENDING_NODE_ASSIGNMENT"}` | Tasks no node has the resources to run |
+| Ray | `ray_actors{State="RESTARTING"}` | Actors being restarted. This gauge is sampled, so it catches a crash loop but can miss a single fast restart |
+| Ray | `ray_resources{Name="GPU"}` by `State` (`USED`, `AVAILABLE`) | GPU allocation per node |
+| Ray | `ray_node_mem_used`, `ray_node_cpu_utilization`, `ray_object_store_memory` | Node resources |
+| Postgres | `sum by (instance) (pg_stat_activity_count{namespace="<release namespace>"}) / sum by (instance) (pg_settings_max_connections{namespace="<release namespace>"})` | Connections against the server limit, per server |
+| Postgres | `pg_stat_activity_max_tx_duration` | Longest open transaction |
+| Postgres | `pg_database_size_bytes`, `pg_locks_count` | Database size, lock contention |
+| Milvus | `milvus_proxy_req_latency`, `milvus_proxy_sq_latency` | Request and search/query latency |
+| Milvus | `milvus_proxy_insert_vectors_count`, `milvus_proxy_search_vectors_count` | Insert and search throughput |
```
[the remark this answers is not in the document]
Ahmath-Gadji: Done in dae59804. The latency row now reads `histogram_quantile(0.99, sum by (le, function_name) (rate(milvus_proxy_req_latency_bucket{namespace="<release namespace>"}[5m])))`, and the same over `milvus_proxy_sq_latency_bucket` by `query_type`. The throughput row applies `sum(rate(...[5m]))` to both vector counters. Both are scoped to the release namespace, like the Postgres row.

I checked it against a live Milvus v3.0.1 scraped by Prometheus 3.5.0. The bare `milvus_proxy_req_latency` returns no series. The new queries return a p99 per `function_name` and `query_type` (the buckets are in milliseconds) and a per-second rate for inserts and searches.

infra/charts/openrag-stack/values.yaml
```diff
@@ -148,6 +175,17 @@ postgresql:
       storageClass: *storageClass
   service:
     port: &pgPort 5432
+  # postgres_exporter sidecar (port 9187) and its Service. Turning it on changes
+  # the StatefulSet's pod template, so the upgrade restarts Postgres.
+  metrics:
+    enabled: false
```
EnjoyBacon7: **`postgresql.metrics.enabled: true` opens the exporter to every namespace, not only to `networkPolicy.metricsFrom`.**

The bitnami sub-chart renders its own NetworkPolicy by default (`primary.networkPolicy.enabled: true`, `allowExternal: true`). Once the sidecar is on, that policy adds 9187 to its port list with no `from`. Rendered with `helm template` against postgresql 18.10.0:

```yaml
# Source: openrag-stack/charts/postgresql/templates/primary/networkpolicy.yaml
  ingress:
    - ports:
        - port: 5432
        - port: 9187
```

NetworkPolicies are additive, so this admits 9187 from any source, whatever `openrag-default-deny` and `openrag-postgresql-metrics` say. With default values the list is only 5432.

So for Postgres, three statements in this PR do not hold:
- the `metricsFrom` comment in values.yaml ("only to these peers"),
- the docs line "Each entry opens only the metrics port, and only on the pods that export it",
- the NOTES.txt warning ("only a Prometheus running in namespace ... can reach these exporters").

The exporter is unauthenticated and exposes `pg_settings_*`, database names and sizes, and connection activity.

Suggested fix: set `postgresql.primary.networkPolicy.enabled: false` in values.yaml and let the chart's own policies govern Postgres like every other pod. `allowExternal: false` would also close it, but it then only admits pods labelled `openrag-postgresql-client: "true"`, which breaks the API if someone sets `networkPolicy.enabled=false`. Either way it also stops 5432 being reachable from every namespace, which is already the case on `develop`. A test asserting the chosen value would keep it from regressing.

Ahmath-Gadji:
#961 merged first, so this PR updates its docs: ead71a32 changes `metrics_reference.md`, `prometheus_metrics.md` and the `ray_metrics.py` docstrings. They no longer describe the Ray target as unscraped, and their example queries use the `openrag_*` names the PodMonitor produces. The compose overlay still has no Ray job. I also merged `develop` in (0f0c02b5): the only conflicts were in `NOTES.txt` and `values.yaml`, where #1021 and this PR each added a block, and I kept both.

infra/charts/openrag-stack/templates/NOTES.txt
```diff
@@ -60,6 +60,33 @@ Collector-side parsing stages: docs/documentation/loki_logs.md
...
+{{- range $key, $labels := $monitors }}
+{{- if not $labels }}
+⚠  {{ $key }} is empty. A Prometheus that selects monitors by label
+   (kube-prometheus-stack matches release: <its release name>) ignores it without any error.
+{{- end }}
+{{- end }}
+{{- if and .Values.networkPolicy.enabled (not .Values.networkPolicy.metricsFrom) }}
+⚠  networkPolicy.metricsFrom is empty, so only a Prometheus running in namespace
+   "{{ .Release.Namespace }}" can reach these exporters; anywhere else their targets report up == 0.
+{{- end }}
```
EnjoyBacon7: **With `monitoring.bundled: true` (from develop), all four of these warnings fire and ask for changes the install does not need.**

The bundled Prometheus runs in the release namespace and selects every monitor in the cluster (`podMonitorSelector: {}`, `serviceMonitorSelector: {}`, as the "Every monitor is scraped" bullet in kubernetes.md already says). So an unlabelled PodMonitor or ServiceMonitor is scraped, and an empty `metricsFrom` is not a problem either.

Output of `helm install --dry-run=client` with `monitoring.bundled: true` and the three monitors on, no labels, no `metricsFrom`:

```
Datastore metrics
-----------------
⚠  milvus.metrics.serviceMonitor.additionalLabels is empty. A Prometheus that selects monitors by label
   (kube-prometheus-stack matches release: <its release name>) ignores it without any error.
⚠  postgresql.metrics.serviceMonitor.labels is empty. ...
⚠  ray.metrics.podMonitor.labels is empty. ...
⚠  networkPolicy.metricsFrom is empty, so only a Prometheus running in namespace
   "rag" can reach these exporters; anywhere else their targets report up == 0.
```

A user following them would add a `release:` label for a Prometheus that does not need one, which trains people to ignore these warnings when they do matter.

Suggestion: skip both checks when `.Values.monitoring.bundled` is on. In the same spirit, the "Monitoring Ray, Postgres and Milvus" section opens with "The chart does not deploy Prometheus", which is no longer true on develop. One sentence there saying that with `monitoring.bundled` only the three `enabled` switches are needed (no labels, no `metricsFrom`) would cover it, since bundled mode turns on the API monitor but not these.

infra/charts/openrag-stack/values.yaml
EnjoyBacon7: Fixed by #1009: `postgresql.primary.networkPolicy.enabled: false`, with a test and a docs note. Checked on 56f909e0.

andyne13 (changes requested):
Checked on `56f909e0` by rendering the chart (defaults, `ray.enabled`, `values-linagora.yaml`, all monitors on with `metricsFrom` set) and diffing against `develop`. The wiring holds: both ServiceMonitors match their Services by label on ports that exist, head and workers export on a port named `metrics` (8090, workers via `rayStartParams`), each metrics policy opens one port only to `metricsFrom`, and #1009 removes the bitnami policy that opened 5432 to every source. Three things before I'd approve:

1. **The Postgres exporter can lose its credentials on template-rendered deploys.** It authenticates as `postgres` with the `postgres-password` key of Secret `openrag-postgresql` (`DATA_SOURCE_USER` falls back to `postgres` when `auth.metricsUsername` is unset). This chart sets no `postgresql.auth.postgresPassword`, so the sub-chart generates a new one on every render: two `helm template` runs give different values. Our `values.yaml` already notes that Argo CD renders without cluster lookups, and the sub-chart only changes a live password through `passwordUpdateJob`, which is off by default. So under Argo CD or `helm template | kubectl apply`, a sync after install can write a Secret that no longer matches the server's superuser password: the exporter then fails to authenticate, and nothing fails at install. Requiring `postgresql.auth.postgresPassword` (or `auth.existingSecret`) when `postgresql.metrics.enabled` is on would prevent it.
2. **Two changes that apply to every install are documented only in the opt-in monitoring section.** The worker port moves to 8090 whether or not a monitor is on, and the PR measured that existing Ray pods keep their UIDs after `helm upgrade`, so workers stay on 8080, which `externalPorts` opens to every source. Likewise, 5432 is no longer reachable from other namespaces whenever `networkPolicy.enabled` is on (the default). Both are explained under "Three things can go wrong" in the monitoring section, and `NOTES.txt` only renders its datastore block when a monitor is on, so an operator upgrading without monitoring sees neither. The description's "with default values the rendered manifests match `develop` exactly" no longer holds since #1009. Separately, the documented `kubectl delete pod -l ray.io/cluster=…` also deletes the head. The chart configures no GCS fault tolerance, so that restarts the cluster: the detached actors (TaskStateManager, the indexer pool, the parser pools) are lost, and under `values-linagora.yaml` so is Ray Serve on the head's port 80. The head already exported on 8090 on `develop`, so recreating only the worker pods closes the exposure; the head needs recreating only for its own scrape target.
3. **Read replicas stay open.** With `postgresql.architecture: replication`, the sub-chart still renders `openrag-postgresql-read` with ingress on 5432 and 9187 and no `from`: only `primary.networkPolicy` is disabled, and `readReplicas.networkPolicy` defaults to `enabled: true, allowExternal: true`. `readReplicas.networkPolicy.enabled: false` closes it.

Non-blocking: `test_helm_datastore_metrics.py` reads the templates as text, so deleting the `ray.metrics.podMonitor` guard, changing the Milvus policy port to 9092, or misspelling its selector all leave its 7 tests passing; rendering under `requires_helm`, as `test_monitoring_delivery.py` does, would catch them. In "What to watch", only the Postgres ratio and the Milvus latency/throughput rows are scoped to the release namespace; the `ray_*`, `pg_stat_activity_max_tx_duration`, `pg_database_size_bytes`, `pg_locks_count`, `milvus_*coord*` and `process_resident_memory_bytes` rows include every release a shared Prometheus scrapes.

infra/charts/openrag-stack/templates/NOTES.txt
Ahmath-Gadji: Done in fd8db02e. Under `monitoring.bundled`, the Datastore metrics block now skips both checks and prints only the targets line. `test_the_notes_do_not_warn_under_the_bundled_prometheus` renders your case (the three monitors on, no labels, no `metricsFrom`) and fails if either warning comes back.

The section no longer opens with "The chart does not deploy Prometheus". It says that with the bundled stack only the `enabled` switches are needed, plus the Postgres superuser password that `postgresql.metrics.enabled` now requires (85f3c849), and that `monitoring.bundled` does not turn them on.

Ahmath-Gadji:
@andyne13 All three reproduced when I rendered them. The fixes are separate commits on top: the branch carries #1009's merge, so I haven't force-pushed. `develop` is merged in as well.

1. **Postgres exporter credentials**, 85f3c849. Reproduced first: two `helm template` runs gave two different `postgres-password` values, with `DATA_SOURCE_USER=postgres`. `postgresql.metrics.enabled` now fails the render unless `postgresql.auth.postgresPassword` or `auth.existingSecret` is set (the `global.postgresql.auth` forms count too). The error and the docs give `kubectl get secret -n <ns> openrag-postgresql -o jsonpath='{.data.postgres-password}' | base64 -d`, because a running release has to set the password the server already has, not a new one.
2. **Changes that apply to every install**, fd66a5d2. The chart is now 0.6.7, and `kubernetes.md` has an "Upgrading to chart 0.6.7" section next to the 0.6.0 one, covering the worker port and the Postgres isolation. On upgrade, `NOTES.txt` prints both whether or not a monitor is on, with the RayCluster name filled in. The recreate command is now `-l ray.io/cluster=<name>,ray.io/node-type=worker`, so it leaves the head alone. The docs say why: no GCS fault tolerance, so deleting the head restarts the cluster. They also say the head needs recreating only for its own scrape target, in a maintenance window. I've corrected the description: with default values, the only rendered difference from `develop` is now the removed bitnami policy, and with `values-linagora.yaml` the RayCluster lines change too. Its upgrade warning uses the worker-only command.
3. **Read replicas**, e8d2d615: `postgresql.readReplicas.networkPolicy.enabled: false`. With `architecture: replication`, the render now has only `openrag-default-deny`.

Non-blocking:
- **Rendered tests**, c9c00046, under `requires_helm`. The policy and replication tests also need the postgresql and milvus archives from `helm dependency build`. They check that:
  - each metrics policy selects at least one pod, and every pod it selects declares the port it opens;
  - the Ray PodMonitor selects head and workers on the port each one's `ray start` exports on;
  - each `fail` guard and the NOTES behave as intended.

  Your three cases each fail a test now: the podMonitor guard deleted, Milvus on 9092, and a misspelled Milvus selector. So do the read replicas reopened, the password guard dropped, the workers back on 8080, and an upgrade command that deletes the head.
- **Namespace scoping**, 8b5be796: every row of "What to watch" now names the release namespace. The memory row also needs `component!=""`. Among this chart's monitors, only Milvus's adds that target label. Without it, the row summed every exporter's `process_resident_memory_bytes` in the namespace.

EnjoyBacon7's `monitoring.bundled` point on `NOTES.txt` is fixed in fd8db02e (reply on that thread).

Checks: `tests/unit` 4413 passed, `tests/unit/infra` passes at each of the six commits, and `helm lint` is clean.

PR #980, opened 2026-09-17
feat: add secure retrieval observability

hedhoud:
## Context

Retrieval benchmarks can show that quality regressed, but final rankings alone do not explain whether the loss came from contextualization, thresholding, hybrid fusion, reranking, or the final cutoff. That makes benchmark tuning slow and difficult to reproduce.

## Why this is useful

This adds opt-in retrieval telemetry that lets benchmark tooling inspect the exact query transformation, candidate progression, rank changes, timings, and effective retrieval configuration from the original request. Developers can identify the first stage that lost a relevant passage, compare pre-rerank and post-rerank behavior, and verify that two runs used compatible index and model settings.

Benchmarks can explicitly bypass query contextualization for one traced request. This makes contextualizer-on versus contextualizer-off evaluation honest: the original-query variant never invokes the rewriting model and uses the exact latest user message, while the rest of the retrieval pipeline stays unchanged.

A benchmark may also apply a request-scoped similarity threshold and retrieval depth. This supports fair threshold sweeps across original and contextualized queries without changing the partition preset or production defaults. Reranking and expansion can be disabled for these diagnostic requests so the sweep measures threshold behavior rather than several changing stages at once.

The snapshot endpoint gives benchmark runs a stable configuration and index identity, so dashboard results can be explained later without rerunning retrieval.

## Security and operational safeguards

Diagnostics are administrator-only, separately rate-limited, and bounded to prevent expensive or oversized requests. Traces exclude document bodies, prompt content, credentials, embeddings, and internal endpoint addresses. Candidate lists, raw search depth, related-document expansion, and optional document-ID snapshots all have explicit limits.

All overrides are request-scoped and require both tracing and forced retrieval. Normal retrieval behavior and configuration remain unchanged when they are omitted.

## Validation

- The focused schema, orchestration, retrieval, and diagnostics suite passes (262 tests).
- The deterministic retrieval-trace integration contract passes.
- Ruff checks pass.
- CodeRabbit raised no issues on the contextualizer-bypass change.
- Independent security follow-up review found no remaining critical or important issues.

## Dependency

This draft targets fix/required-retrieval because the observability flow depends on its retrieval behavior. It can be rebased onto develop after that prerequisite lands.


openrag/services/persistence/document_repo.py
```diff
@@ -141,6 +144,36 @@ async def list_indexed_documents(
...
+                    partition,
+                    prefetch=1000,
+                )
+                async for row in rows:
+                    indexed_at = row["indexed_at"]
+                    identity = [
+                        row["file_id"],
+                        indexed_at.isoformat() if indexed_at is not None else None,
+                        row["content_sha256"],
+                        row["chunk_count"],
```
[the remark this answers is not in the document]
hedhoud: Fixed. The corpus identity now includes relationship and parent links, so expansion-affecting metadata changes invalidate the snapshot fingerprint.

openrag/services/orchestrators/retrieval_snapshot_service.py
```diff
@@ -0,0 +1,144 @@
...
+            "indexed_corpus_count": corpus_state.count,
+            "indexed_corpus_digest": corpus_state.digest,
+            "created_at": self._json_value(detail.get("created_at")),
+        }
+        index: dict[str, object] = {
+            **index_base,
+            "fingerprint": canonical_fingerprint(index_base),
+        }
+        if include_document_ids:
+            document_ids, truncated = await self._document_ids(partition)
```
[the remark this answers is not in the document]
hedhoud: Fixed. Optional document IDs are now collected with count and digest inside the same repeatable-read transaction, remain bounded, and report truncation without a second catalog read.

openrag/core/retrieval/trace.py
```diff
@@ -0,0 +1,306 @@
...
+    if isinstance(value, Enum):
+        return _safe_public_value(value.value, context=context, key=key)
+    if isinstance(value, (datetime, date)):
+        return value.isoformat()
+    if isinstance(value, float) and not math.isfinite(value):
+        return _OMITTED
+    if value is None or isinstance(value, (bool, int, float, str)):
+        if key in {"error", "message"} and isinstance(value, str):
+            return REDACTED_ERROR_MESSAGE
+        return value
```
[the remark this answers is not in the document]
hedhoud: Fixed. Query telemetry now redacts embedded Bearer, Basic, and named token credentials, including quoted multiline values, before serialization. Regression coverage exercises original, contextualized, and child queries.

openrag/services/orchestrators/query_service.py
```diff
@@ -588,25 +725,66 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
...
         web_results: list = []
         if partition is not None and use_websearch:
-            chunks, web_lists = await self._gather_rag_and_web(queries, partition, top_k, filter_params)
+            chunks, web_lists = await self._gather_rag_and_web(
+                queries,
+                partition,
+                top_k,
+                filter_params,
+                trace=trace,
+            )
```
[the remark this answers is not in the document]
hedhoud: Fixed on the latest head. The web-search document branch now receives the same allowlisted retrieval overrides as the normal retrieval path, so the applied threshold, reranker, and expansion settings match the trace. The regression test covers all supported overrides.

openrag/services/orchestrators/retrieval_snapshot_service.py
```diff
@@ -0,0 +1,127 @@
...
+                "model",
+                "enabled",
+                "top_n",
+            ),
+            "contextualizer": self._keys(
+                partition_config.get("contextualizer"),
+                "name",
+                "model",
+                "prompt_name",
+            ),
```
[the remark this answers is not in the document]
hedhoud: Fixed on the latest head. The snapshot now includes only the resolved prompt name, source, and content hash, and snapshot, chat-trace, and raw-search fingerprints use the same prompt-aware identity. Prompt content is never serialized, and changing content under the same name changes both snapshot fingerprints.

hedhoud:
Addressed the final CodeRabbit follow-up: when one partition group fails and another survives, both retrieval paths now record the surviving candidates in the final trace stage. Returned documents are unchanged; the diagnostics now remain complete. Regression coverage exercises both paths.

hedhoud:
@coderabbitai review

hedhoud:
Fixed the remaining fingerprint gap. Retrieval snapshots now account for RRF and query-expansion prompt selection, while raw search traces include the effective request depth, threshold, and expansion settings. This prevents benchmark runs with different retrieval behavior from appearing equivalent. The full unit suite and Ruff checks pass locally.

hedhoud:
@coderabbitai review

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -170,6 +186,166 @@ async def _resolve_query_template(self, prompt_type: str, name: str | None, disk
...
+                    "name": partition_name,
+                    "embedder": self._public_endpoint(self._config, "embedder", embedder_name),
+                    "retrieval": {
+                        "type": getattr(retrieval, "type", None),
+                        "top_k": getattr(retrieval, "top_k", None),
+                        "similarity_threshold": getattr(retrieval, "similarity_threshold", None),
+                        "rrf_k": getattr(retrieval, "rrf_k", 60),
+                        "hyde_prompt_name": getattr(retrieval, "hyde_prompt_name", None),
+                        "multi_query_prompt_name": getattr(retrieval, "multi_query_prompt_name", None),
+                    },
```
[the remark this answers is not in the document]
hedhoud: Fixed in d334cc41. The fingerprint and snapshot now include these effective pipeline settings, so changing any of them produces a different configuration identity. Regression coverage was added for each setting; the full unit suite and Ruff checks pass.

hedhoud:
@coderabbitai review

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -170,6 +186,170 @@ async def _resolve_query_template(self, prompt_type: str, name: str | None, disk
...
+                        "type": getattr(retrieval, "type", None),
+                        "top_k": getattr(retrieval, "top_k", None),
+                        "similarity_threshold": getattr(retrieval, "similarity_threshold", None),
+                        "rrf_k": getattr(retrieval, "rrf_k", 60),
+                        "k_queries": self._legacy_retriever_value("k_queries", 3),
+                        "combine": self._legacy_retriever_value("combine", False),
+                        "with_surrounding_chunks": self._legacy_retriever_value("with_surrounding_chunks", False),
+                        "allow_filterless_fallback": self._legacy_retriever_value("allow_filterless_fallback", True),
+                        "hyde_prompt_name": getattr(retrieval, "hyde_prompt_name", None),
+                        "multi_query_prompt_name": getattr(retrieval, "multi_query_prompt_name", None),
```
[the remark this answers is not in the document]
hedhoud: Fixed in 02a13630. HyDE and multi-query fingerprints now use the effective resolved prompt identity (name, source, and content hash) without storing prompt text. Regression coverage confirms that a same-name prompt content change updates the fingerprint; legacy resolvers no longer claim provenance they cannot provide.

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -170,6 +186,170 @@ async def _resolve_query_template(self, prompt_type: str, name: str | None, disk
...
+            "model": getattr(endpoint, "model_name", None),
+        }
+
+    def public_retrieval_configuration(self, partitions: Sequence[str]) -> dict[str, object]:
+        """Build the stable public retrieval-setting subset used for fingerprints."""
+        configured_partitions = self._partition_configs()
+        public_partitions: list[dict[str, object]] = []
+        for partition_name in sorted(set(partitions)):
+            partition = configured_partitions.get(partition_name)
+            retrieval = partition.retrieval if partition is not None else self._config.retriever
```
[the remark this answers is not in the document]
hedhoud: Fixed in 02a13630. The wildcard now fingerprints the same concrete partition set used by retrieval, while deployments without configured partitions retain the legacy fallback. Regression coverage confirms that changing one selected partition changes the wildcard fingerprint.

hedhoud:
@coderabbitai review

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -156,20 +172,231 @@ def _build_retriever(
...
+            content = await self._prompt_service.resolve_prompt("query_contextualizer", names=[prompt_name])
+            source = "named" if prompt_name else "default"
+        else:
+            content = load_template_by_key(
+                self._config.paths.prompts_dir,
+                self._config.prompts,
+                "query_contextualizer",
+            )
+            source = "disk-seed"
+            prompt_name = None
```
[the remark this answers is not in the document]
hedhoud: Fixed in 4cb0ede6. Legacy contextualizer resolvers now keep the resolved content hash but leave name and source unknown, because content-only resolution cannot prove whether the requested prompt or a fallback was used. The regression covers a missing requested name resolving to default content.

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -156,20 +172,231 @@ def _build_retriever(
...
+        }
+
+    @staticmethod
+    def _public_prompt_identity(prompt: object) -> dict[str, str | None]:
+        source = prompt if isinstance(prompt, dict) else vars(prompt)
+        return {
+            "name": source.get("name"),
+            "source": source.get("source"),
+            "content_hash": source.get("content_hash"),
+        }
```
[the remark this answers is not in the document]
hedhoud: Fixed in 4cb0ede6. Prompt identity extraction now supports mappings and attribute-based objects, including the production slotted ResolvedPrompt type. The regression exercises the real resolved-configuration path so snapshot and fingerprint generation cannot fail on this type.

hedhoud:
@coderabbitai review

openrag/services/orchestrators/query_service.py
```diff
@@ -684,19 +874,136 @@ async def _existing_file_ids(self, file_ids: list[str], partitions: list[str]) -
...
+            "stages": finished["stages"],
+            "timings": finished["timings"],
+            "errors": finished["errors"],
+            "configuration_fingerprint": finished["configuration_fingerprint"],
+        }
+
+    def _effective_retrieval_options(self, metadata: dict) -> dict[str, object]:
+        options: dict[str, object] = {}
+        top_k = metadata.get("retrieval_top_k")
+        if top_k is None and metadata.get("use_map_reduce") is True:
```
Ahmath-Gadji: This changes retrieval for regular map-reduce requests, not only traced ones. Before, `use_map_reduce` set the *final* cutoff (`max_total_documents`, 20) and the retriever still searched at the preset's `top_k`. Now the same value goes through `top_k` into `_pipeline_groups_for_partitions` → `_pipeline_for_partition`, where it replaces `pipeline_cfg.top_k` as the Milvus search depth (`retrieval_service.py:524`, and `:112` on the legacy pipeline).

With the default presets (`top_k=50`, `top_n=10`, reranker on), calling `retrieve_multi(..., top_k=20)` the way map-reduce does:

| | Milvus depth | reranker input | returned |
|---|---|---|---|
| `develop` | 50 | 50 | 20 |
| this branch | 20 | 20 | 20 |

Chainlit's DeepSearch command sends `use_map_reduce: True` (`app_front.py:705`), so every DeepSearch query now reranks 20 candidates instead of 50. On deployments without per-partition configs it also builds a new legacy pipeline on each such request instead of reusing `self._pipeline`.

I'd keep the two apart: `retrieval_top_k` as a search-depth override that is the only thing reaching pipeline construction, and map-reduce's cap as the final cutoff passed to `retrieve_docs`, as before.

openrag/core/retrieval/pipeline.py
```diff
@@ -156,19 +284,35 @@ async def get_relevant_docs(
         search_queries: SearchQueries,
         top_k: int | None = None,
         filter_params: dict | None = None,
+        trace: RetrievalTraceBuilder | None = None,
     ) -> list[Chunk]:
         """Run every sub-query in parallel and fuse the per-query rankings via RRF."""
+        query_trace = trace if len(search_queries.query_list) == 1 else None
```
Ahmath-Gadji: With more than one sub-query the trace isn't passed down, so the per-query stages never get recorded. Same one level up with more than one partition (`retrieval_service.py:748` and `:791`, `pipeline_trace = trace if len(groups) == 1 else None`).

Running `RetrievalService.retrieve_multi` with a trace and a searcher that records the store stages:

| request | reranker calls | `dense_*`, `sparse`, `hybrid_fused`, `pre_rerank`, `post_rerank` | `query_traces` |
|---|---|---|---|
| 1 partition, 1 sub-query | 1 | `complete` | 0 |
| 1 partition, 2 sub-queries | 2 | `not_run` | 0 |
| 2 partitions, 1 sub-query | 2 | `not_run`, except `hybrid_fused` (see the comment on `fuse`) | 0 |

The contextualizer splits compound questions into several sub-queries, and `openrag-all` spans partitions, so these are common chat paths. On them the trace reports stages that did run as `not_run`, which is the opposite of what's needed to find the stage that lost a passage.

`VectorStoreSearcher.multi_query_search` already handles this with one child trace per query plus `_merge_store_traces`; doing the same here (per sub-query, per partition group) would fill `query_traces`. If that's out of scope for this PR, marking those stages `unavailable` rather than leaving them `not_run` would at least not mislead.

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -491,28 +818,66 @@ async def retrieve_per_query(
...
         single partition's ``rrf_k`` applies. Per-partition ``rrf_k`` is honoured
         one layer down, in ``RetrieverPipeline.get_relevant_docs`` (#707).
         """
-        fused = rrf_reranking(doc_lists, key_fn=_chunk_key)
-        return fused[:top_k] if top_k is not None else fused
+        fused = rrf_reranking(
+            doc_lists,
+            key_fn=_chunk_key,
+            top_k=top_k,
+            trace=trace,
```
Ahmath-Gadji, line 898: `fuse()` merges lists from different partitions, but it records under `rrf_reranking`'s default stage name, `hybrid_fused`, which is the dense+sparse fusion stage. On a multi-partition request it's the only writer of that stage (the store never receives the trace there), so `hybrid_fused` shows `complete` with cross-partition RRF candidates, even on a dense-only deployment where no hybrid fusion happened.

It's the same mislabel CodeRabbit raised for the pipeline, which now uses `multi_query_fused`; this call site needs its own stage name as well (e.g. `partition_fused`). `test_fuse_trace_records_duplicates_scores_and_public_cutoff` currently asserts the `hybrid_fused` name.

openrag/services/orchestrators/query_service.py
```diff
@@ -796,22 +1103,32 @@ async def chat(
         model_name: str,
     ) -> dict:
         """Non-streaming chat completion → finalized OpenAI dict."""
+        request_started = time.perf_counter()
         metadata = payload.get("metadata") or {}
         include_all_retrieved = metadata.get("include_all_retrieved_sources") is True
+        include_trace = metadata.get("include_retrieval_trace") is True
+        original_query = _latest_user_query(payload.get("messages", []))
+        trace = (
+            RetrievalTraceBuilder(request_id=str(uuid.uuid4()), original_query=original_query)
```
Ahmath-Gadji: The chat trace gets a fresh `uuid4()` here (and at `:1182` for streaming), while the search routes use `request.state.request_id` from `RequestIdMiddleware`. So a chat trace can't be matched to its log lines or to the `X-Request-ID` response header. Passing the request id from the router into `chat` / `chat_stream` would make both paths consistent.

openrag/core/retrieval/trace.py
```diff
@@ -0,0 +1,341 @@
...
+    "removal_reason": "removal_reason",
+    "errors": "trace_error",
+    "error": "trace_error",
+    "timings": "timings",
+    "comparisons": "comparisons",
+    "query_traces": "query_trace",
+}
+_OMITTED = object()
+_REDACTED_CREDENTIAL = "[redacted]"
+_AUTH_SCHEME_CREDENTIAL = re.compile(
```
Ahmath-Gadji, line 165: This redaction also fires on ordinary questions. `_scrub_query_credentials` on a few:

| query | in the trace |
|---|---|
| `What is the basic idea of RAG?` | `What is the basic [redacted] of RAG?` |
| `Explain basic authentication` | `Explain basic [redacted]` |
| `How do I reset my password: step by step?` | `How do I reset my password: [redacted] by step?` |
| `What does the token: limit mean for the API?` | `What does the token: [redacted] mean for the API?` |
| `Bearer bonds vs stocks` | `Bearer [redacted] vs stocks` |

A benchmark that joins traces back to its dataset by query text, or compares the original and contextualized queries, gets altered strings.

The protection it buys is also thin: the trace is admin-only and goes back to whoever sent the query, and the same text already reaches the LLM and the DEBUG request log. I'd drop the scrubber, or at most keep it for token-shaped values (a JWT, `or-` + hex, a long opaque run after `Bearer`) rather than any word following `basic`, `token:` or `password:`.

openrag/api/routers/user/search.py
```diff
@@ -50,14 +59,18 @@ class CommonSearchParams:
     def __init__(
         self,
         text: str = Query(..., description="Text to search semantically"),
-        top_k: int = Query(5, ge=1, description="Number of top results to return"),
+        top_k: int = Query(5, ge=1, le=200, description="Number of top results to return"),
```
Ahmath-Gadji: These limits apply to every `/search` caller, not only traced requests, so they're API changes rather than diagnostics safeguards:
- `top_k` above 200 now returns 422 (it was unbounded);
- `related_limit` above 100 and `max_ancestor_depth` above 50 return 422;
- `max_ancestor_depth=None` now means 50 rather than unlimited (`bounded_ancestor_depth`), while `API.mdx:506` still says "None means unlimited";
- a configured `retriever.related_limit` above 100 is clamped silently in `BaseRetriever.__init__`, and expansion stops at 1,000 chunks.

The limits also don't line up with each other: presets accept `top_k` up to 1000 (`retrieval_pipeline.py:24`), chat's `retrieval_top_k` stops at 100, and `/search` at 200. A partition configured at `top_k=300` can't be reproduced or swept through the override, and evaluating `/search` at a larger depth is no longer possible.

I'd align the admin-only override with the preset bound (1000), and make the `/search` limit a config value the way MCP does with `mcp.max_top_k` (`OPENRAG_MCP_MAX_TOP_K`) rather than a hard-coded 200. Either way this needs the `breaking-change` label and a release-note line.

Ahmath-Gadji:
A few design points beyond the inline comments.

**Surface area.** This adds seven chat metadata keys (`include_retrieval_trace`, `compare_original_query`, `bypass_query_contextualization`, `retrieval_similarity_threshold`, `retrieval_top_k`, `retrieval_disable_reranker`, `retrieval_disable_expansion`) with cross-key rules between them, a trace flag on three search routes, a snapshot route with `include_document_ids`, and two env vars. Because the diagnostics live on `/v1/chat/completions`, every benchmark query also pays for a full LLM answer it then discards, and the knobs are untyped metadata kept consistent by model validators.

I'd move this to one admin-only retrieval dry-run route (contextualize → retrieve → rerank, no generation) with a typed body, e.g. `contextualize: bool` plus an `overrides` object. That leaves the OpenAI-compatible schema untouched and removes most of the validator rules. Within that:
- `compare_original_query` can go: two calls, with and without contextualization, give the same comparison without the shadow retrieval and without the mutual-exclusion rule against `bypass_query_contextualization`;
- `include_document_ids` can go unless something consumes it: `indexed_corpus_count` and `indexed_corpus_digest` already tell whether the corpus changed.

Which of these options does the benchmark actually use? I couldn't find a consumer in the repo; shipping only those would shrink this considerably.

**Contract.** Trace schema v1 publishes internal stage names (`dense_before_threshold`, `hybrid_fused`, `multi_query_fused`, `pre_rerank`, …) and removal codes as a versioned API, and `trace` plus four overrides are threaded through the router, query service, retrieval service, pipeline, retriever, catalog searcher, vector-store searcher and Milvus store. Any later change to how retrieval is staged becomes a contract change for the benchmark tool. Marking the trace as experimental would avoid that.

**Smaller things:**
- `OPENRAG_COMMIT` isn't set by the Dockerfile, compose or the chart, so the snapshot's `openrag.commit` is always `null`.
- `RETRIEVAL_DIAGNOSTICS_RATE_LIMIT` is documented in `env_vars.md` but missing from `infra/compose/.env.example` (and its copy, `docs/assets/env_example.env`).
- The description still says this is a draft targeting `fix/required-retrieval`; the base is `develop` and `require_retrieval` is already there.

@andyne13 could you take a look at this PR as well?

openrag/core/retrieval/pipeline.py
hedhoud: Fixed. Multi-query and multi-partition fan-out now use isolated child traces and persist them in query_traces. Aggregate stages are marked unavailable when their measurements only exist in child traces, so they no longer appear as not_run.

openrag/services/orchestrators/retrieval_service.py
hedhoud: Fixed. Cross-partition RRF now records partition_fused, leaving hybrid_fused exclusively for dense and sparse fusion.

openrag/services/orchestrators/query_service.py
hedhoud: Fixed. Both streaming and non-streaming chat paths now pass the middleware request ID into the trace, so it matches logs and the X-Request-ID response header.

openrag/api/routers/user/search.py
hedhoud: Fixed the accidental behavior changes: expansion parameters keep their previous ranges, None remains unlimited, configured values are no longer silently clamped, and expansion output is not truncated. Search and admin-only chat overrides now share the existing preset ceiling of 1000. The remaining raw-search ceiling is intentional resource hardening, so I added the breaking-change label and updated the API contract.

openrag/services/orchestrators/query_service.py
hedhoud: Fixed. Map-reduce now uses max_total_documents only as the final result cutoff, while the configured retrieval depth remains unchanged. An explicit retrieval_top_k still overrides both search depth and the requested diagnostic result depth.

openrag/core/retrieval/trace.py
hedhoud: Fixed. Unquoted values are now redacted only when they are credential-shaped, while quoted named credentials remain protected. The ordinary-language examples now round-trip exactly and the existing credential-leak tests still pass.

hedhoud:
@coderabbitai review

openrag/core/retrieval/pipeline.py
Ahmath-Gadji: Thanks, the flat cases work now. Stages are still lost when fan-outs nest: `merge_child_traces` (`trace.py:304`) forwards a child's own `query_traces` and drops the child's stages whenever it has any.

Two paths where that happens, both run with a trace against `f3260d6b`:
- a partition on the `multiQuery` (or `hyde`) preset, with a question the contextualizer splits in two: the reranker runs twice, but no trace anywhere has `pre_rerank`, `post_rerank` or `final`. The top level marks them `unavailable`, and `query_traces` only holds the four store stages of each generated variant (`a-x`, `a-y`, `b-x`, `b-y`);
- two partitions × two sub-queries, on the default preset too: each partition's `multi_query_fused` and its `top_n` cut (`final`) are dropped, so there's no way to see what each partition contributed to `partition_fused`.

The rerank step is the one most needed to find where a passage was lost. Keeping each child as a node with its own stages and nesting its `query_traces` under it, instead of flattening them into the parent, would keep every level.

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -451,23 +765,50 @@ async def retrieve(
...
+        groups = await self._pipeline_groups_for_partitions(
+            partitions,
+            top_k=retrieval_top_k,
+            similarity_threshold=similarity_threshold,
+            disable_reranker=disable_reranker,
+            disable_expansion=disable_expansion,
+        )
+        child_traces = (
+            [
+                RetrievalTraceBuilder(f"{trace.request_id}:partition:{index}", trace.original_query)
```
Ahmath-Gadji: The per-partition child traces can't be told apart in the output. `QueryRetrievalTrace` only carries `query`, `stages`, `timings` and `errors`, so with two partitions and one sub-query `query_traces` holds two entries with the same `query`, and with two sub-queries it's `a, b, a, b`, with nothing saying which partition each came from.

Here the child is also named after `trace.original_query`, the user's raw message, rather than the query that was searched (`retrieve` at `:753` uses `query.query`). For a follow-up like "and in 2023?", the entry shows that text instead of the contextualized query.

Adding a `partition` field to `QueryRetrievalTrace`, set on partition fan-out, and naming the child after the searched query would make these entries usable.

openrag/core/retrieval/trace.py
Ahmath-Gadji: Thanks, the examples I listed pass now, but the same words with trailing punctuation are still redacted. The unquoted capture `[^\s,;]+` keeps a final `?`, `.` or `!`, and then the "8+ characters containing a non-letter" rule in `_looks_like_credential` fires:

| query | in the trace |
|---|---|
| `What is basic authentication?` | `What is basic [redacted]` |
| `Explain basic concepts.` | `Explain basic [redacted]` |
| `Reset password: immediately!` | `Reset password: [redacted]` |
| `the token: "expired" error` | `the token: "[redacted]" error` |

The punctuation is swallowed as well (the `?` is gone in the first row). The last row is the quoted branch, which still redacts any quoted value.

Leaving trailing punctuation out of the capture (or checking the value without it), and applying the same credential-shape check to quoted values, would cover these.

andyne13 (commented):
Reviewed `f3260d6b` from scratch, focused on the security claims, and verified each one by running it rather than reading it. The sanitizer is genuinely well built — the design of "allowlist per context, `empty` for anything unrecognised" is the right shape, and it holds under pressure.

**Six adversarial probes through `safe_public_value`, all held:**

| probe | result |
|---|---|
| candidate carrying `text` / `content` / `metadata.source` | all three dropped |
| stage carrying `text`, with a polluted candidate nested | both dropped, candidate re-sanitised |
| `snippet` smuggled inside `scores` | dropped |
| `chunk_text` + `embedding` at root | dropped, `original_query` kept |
| `embedding` nested under an allowed stage | dropped |
| error message `"conn to 10.0.0.5:19530 failed"` | replaced with `redacted` |

I also checked the authorization surface rather than assuming it:

- **No guard bypass.** All three `_new_trace` sites in `search.py` (261, 347, 421) are immediately preceded by `authorize` (260, 346, 420) under the same flag, and the snapshot endpoint guards unconditionally at 169. `chat.py` guards all three diagnostic flags at 511, and the sibling endpoint rejects them outright with a 400.
- **`require_partition_viewer` returns the `user` dict**, so `authorize(partition_viewer)` reads a real `is_admin` — the two routers' guard calls are equivalent, which was not obvious from the call sites.
- **The override chain is sound.** `_validate_retrieval_diagnostic_overrides` requires `include_retrieval_trace` whenever an override is *active*, and that flag triggers the admin guard — so the overrides are transitively admin-only, as the description claims. `active` correctly treats `threshold=0` as set and a `False` disable-flag as a no-op.

## Confirmed fixed — please resolve these threads

`develop` requires every conversation resolved, so these three are currently blocking merge on their own:

- **The over-eager credential scrubber** is fixed. All five of the examples in that thread now pass through unchanged (`basic idea`, `Explain basic authentication`, `password: step by step`, `token: limit`, `Bearer bonds`), while `api_key = or-…`, a JWT after `Bearer`, and `password: hunter2!!x9` still redact. `_looks_like_credential` is what does it, and `test_trace_preserves_natural_language_that_mentions_authentication` pins exactly those five.
- **The trace not reaching sub-queries / partitions** is fixed. The `pipeline_trace = trace if len(groups) == 1 else None` gate is gone, replaced by per-group `child_traces` merged through `merge_child_traces`.

## Still open — the partition identity is lost, and it's lost in one line

This one is real and I'd fix it before merge, because it defeats the PR's own stated purpose ("identify the first stage that lost a relevant passage") exactly when there is more than one partition.

The partition index *is* generated — `RetrievalTraceBuilder(f"{trace.request_id}:partition:{index}", …)` at `retrieval_service.py:753` and `:806`. It is then discarded at `trace.py:352`:

```python
QueryRetrievalTrace(
    query=child.original_query,   # the only identifying field
    stages=..., timings=..., errors=...,
)
```

`request_id` is not carried over, and `QueryRetrievalTrace` has no `partition` field, so two partitions produce two entries distinguishable only by list position. With two partitions *and* two sub-queries, `merge_child_traces` extends rather than nests (`trace.py:305`), giving `a, b, a, b` with nothing saying which partition each came from.

Two smaller things in the same area:
- The two call sites disagree on what the child is named: `:753` uses `query.query` (the searched query) while `:806` uses `trace.original_query`. For a follow-up like "and in 2023?", the `retrieve_multi` path labels the child with the raw message instead of the contextualised query.
- Adding `partition` to `QueryRetrievalTrace` is safe for the sanitizer as long as it also goes into the `query_trace` allowlist — it's `frozenset({"query", "stages", "timings", "errors"})` today, so a new field would be silently dropped. Worth mentioning because the failure would be invisible.

## New finding: the sanitizer is the last line of defence, and three of its four protections are unpinned

`finish()` ends with `return safe_public_value(trace)`, so this really is the final gate. I mutated it four ways and ran the retrieval + chat-diagnostics suites:

| mutation | result |
|---|---|
| add `"text"` to the `candidate` allowlist | **110 passed** |
| drop the `REDACTED_ERROR_MESSAGE` substitution | **110 passed** |
| `_child_context` default `"empty"` → `"root"` | **110 passed** |
| drop `_scrub_query_credentials` | 6 failed ✓ |

To be fair about severity: none of these is a live leak, because each is backed further upstream — `candidates_from_chunks` projects only id/document_id/rank/scores, `TraceCandidate` is `extra="forbid"`, and `record_error` stores the redacted constant at source rather than the exception text. So this is defence-in-depth going untested, not an open door.

But the allowlist is the layer that makes the security claim *general* — it's what protects the paths that aren't models, and `comparisons` is exactly that (`dict[str, Mapping[str, object]]`, not a model). The `"empty"` default is the deny-by-default property itself, and flipping it to `"root"` costs nothing today. Three assertions on `safe_public_value` output would close all of it.

## One operational note

`RetrievalDiagnosticsGuard` uses `MemoryStorage()`, so the rate limit is per process. Under Ray Serve or multiple API replicas the effective limit is `120/minute × replicas`, not `120/minute`. Fine for an admin-only benchmark path — worth a line in the docs so nobody sizes a sweep against the configured number.

Also: the description still says "This draft targets `fix/required-retrieval` … can be rebased onto develop after that prerequisite lands." The base is `develop` now, so that paragraph is stale. The PR is `BEHIND` and will need a sync before it can merge.

Not blocking on my side — the partition finding is already open as a thread from another reviewer, and everything I added is test coverage. Happy to re-review once the partition field lands.

andyne13 (commented):
_Duplicate of the review above — posted three times by mistake. Content is unchanged there._

andyne13 (commented):
_Duplicate of the review above — posted three times by mistake. Content is unchanged there._

hedhoud:
@Ahmath-Gadji @andyne13 I rechecked the review points against `f3260d6b`. Here is the current assessment.

**Valid and in scope for this PR**

- Partition identity is missing from child traces, and one path labels them with the original message rather than the query actually searched. The trace should retain the partition and effective searched query so multi-partition results can be attributed correctly.
- Nested fan-outs still flatten child traces and can lose reranking or final stages. The trace should preserve each child node and nest its subquery traces instead of keeping only the deepest entries.
- Trailing punctuation still causes normal language such as `authentication?` to be treated as a credential. Punctuation should be separated from the value before credential detection and preserved in the serialized query.
- The sanitizer test-coverage gap is valid. This is defence-in-depth rather than a current leak; direct tests should pin candidate-field filtering, deny-by-default nested contexts, and final error redaction.
- The diagnostics rate limit is per process. That behavior should be documented, along with correcting the stale dependency paragraph and synchronizing the branch with `develop`.

**Partly valid**

The quoted-value false positive is real, but applying the same permissive shape check to every quoted value could expose short real credentials. The fix should preserve normal quoted language without weakening protection for explicitly named passwords, tokens, or secrets.

**Not current bugs / not changes to make in this PR**

- I found no active document-content, embedding, internal-error, credential, or authorization-guard leak in the current implementation. The additional sanitizer assertions are regression hardening.
- Replacing this work with a new dry-run endpoint is a larger architectural alternative, not a correctness fix required here.
- The comparison and document-identity capabilities are consumed by the separate OpenRAG Evals project, so their absence from this repository is not evidence that they are unused.

The GitHub resolved state is not sufficient evidence for the punctuation and nested-fan-out points; those behaviors remain in the current code and should stay part of the follow-up.

openrag/services/orchestrators/retrieval_service.py
hedhoud: @Ahmath-Gadji Valid and fixed in the latest head. Each partition trace now carries its partition identity, and single-query partition traces use the effective searched query rather than the original raw message. Multi-query partition nodes contain the effective queries as ordered children, so their origin is no longer inferred from list position.

openrag/core/retrieval/pipeline.py
hedhoud: @Ahmath-Gadji Valid and fixed in the latest head. Fan-out traces now keep each partition or effective query as a parent node, with generated queries nested below it. This preserves the parent fusion, reranking, and final-cutoff stages instead of flattening them away. Regression coverage includes nested partition and query fan-out.

openrag/core/retrieval/trace.py
hedhoud: @Ahmath-Gadji Partly valid and fixed for trailing sentence punctuation: ordinary unquoted text such as authentication? and concepts. now round-trips unchanged. Quoted values after explicit credential labels remain conservatively redacted by design because a short quoted secret cannot be distinguished reliably from ordinary text once telemetry is persisted. Tests cover both boundaries.

openrag/services/orchestrators/retrieval_service.py
```diff
@@ -156,20 +174,234 @@ def _build_retriever(
...
+            embedder_name = partition.embedder if partition is not None else "default"
+            reranker_enabled = bool(
+                getattr(
+                    retrieval,
+                    "enable_reranker",
+                    getattr(getattr(self._config, "reranker", None), "enabled", False),
+                )
+            )
+            reranker_name = getattr(retrieval, "reranker", None) or ("default" if reranker_enabled else None)
+            contextualizer_name = (
```
Ahmath-Gadji: **[Major] Snapshot/fingerprint report the wrong contextualizer model.**

Contextualization runs on `QueryService._resolve_llm(partitions)`, i.e. the partition's `chat_llm` (or the default endpoint). `retrieval.llm` is the multiQuery/HyDE LLM and never contextualizes anything. On a partition that sets `retrieval.llm`, `/snapshot` and the configuration fingerprint name a model that did not run, and two setups contextualizing with different models can share a fingerprint. `llm_override.model` is also not part of the fingerprint. For a reproducibility feature this should follow `resolve_partition_chat_llm` exactly.

docs/content/docs/documentation/env_vars.md
```diff
@@ -439,8 +439,8 @@ The retriever fetches relevant documents from the vector database based on query
 | `WITH_SURROUNDING_CHUNKS` | `bool` | false | When enabled, retrieves adjacent chunks (preceding and following) for each matched document to provide additional context.|
 | `INCLUDE_RELATED` | `bool` | true | Expand results with chunks from files sharing the matched file's `relationship_id` (see [Linked files](/openrag/documentation/linked_files/)). |
 | `INCLUDE_ANCESTORS` | `bool` | true | Expand results with chunks from ancestor files in the parent/child file hierarchy (see [Linked files](/openrag/documentation/linked_files/)). |
-| `RELATED_LIMIT` | `int` | 10 | Maximum number of related/ancestor chunks fetched per matched result when expansion is enabled. |
-| `MAX_DEPTH` | `int` | 10 | Maximum ancestor depth traversed when `INCLUDE_ANCESTORS` is enabled. |
+| `RELATED_LIMIT` | `int` | 10 | Maximum number of related/ancestor chunks fetched per matched result when expansion is enabled. Values above 100 are capped at 100. |
```
Ahmath-Gadji: **[Major] Documented caps are not implemented.**

Neither this PR nor `develop` caps `RELATED_LIMIT` at 100 or `MAX_DEPTH` at 50. The only bound is `max_ancestor_depth_cap = 1000` (`core/config/retrieval.py`), and `/search`'s `related_limit` has `ge=0` with no upper bound. Either implement the caps (config validator + `Query(le=...)`) or drop these sentences.

openrag/core/retrieval/pipeline.py
```diff
@@ -122,13 +154,55 @@ async def retrieve_docs(
                 query=query.query,
                 filter=None,
                 filter_params=filter_params,
+                **trace_kwargs,
```
Ahmath-Gadji: **[Major] The filterless fallback is invisible in the trace.**

The retry reuses the same builder, so every store stage (`dense_*`, `sparse`, `hybrid_fused`) and the `embedding`/`dense_search` timings from the filtered attempt are overwritten. Nothing records that the temporal filter was dropped, so the trace reads as an ordinary unfiltered search. That's exactly the "which stage lost the passage" case this feature targets. Suggest tracing the first attempt separately (child trace or a `fallback` marker) and flagging `temporal_filter` removals.

openrag/core/retrieval/trace.py
```diff
@@ -0,0 +1,389 @@
...
+    ContextualizationTrace,
+    QueryRetrievalTrace,
+    TraceCandidate,
+    TraceError,
+    TraceStage,
+)
+from pydantic import BaseModel
+
+TRACE_SCHEMA_VERSION = 1
+MAX_TRACE_CANDIDATES_PER_STAGE = 200
```
Ahmath-Gadji: **[Major] Trace size is bounded per stage only.**

`query_traces` nest per partition × sub-query × store sub-query, 11 stages each. An `openrag-all` request over ~50 partitions with multiQuery (3 queries) can reach roughly 50 × 3 × 11 × 200 ≈ 330k candidates in one response. `_merge_store_traces` also both concatenates child candidates into the parent stages *and* keeps them in `query_traces`, which doubles the payload. The PR describes traces as "bounded". It needs a global cap (total candidates or bytes) or a nesting limit.

openrag/services/storage/milvus_store.py
```diff
@@ -1035,6 +1046,287 @@ def _parse_search_response(self, response: Any) -> list[dict[str, Any]]:
...
+        advanced: list[TraceCandidate],
+        *,
+        file_scope: object | None,
+        temporal_relaxed: bool,
+        similarity_threshold: float | None,
+    ) -> list[TraceCandidate]:
+        advanced_ids = {candidate.id for candidate in advanced}
+        if isinstance(file_scope, (list, tuple, set)):
+            allowed_file_ids = {str(value) for value in file_scope}
+            file_reason: RemovalReasonCode = "workspace_filter"
```
Ahmath-Gadji: **[Minor] Removal reasons are mislabelled.** Scope *shape* (list vs scalar) is used as a proxy for scope *kind*. Chat attachments pass a list (`filter_params = {"file_id": indexed_attachment_ids}`), so they're reported as `workspace_filter`. `/search/partition/{p}/file/{file_id}` passes a scalar, so it's reported as `attachment_filter`. The caller knows the kind, so it should pass it down explicitly.

openrag/services/orchestrators/query_service.py
```diff
@@ -835,6 +1159,15 @@ async def chat(
         if metadata.get("attachments"):
             # Indicate which attachments were actually searched to generate the answer.
             extra["attachments"] = attachments
+        if trace is not None:
+            if trace.contextualization is None:
```
Ahmath-Gadji: **[Minor] This fabricates a contextualization for direct-LLM requests.** If contextualization never ran (direct-LLM model with `include_retrieval_trace`), this calls `_record_contextualization` with a synthetic query, which marks `contextualized_query` as `complete`. It should be recorded as `not_run` (same in the streaming path below).

openrag/core/retrieval/pipeline.py
```diff
@@ -137,7 +211,37 @@ async def retrieve_docs(
             if len(expanded) > len(head):
                 chunks = expanded
                 if self.reranker_enabled:
+                    pre_rerank_chunks = list(chunks)
```
Ahmath-Gadji: **[Minor] The second rerank overwrites the first.** After expansion, `pre_rerank`/`post_rerank` are overwritten, so the first pass's `reranker_top_n` removals are lost and expanded chunks appear in `pre_rerank` with no stage explaining where they came from. Expansion itself is never traced as a stage.

openrag/core/utils/source_filtering.py
```diff
@@ -321,7 +323,11 @@ async def stream_with_source_filtering(
         finish_chunk = copy.deepcopy(template)
         finish_chunk["choices"][0]["delta"] = {}
         finish_chunk["choices"][0]["finish_reason"] = last_finish_reason or "stop"
-        finish_chunk["extra"] = extra_payload
+        finish_extra = dict(extra_payload)
+        if terminal_extra_fields:
+            finalized_fields = terminal_extra_fields() if callable(terminal_extra_fields) else terminal_extra_fields
```
Ahmath-Gadji: **[Minor] Streaming can lose the trace, or break the stream.**

- The trace is only attached to the finish chunk, which exists only `if template:`. If upstream fails before sending any chunk, the trace is silently dropped.
- `terminal_extra_fields()` isn't guarded, so an exception here ends the stream after the content without `[DONE]`. Telemetry shouldn't be able to do that.

openrag/core/retrieval/trace.py
```diff
@@ -0,0 +1,389 @@
...
+            "embedding",
+            "dense_search",
+            "sparse_search",
+            "fusion",
+            "reranking",
+            "total",
+        }
+    ).union(TRACE_STAGE_NAMES),
+    "comparisons": frozenset({"original_query"}),
+    "comparison": frozenset({"status", "stages", "timings", "errors", "configuration_fingerprint"}),
```
Ahmath-Gadji: **[Minor] The comparison allowlist has no `query_traces`.** On multi-partition requests, `comparisons.original_query` loses all per-partition stages and keeps only the stages `merge_child_traces` marked `unavailable` plus `partition_fused`.

openrag/api/routers/user/search.py
```diff
@@ -88,6 +100,76 @@ def _documents(request: Request, chunks) -> list[dict]:
...
+    service,
+    partitions: list[str],
+    effective_options: dict[str, object],
+) -> dict[str, object]:
+    payload: dict[str, object] = {"documents": _documents(request, chunks)}
+    if trace is None:
+        return payload
+    try:
+        resolved_fingerprint = getattr(service, "resolved_configuration_fingerprint", None)
+        stored_fingerprint = (
```
Ahmath-Gadji: **[Minor] Search fingerprints can't be compared with snapshots.**

`RetrievalService.search` doesn't use the partition preset (retrieval type, reranker, contextualizer, HyDE/multi-query prompts), yet the fingerprint is built from it. It's also always wrapped with the request options, so it can never equal `/snapshot`'s `retrieval_configuration_fingerprint`. It also costs extra prompt-DB lookups per traced search. A search-specific fingerprint (embedder + hybrid + request options) would be more accurate.

Ahmath-Gadji (commented):
Review pass. Tracing-off behaviour looks untouched: every new path is gated on `trace is not None` or an explicit override. Focused unit suites (618) and ruff pass locally on 84a5da0f. Findings are inline, labelled Major/Minor/Nit.

Two points that don't fit on a line:

- **[Minor] Breaking change isn't described.** The PR is labelled `breaking-change`, but the body doesn't say what breaks. From the diff: non-admins sending `include_retrieval_trace` / `compare_original_query` / `bypass_query_contextualization` now get 403 (these keys were previously ignored); `/search` `top_k > 1000` now 422s; `extra` gains `retrieval_trace`. Please nest that under the PR's bullet for the release notes.
- **[Nit] Stale test fakes.** `VectorStore.query_chunks_by_filter` gained `limit`, and `VectorStoreSearcher` now always passes it. The fakes in `tests/unit/conftest.py`, `test_pipeline_stages.py`, `test_partition_service.py`, `test_mcp_service.py` and `test_conversion_service.py` don't accept it and would `TypeError` if a test ever reached expansion through them.

openrag/services/orchestrators/query_service.py
```diff
@@ -576,10 +700,28 @@ async def _prepare_chat(self, partition: list[str] | None, payload: dict, llm: L
                 casual_policy = CasualMessagePolicy(queries.intent, language or "en")
             elif not usable_queries:
                 queries = SearchQueries(query_list=[Query(query=last_user_message)])
+                self._mark_contextualization_fallback(trace, queries)
             elif len(usable_queries) != len(queries.query_list):
                 queries = queries.model_copy(update={"query_list": usable_queries})
 
+        if trace is not None:
+            contextualizer_prompt = trace.contextualization.prompt if trace.contextualization is not None else None
```
Ahmath-Gadji: **[Minor] The fingerprint includes a prompt that wasn't used.** On the bypass, SIMPLERAG and casual paths `contextualizer_prompt` is `None`, so `_configuration_fingerprint` re-resolves the contextualizer prompt and folds it in. Editing that unused prompt changes the fingerprint of runs that never used it.

openrag/services/orchestrators/query_service.py
```diff
@@ -412,22 +424,30 @@ async def generate_query(
...
+            self._record_contextualization(trace, queries, original_query=last_user)
+            return queries
 
         chat_history = "".join(f"{m['role']}: {m.get('content') or ''}\n" for m in messages)
-        contextualizer = await self._prompt_service.resolve_prompt(
-            "query_contextualizer",
-            names=[self._retrieval_prompt_name("query_contextualizer_prompt_name", partition)],
-        )
+        prompt_name = self._retrieval_prompt_name("query_contextualizer_prompt_name", partition)
+        resolve_with_identity = getattr(self._prompt_service, "resolve_prompt_with_identity", None)
```
Ahmath-Gadji: **[Nit]** This `getattr(..., None)` duck typing (also `resolved_configuration_fingerprint` in `search.py`/`query_service.py`, and in `retrieval_service.py`) only serves test mocks, because every real service implements the method. The fallback branch also mislabels the source (`named`/`default` even when the disk seed was used). I'd call the methods directly and update the mocks.

openrag/api/routers/user/search.py
```diff
@@ -88,6 +100,76 @@ def _documents(request: Request, chunks) -> list[dict]:
...
+    partition: str,
+    include_document_ids: bool = Query(
+        False,
+        description="Include sorted indexed document identifiers for index-health checks.",
+    ),
+    partition_viewer=Depends(require_partition_viewer),
+    service=Depends(get_retrieval_snapshot_service),
+    diagnostics_guard=Depends(get_retrieval_diagnostics_guard),
+):
+    await diagnostics_guard.authorize(partition_viewer)
```
Ahmath-Gadji: **[Nit]** `require_partition_viewer` runs before the admin check, so an admin who isn't a member is refused unless `SUPER_ADMIN_MODE` is on. That's fine if intended, but the docs only say "administrator-only".

openrag/api/dependencies/retrieval_diagnostics.py
```diff
@@ -0,0 +1,59 @@
...
+
+class RetrievalDiagnosticsGuard:
+    """Keep expensive, sensitive retrieval diagnostics admin-only and bounded."""
+
+    def __init__(self, rate_limit: str | None = None) -> None:
+        configured = rate_limit or os.getenv("RETRIEVAL_DIAGNOSTICS_RATE_LIMIT", _DEFAULT_RATE_LIMIT)
+        try:
+            self._limit = parse(configured)
+        except ValueError:
+            logger.bind(configured_value=configured).warning(
```
Ahmath-Gadji: **[Nit]** A malformed `RETRIEVAL_DIAGNOSTICS_RATE_LIMIT` silently falls back to the default with only a warning. That differs from the general `RATE_LIMIT_*` handling and isn't in the env-var docs.

infra/docker/api.Dockerfile
```diff
@@ -63,7 +63,9 @@ COPY openrag/ .
 # Copy assets and config (prompt templates ship inside the package under openrag/prompts/)
 COPY scripts/ /app/scripts/
 COPY conf/ /app/conf/
-ENV PYTHONPATH=/app/openrag/
+ARG OPENRAG_COMMIT
+ENV PYTHONPATH=/app/openrag/ \
+    OPENRAG_COMMIT=${OPENRAG_COMMIT}
```
Ahmath-Gadji: **[Nit]**

- Setting the commit as `ENV` here invalidates the `useradd`/`chown` layers below on every commit. It's cheaper right before `USER`.
- `_installed_version()` (`package_version("openrag")`) likely returns `None` in this image, since the project isn't installed as a distribution. `openrag.version` in snapshots will usually be null.

PR #983, opened 2026-09-17
fix(indexing): fail documents with zero chunks

### LINKED ISSUE #946: A document that parses to zero chunks is reported as successfully indexed
## Summary

If parsing or chunking yields no chunks, indexing **succeeds**: the Milvus upsert is a
no-op, the Postgres catalog row is written, and the task reaches `COMPLETED`. The file
is listed as indexed and can never be retrieved. Nothing records how many chunks a file
produced, so the condition is invisible in aggregate.

## Evidence (`develop`)

**Both chunkers return `[]` for empty or whitespace-only content, without raising.**
Verified directly:

```
structured_section     empty      -> 0 chunks
structured_section     whitespace -> 0 chunks
recursive_splitter     empty      -> 0 chunks
recursive_splitter     whitespace -> 0 chunks
```

**`store_stage` accepts an empty chunk list.** The `if chunks:` guard at
`services/workers/stages/store.py:38` skips `ensure_collection` and the
missing-embedding check, and `MilvusVectorStore.upsert` returns early
(`services/storage/milvus_store.py:914`: `if not chunks: return 0`). The row ends with
`stored_count = 0`, `stage = "stored"`, and no exception.

**The catalog row is then written unconditionally.**
`IndexerWorker.process_file` (`services/workers/indexer_actor.py:118-152`) calls
`_write_catalog_record` on any successful pipeline run and sets `COMPLETED`. The
vector-cleanup path (`_cleanup_indexed_vectors`) only runs inside `except`.

**The behaviour is already known on the re-index path.**
`services/workers/pipeline_builder.py:278-289` documents it verbatim — *"`store_stage`
completes successfully even when it stores zero chunks — an empty/whitespace-only file,
a parser that extracts no text, etc. all legitimately chunk down to `[]` without
raising"* — and gates stale-chunk deletion on `stored_count` so a re-index cannot empty
a file that was previously populated. The **first-index** path has no equivalent guard.

**Nothing records the outcome.** `files` has no chunk-count column
(`services/persistence/schema.py`). `GET /partition/{p}/file/{id}` returns
`{"metadata": {}, "documents": []}` for such a file, but only for someone who already
suspects that one file.

## Why it matters

- A scanned PDF whose OCR produced nothing, a parser returning an empty string, content
  fully stripped by preprocessing — all report success to the uploader and to any
  `callback_url`, which receives `"status": "success"`.
- It makes the "catalog row with no chunks" direction of a Postgres↔Milvus
  reconciliation report (#924) unusable: a legitimate empty parse is indistinguishable
  from chunks that were lost. That is exactly the false-positive problem #924 warns
  will get the report ignored.
- #27 asked for *"a human-friendly error … explaining that the provided file is
  empty"*. The crash it reported was removed; that outcome was never delivered.

## Proposal

1. **Record the chunk count produced by the run on the `files` row.** Cheap, gives #924
   the discriminator it needs, and gives the admin UI a column.
2. **Stop reporting a bare success for zero chunks** — either fail the task with a
   specific reason, or settle it in a degraded state carrying the reason. A deployment
   that legitimately ingests empty placeholders can opt out by config; the default
   should not claim success.

## Acceptance

- Uploading an empty `.txt`, and a PDF from which no text is extracted, does not report
  a plain success.
- The chunk count for a file is queryable without scanning Milvus.

hedhoud:
## Context

A parser or chunker can legitimately produce no chunks. OpenRAG previously reported that outcome as successful even though the document could never be retrieved.

## Expected behavior

Zero-chunk attempts now fail with a clear reason before the catalog is changed. A failed re-index keeps the previous retrievable document intact. Successful catalog entries expose their chunk count from Postgres, while historical rows remain unknown rather than receiving an invented value.

## Validation

- 3,322 unit tests passed
- 17 Postgres repository integration tests passed
- migration upgrade, downgrade, and re-upgrade passed
- Ruff passed
- CodeRabbit raised 0 issues

Closes #946

## Summary by CodeRabbit

- **New Features**
  - Document records now track the number of indexed chunks.
  - Chunk counts are preserved when files are created, updated, copied, or replaced.
  - Chunk counts must be zero or greater.

- **Bug Fixes**
  - Indexing now reports a clear, standardized failure when no indexable content is extracted.
  - Empty or scanned-only files no longer complete successfully or overwrite existing catalog and vector data.
  - These failures return a 422 status with a dedicated error code.


openrag/services/persistence/document_repo.py
```diff
@@ -867,6 +875,7 @@ def _row_to_dict(row: asyncpg.Record) -> dict:
             "parent_id": row["parent_id"],
             **metadata,
             "content_sha256": row.get("content_sha256"),
+            "chunk_count": row.get("chunk_count"),
```
Ahmath-Gadji: `chunk_count` is server-computed now, but it isn't in `PROTECTED_METADATA_KEYS` (`core/utils/consts.py`), so an upload with `metadata={"chunk_count": 999}` keeps it: it is saved in `file_metadata` and copied onto every Milvus chunk. This line makes the column win in the file listing, but anything reading chunk metadata (e.g. the `chunk` object of a search source) still shows `999`.

Adding `"chunk_count"` to that set, next to `content_sha256` / `indexed_at`, would close it.

openrag/services/workers/indexer_actor.py
```diff
@@ -158,6 +162,9 @@ async def process_file(
             if resolved_prompts:
                 row.update(resolved_prompts)
             row = await self._pipeline.run(row)
+            stored_count = row.get("stored_count", 0)
+            if stored_count == 0:
+                raise NoIndexableContentError("No indexable content was extracted from this document.")
```
Ahmath-Gadji: Side effect on `parent_id` chains: a file that yields zero chunks now gets no `files` row. If that file is a parent (A ← B ← C, with B empty or text-less), the recursive CTE in `get_file_ancestors` (`document_repo.py`, joins `f.file_id = a.parent_id`) stops at the missing B, so A silently drops out of the ancestor context for C and everything below it. Before this PR, B had a row and the chain held.

Probably acceptable, but worth making it a deliberate choice (or at least documenting it).

openrag/services/workers/indexer_actor.py
```diff
@@ -29,6 +29,10 @@ class _TaskCancelledBeforeStart(Exception):
     outage and does need both)."""
 
 
+class NoIndexableContentError(RuntimeError):
```
Ahmath-Gadji: Nit: by the repo convention (CLAUDE.md, *Custom Exceptions*), custom exceptions inherit from `OpenRAGError` and are defined in `core/utils/exceptions.py`, not next to the code that raises them. `NoIndexableContentError` could move there as a `PipelineError` subclass, and be added to the hierarchy diagram at the top of that file. That also gives it a `code` / `status_code` like the rest of the hierarchy.

Ahmath-Gadji (dismissed):
Approving. The fix does what #946 describes, and a failed re-index keeps the previous document intact since the check runs before the catalog write and the old-chunk delete. The inline comments are non-blocking.

Labelled `breaking-change`, since API clients now see a different outcome for the same upload. For a file that yields no chunks (e.g. an empty or whitespace-only file, or a document the parser extracts no text from):

| | Before | After |
|---|---|---|
| Task status | `COMPLETED` | `FAILED` — `No indexable content was extracted from this document.` |
| Indexing callback | `status: "success"` | `status: "error"` |
| `files` row | created, file listed as indexed | not created |
| Uploader's `file_count` | incremented | unchanged |
| `PUT` with an empty new version | row updated, old chunks kept | fails, previous version kept as is |

Clients that act on the task status or the callback (e.g. cozy-stack) will get an error for these files where they used to get success.

openrag/services/workers/indexer_actor.py
hedhoud: Fixed in ff0a7d48. The zero-content failure now uses the shared pipeline exception hierarchy with a stable code and status. Existing failed-task, callback, and replacement-safety behavior is unchanged.

Ahmath-Gadji (approved):
Re-approving on ff0a7d48. `NoIndexableContentError` now lives in `core/utils/exceptions.py` under `PipelineError`, which addresses my comment. The develop merge is clean, and the breaking-change notes in my previous review still apply.

Issue #702, opened 2026-07-20
Enrichment stage failures (caption/contextualize/topic-tag) abort the entire indexing pipeline


Ahmath-Gadji:
## Problem

Any error in an enrichment stage — image captioning, chunk contextualization, or topic
tagging — aborts the **entire** indexing pipeline for that file. The file is not
indexed at all, even though parsing (and often chunking) already succeeded and the
base text content was ready to store.

## Where this happens

`caption_stage`, `contextualize_stage`, and `topic_tag_stage`
(`openrag/services/workers/stages/{caption,contextualize,topic_tag}.py`) share the same
pattern — tag the row for diagnostics, then re-raise:

```python
except Exception as exc:
    row["stage"] = "caption_failed"   # or contextualize_failed / topic_tag_failed
    row["error"] = str(exc)
    raise
```

`IndexingPipeline.run()` (`openrag/services/workers/pipeline_builder.py`) awaits each
stage inside a single `try/finally` with **no `except`** — nothing catches these, so the
exception propagates straight out of `run()`, past `chunk`/`embed`/`store`, which never
run.

`IndexerWorker.process_file` (`openrag/services/workers/indexer_actor.py:127`) catches
it only to mark the task `FAILED` and re-raise:

```python
except Exception:
    ...
    await self._tsm.set_failed_if_not_cancelled.remote(task_id, tb)
    raise
```

No catalog record is written, no chunks are stored — the file is a total loss for a
failure in an *enrichment* step, not the core parse/chunk/embed/store path.

## Existing precedent in the codebase for graceful degradation

`IndexingPipeline._select_vlm` / `_select_contextualizer` / `_select_topic_tagger`
already catch `KeyError` when a *named endpoint can't be resolved* and skip the stage
with a warning instead of failing the file — see the docstring on `_select_vlm`:

> "Captioning is an enrichment step, so an unresolvable endpoint name must not fail the
> file — same rationale as `_select_contextualizer` / `_select_topic_tagger`."

That reasoning currently only covers *resolution* failures (unknown/renamed endpoint
name). It does not extend to failures during actual *invocation* — timeouts, connection
errors, malformed VLM/LLM responses — which is exactly the case that fails the whole
file today.

## Related issue

#630 (VLM caption failure leaks a `DistributedSemaphore` permit and eventually
deadlocks the cluster) independently landed on the same idea in its suggestions list:
"Consider not failing (and not cancelling) the entire document when a single image
[...]

PR #986, opened 2026-09-17
fix(indexing): surface degraded enrichment outcomes

### LINKED ISSUE #947: Enrichment degradation is recorded only in a log line: degraded_stages is written and never read
## Summary

Enrichment failures (captioning, contextualization, topic tagging) no longer lose the
file — the pipeline skips the stage and stores the base content (#702 / #918). The
degradation itself is recorded **only** as a warning log line and a key on the
in-memory pipeline row, which is discarded when the task ends. The task reports
`COMPLETED`, indistinguishable from a fully enriched run.

## Evidence (`develop`)

- `services/workers/pipeline_builder.py:166` —
  `row.setdefault("degraded_stages", {})[name] = str(exc)`, plus a `logger.warning`.
- `grep -rn "degraded_stages" openrag/` returns **that one write and nothing else**.
  There is no reader in the indexer actor, the task state manager, the catalog row, the
  indexing callback, the jobs API or the UI.
- `IndexerWorker.process_file` sets `COMPLETED` without consulting it
  (`services/workers/indexer_actor.py:152`), and `send_indexing_callback` posts
  `"status": "success"`.

## Why it matters

#702 listed surfacing the degradation as a distinct part of the fix — *"record which
stage(s) failed on the file (catalog metadata and/or the task-state error field) so
it's visible via the jobs API / admin UI instead of silently shipping a lesser-quality
result, and so a later reconciliation/retry pass has something to act on"*. Only the
non-fatal half landed. Without the other half:

- An operator cannot tell which documents are indexed without captions, context or
  topic tags.
- There is no way to **select** the affected files for a re-index once the endpoint
  recovers. A VLM or LLM outage during a bulk import silently produces a
  lower-quality corpus with no record of which files are affected — and the outage is
  precisely when it is most likely to happen at scale.
- The jobs list shows these as ordinary successes. #934 adds a failure-reason column
  for failed tasks; a degraded run has no failure and no reason to show.

## Proposal

Persist the degraded stage names with the file and with the task:

- on the catalog row (a natural pair with a per-run chunk count, see #946), so a
  re-index pass can query for them;
- in the task details, so the jobs API can report a `completed_degraded` outcome with
  the stage names rather than a bare `COMPLETED`.

## Acceptance

A file indexed while the VLM endpoint is down is reported as degraded (with the stage
named) through the jobs API, and can be found later by querying the catalog — without
grepping worker logs.

hedhoud:
Closes #947
Refs #702

Enrichment failures no longer discard a document, but the degraded result previously disappeared when the worker finished. Operators therefore saw a normal completion and could not identify files that needed re-indexing after an inference service recovered.

This change persists bounded degraded stage names with the task and catalog record. The jobs API exposes a completed_degraded outcome, and the catalog API can filter affected files by stage. Raw provider errors remain in structured logs and are not exposed through these APIs.

Admin UI rendering remains tracked by #702 and will be delivered in a separate PR.

Validation:
- 3,446 unit tests pass
- API, integration, UI, lint, security, and CodeQL checks pass
- Local and hosted CodeRabbit reviews report no findings


openrag/services/persistence/migrations/alembic/versions/2a3b4c5d6e7f_add_job_degraded_stages.py
```diff
@@ -0,0 +1,37 @@
...
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from schema_helpers import column_exists, table_exists
+from sqlalchemy.dialects import postgresql
+
+revision: str = "2a3b4c5d6e7f"
+down_revision: str | Sequence[str] | None = "1f53920217de"
```
Ahmath-Gadji: **Two Alembic heads after merging `develop`.** #983 added `09f6c4b8a2d1`, which also revises `1f53920217de`. Git won't flag it, but `command.upgrade(cfg, "head")` (`connection.py:207`) refuses to run with two heads, so startup fails. Suggest `down_revision = "09f6c4b8a2d1"` when merging develop.

openrag/services/workers/indexer_actor.py
```diff
@@ -248,9 +276,11 @@ async def _write_catalog_record(
     indexed_at: datetime | None = None,
     require_existing_partition: bool = False,
     workspace_ids: list[str] | None = None,
+    degraded_stages: list[str] | None = None,
 ) -> bool:
     file_id = metadata.get("file_id", "")
     file_metadata = {key: value for key, value in metadata.items() if key != "page"}
+    file_metadata["degraded_stages"] = list(degraded_stages or [])
```
Ahmath-Gadji, line 312: **Lost on metadata PATCH / copy.** This key lives only in the catalog, but `WorkerDispatcher.update_file_metadata` (`dispatcher.py:601-603`) and `copy_file` (L653) rebuild `file_metadata` from a chunk + the patch and overwrite the column. A title-only PATCH drops the file out of `?degraded_stage=` while its chunks are still un-enriched, and a copy looks clean. Those paths should carry the existing value over.

openrag/core/models/catalog.py
```diff
@@ -24,6 +24,22 @@ class DocumentStatus(str, Enum):
 # re-declaring the set.
 TERMINAL_TASK_STATES = frozenset({DocumentStatus.COMPLETED, DocumentStatus.FAILED, DocumentStatus.CANCELLED})
 
+# Enrichment can fail without making the base-content index unusable. Keep the
+# names bounded and stable because they are persisted and exposed through APIs.
+DEGRADABLE_ENRICHMENT_STAGES = frozenset({"caption", "contextualize", "topic_tag"})
```
Ahmath-Gadji, line 29: **`degraded_stages` is caller-writable.** It isn't in `PROTECTED_METADATA_KEYS`, so a PATCH with `{"degraded_stages": ["caption"]}` writes it to the catalog and to every chunk (it then shows up in `extra.sources[].chunk`, and later PATCHes/copies copy it back from the chunk). Upload metadata lands on the chunks the same way. Suggest adding it to `PROTECTED_METADATA_KEYS`.

openrag/services/workers/indexer_actor.py
```diff
@@ -96,6 +96,27 @@ async def _record_started(
...
+            await retry_idempotent_ray_actor_method(
+                lambda: remote(task_id, stages),
+                task_description=f"set_degraded_stages({task_id})",
+            )
+        except Exception as exc:
+            logger.warning(
+                "Failed to record enrichment degradation in task state",
+                task_id=task_id,
+                error=str(exc),
+            )
```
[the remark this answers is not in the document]
Ahmath-Gadji: Agree. A rolling deploy triggers this on every degraded file, not only on exhausted retries: `_supports_task_state_recovery` (`bootstrap.py:109`) doesn't check for `set_degraded_stages`, so an upgraded deployment keeps the old TaskStateManager and jobs report `completed`. Workers also stay on protocol `v9`, so on a Ray cluster that outlives the API deploy the old worker code writes no catalog stages either. Worth a version bump, or at least a release note.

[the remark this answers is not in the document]
hedhoud: Fixed in dcf8db2e. Completion now persists normalized degraded stages and COMPLETED under one TaskStateManager lock. The operation is idempotent, rejects unknown or fenced terminal tasks, and the worker never falls back to split writes. Rolling deployments replace actors and trackers that lack the new capability and use indexer protocol v10. Tests cover retry exhaustion, cancellation, idempotency, and callback behavior.

openrag/core/models/catalog.py
hedhoud: Fixed in dcf8db2e. degraded_stages is now server-managed metadata. Upload, REST PATCH, MCP update, and copy all remove caller-supplied values, while the indexer can still persist the normalized outcome. The separately authorized partition move control remains available and keeps its destination permission check.

openrag/services/persistence/migrations/alembic/versions/2a3b4c5d6e7f_add_job_degraded_stages.py
hedhoud: Fixed in dcf8db2e. The degradation migration now follows 09f6c4b8a2d1, and a repository-level migration graph test fails if Alembic ever has more than one head. The local Alembic check reports only 2a3b4c5d6e7f as the head.

openrag/services/workers/indexer_actor.py
hedhoud: Fixed in dcf8db2e. Metadata PATCH and copy now read the authoritative file metadata from PostgreSQL before changing vectors. The degradation list is preserved for both degraded and clean files, and a missing catalog row aborts the mutation instead of rebuilding catalog data from an orphan chunk. Regression tests cover all of those cases.

openrag/core/models/catalog.py
```diff
@@ -24,6 +24,22 @@ class DocumentStatus(str, Enum):
...
+
+def normalize_degraded_stages(value: Any) -> list[str]:
+    """Return a stable, bounded stage-name list from pipeline or API data."""
+    if isinstance(value, dict):
+        candidates = value.keys()
+    elif isinstance(value, list | tuple | set | frozenset):
+        candidates = value
+    else:
+        return []
+    return sorted({stage for stage in candidates if stage in DEGRADABLE_ENRICHMENT_STAGES})
```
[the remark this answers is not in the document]
hedhoud: Fixed in 3293163b. Non-string values are ignored before allowlist membership, so malformed provider data cannot fail an otherwise usable index. A regression test covers unhashable values.

openrag/services/workers/dispatcher.py
```diff
@@ -579,6 +566,9 @@ async def update_file_metadata(
         partition: str,
         user: dict | None,
     ) -> None:
+        file_metadata = await self._document_repo.get_file_metadata(file_id, partition)
```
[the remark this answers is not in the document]
hedhoud: Fixed in 3293163b. Metadata updates now merge atomically into the current catalog value, while degraded stages remain server-owned. An interleaving regression test confirms a concurrent degradation write survives.

openrag/api/routers/admin/partitions.py
```diff
@@ -139,11 +140,12 @@ async def list_files(
     request: Request,
     partition: str,
     limit: int | None = None,
+    degraded_stage: Literal["caption", "contextualize", "topic_tag"] | None = Query(default=None),
```
Ahmath-Gadji, line 143: **`contextualize` and `topic_tag` are never recorded as degraded.** `_timed_enrichment` only records a stage that raises, but neither of these raises when the LLM is down: `ChunkContextualizer._generate_context` (`contextualize.py:72-79`) turns each chunk error into `""`, and `TopicTagger.tag` (`topic_tags.py:65-76`) returns `[]` on `InferenceError`/`CircuitBreakerOpenError`. I checked with an unreachable LLM: both stages finish cleanly (`stage=contextualized` with empty contexts, `topic_tags=[]`). The job reads `completed` and this filter finds nothing; only a stage-wide timeout is caught. Contextualization is on by default, so this is the common outage. Either report the failures from those two classes, or limit this to `caption` and track the other two in a follow-up.

openrag/services/workers/indexer_actor.py
```diff
@@ -193,10 +195,12 @@ async def process_file(
...
                 )
-            await retry_idempotent_ray_actor_method(
-                lambda: self._tsm.set_state.remote(task_id, "COMPLETED"),
-                task_description=f"set_state({task_id}, COMPLETED)",
+            completion_accepted = await retry_idempotent_ray_actor_method(
+                lambda: self._tsm.complete_with_degraded_stages.remote(task_id, degraded_stages),
+                task_description=f"complete_with_degraded_stages({task_id})",
             )
+            if completion_accepted is False:
+                raise RuntimeError(f"Task state manager rejected completion for task {task_id}")
```
Ahmath-Gadji, line 246: **An unknown task gets an error callback for a file that is indexed.** `complete_with_degraded_stages` returns `False` when the TaskStateManager doesn't know the task. This raise then runs after `catalog_written = True`, so the vectors are kept, and `set_failed_if_not_cancelled` returns `True` for unknown ids. The client gets `"error"` while the file is live. The old `set_state` recreated the record and sent `"success"`. It only happens if the TaskStateManager loses its stored state, but the callback then disagrees with the catalog. Also, a task cancelled mid-run now logs this as an error traceback.

hedhoud: Fixed. Atomic completion now distinguishes completed, cancelled, missing, and conflicting state. A cancellation after the catalog commit ends quietly without an error callback; a missing in-memory receipt repairs the durable completed job and reports success; genuine conflicts still fail. The actor compatibility guard and protocol generation prevent old Boolean semantics from surviving a rollout.

openrag/api/routers/admin/partitions.py
hedhoud: Fixed. Contextualization and topic tagging keep their best-effort fallback, but now report provider, timeout, and malformed-response failures to the pipeline. The completed file therefore records the affected stage as degraded instead of appearing clean. The tests exercise the real helper fallback paths.

Ahmath-Gadji: @hedhoud Verified on `3c3e8fb5`. With the LLM down, with one of three chunks failing, and with a malformed tag response, the stage now shows up in `degraded_stages` (`contextualize` / `topic_tag`). Resolved on my side.

openrag/services/workers/indexer_actor.py
Ahmath-Gadji: **Nit (non-blocking)** @hedhoud The callback is fixed. One leftover on `missing`: that outcome means the TaskStateManager lost its records, and `recover()` then runs `reconcile_jobs`, which marks the durable row FAILED (orphaned) once it is past the grace window. `_record_completed` can't undo that because `upsert_job` keeps terminal rows. So the jobs API can still show FAILED while the callback says success. Fine as a follow-up.

Ahmath-Gadji:
@hedhoud Is surfacing this in the admin UI planned for a separate PR? Right now it is API-only:
- the documents list receives `degraded_stages` but doesn't show it and has no filter;
- the jobs list badge still reads COMPLETED because it ignores `outcome`;
- the document detail page can't show it, because `GET /partition/{p}/file/{id}` builds its metadata from the first Milvus chunk, not from the catalog.

#702 asks for this to be "visible via the jobs API / admin UI". If the UI comes later, a follow-up issue is probably better than closing #702 here.

hedhoud:
@Ahmath-Gadji Thanks for flagging this. The admin UI follow-up is now #990. PR #986 keeps the durable catalog and jobs API contract and closes #947; #702 remains open until #990 lands. The follow-up covers the degraded jobs badge and stage names, document filtering, and catalog-backed document details.

Ahmath-Gadji (approved):

PR #989, opened 2026-09-18
fix(workers): make parser pools follow an admin restart, and stop listing the replaced actor

### LINKED ISSUE #988: Restarting a parser pool from the admin API breaks indexing until the stack restarts
Line numbers are as of `b1e4fb39` (`develop`).

## Summary

After `POST /actors/MarkerPool/restart` (the Restart button under System > Actors in the admin UI), every PDF upload fails with `ActorDiedError` until the whole stack is restarted. `DoclingPool` and `WhisperPool` are built the same way. Separately, every restart leaves the killed actor listed as `DEAD` next to its replacement.

## 1. Indexing keeps calling the killed pool

Reproduced on a compose stack:

1. Index a PDF: `COMPLETED`.
2. `POST /actors/MarkerPool/restart`: `200`. The new pool starts and its `MarkerWorker` initializes.
3. Index another PDF: `FAILED`. Every later PDF fails the same way.

The indexer worker is still calling the old pool:

```
ray.exceptions.ActorDiedError: The actor died unexpectedly before finishing this task.
	class_name: MarkerPool
	actor_id: f4f95f5d0ec361eb4e346d6301000000
The actor is dead because it was killed by `ray.kill`.
...
core.utils.exceptions.ServiceUnavailableError: RAY_ACTOR_UNAVAILABLE: Worker service is temporarily unavailable
```

At the same moment `GET /actors/` listed the replacement `a3c9962f…` as `ALIVE` and `f4f95f5d…` as `DEAD`.

Cause:

- `MarkerLoader.__init__` resolves the pool handle once (`services/workers/parsers/marker_workers.py:528`).
- The parser dispatcher keeps that loader for the life of the `IndexerWorkerActor` (`services/workers/parsers/parser_dispatcher.py:150-154`).
- `restart_ray_actor` kills the pool with `no_restart=True` and creates a new actor with a new id (`di/workers.py:103`). Nothing refreshes the cached handle.

`DoclingLoader` (`docling_workers.py:149`) and `LocalWhisperLoader` (`whisper_workers.py:207`) cache their handle the same way. `TaskStateManager` got a dedicated in-place restart for exactly this reason (`di/workers.py:75`); the parser pools never did. The distributed semaphores are not affected, because they look their actor up by name on every call (`services/inference/distributed_semaphore.py:104`).

**Impact.** Indexing for that file type stays down until the API container restarts. The error says "temporarily unavailable", but it never recovers on its own. The trigger is the restart button, which is what an admin reaches for when a pool looks stuck.

## 2. A restarted actor is listed twice

`list_ray_actors` (`di/workers.py:13-39`) returns every record Ray holds in the `openrag` namespace, and Ray keeps a killed actor's record as `DEAD`. After restarting `vlmSemaphore`:

```
ALIVE    vlmSemaphore     4308989725eb340c57e3d5e801000000
DEAD     vlmSemaphore     875ae309d52b185ab037b38501000000
```

The admin overview then reports "8/9 alive" while every actor is healthy. The dead records pile up with each restart and share Ray's default `limit=100` result window with the live actors.

A `DEAD` row is still worth listing when nothing replaced it: that actor is really down and needs a restart.

Ahmath-Gadji:
Fixes both problems in #988.

**Parser pools follow a restart.** `MarkerLoader`, `DoclingLoader` and `LocalWhisperLoader`
resolved their pool handle once, in `__init__`, and the parser dispatcher keeps each loader
for the life of the indexer worker. After `POST /actors/{pool}/restart` that handle pointed at
the killed actor, so every later file failed with `ActorDiedError`. Each dispatch now looks the
pool up by name through `get_or_create_actor`, the way the distributed semaphores already do.

**The actor list drops the record a restart replaced.** `GET /actors/` lists a `DEAD` record
only when no actor holding its name is `ALIVE`, and once per name.

Worth knowing when reviewing:

- **The handle goes into a local before the call.** Ray's `ActorMethod` holds its handle
  weakly, so `self._pool().process_pdf.remote()` raises "Lost reference to actor", the same
  trap as dd5d67f0. `worker` became a `_pool()` method so the one-line form stops looking safe.
  The new loader test dispatches through each loader's real call site with a fake handle that
  behaves the same way; it fails on the one-line form.
- **One name lookup per file.** When the pool exists, `get_or_create_actor` is a GCS name
  lookup, which is negligible next to a parse. The pool is now created on first dispatch rather
  than when the loader is built, so building the `.eml` parser no longer starts Marker for a
  message that has no PDF attachment.
- **The dead row stays until the replacement is `ALIVE`.** While the new actor is
  `PENDING_CREATION`, `DEPENDENCIES_UNREADY` or `RESTARTING`, or if it failed, one `DEAD` row
  is still listed, so a restart that hasn't taken is visible. Observed on a compose stack:

  | restart | old row `DEAD` next to `PENDING_CREATION` | only the new row, `ALIVE` |
  |---|---|---|
  | `MarkerPool` | +4.5 s | +9.3 s |
  | `vlmSemaphore` | +0.05 s | +0.14 s |

  Ray's `ActorState` carries no death time, so when a name has several dead records the one
  listed is arbitrary. It can show an older instance's id rather than the one just killed.
- **An actor that is down stays listed.** With `vlmSemaphore` killed and nothing replacing it,
  the list showed one `DEAD` row, although Ray held two dead records for that name.
- **One query per state, explicit limit.** Each actor state is fetched on its own with a
  `state =` filter and `limit=10_000`, the most Ray's API server accepts. Ray's default of 100
  dropped records past the first 100 by actor id without an error, so a down actor could vanish
  from the list. Ray's data source reads at most 10k records, in no particular order, and applies
  only equality filters before that cap. It keeps up to 100k dead records, some still among its
  registered actors. So one `state != DEAD` query could lose live actors behind dead records; a
  per-state query only counts records in that state.
- **`GET /actors/` gains `complete`.** It is `false` when a query still comes back short, or fills
  the whole page. The actors Ray did return are listed, and a warning is logged. Existing readers
  only use `actors`. The admin UI doesn't show the flag yet.
- **No actor protocol bump.** `process_file`'s remote contract is unchanged. The loader fix runs
  inside the detached indexer workers, though, so on a deployment with a persistent Ray cluster
  it only takes effect once the indexer workers are replaced by a new set, either by a Ray
  restart or by a bump. Compose's embedded Ray restarts with the API container. Say if you'd
  rather bump to v10 here.

Verification:

- On a compose stack (GPU, Marker), indexing a one-page PDF each time:

  | step | `develop` | this branch |
  |---|---|---|
  | index before any restart | `COMPLETED` | `COMPLETED` |
  | restart `MarkerPool`, index | `FAILED`: `ActorDiedError` on the old actor id | `COMPLETED` |
  | restart it again, index | `FAILED` | `COMPLETED` |

  The log showed each file after a restart allocated on the new pool's process, with no
  `ActorDiedError` or "Lost reference" anywhere.
- On a local Ray 2.47.1:
  - With 150 dead records, a killed `Down` actor is listed. Before, it dropped out of the list.
  - With Ray's data-source cap lowered to 50 and 80 owned actors that died while their handles
    were still held, all 5 live actors are listed, with `complete: false`. A single
    `state != DEAD` query listed 2 of them, with no error.
- Unit tests: across `tests/unit/services/workers`, `tests/unit/di` and `tests/unit/api`, 1151
  pass. Ruff and the layer import guard are clean.

Closes #988

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved actor status visibility during restarts, retaining dead actors until a live replacement is available.
  * Enhanced actor listing for large clusters and partial results, including an indicator when some statuses may be unavailable.
  * Parser workers now reconnect to restarted processing pools automatically, preventing requests from being sent to stale actors.

* **Tests**
  * Added coverage for restart states, large actor collections, incomplete results, and parser recovery after pool restarts.


openrag/di/workers.py
```diff
@@ -23,9 +23,31 @@ def list_ray_actors() -> list[dict[str, str | None]]:
...
+    accumulated dead records must not push live actors out of the result window.
     """
     from ray.util.state import list_actors
 
+    in_namespace = ("ray_namespace", "=", OPENRAG_NAMESPACE)
+    actors = list(list_actors(filters=[in_namespace, ("state", "!=", "DEAD")]))
+    # Names whose dead records add nothing: an ALIVE actor holds the name, or a
+    # dead record for it is already listed.
+    covered_names = {actor.name for actor in actors if actor.name and actor.state == "ALIVE"}
+    for actor in list_actors(filters=[in_namespace, ("state", "=", "DEAD")]):
```
hedhoud: Splitting live and dead actors into separate queries does not remove the State API limit. After 100 accumulated dead records, this call can raise because `list_actors` defaults to `limit=100` and `raise_on_missing_output=True`, so repeated restarts can make the admin actor endpoint fail before deduplication runs. Please use an explicit supported limit, define safe truncation behavior for both queries, and add a regression test with more than 100 dead records.

hedhoud (commented):
The parser-pool reconnection changes are sound, but the actor-list cleanup still has one truncation failure.

tests/unit/services/workers/parsers/test_pool_loaders.py
```diff
@@ -41,5 +81,38 @@ def fake_get_or_create_actor(name, cls, **options):
...
+
+    Dispatches through the real call site with Ray's weak ``ActorMethod``
+    semantics: looking the pool up per call must not leave the handle unstored
+    (``self._pool().process_pdf.remote()`` raised "Lost reference to actor" on
+    every file)."""
+    import services.workers.bootstrap as bootstrap
+
+    module = importlib.import_module(module_name)
+    handles = iter([_ActorHandle("pool-before-restart"), _ActorHandle("pool-after-restart")])
+    monkeypatch.setattr(bootstrap, "get_or_create_actor", lambda name, cls, **options: next(handles))
```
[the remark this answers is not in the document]
Ahmath-Gadji: Right. The iterator keeps its list, and the list kept both handles alive, so the one-line form passed. I checked by rewriting all three loaders to `self._pool().<method>.remote(...)`. With the old stub, all 3 cases passed. With this one, all 3 fail with `Lost reference to actor.` and the real code passes.

Fixed in 868a1c1f.

openrag/di/workers.py
Ahmath-Gadji: Agreed that it needs fixing. The failure is different from what you describe, though: two failures, at two limits. On Ray 2.47.1, `limit` is applied by the API server *after* the data source has read, and exceeding it doesn't raise: it keeps the first 100 records by actor id and drops the rest. `raise_on_missing_output` only fires when the data source (GCS) itself stops short, and that happens at `RAY_MAX_LIMIT_FROM_DATA_SOURCE` = 10k records across all namespaces. So:

- **More than 100 dead `openrag` records.** A down actor can disappear from the list, with no error. On a local Ray 2.47.1: after 149 restarts of `Pool` plus one killed `Down`, the previous commit listed only `Pool`.
- **More than 10k actor records in the GCS** (Ray keeps up to 100k dead ones). Both queries raise and the endpoint returns a 500. With `RAY_MAX_LIMIT_FROM_DATA_SOURCE=50`, the previous commit raised `RayStateApiException`.

Fixed in 589bc083:

- Both queries pass `limit=10_000`, the most Ray's API server accepts. A higher limit is rejected.
- **Live query:** `raise_on_missing_output=False`. The GCS reads every registered actor before any destroyed one (`GcsActorManager::HandleGetAllActorInfo` in ray-2.47.1), so anything cut at the cap is a dead record, and this query drops those anyway. The live list stays complete.
- **Dead query:** if it raises, a warning is logged and the query retries with `raise_on_missing_output=False`, listing what Ray returned. Past the cap that is an arbitrary subset, so a down actor can lose its row, but the endpoint still answers. A real API-server error fails the retry too and raises as before.

Tests: the fake `list_actors` now behaves like Ray's state API: default `limit=100`, the data-source cap, GCS read order, and the raise. There's a regression test with 150 dead records and a down actor that sorts after them, and one past the data-source cap. Both fail on the previous commit.

The same checks on a local Ray:
- With default caps, `Down` is listed.
- With the cap at 50, both live actors are listed and the warning is logged.

openrag/di/workers.py
```diff
@@ -19,12 +28,61 @@ def list_ray_actors() -> list[dict[str, str | None]]:
...
+    from ray.util.state.exception import RayStateApiException
+
+    in_namespace = ("ray_namespace", "=", OPENRAG_NAMESPACE)
+    # Whatever Ray's data source cuts here is dead records, which this query
+    # filters out anyway, so its "missing output" error would be a false alarm.
+    actors = list(
+        list_actors(
+            filters=[in_namespace, ("state", "!=", "DEAD")],
+            limit=_ACTOR_LIST_LIMIT,
+            raise_on_missing_output=False,
```
hedhoud: Suppressing missing-output errors here is only safe if Ray always returns every non-DEAD actor before the data-source cap. The GCS does iterate `registered_actors_` before destroyed actors, but that map can itself contain restartable actors in `DEAD` state, and its order does not guarantee that ALIVE or pending actors come first. The test fake currently sorts non-DEAD first and therefore bakes in the assumption being tested. Please query each non-DEAD state with an equality filter while retaining completeness checks, or explicitly surface an incomplete live inventory instead of silently omitting active actors.

hedhoud (commented):
The previous limit and weak-reference test concerns are addressed. One completeness assumption in the live-actor query still needs attention.

openrag/di/workers.py
Ahmath-Gadji: You're right, and I overstated it. When an owned actor dies with no restarts left, `RestartActor` marks it `DEAD` but leaves it in `registered_actors_` until its owner drops the handle, and that map has no order. So a live actor can sit behind dead records, and the fake's live-first sort assumed exactly what the test was meant to check.

Reproduced on a local Ray 2.47.1 with the data-source cap lowered to 50. I made 80 owned actors crash with `max_restarts=0` while the driver still held their handles, alongside 5 live detached `openrag` actors. 589bc083 listed 2 of the 5 live actors, with no error.

Fixed in 6b03544a with both options:

- **One query per state** (`DEPENDENCIES_UNREADY`, `PENDING_CREATION`, `ALIVE`, `RESTARTING`, `DEAD`), each with a `state =` filter. Ray applies that filter at the data source, so only records in that state count toward the cap. The completeness check is back on for every query.
- **Incomplete results are surfaced.** If a query still comes back short, the listing returns what Ray sent, sets `"complete": false` in the response and logs a warning. A full page also counts as incomplete, in case the data-source cap was raised above our limit. The route description documents the field.

The fake now reads records in the order the test lists them, and only matching records count toward the cap. The cap test puts the dead records first; against 589bc083's logic, neither live actor is listed. In the same scenario on real Ray, all 5 live actors are listed, with `complete: false` because the dead query hit the cap.

openrag/di/workers.py
```diff
@@ -19,12 +28,61 @@ def list_ray_actors() -> list[dict[str, str | None]]:
...
+        )
+    )
+    dead_filters = [in_namespace, ("state", "=", "DEAD")]
+    try:
+        dead = list_actors(filters=dead_filters, limit=_ACTOR_LIST_LIMIT)
```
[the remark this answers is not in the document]
Ahmath-Gadji: Done in 6b03544a. `GET /actors/` now returns `{"actors": [...], "complete": bool}`:

- `complete` is false when any of the per-state queries reports that Ray stopped short, or fills the whole page.
- The actors Ray did return are still listed, so the endpoint answers instead of failing.
- The route description documents the field.

Existing readers (the admin UI and the integration and robot tests) only read `actors`, so they're unaffected. The admin UI doesn't display the flag yet.

hedhoud (approved):
The follow-up addresses the actor-list completeness concerns: Ray is queried by explicit state, truncation is surfaced through the response contract, and the weak-reference regression test now exercises the real failure mode. All CI checks are green.

Issue #958, opened 2026-09-14
Per-embedder vector fields: follow-ups to #762


Ahmath-Gadji:
Follow-ups to #762. #939 makes an embedder swap visible and #952 gives every embedder its own vector field, so a swap no longer mixes two models' vectors in one index. These are what that work surfaced and left out, most urgent first.

### 1. `/search` and MCP search ignore the partition's embedder

`RetrievalService.search` (`openrag/services/orchestrators/retrieval_service.py`) searches through the global `self._searcher`, bound to the **default** embedder, instead of building a per-partition searcher the way chat retrieval already does (`_searcher_factory(partition_cfg.embedder)`). Affected: the three `GET /search` routes and the MCP search tool. Chat and indexing are correct.

Schema v3 (#952) changes the symptom, not the cause. Before, such a search embedded the query with the wrong model and scored it against every vector; now it reads only the default embedder's field, so a partition on any other embedder gets **no results** from `/search`, with no error. Multi-partition search across embedders needs one query embedding per embedder, then a merge.

### 2. New partitions still follow set-default

`create_partition` defaults `embedder` to the `default` alias, and so does every path that does not name one: the API without `embedder`, the legacy create route, implicit creation on upload, MCP, and the seeded partition. The UI pins the concrete name.

SQL revision `c4e8f2a6b913` (#952) pins partitions that were on the alias at upgrade time, but nothing stops new ones from storing it. A later set-default then moves such a partition to another model's field: new uploads land there, and every file indexed before stops being returned until it is re-embedded. #939's drift warning reports it after the fact; nothing prevents it.

Proposed: resolve the alias to the concrete embedder name when a partition is written.

### 3. Re-embed a partition into a new embedder (F4)

#952 makes the swap safe and #939 announces it, but getting the partition's existing files back still takes a full re-upload. Proposed: re-embed each chunk into the new embedder's field while the old field keeps serving searches, with uploads to the partition blocked — and the upload button disabled with an explanation — until it completes. Chunk IDs, text, metadata and BM25 vectors stay untouched, the same partial-upsert approach migration 3 uses.

### 4. An existing vector field is never checked against the embedder's dimension

[...]

PR #994, opened 2026-09-18
feat(partitions): give each embedder its own vector field (2/2, #762)

### LINKED ISSUE #762: Partition embedder/endpoint changes have no guardrails — can silently corrupt retrieval quality for already-indexed partitions

Ahmath-Gadji:
Part 2 of 2 for #762, following #939. Closes #762; follow-ups are in #958.

#939 makes an embedder swap visible; this makes it safe. Every embedder wrote into one shared `vector` field, so a partition that changed model searched two vector spaces in one index. Now each embedder owns a field and a search reads only its own, and a partition that changes embedder re-embeds its files into the new field before switching to it.

### Changes

- **A field per embedder.** `model_endpoints.vector_field` (`vector_<name>`, suffixed on collision) is allocated server-side when an embedder is created, is never settable by a client, and survives renames. SQL revision `c4e8f2a6b913` allocates one for each existing embedder. Partitions on the `default` alias that hold files are pinned to the embedder that is default at upgrade time; empty ones keep following the default, as #939 does at runtime.
- **Fields are added to a live collection.** `ensure_vector_field` adds a nullable, HNSW-indexed field on an embedder's first write. Search skips null rows instead of scoring them as zero.
  *Milvus 3.0.1: 100+ concurrent searches with zero errors while a field was added. A null costs ~4 bytes a row; a zero placeholder costs a full vector plus an extra index.*
- **Reads and writes use the partition's embedder field (schema v3).** Indexing writes into it and every searcher reads from it; BM25 stays shared. The `vector` field is gone, with no fallback: an embedder without a field raises.
- **Migration 3** moves existing vectors into their embedder's field in place (partial upsert, no re-embedding), verifies, then drops `vector`. It refuses up front if any partition can't be routed, aborts if rows change mid-run, and is safe to re-run.
  *Milvus 3.0.1: a synchronous `create_index` on a new field over existing data never finished, so indexes build asynchronously; `IS NULL` is rejected on vector fields, so verification scans.*
- **A partition can move to another embedder.** `POST /partition/{partition}/embedder-swap` walks the partition file by file, embeds each chunk's stored text with the new embedder and writes it into that embedder's field. Nothing is re-parsed or re-chunked; chunk ids, text, metadata and BM25 vectors are untouched. The old field keeps serving searches until every file is done — the partition switches in the transaction that marks the swap complete. Progress is a row in `partition_embedder_swaps`, updated once per file, which doubles as the cancellation check and lets a resumed swap skip what it did; `GET` on the same path returns it, or `null` for a partition that never swapped. Uploads, metadata updates and copies into the partition are fenced out while it runs, and the target endpoint cannot be deleted or renamed under it.
  *Milvus 3.0.1: 500 rows partial-upserted into a field added to an already-sealed collection were all searchable 100 ms after the last write, exact top hit — the index is created on the swap's first write, and rows not yet in a built segment are brute-forced, so completion needs no waiting period.*
- **A copy into a partition on another embedder re-embeds** its chunks into that embedder's field, instead of carrying vectors no search there reads.
- **Vector fields stay out of chunk metadata**, and are refused as metadata keys so a metadata update can't overwrite them.
- **Deleting an embedder drops its field**, freeing one of Milvus's 10 vector-field slots.
  *Milvus 3.0.1: the drop works on a loaded collection without a reload. A write to a dropped field is silently stored in the dynamic field, so `ensure_vector_field` no longer caches which fields exist.*

### Breaking change

Collections below schema v3 are refused for search and indexing (`VDB_SCHEMA_MIGRATION_REQUIRED`), and tools that read Milvus's `vector` field directly stop working. To upgrade:

1. Run the SQL migrations (on startup, or `python -m services.persistence.migrations.run`).
2. Stop OpenRAG and, with `rdb` and `milvus` up, run `services/persistence/migrations/milvus/migrate.py --dry-run`, then without `--dry-run`.
3. Start OpenRAG.

Changing `embedder` through `PATCH /partition/{partition}` on a partition that holds indexed files now returns `409 EMBEDDER_SWAP_REQUIRED` — repointing the field without re-embedding would hide every indexed file. Use the swap endpoint, which re-embeds first; the edit is still accepted on an empty partition.

### Still open (#958)

`/search` and MCP search ignore the partition's embedder; no dimension check on an existing field; docs for migration 3.


hedhoud (commented):
I found three issues that should be resolved before this draft is ready to merge.

openrag/services/orchestrators/model_endpoint_service.py
```diff
@@ -753,7 +757,10 @@ async def delete_model_endpoint(self, name: str, model_type: str) -> None:
         promoted default. A ``chat_llm`` reference is cleared to NULL,
         which is precisely the request-time default it would have fallen back
         to anyway. See ``PgModelEndpointRepository._settle_partition_references``.
+
+        A deleted embedder's vector field is dropped once the delete commits.
         """
+        doomed = await self._repo.get(name, model_type) if model_type == "embedder" else None
```
hedhoud: Please get the vector field from the same locked transaction that deletes the endpoint. This read happens before the lock, so a concurrent rename followed by reuse of the old name can make the transaction delete the new endpoint while this code drops the renamed endpoint's live field. That would make its indexed partitions unsearchable. Returning the exact deleted row's field from the repository transaction would keep the identities consistent.

openrag/di/container.py
```diff
@@ -525,11 +526,17 @@ def retrieval_service(self) -> RetrievalService:
...
+            def _vector_field_for(embedder_name: str) -> str | None:
+                endpoint_cfg = settings.models.embedder.get(embedder_name)
+                return endpoint_cfg.vector_field if endpoint_cfg is not None else None
+
             searcher = VectorStoreSearcher(
                 vector_store=self.vector_store,
                 embedder=embedder,
                 document_repo=self.document_repo,
                 collection=settings.vectordb.collection_name,
+                vector_field=lambda: _vector_field_for("default"),
```
hedhoud, line 539: This default-field searcher is still used directly by RetrievalService.search, which powers the /search routes and MCP search. Those paths do not use searcher_factory, so a partition indexed with a non-default embedder now returns no results. Please route raw search through each partition's embedder and merge the results before schema v3 can be merged.

openrag/services/storage/milvus_store.py
```diff
@@ -1337,24 +1412,120 @@ async def ensure_collection(self, name: str, dimension: int, **kwargs: Any) -> N
...
+        another process may have dropped it, and Milvus silently stores a write
+        to a missing field in the dynamic field.
+        """
+        async with self._vector_field_lock:
+            return await asyncio.to_thread(self._ensure_vector_field_sync, field, dimension)
+
+    def _ensure_vector_field_sync(self, field: str, dimension: int) -> bool:
+        description = self._client.describe_collection(self._collection_name)
+        created = False
+        if field not in _dense_fields(description):
```
hedhoud: Please validate the dimension when this field already exists. A failed field drop is intentionally only logged, so recreating an embedder with the same name and a different model can reuse a stale field. This branch accepts it, and every later upload fails at Milvus insertion. Refusing the mismatch here would make the failure immediate and actionable.

openrag/services/orchestrators/model_endpoint_service.py
Ahmath-Gadji: Done. `delete_and_promote_default` now returns the deleted row's `vector_field` (`DELETE … RETURNING`) from its locked transaction, and the service drops that one. The pre-lock read is gone; a test replays the rename race.

openrag/di/container.py
Ahmath-Gadji: I'd keep this for #958 (§1), where it's tracked. On develop these partitions were already ranked with the wrong model's vectors. Now, with hybrid search (the default), they get BM25-only results; they're empty only with hybrid off, or before anything is indexed with the default embedder. Chat already routes per partition.

openrag/services/storage/milvus_store.py
Ahmath-Gadji: Done. `ensure_vector_field` now refuses an existing field of another dimension, telling the admin to register the model under another name, before anything is written.

docs/content/docs/documentation/API.mdx
```diff
@@ -665,15 +665,15 @@ Changing an **embedder**'s `endpoint`, `model_name`, `extra.implementation` or `
 
 Files still indexing when such an edit is saved are not counted, since a file is only counted once indexing finishes. Instead, when the file is recorded, the indexer checks that the partition's embedder still has the configuration it embedded with. If it doesn't, the file fails with `EMBEDDER_CHANGED_DURING_INDEXING`, its vectors are removed, and it has to be indexed again. Indexers reload an edited endpoint before embedding, so only files already embedding during the edit are affected.
 
-`DELETE` of an embedder returns `409` while a partition names it or, for the default embedder, while a partition following the `default` alias already holds indexed files. Reassign those partitions first.
+`DELETE` of an embedder returns `409` while a partition names it or, for the default embedder, while a partition following the `default` alias already holds indexed files. Reassign those partitions first. Deleting an embedder also drops its vector field from the collection.
```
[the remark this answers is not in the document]
Ahmath-Gadji: Done: documented as best effort.

openrag/services/persistence/partition_repo.py
```diff
@@ -162,6 +163,37 @@ async def partition_operation_lock(self, name: str) -> AsyncIterator[_PartitionO
                     name,
                 )
 
+    @asynccontextmanager
+    async def copy_lock(self, name: str) -> AsyncIterator[None]:
+        """Mark a copy into partition *name* as in flight, for :meth:`copy_in_progress`.
+
+        Shared: copies never wait on each other, and uploads never take it.
+        """
+        async with self.pool.acquire() as conn:
```
hedhoud: This session-level advisory lock holds a normal Postgres pool connection until the whole copy and re-embedding operation finishes. Several large concurrent copies can therefore exhaust the application pool and make unrelated requests wait or fail. Please keep the in-flight protection, but move it to a dedicated lock connection/pool or another durable marker that does not occupy the request pool for the copy lifetime, and add a regression test for pool exhaustion.

Ahmath-Gadji: Done. A process's copy locks now live on one dedicated connection, outside the request pool: each copy takes and releases one hold on it, so any number of copies costs one connection. If that connection drops, the running copies' holds are taken back on a new one. Integration tests cover both, including a 2-connection pool with 5 copies in flight.

openrag/services/persistence/partition_repo.py
```diff
@@ -134,11 +139,81 @@ async def pin_default_embedder(self, name: str) -> str | None:
...
+
+    def _discard(self) -> None:
+        """Drop the session, and every hold with it; the next use takes back the others'."""
+        if self._conn is not None:
+            self._conn.terminate()
+        self._conn = None
+
+    async def _connection(self) -> asyncpg.Connection:
+        if self._conn is None or self._conn.is_closed():
+            conn = await self._connect()
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed. Losing the lock session now stops every copy holding a lock on it: asyncpg's termination listener cancels them, and they fail with 503 `COPY_INTERRUPTED` instead of finishing unprotected. Nothing re-takes their locks, so there is no gap left. The regression test terminates the lock backend mid-copy, checks that the copy stops, then that `copy_in_progress("p1")` is false and the next copy locks on a new session.

openrag/services/persistence/partition_repo.py
Ahmath-Gadji: Correction to the above: when the lock connection drops, its running copies are now stopped (503 `COPY_INTERRUPTED`) rather than having their holds re-taken later, which left a gap (see CodeRabbit's thread below).

openrag/services/persistence/partition_repo.py
```diff
@@ -134,11 +139,107 @@ async def pin_default_embedder(self, name: str) -> str | None:
...
+            try:
+                conn = await self._connection()
+                await conn.execute(_HOLD_COPY_LOCK_SQL, _PARTITION_COPY_LOCK_NAMESPACE, name)
+            except Exception:  # noqa: BLE001 - most likely a lost session: retry once on a new one
+                self._discard()
+                conn = await self._connection()
+                await conn.execute(_HOLD_COPY_LOCK_SQL, _PARTITION_COPY_LOCK_NAMESPACE, name)
+            hold = _CopyHold(name, conn, task, task.cancelling())
+            self._holds.add(hold)
+            return hold
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed. `hold()` now takes the lock in a shielded task. If the caller is cancelled meanwhile, it waits for that task, releases the lock if it was granted, then re-raises the cancellation. Release was already shielded in `copy_lock`. A unit test cancels a copy after its lock is granted but before the reply comes back, and checks the lock is released; it fails on the previous code.

openrag/services/workers/indexer_pool.py
```diff
@@ -1086,6 +1089,25 @@ def resolve(name: str = "default") -> int | None:
...
+    """
+    models = getattr(cfg, "models", None)
+    named_embedders = models.embedder if models is not None else {}
+    fallback_cfg = _global_embedder_endpoint_config(cfg)
+
+    def resolve(name: str = "default") -> str | None:
+        model_cfg = named_embedders.get(name)
+        if model_cfg is None and name == "default":
+            model_cfg = fallback_cfg
+        return getattr(model_cfg, "vector_field", None) if model_cfg is not None else None
```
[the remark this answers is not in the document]
Ahmath-Gadji: Fixed by refusing it clearly. A stable field for the fallback would split vectors: the global config is the same model as the seeded default, whose registered field is the one search reads.

The state needs no default embedder in the DB. Startup seeds one from env, and the API can neither unset the default nor delete the last embedder. If it happens anyway (a hand edit, or a failed registry load), the vector-field resolver now raises "No embedder endpoint is marked as the default…" before the file is parsed, and a missing default reloads the registry instead of falling back to the global config.

Checked on a live instance with the default flag cleared: the previous build embedded the file, then failed with the misleading migration error; this one fails before parsing with the new message, and indexes again once a default is set. Unit tests cover the resolver, the reload and the early failure; they fail on the previous code.

openrag/services/workers/dispatcher.py
```diff
@@ -642,6 +647,8 @@ async def copy_file(
                 entity["indexed_at"] = indexed_at.isoformat()
                 entities.append(entity)
 
+            if vector_field is not None:
+                await self._move_to_vector_field(entities, vector_field, embedder)
```
hedhoud: Cross-embedder copies re-embed chunks into the target field here, but the catalog insert below omits the target run’s embedder provenance and fingerprint. The source lookup only returns file metadata, so the new row is stored without `indexation_config`. As a result, `count_files_by_embedder` reports every cross-embedder copy as unknown and drift reporting cannot identify the model that produced its vectors. Please record the target embedder provenance in the catalog write and add a regression assertion for the copied row.

Ahmath-Gadji: Done. The copy's catalog row now keeps the source file's `indexation_config`, which still describes vectors copied as they are. A copy that re-embeds records the target embedder there instead (reference, model, endpoint, dimension), and its catalog write passes the target endpoint's fingerprint, so an endpoint edited meanwhile refuses it with `EMBEDDER_CHANGED_DURING_INDEXING`. The copy's chunks carry their own marker and are removed on that refusal.

Unit tests assert the copied row's record in both cases, the fingerprint, and the cleanup; they fail on the previous code. Checked live: after a copy into a partition on another embedder, `indexed_embedders` reports the target embedder with its model and dimension, and search finds the copy.

hedhoud (approved):

Issue #975, opened 2026-09-16
Chart: split the reranker's HF cache so no first-party PVC needs ReadWriteMany


andyne13:
`persistence.accessMode` is `ReadWriteMany` and governs the three first-party PVCs —
`model-weights`, `data` and `venv`. With `ray.enabled: false` (the chart default, where Ray
runs inside the openrag process) two of them have exactly one mounter, and the third has
two only by accident:

| PVC | Mounted by | Needs RWX |
| --- | --- | --- |
| `venv` | openrag | No — and the volume goes away entirely with #972 |
| `data` | openrag | No — and it only exists with `SAVE_UPLOADED_FILES=true` |
| `model-weights` | openrag (`templates/openrag.yaml`) + reranker (`templates/infinity.yaml`) | Yes, today |

The third is the only real case, and the sharing is not deliberate: openrag mounts the
volume at `/app/model_weights` (its `HF_HOME` / `HF_HUB_CACHE` / `XDG_CACHE_HOME`, where
Marker's weights land), and the reranker mounts the *same* volume at
`/app/.cache/huggingface`. Different paths, different content — one PVC doing two unrelated
jobs, which is what forces `ReadWriteMany` on a 50Gi volume.

Giving the reranker its own small PVC leaves every first-party volume single-mounter, and
`persistence.accessMode` can then default to `ReadWriteOnce`.

This matters beyond tidiness. RWX is a genuine constraint on managed Kubernetes: on the
target platform RWX classes are in beta with a 150Gi minimum per volume, and the
alternatives an operator would reach for — Longhorn, or NFS over an RWO volume — each carry
their own failure mode. Nothing in this chart currently needs any of that.

Two things to keep in mind when doing it:

- The reranker's volume is a Hugging Face cache, so it can be sized far below 50Gi, and
  losing it costs a re-download rather than data.
- **`ray.enabled: true` brings the requirement back**: the RayCluster mounts `data`,
  `model-weights` and `venv` on the head and every worker. So RWO should be the default
  with RWX documented as required for the KubeRay path, not removed as an option.

Scope: `infra/charts/openrag-stack/templates/{infinity,pvc}.yaml`, `values.yaml`
(`persistence.accessMode`, `persistence.volumes`).

PR #995, opened 2026-09-18
fix(indexing): carry the upload's own path on Document

### LINKED ISSUE #911: Pooled parsers pass a node-local temp file path across the actor boundary, so the parser tier cannot scale to multiple Ray nodes
## Summary

The pooled parsers hand their worker actors a **filesystem path** that was written
to node-local temporary storage by the caller. Every parse therefore only works
when the calling actor and the worker actor happen to share a filesystem.

On a single-node deployment that is always true, so the defect is invisible. On a
multi-node Ray cluster it makes the parser tier the one part of the pipeline that
cannot be scaled horizontally — which is also the only part that is slow enough to
be worth scaling.

## Trace

`MarkerLoader.parse` materialises the document to a temp file and passes the path
across the actor boundary:

```python
# services/workers/parsers/marker_workers.py
async with document.as_temporary_file() as path:      # written locally
    markdown, images = await self._dispatch(str(path))

async def _convert_pdf(self, file_path: str):          # :480
    return await call_ray_actor_with_timeout(
        self.worker.process_pdf.remote(file_path),     # :482 — only the string crosses
        ...
    )
```

`Document.as_temporary_file` writes through `tempfile.NamedTemporaryFile`
(`core/models/document.py:186`) with no `tempfile.tempdir` / `TMPDIR` override
anywhere in the repository, so the file lands in the process's own temp directory.

`MarkerPool.process_pdf` then reads that same path repeatedly — once for the page
count, then once per page-range chunk, each dispatched to a different worker
actor:

```python
async def process_pdf(self, file_path: str):           # :355
    total_pages = self._get_page_count(file_path)
    ...                                                # -> _process_chunk(file_path, page_range)
```

So each `MarkerWorker` independently reopens a path it may not be able to see.

Same pattern in the other pooled parsers:

- `services/workers/parsers/docling_workers.py:161,177`
- `services/workers/parsers/whisper_workers.py:219,221`

## Why the rest of the pipeline is unaffected

The indexer actors receive a path under `config.paths.data_dir`
(`api/routers/admin/indexing.py:204`), which is a shared volume in every deployment
shape — so chunking, embedding and storage already place freely across nodes. The
temp-file path is the only handoff that is node-local.

## Impact

- **The parser tier cannot scale horizontally.** Adding a second GPU node yields a
  node that runs indexer actors — the fast stages — while parsing stays pinned to
  wherever the caller runs.
- **The failure is intermittent and placement-dependent.** Ray may put a worker on
  the caller's node (works) or elsewhere (`FileNotFoundError`), so the same
  document can succeed and fail across retries.
- Latent today: single-node deployments share a filesystem and never see it.

## Proposed fix

Stop materialising on the caller side — but pass the **path the file already
has**, not the bytes.

Uploads land under `config.paths.data_dir` (`api/routers/admin/indexing.py:198`
and `:204`), which is node-shared in both deployment shapes — Helm
`persistence.accessMode: ReadWriteMany` on the shared `data` volume, Compose
`${DATA_VOLUME:-../../data}:/app/data` — and keeps the original extension
(`make_unique_filename` only prepends a timestamp and random hex). The indexer
actor already receives that path. It is only `Document.as_temporary_file` that
introduces a node-local one.

So `Document` should carry a `source_path`, set where the file is loaded
(`indexer_actor.py:393`, `file_serializer.py:64`), and `as_temporary_file`
should yield it directly when set — writing a temp file only for **derived**
documents that have no file of their own.

This also closes the residency half of #846 in the same change: it removes the
full `read_bytes` *and* the temp-file write, instead of making residency a
permanent requirement.

**Trap — exactly one site.** `model_copy` propagates every field, so
`doc_parser.py:54` (`document.model_copy(update={"raw_bytes": docx_bytes,
"content_type": DocumentType.DOCX})`) would hand `docx_parser` the original
`.doc` path and it would open the wrong file. Derived documents must clear
`source_path` explicitly. The EML sub-documents (`eml_parser.py:200`, `:249`)
are fresh `Document(...)` constructions, so they inherit nothing and fall back
to the bytes path safely.

### Rejected: pass the bytes

The original proposal here was to pass `raw_bytes` across the boundary and let
each worker materialise its own temp file, relying on Ray's object store. That
fixes placement but is the wrong direction on three counts:

- it makes `raw_bytes` a **hard requirement** rather than incidental, foreclosing
  the #846 residency work;
- it adds a per-node object-store replica of every file;
- combined with Marker's page-range chunking it produces **one temp file per
  worker** instead of one per document — a 2,000-page PDF at
  `marker_chunk_size: 10` fans out to 200 chunks.

## Scope

Smaller than first assessed. The three pooled parsers need **no change**: they
already take a path, so they are fixed by what `as_temporary_file` yields.

- `core/models/document.py` — add `source_path`; yield it from `as_temporary_file`
- `services/workers/indexer_actor.py` — set it when loading the document
- `services/workers/parsers/file_serializer.py` — set it on the extract path
- `core/indexing/parsers/doc_parser.py` — clear it on the derived `.docx`

`marker_workers.py`, `docling_workers.py` and `whisper_workers.py` are unchanged.

## Not in scope

Making `marker_num_gpus` reflect real GPU capacity. It defaults to `0.01`, which
tells Ray that 100 Marker actors fit on one GPU, so actors pack onto the first
card rather than spreading. That is a separate prerequisite for multi-node
parsing and deserves its own issue.

andyne13:
The pooled parsers (Marker, Docling, Whisper) hand a filesystem path to worker actors that may be scheduled on another Ray node, but that path came from `Document.as_temporary_file`, which writes a `NamedTemporaryFile` in the calling actor's own temp directory. On a single node this always works; on a multi-node cluster it makes the parser tier the one stage that cannot be scaled horizontally — and it is the only stage slow enough to be worth scaling.

`Document` now carries `source_path`, set where the file is loaded (`_load_document`, `ParserFileSerializer.serialize`), and `as_temporary_file` yields it directly when set. Uploads already live under `config.paths.data_dir`, which is node-shared in both deployment shapes — Helm `persistence.accessMode: ReadWriteMany`, Compose `${DATA_VOLUME}:/app/data` — so the path resolves from any node. Temp files are still written for derived documents that have no file of their own.

The three pooled parsers are unchanged: they already take a path, so they are fixed by what `as_temporary_file` yields.

### Two guards

The yielded `source_path` is **never unlinked** — it is the caller's file, not ours. The upload is purged by `indexer_pool` in a `finally` after indexing settles, so it outlives the pipeline even when `save_uploaded_files` is off.

`model_copy` propagates every field, so `doc_parser` clears `source_path` on the converted `.docx`; otherwise `DocxParser` would reopen the original `.doc` and parse the wrong file. That is the only `model_copy` on a `Document` in the tree — the EML sub-documents are fresh constructions that inherit nothing.

A `source_path` whose extension does not match the requested suffix falls back to writing the bytes, since the sync libraries dispatch on it. One consequence worth knowing: an upload stored without an extension takes the fallback, so the temp-file write still happens for it.

### `raw_bytes` is deliberately untouched

Fourteen parsers gate on `document.raw_bytes`, so dropping the eager read would make each of them silently return an empty `ProcessedDocument`. That is the residency half of #846 and needs a lazy accessor on `Document` first — separate change, and why this is `Closes #911` but only `Refs #846`.

### Verification

11 new tests. Six-mutation matrix, all caught, including "the source branch also unlinks the upload" — the one that would delete a user's file mid-pipeline. 3703 unit tests, `ruff check`, `ruff format --check` and `check_layer_imports.py` clean.

**Caveat, stated rather than implied:** the defect is multi-node-only and cannot be reproduced or end-to-end verified on a single node — on one node the old code works too, which is why it survived. Everything above is unit-level plus mutation testing. The change rests on the shared-volume premise, which I did verify (`ReadWriteMany`, the bind mount, and the upload outliving the pipeline), and that premise is the load-bearing part of the review.

Draft until #984 clears review — the two do not share files, but this is the larger architectural change of the two.


andyne13:
One qualification on the premise, because a chart change landing in parallel undercuts it.

This PR justifies the path handoff by saying `config.paths.data_dir` is node-shared,
citing Helm `persistence.accessMode: ReadWriteMany`. That is true of the chart today, but
**#975 changes that default to `ReadWriteOnce`**. With `ray.enabled: false` — the chart
default, and the configuration the first production deployment is being planned around —
only `model-weights` has two mounters, and that is only because openrag and the reranker
share one PVC for two unrelated caches. Splitting them leaves every first-party volume
single-mounter, so RWO becomes correct, and the target platform's RWX classes are in beta
with a 150Gi minimum per volume — which is why the operator has committed to an RWO-only
cluster.

Nothing here breaks. With Ray embedded in the openrag process there is no cross-node
handoff, so this PR's fix still works; its multi-node benefit simply is not exercised.

Two things worth writing down in the PR body rather than leaving implicit:

- the path handoff needs a **shared** `data` volume, i.e. the RWX case;
- so with the RWO default, enabling `ray.enabled: true` later would leave Ray workers unable
  to mount `data` from another node — the parser tier this PR and #911 exist to make
  scalable would be blocked on a storage-class decision, not on code.

#975 already carries the matching caveat in the other direction ("`ray.enabled: true` brings
the requirement back, so RWO should become the *default*, not the only option"). Stating it
on both sides keeps the two from contradicting each other for whoever reads them next.

hedhoud (approved):
Reviewed the full diff and path lifecycle, including Ray handoff, cleanup, DOC conversion, and the shared-volume assumption. CodeRabbit reported 0 issues. Focused tests (67), Ruff, formatting, layer guard, and all GitHub checks pass.

